PR #27, opened 2019-05-06
Add linked option/lockstep

### LINKED ISSUE #7: Add support for a monorepo that requires everything remain on version lockstep
Required to support emotion

mitchellhamilton:
Closes #7 

I _think_ we ended up deciding on calling the option `linked`, not sure if I'm remembering correctly though.

The behavior here is that nothing changes for adding changesets. You still write changesets for individual packages. During bumping, for each set of linked packages, it'll find the highest bump type and the highest current version of all the packages in the set and and set the new version for all the packages in the set that are being released to highest current version + highest bump type.

# An example 

I have three packages, `pkg-a`, `pkg-b` and `pkg-c`. `pkg-a` and `pkg-b` are linked but `pkg-c` is not so the config looks like this.
```jsx
// stuff...

const versionOptions = {
  ...otherStuff,
  linked: [['pkg-a', 'pkg-b']]
}

// stuff...
```

- `pkg-a` is at `1.0.0`
- `pkg-b` is at `1.0.0`
- `pkg-c` is at `1.0.0`

I have a changeset with a patch for `pkg-a`, minor for `pkg-b` and major for `pkg-c` and I do a release, the resulting versions will be:

- `pkg-a` is at `1.1.0`
- `pkg-b` is at `1.1.0`
- `pkg-c` is at `2.0.0`

I now have another changeset with a minor for `pkg-a` and I do a release, the resulting versions will be:

- `pkg-a` is at `1.2.0`
- `pkg-b` is at `1.1.0`
- `pkg-c` is at `2.0.0`

I now have another changeset with a minor for `pkg-b` and I do a release, the resulting versions will be:

- `pkg-a` is at `1.2.0`
- `pkg-b` is at `1.3.0`
- `pkg-c` is at `2.0.0`

I now have another changeset with patches for all three packages and I do a release, the resulting versions will be:

- `pkg-a` is at `1.3.1`
- `pkg-b` is at `1.3.1`
- `pkg-c` is at `2.0.1`

I'm pretty sure this example shows all the cases for lockstep and it's the behavior I'm pretty sure I want for emotion.


packages/cli/src/__fixtures__/linked-packages/.changeset/config.js
```diff
@@ -0,0 +1,88 @@
+/*
+Hey, welcome to the changeset config! This file has been generated
```
Noviny: It sounds like you didn't want this config file from the replacement readme text?

mitchellhamilton: Whoops, the readme text is wrong, I just copied this fixture from another one and updated the part of the config I needed to.

packages/cli/src/utils/createRelease/index.js
```diff
@@ -35,24 +35,41 @@ import flattenChangesets from "./flattenChangesets";
...
-  const flattenedChangesets = flattenChangesets(changesets);
+  const flattenedChangesets = flattenChangesets(changesets, allLinkedPackages);
+
+  let currentVersions = new Map();
+
+  for (let pkg of allPackages) {
+    currentVersions.set(pkg.name, pkg ? pkg.config.version : null);
+  }
+
+  for (let linkedPackages of allLinkedPackages) {
```
Noviny: What's the reason for iterating over allLinkedPackages both here and in flattenChangesets?

mitchellhamilton: Here, it's finding the highest current version. In flattenChangesets, it's finding the highest bump type.

packages/cli/src/commands/bump/index.js
```diff
@@ -39,7 +39,11 @@ export default async function version(opts) {
   removeEmptyFolders(changesetBase);
 
   const unreleasedChangesets = await getChangesets(changesetBase);
-  const releaseObj = createRelease(unreleasedChangesets, allPackages);
+  const releaseObj = createRelease(
+    unreleasedChangesets,
+    allPackages,
+    config.linked
```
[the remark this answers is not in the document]
mitchellhamilton: Probably, I probably would have done that if we were using a type system here but without a type system, I don't like passing objects with a bunch of stuff because it means changing the object is really hard.

packages/cli/src/utils/createRelease/index.js
```diff
@@ -35,24 +35,41 @@ import flattenChangesets from "./flattenChangesets";
...
+    let highestVersion;
+    for (let linkedPackage of linkedPackages) {
+      let version = currentVersions.get(linkedPackage);
```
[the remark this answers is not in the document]
taras: We need a support group for people who have been affected by `prefer-const` eslint rule.

PR #37, opened 2019-05-09
add validate command

Noviny:
image

So, uh, I wrote this thing, that validates projects for linking like bolt does, but for yarn workspaces, because I couldn't find an easy way to do this in yarn workspaces (the info command tells you mis-aligned things, but not how and how to fix)

The more I did this, the less this feels like changesets. @mitchellhamilton would preconstruct be interested in doing something like this? I feel like preconstruct fix would be nice here. But also, maybe it can't be in preconstruct because it cares about all workspaces?

### DIFF
--- a/packages/cli/src/commands/validate/index.js
+++ b/packages/cli/src/commands/validate/index.js
@@ -1,0 +1,37 @@
+import getWorkspaces from "get-workspaces";
+import semver from "semver";
+import chalk from "chalk";
+
+import { getDependencyVersionRange } from "../../utils/bolt-replacements/getDependencyInfo";
+import logger from "../../utils/logger";
+import versionRangeToRangeType from "../../utils/bolt-replacements/versionRangeToRangeType";
+
+export default async function bumpReleasedPackages(config) {
+  let allPackages = await getWorkspaces(config);
+
+  let errors = [];
+
+  for (let pkg of allPackages) {
+    for (let pkg2 of allPackages) {
+      let depRange = getDependencyVersionRange(pkg2.name, pkg.config);
+      if (!depRange) continue;
+
+      let cannotSymlink = !semver.satisfies(pkg2.config.version, depRange);
+      if (cannotSymlink) {
+        errors.push(
+          chalk`- Update {green ${pkg.name}} to depend on {yellow "${
+            pkg2.name
+          }": "${versionRangeToRangeType(depRange)}${pkg2.config.version}"}`
+        );
+      }
+    }
+  }
+  if (errors.length) {
+    logger.error(
+      "Oh no! Not everything can ge linked! Here's what you should do:"
+    );
+    errors.forEach(a => logger.error(a));
+  } else {
+    logger.success("Everything should link together nicely!");
+  }
+}
--- a/packages/cli/src/index.js
+++ b/packages/cli/src/index.js
@@ -7,6 +7,7 @@ import add from "./commands/add";
 import bump from "./commands/bump";
 import release from "./commands/release";
 import status from "./commands/status";
+import validate from "./commands/validate";
 
 const { input, flags } = meow(
   `
@@ -70,6 +71,10 @@ const cwd = process.cwd();
     } = flags;
 
     switch (input[0]) {
+      case "validate": {
+        await validate({ cwd });
+        return;
+      }
       case "init": {
         await init({ cwd });
         return;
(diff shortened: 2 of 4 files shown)


mitchellhamilton:
I very much want this from some tool. Going back to yarn workspaces after using bolt and not having that validation scares me like going from typed to untyped code.(to a lesser extent though)

This only does one of bolts checks right? It's checking that when a workspace(A) depends on another workspace(B), the version of B in the repo is within the range that A specifies but not that all workspaces that depend on an external package depend on the same range?

> The more I did this, the less this feels like changesets. @mitchellhamilton would preconstruct be interested in doing something like this? I feel like preconstruct fix would be nice here. But also, maybe it can't be in preconstruct because it cares about all workspaces?

That's really interesting, I would have said that of all the current tools we have, changesets would be the best place for this since changesets inherently cares about the versions of packages. As you say, preconstruct probably isn't the best place for it because it cares about all workspaces. preconstruct doesn't really care about how you link packages, it reads yarn workspaces to  and it checks if you are using bolt to know what package manager to use to auto-install packages but they're not really fundamental to what preconstruct does and are just little extras. Someone could use preconstruct with lerna even and it would work. (I'm not trying to say make changesets support lerna, please don't do that)

> I feel like preconstruct fix would be nice here

Changesets could have a fix command too that fixes this, it's not really that complex. This thing is especially easy to fix since there's only one thing you would want to do.


But also, isn't Decombolt is supposed to do things exactly like this? Or is that just an idea that might take a while to become a real thing?

---

Unrelated to this behavior but I had an idea for something that changesets should validate. To ensure that if people modify `changes.json` things aren't broken, some validation could happen in a validate command and  the version command that goes through all the changesets and re-generates the `dependents` and checks that it's not different from what's in `changes.json`. This could also be auto-fixed if changesets had a fix command, though I guess that could encourage people to edit `changes.json` which is probably bad but also, if we know it's valid, what's the harm in people editing it?

packages/cli/src/commands/validate/index.js
```diff
@@ -0,0 +1,37 @@
...
+import logger from "../../utils/logger";
+import versionRangeToRangeType from "../../utils/bolt-replacements/versionRangeToRangeType";
+
+export default async function bumpReleasedPackages(config) {
+  let allPackages = await getWorkspaces(config);
+
+  let errors = [];
+
+  for (let pkg of allPackages) {
+    for (let pkg2 of allPackages) {
```
mitchellhamilton: Performance nit that probably won’t matter but might matter at atlaskit scale but even then probably won’t be a big deal: This isn’t as efficient as it could be since this’ll happen `n^2`  (where n is the number of packages)times when it could be `2n` if you went through all the packages and found the versions for all the packages and put them in a Map and iterated through the packages again to do the checks.

Noviny:
> This only does one of bolts checks right?

Yeah. The other one definitely doesn’t belong in changesets though. A similar hacky implementation can probably be made for the other one. Maybe just make both their own package called `bolt-lint` or something?

>  isn't Decombolt is supposed to do things exactly like this?

Yes, but I don’t know when it’s happening and wanted implementations of these two checks I could throw in projects now. Changesets implicitly does this during the bump command (and throws an error if this doesn’t pass) so this was the easy check to get working.

Currently thinking temporary bolt-check package (or something near that name) until decomp work happens with just these two checks. Probably going to write it in the changesets repo for now and decompose any bits of changesets this package needs.

packages/cli/src/commands/validate/index.js
Noviny: Yep, this needs to be simplified, also tests. This is based on a function in `bump` which has been minimized to just this element which after doing this I also want to refactor!

Pushed without fixing for basically the reason you noted that we will probably never get perf errors with this as it won’t do enough.

mitchellhamilton:
`bolt-check` or something sounds very nice, I would 100% use it. lmk if you want help implementing it.

mitchellhamilton:
Related to this: Yarn v2 is looking really interesting and has this constraints feature which I think would let you express bolt's checks easily. https://yarnpkg.github.io/berry/features/constraints (it's very much not ready yet but might be a thing to think about for the future)

Noviny:
Yeah, it would be able to implement the things we need. Not a superfan of prolog as the definition format but maybe that's fine.

Definitely this package is not intended as the end package.

Noviny:
Closing in favor of #55

8bnns:
.

PR #41, opened 2019-05-14
Convert get-workspaces to TypeScript

mitchellhamilton:
The first of likely many PRs, I just wanted to convert a small thing first.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -6,6 +6,6 @@ module.exports = {
         targets: { node: 8 }
       }
     ],
-    "@babel/preset-flow"
+    "@babel/preset-typescript"
   ]
 };
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,7 +5,10 @@ module.exports = {
       {
         targets: { node: 8 }
       }
-    ],
-    "@babel/preset-typescript"
+    ]
+  ],
+  overrides: [
+    { test: "**/*.ts", presets: ["@babel/preset-typescript"] },
+    { test: "**/*.js", presets: ["@babel/preset-flow"] }
   ]
 };
--- a/packages/get-workspaces/src/index.test.ts
+++ b/packages/get-workspaces/src/index.test.ts
@@ -68,7 +68,7 @@ describe("get-workspaces", () => {
       await getWorkspaces({ cwd });
     } catch (err) {
       expect(err.message).toBe(
-        'The package.json at packages/pkg-a/package.json is missing a "name" field.'
+        'The following package.jsons are missing the "name" field:\npackages/pkg-a/package.json\npackages/pkg-b/package.json'
       );
       return;
     }
--- a/packages/get-workspaces/src/index.test.ts
+++ b/packages/get-workspaces/src/index.test.ts
@@ -5,35 +5,54 @@ describe("get-workspaces", () => {
   it("should resolve yarn workspaces if the yarn option is passed", async () => {
     let cwd = await getFixturePath(__dirname, "yarn-workspace-base");
     const workspaces = await getWorkspaces({ cwd, tools: ["yarn"] });
+    if (workspaces === null) {
+      return expect(workspaces).not.toBeNull();
+    }
     expect(workspaces[0].name).toEqual("yarn-workspace-base-pkg-a");
     expect(workspaces[1].name).toEqual("yarn-workspace-base-pkg-b");
   });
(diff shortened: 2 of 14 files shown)


packages/get-workspaces/src/index.ts
```diff
@@ -6,7 +6,14 @@ import fs from "fs-extra";
...
 
-export default async function getWorkspaces(opts) {
+type Options = {
+  cwd?: string;
+  tools?: Array<"yarn" | "bolt" | "root">;
+};
+
+export default async function getWorkspaces(
+  opts: Options = {}
+): Promise<Array<{ config: object; name?: string; dir: string }> | null> {
```
Noviny: Why is name optional here?

tsconfig.json
```diff
@@ -0,0 +1,63 @@
+{
+  "compilerOptions": {
```
Noviny: What's with all the commented out options? What's the reason for adding these?

packages/get-workspaces/src/index.test.ts
```diff
@@ -5,35 +5,54 @@ describe("get-workspaces", () => {
...
+      return expect(workspaces).not.toBeNull();
+    }
     expect(workspaces[0].name).toEqual("bolt-workspace-pkg-a");
     expect(workspaces[1].name).toEqual("bolt-workspace-pkg-b");
   });
   it("should resolve main package if root option is passed", async () => {
     let cwd = await getFixturePath(__dirname, "root-only");
     const workspaces = await getWorkspaces({ cwd, tools: ["root"] });
+    if (workspaces === null) {
+      return expect(workspaces).not.toBeNull();
```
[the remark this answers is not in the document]
mitchellhamilton: Yeah and sadly you can't express the type `expect(workspaces).not.toBeNull()` as refining `workspaces` to not be null.

packages/get-workspaces/src/index.ts
```diff
@@ -6,7 +6,14 @@ import fs from "fs-extra";
 import path from "path";
 import globby from "globby";
 
-export default async function getWorkspaces(opts) {
+type Options = {
+  cwd?: string;
+  tools?: Array<"yarn" | "bolt" | "root">;
```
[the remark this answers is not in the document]
mitchellhamilton: I doubt it's possible but maybe there's a weird TS thing.¯\\\_(ツ)_/¯

packages/get-workspaces/src/index.ts
mitchellhamilton: Because it could technically not be in the package.json. Maybe we should check that it's there and throw if it's not?

tsconfig.json
mitchellhamilton: I just ran `tsc init` and it wrote all that. I can remove them if you want?

packages/get-workspaces/src/index.ts
Noviny: I'd throw here. We heavily depend on this existing downstream from this.

tsconfig.json
Noviny: I'm fine with this for now. Might come back and Have Opinions about it later, as I keep hearing some ts settings can be footguns, but can never remember which ones they are.

Noviny:
If we could add a throw or more serious erroring when there is no packageName, then I think this would be good to merge.

This should probably have its own changeset as well.

babel.config.js
```diff
@@ -5,7 +5,10 @@ module.exports = {
       {
         targets: { node: 8 }
       }
-    ],
-    "@babel/preset-flow"
+    ]
+  ],
+  overrides: [
+    { test: "**/*.ts", presets: ["@babel/preset-typescript"] },
+    { test: "**/*.js", presets: ["@babel/preset-flow"] }
```
mitchellhamilton: Flow is still used in at least one place it looks like so I'll look at that in another PR but using both of the presets is easy so I've just done that.

packages/get-workspaces/src/index.ts
```diff
@@ -53,6 +53,14 @@ export default async function getWorkspaces(
       .map(async dir =>
         fs.readFile(path.join(dir, "package.json"), "utf8").then(contents => {
           const config = JSON.parse(contents);
+          if (!config.name) {
+            throw new Error(
```
Noviny: Just thinking of an annoyance I’ve had with bolt - can we collect all these failures then throw all of them?

PR #48, opened 2019-05-21
Fix eslint errors

mitchellhamilton:
See inline comments for things


.eslintrc
```diff
@@ -1,7 +1,11 @@
 {
-  "extends": ["standard", "prettier"],
-  "plugins": ["prettier"],
-  "parser": "babel-eslint",
+  "extends": [
+    "plugin:@typescript-eslint/eslint-recommended",
```
mitchellhamilton: This config seemed to be the least obnoxious config and pretty much just disables rules that TypeScript handles like `no-undef` and etc.

.eslintrc
```diff
@@ -1,7 +1,11 @@
...
+    "standard",
+    "prettier"
+  ],
+  "plugins": ["prettier", "@typescript-eslint"],
+  "parser": "@typescript-eslint/parser",
```
mitchellhamilton: This seems to work for JS as well and I think it's necessary to parse the `.d.ts` file.

.eslintrc
```diff
@@ -30,6 +34,12 @@
       "env": {
         "jest": false
       }
+    },
+    {
+      "files": ["*.ts"],
+      "rules": {
+        "prefer-const": "off"
```
mitchellhamilton: `plugin:@typescript-eslint/eslint-recommended` has this.

PR #56, opened 2019-05-29
human readable ids

Noviny:
So many broken tests!

I suspect we will need to mock the id library in testing because randomness is very upsetting.


.changeset/cuddly-eggs-burn/changes.json
```diff
@@ -0,0 +1,4 @@
+{
```
mitchellhamilton: I already love these names.

PR #61, opened 2019-05-31
Noviny/add prettier on changelog write

Noviny:


packages/cli/src/utils/updateChangelog/index.js
```diff
@@ -1,11 +1,14 @@
...
-
 import path from "path";
 import util from "util";
+import prettier from "prettier";
+
 import generateMarkdownTemplate from "./template";
 import logger from "../logger";
 import * as bolt from "../bolt-replacements";
 
+prettier.resolveConfig(process.cwd());
```
mitchellhamilton: What's this for?

Noviny: This gets the local prettier config, so we don't accidentally fight whatever their prettier settings are.

mitchellhamilton: Does it magically set the options or are you meant to use the return value?

Noviny: No, I just misread the API docs.

PR #62, opened 2019-05-31
Add 🦋 to the CLI

mitchellhamilton:
I thought it'd be cool if the CLI had some butterflies.
I'm not 100% sure about the emoji on the questions, maybe it should stay as the question mark? or `🦋 ?` but then it's not aligned so ¯\\\_(ツ)_/¯
Also, maybe all the logged things should be split by `\n` and the emoji should be added before each line for the things that log multiple lines?
changeset cli with butterflies

### DIFF
--- a/.changeset/young-buttons-press/changes.json
+++ b/.changeset/young-buttons-press/changes.json
@@ -1,0 +1,4 @@
+{
+  "releases": [{ "name": "@changesets/cli", "type": "patch" }],
+  "dependents": []
+}
--- a/.changeset/young-buttons-press/changes.md
+++ b/.changeset/young-buttons-press/changes.md
@@ -1,0 +1,1 @@
+Add butterfly emoji prefix to CLI output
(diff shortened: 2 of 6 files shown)


Noviny:
Changeset bot is really getting out of hand.

Noviny:
> I'm not 100% sure about the emoji on the questions

My 'probably too cute' thought is to put 🐛next to questions, and 🦋next to other things.

I'm happy just with the replacement though. Thanks for this! This brings just that extra bit of joy to this project.

mitchellhamilton:
> My 'probably too cute' thought is to put 🐛next to questions, and 🦋next to other things.

I really like the cuteness of that, IMO the biggest problem with 🐛  is that 🐛 isn't the caterpillar emoji(though Apple happens to make it one), it's the bug emoji and bugs have a very different meaning in a software context which could confuse people.

Also, I've updated things so it prefixes multiline strings and everything is aligned so it looks like this:

changeset cli output

Noviny:
Yeah, bug emoji might trick some people.

This looks good! Am I good to merge it?

mitchellhamilton:
Yep, good to merge.

PR #64, opened 2019-06-01
bolt-check --fix WIP

Noviny:
Got one check written, need to finish off the other two (which will probably be one step)

Also, because I am a chaos writer, I have not written or done testing yet, so will be coming back for that.


Noviny:
Amused that this is failing linting because I added bolt-check to it.

PR #74, opened 2019-06-03
Add check banning devDependencies in root package.jsons in bolt-check

### LINKED ISSUE #70: Ban devDeps in the root of monorepos in bolt-check
The distinction does not matter for the root package.json and by letting people choose where to put them, it only creates confusion because people aren’t sure where a dep should go and there isn’t really a “right” answer because it doesn’t affect anything.

mitchellhamilton:
Closes #70

### DIFF
--- a/.changeset/modern-buckets-sneeze/changes.json
+++ b/.changeset/modern-buckets-sneeze/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "bolt-check", "type": "minor" }], "dependents": [] }
--- a/.changeset/modern-buckets-sneeze/changes.md
+++ b/.changeset/modern-buckets-sneeze/changes.md
@@ -1,0 +1,1 @@
+Add check banning devDependencies in root package.jsons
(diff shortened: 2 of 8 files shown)


Noviny:
Nice! Just a quick question, what’s the argument for only banning dev deps? Why not also ban all non main deps?

mitchellhamilton:
I don't think `optionalDependencies` should be banned because there's a legitimate use case for it, a dependency that is only required on certain operating systems, e.g. fsevents.

I don't think `peerDependencies` should be banned because I don't think bolt-check should even look at `peerDependencies` for any package because `peerDependencies` are very different to deps and devDeps, they generally need different ranges than what `dependencies` and `devDependencies` and they also don't affect what version of a dependency is installed. I was trying bolt-check on emotion today and it changed the `peerDependencies` of packages which isn't at all what I wanted because some packages need to have a peer dependency on react but it needs to be a wider range than the locally installed react version. It's also worth noting that bolt and bolt-check only check the devDep if there's a package that is a devDep and peerDep (and similar for other cases) so this hasn't really caused problems in the past but with an autofix, the problem has properly emerged.

And `bundledDependencies` isn't a thing most people know about/ever use/is handled by other parts of bolt-check so I don't think this should care about it.

PR #75, opened 2019-06-04
Change how we ask questions

### LINKED ISSUE #52: Change how we ask questions
Currently we ask question as:

```
1. Select packages to update: [list of all packages]
=>
2. Select bump type for package [patch, minor, major]
=>
Repeat step two for each package
```

This means that the number of questions you have to think about is 1 + packages-bumped, and scales up with larger system changes.

Proposed new format is:

```
1. Select packages to update: [list of all packages]
=>
2. Which of these packages are a major change? [list of selected packages]
=>
3. Which of these packages are a minor change? [list of selected packages - major changed packages)
=>
4. The rest (list them) will be a minor change - is this okay?
```

This reduces the questions to always 4.

---

Other concerns:

**This makes it too easy to mass-update**
We previously removed a bunch of questions because while being the most correct in querying individual information is good, decision overloading users gets worse results than being a bit smarter.

**If I am updating less than three packages, I will get asked more questions**
We could theoretically avoid this by switching to the old format for 1 or 2 packages, but I would like to have a consistent predictable experience when updating packages.

**Could we ask one question only?**

If we had one Very Complicated question that looked like:

```
Select the bump type of all packages:
pkgA *none* | patch | minor | major
pkgB *none* | patch | minor | major
```

and could cursor both up+down and left+right, you could just use one set of selects to choose EVERYTHING.

I... don't actually like this. It feels like to get all the answers right you need to hold too much information in your head at once, once again upping the complexity cost of getting your changeset 'right'.

mitchellhamilton:
Closes #52 

~~Still WIP mainly because I haven't fixed the tests yet but thought I'd open this to get feedback~~ The tests are fixed now.

changeset questions


.changeset/fast-chicken-beam/changes.md
```diff
@@ -0,0 +1 @@
+Improve how the CLI asks questions
```
Noviny: Can we expand this a little more into how it's changed? I'd almost add the gif, but also the gif is 7MB and I don't really want to carry that around with this repo.

packages/cli/package.json
```diff
@@ -17,21 +17,19 @@
...
     "@babel/runtime": "^7.4.4",
-    "@types/inquirer": "^6.0.3",
     "@types/semver": "^6.0.0",
     "@types/uuid": "^3.4.4",
     "bolt": "^0.22.1",
     "boxen": "^1.3.0",
     "chalk": "^2.1.0",
     "cli-table": "^0.3.1",
     "detect-indent": "^6.0.0",
+    "enquirer": "^2.3.0",
```
Noviny: *this* is why things are nicer. I've been wondering why all the changes.

packages/cli/src/commands/add/createChangeset.js
```diff
@@ -162,16 +136,105 @@ export default async function createChangeset(
...
+        formatPkgNameAndVersion(pkgName, pkgJsonsByName.get(pkgName).version)
+      );
+    });
+    let shouldTheRestOfThePkgsBeMinorBumped = await cli.askConfirm(
+      bold(`Are you okay with these packages having a ${blue("patch")} bump?`)
+    );
+    while (!shouldTheRestOfThePkgsBeMinorBumped) {
+      logger.error("The rest of the selected packages must be patch bumped");
+      shouldTheRestOfThePkgsBeMinorBumped = await cli.askConfirm(
+        bold(`Are you okay with these packages having a ${blue("patch")} bump?`)
```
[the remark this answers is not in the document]
mitchellhamilton: I've gotten rid of the question and kept the logging of the packages.

.changeset/fast-chicken-beam/changes.md
mitchellhamilton: Updated the text and embedded to the gif from the GH user content url so it's not in the repo and unless GH dies, it'll be accessible.

PR #77, opened 2019-06-05
Stop checking peerDependencies in bolt-check

### LINKED ISSUE #69: bolt-check --fix improvements
I'm going to test bolt-check on some monorepos and find things that can be improved in the auto fix.

Other fixes for this:
- [ ] Stop looking at peerDeps

mitchellhamilton:
Making this close #69 because I've tried bolt-check on emotion, preconstruct and react-select and `--fix` did what I wanted mostly except for some cases where the correct version of a module was in a workspace and a different wrong version was in the root but it was really easy to change it in the root and run `--fix` again and I think that's a pretty good workflow.


packages/bolt-check/src/bolt-check.ts
```diff
@@ -14,7 +14,6 @@ function versionRangeToRangeType(versionRange: string) {
 const DEPENDENCY_TYPES = [
   "dependencies",
   "devDependencies",
-  "peerDependencies",
```
Noviny: Do we have thoughts on peerDeps checking for internal packages? I know that AK has some packages that are peerDeps of other packages, and we don't completely want to lose track of that.

Going to merge this, but might open an issue for bolt-check supports internal peerDeps.

mitchellhamilton: Is that a thing that yarn checks?

mitchellhamilton: I.e. shouldn't yarn require that when a package has a peerDependency on a package, it also has a devDependency on that package that is in the range of the peerDependency range and bolt-check should check the devDependency. (not sure if yarn actually does those checks though)

Noviny: Same - don't really know. I'll do some digging.

PR #81, opened 2019-06-06
Fix tables in status command from flowing over to the next line

mitchellhamilton:
changeset status --verbose output


Noviny:
Aside, I love that we're adding images to summaries and have a neat solution to stop them bloating the repo. Should make a huge difference.

PR #83, opened 2019-06-06
Add changeset from GitHub comment

emmatown:
Still very WIP but I'm at the stage where I can create a changeset for a given GH repo directly from information from the GH API.

UPDATE: I got it to commit a changeset to a GH repo.


Noviny:
<3 Can't wait.

Just a thing - does it make sense to merge this into the existing bot and have it handle both things?

Noviny:
Glad to see splitting up changeset operations into different packages too. It's going to make changesets a lot easier for other people to work with/build ideas on top of.

Are you mostly doing this ad-hoc as needed? Should proper decomposition (and probably code review) be a task we schedule outside of this work?

mitchellhamilton:
> does it make sense to merge this into the existing bot and have it handle both things?

Yes, that's what I was planning. I'm writing this so that it should be easy to plug it into the current bot without much work.

> Are you mostly doing this ad-hoc as needed?

Yes.

> Should proper decomposition (and probably code review) be a task we schedule outside of this work? Our two biggest code problems are probably still: things built around bolt, things extracted from bolt, mostly wholesale, and the ~2 places we are still calling bolt functions.

Probably, yes. Do you want to open issue with the things you think need to be reviewed, refactored and etc. and then I can add to it?

Noviny:
Random brain splot on how this works - how does this work for PRs from forks? You probably don't have commit permissions to the fork, so can't add it on.

This seems quite relevant because one of the goals of this is "I have a good external contribution, and don't want to delay it by making them write a changeset"

mitchellhamilton:
I was thinking about that and I’m not totally sure, I would assume if the allow maintainers to edit stuff option is enabled on a PR(which it is by default), all the people/bots that can write to the base repo could write to the fork but I need to test that.

Noviny:
BTW, package splitting now has its own issue at #86

mitchellhamilton:
I have some good news, some not so good news and some good news on this.

I got it working with a test GitHub App and it commits things from a comment correctly. (I made it just write the contents of the changeset code block since it's the new format)

And then I tried it with a PR from a fork and it didn't work 😞 

But I had a new idea, since we have single file changesets now, users can themselves commit the changeset from the GH UI and we could provide a link to do that. I experimented a little bit with query paramaters and links like `https://github.com/test-mitchellhamilton/changeset-test/new/master?filename=.changeset/some-human-id.md` work (note the `?filename=`) so we could provide a link like this in the current GH bot comment so maintainers can click a link, write a changeset in GH's editor and commit it.

Noviny:
> I have some good news, some not so good news and some more good news on this.

Nice start - this sounds like a pretty good solution. I'll add it to the list of 'things blocked by me finishing the 2.0 work'.

Some additional thoughts on this approach - can we fill in any details for someone using the API? Like, could we have checkboxes for packages and it auto-adds appropriate yaml, or is that out of scope?

mitchellhamilton:
> can we fill in any details for someone using the API? Like, could we have checkboxes for packages and it auto-adds appropriate yaml, or is that out of scope?

_potentially_, from doing this stuff, I now understand executing globs directly on GitHub APIs is slowish and annoying and the implementation in this PR is different from globby(this one doesn’t allow negations and probably other things) so that would have to be fixed which would be annoying to do.

Also, the bigger blocker is I’m not sure how we can autofill the content of the file, I found the filename query thing by guessing and it worked but I tried content and body and neither worked, I tried some searching but couldn’t find anything either.

So TLDR: not right now but maybe later if we can figure out how to autofill the content.

Noviny:
Fair. Autofill can definitely be a post-launch feature.

mitchellhamilton:
Closing this since this thing won't exist as the code is in this PR, will update the issue for this with the new plan.

Issue #59, opened 2019-05-31
get-workspaces order is not deterministic


Noviny:
When running our testing suite, sometimes `get-workspaces`, or tests that rely on the output of `get-workspaces` fail, because the order of packages changes. This is very bad for tests, but also probably bad for the package.

We need to track down the root cause and then resolve.

PR #89, opened 2019-06-07
Making get-workspaces order deterministic

highvoltag3:
Making sure the arrays are sorted before using them


highvoltag3:
Fixes #59

packages/get-workspaces/src/index.ts
```diff
@@ -67,6 +68,7 @@ export default async function getWorkspaces(
               path.relative(cwd, path.join(dir, "package.json"))
             );
           }
+          pkgJsonsMissingNameField.sort();
```
Noviny: This sort isn't assigning anything, so I believe isn't doing anything.

mitchellhamilton: `sort` mutates the array. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Noviny: 🤦‍♀

In that case, can we move this sort to outside the mapping, since we don't need to sort on each pass.

highvoltag3: > Thanks for this! It should make the library more reliable.
> 
> I am kind of curious on this - I don't know why it isn't deterministic. Did you do any digging to try and figure out why it wasn't consistent?

@Noviny the way it was the order of the packages going in was not guaranteed, so we need to sort the array in order to ensure it's always the same.


Regarding:
> 🤦‍♀
> 
> In that case, can we move this sort to outside the mapping, since we don't need to sort on each pass.

Yeah I thought of creating another array and using the spread operator to bring the array values in there then `.sort()` them, so I can do that but if we don't `.sort()` here the one of the tests still fails every now and then so it "does something"

Happy to get feedback.

mitchellhamilton: Couldn’t you move the sort to where pkgJsonsMissingNameField is used so it’ll only ever happen once?

highvoltag3: Done thanks for the feedback @mitchellhamilton definitively could have handled that better 🙈  I've removed the unnecessary `.sort()` and just fixed the test instead, which is what I should have done in the first place.

mitchellhamilton: IMO the items should still be sorted where they’re being joined in the error message because I think this should ideally be as deterministic as possible and `checking an exact match in a test > checking a partial match` if possible

highvoltag3: Roger that, rolling back the last commits and moving the sort on `pkgJsonsMissingNameField`

packages/get-workspaces/src/index.test.ts
```diff
@@ -67,8 +67,8 @@ describe("get-workspaces", () => {
     try {
       await getWorkspaces({ cwd });
     } catch (err) {
-      expect(err.message).toBe(
-        'The following package.jsons are missing the "name" field:\npackages/pkg-a/package.json\npackages/pkg-b/package.json'
+      expect(err.message).toMatch(
+        /^The following package\.json's are missing the "name" field/
```
mitchellhamilton: Could you change this back to `toBe`?

highvoltag3: Should have happened on the rollback https://github.com/changesets/changesets/compare/4182c4df8bfc7fd1cabb64117aafb20da932d2f9..1ea1d66cc73550f6de62aa541859294bcbbd9e5d

packages/get-workspaces/src/index.ts
highvoltag3: thanks for reviewing :)

PR #91, opened 2019-06-11
Convert more things to TS

mitchellhamilton:
Mostly things from #83

### DIFF
--- a/packages/cli/src/utils/bolt-replacements/versionRangeToRangeType.ts
+++ b/packages/cli/src/utils/bolt-replacements/versionRangeToRangeType.ts
@@ -1,4 +1,6 @@
-export default function versionRangeToRangeType(versionRange) {
+export default function versionRangeToRangeType(
+  versionRange: string
+): "^" | "~" | "" {
   if (versionRange.charAt(0) === "^") return "^";
   if (versionRange.charAt(0) === "~") return "~";
   return "";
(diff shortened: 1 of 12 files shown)


packages/cli/src/utils/bolt-replacements/versionRangeToRangeType.ts
```diff
@@ -1,4 +1,6 @@
-export default function versionRangeToRangeType(versionRange) {
+export default function versionRangeToRangeType(
+  versionRange: string
+): "^" | "~" | "" {
```
Noviny: Unrelated question: we talk about rangeType here, BUT we only support 3 of the [many] types. This hasn't burned us so far, but probably will at some point? Is it worth us trying to proactively fix this?

mitchellhamilton: We might want to improve it but we physically can't handle every case and other types generally aren't used that often so it's probably not something we should spend much time on.

From https://yarnpkg.com/lang/en/docs/cli/upgrade/#toc-yarn-upgrade-package-latest-l-caret-tilde-exact-pattern

> The package.json file will be updated to reflect the latest version range. By default, the existing range specifier in package.json will be reused if it is one of: ^, ~, <=, >, or an exact version. Otherwise, it will be changed to a caret (^).

I think if we had the same behavior as yarn, that would probably work well.

PR #92, opened 2019-06-11
Catch Promise rejection on SIGINT

### LINKED ISSUE #88: Catch the error where you exit the changeset flow halfway through:
image

Seeing this error for a deliberate action is causing me a few moments of confusion each time.

highvoltag3:
Open to feedback fixes #88


packages/cli/src/utils/cli.ts
```diff
@@ -13,6 +13,11 @@ import { prompt } from "enquirer";
 
 const limit = Math.max(termSize().rows - 5, 10);
 
+function handlePromiseOnSigint(message: string) {
+  console.log(message);
```
mitchellhamilton: Could this use `logger.error` from `./logger`?

packages/cli/src/utils/cli.ts
```diff
@@ -30,7 +35,11 @@ async function askCheckboxPlus(
     choices,
     format,
     limit
-  }).then((responses: any) => responses[name]);
+  })
+    .then((responses: any) => responses[name])
+    .catch(() => {
+      handlePromiseOnSigint("Try choosing at least one item");
```
mitchellhamilton: Could this say `Please choose at least one item` for consistency with the other messages?

packages/cli/src/utils/cli.ts
```diff
@@ -59,7 +72,11 @@ async function askConfirm(message: string): Promise<boolean> {
...
     }
-  ]).then((responses: any) => responses[name]);
+  ])
+    .then((responses: any) => responses[name])
+    .catch(() => {
+      handlePromiseOnSigint("Please type Y or N");
+    });
 }
 
 async function askList(
```
mitchellhamilton: I don't think we use askList anywhere now so we should probably remove it.

packages/cli/src/utils/cli.ts
```diff
@@ -44,7 +53,11 @@ async function askQuestion(message: string): Promise<string> {
       // @ts-ignore
       prefix
     }
-  ]).then((responses: any) => responses[name]);
+  ])
+    .then((responses: any) => responses[name])
+    .catch(() => {
+      handlePromiseOnSigint("Please make sure to answer the question(s)");
```
mitchellhamilton: When will someone see this message when they haven't answered multiple questions? Won't it only ever be a single question and then they'll see this message?

packages/cli/src/utils/cli.ts
highvoltag3: Totally can, I thought about it and then I thought I saw I file that didn't use it so I backed out. Will change it

packages/cli/src/utils/cli.ts
highvoltag3: Yeah, I noticed but felt like I should cover it in case it was meant to be used in the future... I'll remove it though.

packages/cli/src/utils/cli.ts
highvoltag3: The user will see it if they CTRL + C while answering questions for example...

mitchellhamilton: Yeah, I mean why does this message say answer the question **(s)**(with an emphasis on the pluralness) They will see this message immediately after they don't answer a question so they will never not answer multiple questions, right?

highvoltag3: Oh, sorry I misunderstood question the first time around. Yes, I can't remember the reason behind it, but I remember doing it on purpose :p yet. You are right though  it will show up after the user doesn't answer a question, so it's redundant I'm removing the `(s)`

packages/cli/src/utils/cli.ts
highvoltag3: Done :)

packages/cli/src/utils/cli.ts
highvoltag3: Done :)

packages/cli/src/utils/cli.ts
highvoltag3: Pushed as well

highvoltag3:
> Thanks for this! I don't know if we need to log anything in this situation, as most CLIs just exit silently.
> 
> Second thing, can we check the type of error and re-throw for errors other than manual process exits to make sure we don't swallow genuine errors.

@Noviny sure, I can add the `err` in the `.catch` and re-throw it in case there's an actual error happening I had done it that way initially but since SIGINT is handled by `process` I figured maybe we don't need it for this use case.
I've re-added that now the question that I have though is, do we want to exit silently or keep the messages for the user? 
In my mind, it's good to provide them with some feedback even if it's redundant (say I know I need to pick a package from the checklist but I want to exit now and hit CTRL + C) re-iterating that a selection is needed or a question needs to be answered kind of sticks to the back of my head if I re-run the command again and go through the flow.

That being said, happy to do it either way. I'll wait for your response before moving forward.

highvoltag3:
> @Noviny sure, I'll make the change.
> 
> Regarding exiting silently vs showing a message In my mind, it's good to provide the user with some feedback even if it's redundant (say I know I need to pick a package from the checklist but I want to exit now and hit CTRL + C) re-iterating that a selection is needed or a question needs to be answered kind of sticks to the back of my head if I re-run the command again and go through the flow.
> 
> That being said, happy to do it either way. I'll wait for your response before moving forward.

Thinking about this a bit more, I think I digressed...

> > Thanks for this! I don't know if we need to log anything in this situation, as most CLIs just exit silently.
IMHO I think it's worth giving the user some context when they are exiting in the middle of the flow (specifically in a step, like answering a question) even though they are quitting purposely. (So, I get why you're saying it) either way is fine with me. :)

> > Second thing, can we check the type of error and re-throw for errors other than manual process exits to make sure we don't swallow genuine errors.
Technically this was already happening, I didn't actually plug into the Ctrl + C (SIGINT) event, rather I addressed the errors you were seeing which were basically that the promises building the prompts/questions didn't have a `.catch()` so ctrl + c'ing out of them threw an error because the execution continued to the next question and since it depends on the previous answers it would 💩

Noviny:
> IMHO I think it's worth giving the user some context

Possibly some context line is good, but the context seems more like `process manually exited - no changeset written`. `Please choose at least one item` suggests the process failed because they did not make a valid selection, which is not what this is an error for. Giving a message about what they should have done suggests that they made a mistake, which is not what we want to communicate.

If we wanted to let them know what step they exited on, we could go with:

```
Process manually exited while selecting packages to bump
```

etc, but that can be seen already in the terminal.

This is also why I want to detect for SIGINT, as it's the only known error and just rethrow other things. The way this is set up, it will display this message for any error that occurs during steps, and if there is a bug that causes the process to crash, it will hide that, and return this message that implies it's the user's fault (in how I am reading this).

highvoltag3:
Fair enough, I'll remove the `.catch()` for the promises in the flow and intercept SIGINT it should be "easier" just need to figure out how to avoid Enquirer from catching CTRL+C or bubbling it up to   `process`

highvoltag3:
Should be all good for another review :)

Issue #84, opened 2019-06-06
The wrong dependents are generated in some rare cases with `linked` and changing to single file changesets


mitchellhamilton:
`changeset add` currently doesn't care about the `linked` option and that creates a problem if a package is bumped at a higher level because of `linked` which means there should be some dependents.

`A`, `B` and `C` are all linked together and they are all currently at 1.0.0. `B` has a dependency on `C`.

A changeset is added:

```
A@major
C@patch
```

There will be no dependents generated. When bumping though, `A` and `C` will both be released as `2.0.0` and `B` won't be released which is wrong since its dependency on `C` is out of range now.

Noviny:
😬

I guess step 1 is having validation in `bump`, then making a call whether to auto-fix this vs warn and stop. Some form of check and check --fix command would help here too right?

Reason I'm thinking of this being later is you also need to cope with these two changes being between two changesets.

mitchellhamilton:
*internal screaming because of all the times when dependents is inaccurate now*

I feel like the problem here is that we're trying to make things like linked and pre-releases fit into the model of dependents being purely determined by their releases when that's not really true in many cases even beyond those two features. Here's another example:

I have one package `A@1.0.0` in a monorepo. I add a changeset with a major release. 

I add a new package `B@1.0.0` which has a dependency on `A`. I add a changeset with a major release on `B`. Will the dependency  on `B` be updated since the changeset for `A` doesn't include the dependent on `B`?

A check and check --fix command would solve this particular case but I feel like having changesets become invalid and need fixing over time is a bad thing and a --fix command wouldn't fix the case where changes are between two changesets because where would the dependents go since the dependents would be determined by multiple changesets?

I feel like the answer to all of this is to remove the dependents field from a changeset. I don't totally know about this and of course it means that a person won't know all the packages that will be released when creating a changeset but as we know, that's sometimes wrong and IMO `not knowing what will be released > being told incorrect information about what will be released`.

Noviny:
> not knowing what will be released > being told incorrect information about what will be released

Probably true. I'd want to modify the `status` command so it would tell you extra stuff it was going to release, but otherwise fine.

I think I want to do this as a breaking change?

This leaves us with two breakin changes to do:

1) remove `dependents` from changeset object - generate dependents to be bumped as part of `bump`, not `add`.
2) Redo the changesets config

Does this sound right?

mitchellhamilton:
[...]

PR #93, opened 2019-06-13
VERY WIP - adding many packages

Noviny:
Output of #84 

- current status: no tests, everything is broken, not worth looking at.

Added `@changesets/parse` and `@changesets/read`


Noviny:
Of note @mitchellhamilton, I've added a types package as a place to store all the types that are used throughout the system. I'm hoisting some non-obvious types into it (`Workspace` for example, so everything that wants types has one dependency, and to try and ensure no circular deps just for types

Noviny:
I'm open that packages may also want to export their own most important things, so people relying on `get-workspaces` don't have to unintuitively pull in `@changesets/types`

Noviny:
Also, considering just scoping all packages to changesets - some of these should be switched to decombolt's namespace once that exists, but don't want to worry about name collisions.

mitchellhamilton:
> types package

👍 

> Also, considering just scoping all packages to changesets - some of these should be switched to decombolt's namespace once that exists, but don't want to worry about name collisions.

By all packages, do you mean all new packages or all packages(as in bolt-check and get-workspaces as well?)

Assuming you mean all new packages, I'm a 👍 to that.

> I'm open that packages may also want to export their own most important things, so people relying on get-workspaces don't have to unintuitively pull in @changesets/types

Assuming get-workspaces will still be unscoped then it should _probably_ re-export the types(though anything on the scope shouldn't re-export types), I've had re-exporting things across packages cause confusion before though (react-emotion and emotion in v9, react-emotion re-exported all of emotion's exported and it confused people as to where they should import things from) so I'm slightly hesitant to say it's a good idea but it's probably fine/it doesn't really matter in this case.

packages/parse/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+): {
+  summary: string;
+  releases: Release[];
+} {
+  if (!contents.startsWith("---")) {
+    throw new Error(
+      `could not parse changeset - invalid frontmatter: ${contents}`
+    );
+  }
+  let [, roughReleases, roughSummary] = contents.split("---");
```
mitchellhamilton: This would break if the summary contains --- or if a package name contains ---. This regex should work better, `/\s*---([^]+?)\n\s*---\n([^]+)/`

Noviny:
Next goals:

a) get `add` tests passing, move a lot of the resolution checking that is absent from `add` into `determine-dependents`
b) update bump to use new functions, get bump tests passing
c) get status working off new functions

There's still some in-package testing and docs to add after that.

Noviny:
Broke this up into multiple other PRs

PR #94, opened 2019-06-14
better docs

Noviny:


packages/cli/README.md
```diff
@@ -6,6 +6,23 @@ single-package repositories too).
 
 This package is intended as a successor to `@atlaskit/build-releases` with a more general focus.
 
+## Getting Started
+
+```
+yarn add --dev @changesets/cli
```
mitchellhamilton: I feel like `--dev` is probably a good thing here because it will do the right thing in multi and single package repos but then maybe bolt-check should ban regular deps at the root and only allow devDeps?

Noviny: Oh no - things working together. Need more better docs. I want to build a website with a question section to guide you to only the info you need for your project.

One day. Add it to the pile.

PR #106, opened 2019-07-01
Remove dep on bolt and fix 2fa stuff

mitchellhamilton:


package.json
```diff
@@ -80,7 +80,10 @@
     "typescript": "^3.4.5",
     "uuid": "^3.3.2",
     "strip-ansi": "^5.2.0",
-    "term-size": "^2.1.0"
+    "term-size": "^2.1.0",
+    "p-limit": "^2.2.0",
+    "spawndamnit": "^2.0.0",
+    "is-ci": "^2.0.0"
```
Noviny: I couldn't find any instance where we are using bolt in our packages - can we remove the dependency from here entirely?

packages/cli/src/commands/bump/index.js
```diff
@@ -130,13 +136,19 @@ async function bumpReleasedPackages(releaseObj, allPackages, config) {
...
         throw new Error(
-          boltMessages.invalidBoltWorkspacesFromUpdate(
-            pkg.name,
-            depName,
-            depRange,
+          `${importantSeparator}
+  ${
+    pkg.name
+  } has a dependency on ${depName} at ${depRange}, however the new version of ${
+            // TODO: look into this, accessing internalDeps[depName] seems wrong
```
Noviny: I think these checks may still make sense, but it's more 'release plan will create invalid state' - We may want to lean on `bolt-check` to do this, and filter the errors we get?

Noviny: Bump is the thing most impacted by my rewrite, so I wouldn't worry about fixing these now, I'll be doing a pass over the messaging around running bumps as part of that.

mitchellhamilton: Yep, I wrote a comment about changing the error message to something like that above, this comment was purely about accessing `internalDeps[depName]` because `internalDeps` is an array and `depName` is a string so the result will always be undefined.

https://github.com/changesets/changesets/blob/5d03edd84bdc049f6dbe0fcb466dc6e2edb89cdf/packages/cli/src/commands/bump/index.js#L139-L141

PR #115, opened 2019-07-09
Add cli entrypoint to published files so the cli isn't broken in bolt-check

mitchellhamilton:


Noviny:
Quick q - if we move the cli code into `src`, does this also solve the problem? Where possible, I'd like to limit adding new things to the `files` field, as it creates bugs like this.

mitchellhamilton:
The cli code is already in src, the entrypoint is outside of src though, I suppose you could move the entrypoint(as in the dir with the package.json in it) into the src dir but that seems _strange_.

I'm thinking of adding some validation that all entrypoints are in the files field(more technically "things that would be put into the tarball that npm pack would create") to preconstruct.

PR #124, opened 2019-07-17
Noviny/assemble release plan

Noviny:
SOME NOTES:

This works in theory - I've thrown some gnarly test cases at it and it didn't break. Next step is to make this thorough. We had a lot of old tests for updating versions and determining dependencies that I want to rewrite for this, since errors here can be extremely costly.

There's some code neatness that needs to be addressed.

The `test-utils` file is an implementation of @mitchellhamilton's idea to generate a whole state to run tests against. We can set up and modify a data set easily now. I'm still not in love with how I am testing the output though.

Next package after this is the one that takes the output of this and applies it to a repository. After that, probably wire it into the CLI


Noviny:
@mitchellhamilton think this is good for review if you want to look at it.

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -0,0 +1,160 @@
...
+          ) {
+            type = "patch";
+          }
+        }
+        return { name: dependent, type, pkgJSON: dependentPkgJSON };
+      })
+      .filter(({ type }) => type)
+      .forEach(
+        // @ts-ignore - I don't know how to make typescript understand that the filter above guarantees this and I got sick of trying
+        ({ name, type, pkgJSON }: Release & { pkgJSON: PackageJSON }) => {
```
mitchellhamilton: I think I figured this out! I remember reading something like you have to make filter into a [type guard](https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html#user-defined-type-guards) so it would be like this
```tsx
.filter((arg: SomeType): arg is SomeOtherType  => isTheArgTheOtherType)
```
(this is what happens when i spend too much time looking at ASTs)

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -0,0 +1,97 @@
+// This is a modified version of the graph-getting in bolt
+
+import fs from "fs-extra";
+import path from "path";
+import semver from "semver";
+import { PackageJSON, Workspace } from "@changesets/types";
+// @ts-ignore
+import * as boltMessages from "bolt/dist/modern/utils/messages";
```
mitchellhamilton: Can we can inline the messages so we don't need a dep on bolt?

packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -0,0 +1,69 @@
...
+        release.changesets.push(changeset.id);
+      }
+
+      releases.set(name, release);
+    });
+  });
+
+  return [...releases.values()].map(release => {
+    // @ts-ignore
+    let newVersion: string = semver.inc(release.oldVersion, release.type);
```
mitchellhamilton: Why is this erroring/why do we need the ignore?

packages/parse/src/indext.test.ts
```diff
@@ -0,0 +1,139 @@
+import outdent from "outdent";
```
mitchellhamilton: I think this is a broken thing from merge since it was renaming in the other stuff.

packages/git/src/index.test.ts~HEAD
```diff
@@ -0,0 +1,245 @@
+import { copyFixtureIntoTempDir } from "jest-fixtures";
```
mitchellhamilton: Same as previous comment. (weird merge thing)

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -0,0 +1,160 @@
...
+          // For things that are being given a major bump, we check if we have already
+          // added them here. If we have, we update the existing item instead of pushing it on to search.
+          // It is safe to not add it to pkgsToSearch because it should have already been searched at the
+          // largest possible bump type.
+
+          if (existing && type === "major" && existing.type !== "major") {
+            existing.type = "major";
+
+            // @ts-ignore
+            existing.newVersion = semver.inc(
```
mitchellhamilton: What's the error of this and the ts-ignore comment a bit down in this file?

packages/get-dependents-graph/src/get-dependency-graph.ts
Noviny: 👍 I believe I copied this before you removed bolt - will fix for this.

packages/assemble-release-plan/src/flatten-releases.ts
Noviny: semver.inc is technically typed as `string | null` - handling this became a pain and didn't feel like it was adding value.

packages/assemble-release-plan/src/determine-dependents.ts
Noviny: semver.inc strikes again

packages/assemble-release-plan/src/determine-dependents.ts
Noviny: Still couldn't solve this one - leaving it alone. Sorry, friend.

Noviny:
@mitchellhamilton updated yah!

Issue #21, opened 2019-05-03
Ensure writing changelogs will succeed before writing files


Noviny:
Currently if there is an error in getting release lines, the changelog will fail to be written, but the package will be updated. This can lead to lossy changelogs.

Fix:
Generate all changelog updates before writing any files, then, if there are no errors, proceed on to writing the files afterwards.

If there are errors bail completely out of consuming changelogs

Noviny:
I think this wants some minor/major changes to how bumping works. Currently file writing gets handed over to whatever function edits the file. Instead, all file edit functions return the new contents and the file path in an object, but don't write.

Finally, an end process does all file writing.

Bonus, we can add a check to make sure in any bump, we only have one instance of a file to write.

New pattern in bump effectively looks like:

```javascript
const bump = (config) => {
  /* some setup stuff */

  let { errors: changelogErrors, files: changelogFiles } = await updateChangelogs(config)
  let { errors: pkgJsonErrors, files: pkgJsonFiles } = await updateVersions(config)

  if (changelogErrors || pkgJsonErrors ) / * handle this */

  await writeFiles([...changelogFiles, ...pkgJsonFiles])
}
```

PR #131, opened 2019-07-18
Noviny/apply release plan

Noviny:
No tests + JANK code, hasn't been run yet. Can I get a Work in Progress up in this PR?

assuming I wrote perfect code and there are no bugs, this is done 😇


packages/apply-release-plan/src/index.ts
```diff
@@ -0,0 +1,205 @@
...
+  // iterate over releases updating packages
+  let finalisedRelease = await Promise.all(
+    releaseWithChangelogs.map(release => {
+      return versionPackage(release, versionsToUpdate);
+    })
+  );
+
+  let prettierConfig = await prettier.resolveConfig(cwd);
+
+  for (let release of finalisedRelease) {
```
mitchellhamilton: Not something we need to do now but post-v2, let's make this happen in parallel and only write after we've got all the new changelog entries #21

packages/apply-release-plan/src/index.ts
```diff
@@ -0,0 +1,205 @@
...
+function getNewChangelogEntry(
+  releaseWithWorkspaces: ModCompWithWorkspace[],
+  changesets: NewChangeset[],
+  changelogConfig: false | readonly [string, any],
+  cwd: string
+) {
+  let getChangelogFunc: ChangelogFunction = () => Promise.resolve("");
+  let changelogOpts: any;
+  if (changelogConfig) {
+    let changelogPath = resolveFrom(cwd, changelogConfig[0]);
```
mitchellhamilton: This should resolve like this
```suggestion
    let changelogPath = resolveFrom(path.join(cwd, ".changeset"), changelogConfig[0]);
```

Noviny:
> Is packages/apply-release-plan/src/test-utils/get-changelog-dep-updated.ts used at all, it shows as an empty file on GitHub?

It goes along with a bunch of skipped tests so I could get this 'done enough' - deliberately added so its discrepancy is obvious, even though it's an empty file.

Noviny:
> only write after we've got all the new changelog entries

This bit's already implemented but not tested (we call getNewChangelogEntry before we start writing, `finalisedRelease` which we're iterating over here has the written changelogs. #21 should be closed when this is implemented.

Parallelising writing makes sense for later.

PR #136, opened 2019-07-30
apply linked was wrong

### LINKED ISSUE #135: Properly resolve `linked`
`generateReleaseObject` handles links, but it handles them WRONG

Currently it changes the version type and sort of assumes that's correct, but I think the current pattern creates ERRORS mostly because it a) should be updating newVersions, NOT types, and it should calculate off all `linked`, while it currently only does releases

I should fix this but making this issue in case I don't so we can check this before v2 launches

Noviny:
Fixes #135 

No changeset needed as the thing we are patching hasn't been published yet.

@mitchellhamilton - are there any other tests around linked packages that you think would be worth doing while we're here?

### DIFF
--- a/packages/assemble-release-plan/src/apply-links.ts
+++ b/packages/assemble-release-plan/src/apply-links.ts
@@ -1,5 +1,5 @@
 import semver from "semver";
-import { Linked, ComprehensiveRelease } from "@changesets/types";
+import { Linked, ComprehensiveRelease, Workspace, VersionType } from "@changesets/types";
 
 /*
   WARNING:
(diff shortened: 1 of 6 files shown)


emmatown (commented):
I hope London is nice and holidaying is going well!!!

So... I thought of another test case(I pushed it to this branch) and I think it represents a larger problem in the way we're doing this. We have ~~derived~~ contrived state with oldVersion + newVersion + type. Some code is assuming that `semver.inc(oldVersion, type) === newVersion` when that's not true so things like the case I added happen.

I think we need to remove the ~~derived~~ contrived state but I'm not totally sure which two things should be the source of truth. I think the only two that we could make functions of the other two are type(`semver.diff`) and newVersion(`semver.inc`) since I can't find a semver function to take a version and reverse a bump to it. This would sort of mean that oldVersion wouldn't really be oldVersion then, it would more mean "the highest version in a linked group before the release"(if you think of linked as the linked option + a group for each package not in a linked group)

emmatown (approved):
I fixed the things, you should probably look over the fix.

I didn't change the error things so we can talk more about it after v2 is out.

PR #140, opened 2019-08-14
pass dependents to getChangelogFunc

Noviny:
I think this is better. It's not hard, but it's not obvious how to derive this.

### DIFF
--- a/packages/apply-release-plan/src/update-changelogs.ts
+++ b/packages/apply-release-plan/src/update-changelogs.ts
@@ -1,8 +1,7 @@
 import {
-  ComprehensiveRelease,
   NewChangeset,
   ChangelogFunction,
-  PackageJSON
+  ModCompWithWorkspace
 } from "@changesets/types";
 
 import resolveFrom from "resolve-from";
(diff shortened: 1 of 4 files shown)


packages/apply-release-plan/src/update-changelogs.ts
```diff
@@ -0,0 +1,128 @@
...
+      let thisDependentsGraph = dependentsGraph.get(release.name)!;
+
+      const dependents = releaseWithWorkspaces.filter(release =>
+        thisDependentsGraph.includes(release.name)
+      );
+
+      let changelog = await getChangelogFunc(
+        release,
+        relevantChangesets,
+        dependents,
```
mitchellhamilton: Isn't this `Array<ModCompWithWorkspace>` but the types say `Array<Release>`?

Noviny: True - will fix later.

Noviny: (before merging)

PR #146, opened 2019-09-13
Update docs for v2

mitchellhamilton:


packages/cli/README.md
```diff
@@ -95,71 +95,54 @@ This command sets up the `.changeset` folder. It generates a readme and a config
...
 
 ```
-changeset release [--public] [--otp={token}]
+changeset publish [--otp={token}]
 ```
 
-Publishes to NPM repo, and creates tags. Because this command assumes that last commit is the release commit you should not commit any changes between calling `version` and `publish`. These commands are separate to enable you to check if release commit is acurate.
-
-`--public` - enables the `--access-public` flag when publishing. This is required if trying to publish public scoped packages.
+Publishes to NPM repo, and creates git tags. Because this command assumes that last commit is the release commit you should not commit any changes between calling `version` and `publish`. These commands are separate to enable you to check if release commit is acurate.
```
Noviny: I don't agree with this, but this is true in 99% of cases so I think this is fundamentally fine.

PR #154, opened 2019-09-13
take 2-2

Noviny:


packages/cli/CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
 # @changesets/cli
 
-## 2.0.1
```
mitchellhamilton: I think this is being deleted in error?

PR #156, opened 2019-09-16
Fix getting commits for old changesets and small things

mitchellhamilton:


packages/cli/src/commands/version/index.ts
```diff
@@ -52,18 +48,6 @@ async function getOldChangesetsAndWarn(
 
   return thing;
 }
-// this function only exists while we wait for v1 changesets to be obsoleted
-// and should be deleted before v3
-async function cleanupOldChangesets(cwd: string, config: Config) {
```
mitchellhamilton: There was already code in applyReleasePlan to remove old changesets so this isn't necessary

PR #163, opened 2019-09-18
Noviny/use manypkg

Noviny:


Noviny:
> Do you want to remove the things that are at the root that don't need to be there?

Yes, but wanted to just land this and remove them as a separate task (since that is likely to be more manual)

PR #164, opened 2019-09-18
Noviny/add website

Noviny:
@mitchellhamilton I was wanting to add a proper website (biggest motivation is so we can use magical-types to document `@changesets/types`

When I ran the dev command on this branch, I got an error declaring it was an error that should never happen.


mitchellhamilton:
The error is happening because you're importing react-select but it's not installed. If there are type errors, magical-types generally won't work. The error message definitely needs to be better for this.

PR #167, opened 2019-09-20
fix since-master

Noviny:


Noviny:
(for future self - the tests for this were pretty bad)

PR #170, opened 2019-09-24
Changelog packages

mitchellhamilton:


packages/changelog-github/src/index.ts
```diff
@@ -0,0 +1,67 @@
+import { ChangelogFunctions } from "@changesets/types";
+// @ts-ignore
+import { config } from "dotenv";
+import { getInfo } from "@changesets/get-github-info";
+
+config();
```
mitchellhamilton: I have mixed feelings about this. Is adding dotenv an opinion that we should have?

packages/get-github-info/src/index.ts
```diff
@@ -1,7 +1,8 @@
+// @ts-ignore
 import fetch from "node-fetch";
```
mitchellhamilton: I was lazy about typing this file because IMO the external API is the important part to be typed here. Typing the internals would just be painful and unhelpful.

packages/changelog-github/src/index.ts
Noviny: Same - I don't have strong opinions. I think this is an easy path forward so happy to Just Merge for now. What do we do in other projects?

Noviny:
(happy if you want to merge this)

packages/changelog-github/src/index.ts
mitchellhamilton: We use it in other places so I'm happy to Just Merge

PR #173, opened 2019-10-01
Catch errors from git being absent and continue on as best possible

ajaymathur:


packages/git/src/index.ts
```diff
@@ -140,8 +140,13 @@ async function getChangedPackagesSinceCommit(commitHash: string, cwd: string) {
 //
 // Don't use this function in master branch as it returns nothing in that case.
 async function getChangedPackagesSinceMaster(cwd: string) {
```
ajaymathur: I was not able to get it tests in unit test properly. If you have encountered this type of condition please let me know the best way to test it.

packages/git/src/index.ts
```diff
@@ -140,8 +145,13 @@ async function getChangedPackagesSinceCommit(commitHash: string, cwd: string) {
...
-  const masterRef = await getMasterRef(cwd);
-  return getChangedPackagesSinceCommit(masterRef, cwd);
+  try {
+    const masterRef = await getMasterRef(cwd);
+    return getChangedPackagesSinceCommit(masterRef, cwd);
+  } catch {
+    // Do nothing when operation fails
+  }
+  return [];
```
ajaymathur: I was not able to get it tests in unit test properly. If you have encountered this type of condition please let me know the best way to test it.

packages/git/src/index.ts
```diff
@@ -82,25 +84,28 @@ async function getChangedChangesetFilesSinceMaster(
...
+    let tester = /.changeset\/[^/]+\.md$/;
+
+    const files = cmd.stdout
+      .toString()
+      .trim()
+      .split("\n")
+      .filter(file => tester.test(file));
+    if (!fullPath) return files;
+    return files.map(file => path.resolve(cwd, file));
+  } catch {}
```
mitchellhamilton: Could we make this specific to the git errors so we don't silently catch errors from actual potential bugs.

Noviny: Good pattern is:

```js
try {
  // some code at some time
} catch (e) {
  if (e.message.includes('this-specific-thing') {
    // handle the error
  } else {
    throw e
  }
}
```

message is one option, but there may be better things to interrogate.

mitchellhamilton: One thing I like doing is this

```js
class GitError extends Error {}


try {
  // some code at some time
  throw new GitError('the message')
} catch (e) {
  if (e instanceof GitError) {
    // handle the error
  } else {
    throw e
  }
}
```

packages/git/src/index.ts
ajaymathur: If we don't wrap this in try catch, then error will bubble up and it will return an empty array they way we want it to.

ajaymathur:
@Noviny @mitchellhamilton Thanks for the review, PR is ready for review again 🙂

packages/git/src/index.ts
mitchellhamilton: Oh, sorry, I thought this function was calling out to the other one that includes a try catch, makes sense.

mitchellhamilton: wrt testing: using `cwd = await copyFixtureIntoTempDir(__dirname, "with-git")` in a test but without running git init should make it fail.

ajaymathur:
@mitchellhamilton Thanks for adding changeset and updating readme.

I actually added the package as `@changesets/errors` since in future there will more that one error type exported from the package. But yeah now I see having it as error is better. I can update it if you are not already doing it. 😉

packages/git/src/index.ts
ajaymathur: Cool, I will add a test.

mitchellhamilton:
> I actually added the package as @changesets/errors since in future there will more that one error type exported from the package. But yeah now I see having it as error is better. I can update it if you are not already doing it. 😉

Totally agree with having @changesets/errors(including the name), that update to the readme was making it consistent with the name in the package.json

ajaymathur:
I read your commit in a wrong way. Thanks for correcting it, 🙂

PR #176, opened 2019-10-04
Change the log type of the default config writing message

mitchellhamilton:
I thought this seemed strange as an error because this part of it isn't really an error.


mitchellhamilton:
Whoops - I was going to use this for screenshots but I forgot this repo doesn't have the changeset bot. It's probably a good change though anyway.

PR #183, opened 2019-10-09
Prereleases

### LINKED ISSUE #9: Prereleases
A lot of large open source projects need to do prereleases to share things with the community before releases so they can provide feedback without having to do the real release and then having to do lots of major versions when they inevitably need to change things.

The way I'm thinking this would work is that you would have all of your changesets and when you run something like `build-releases version-packages --prerelease=beta`, the tool would calculate all of the versions and then append something like `-beta.0` to them and write all the versions. I'm thinking choosing the npm dist tag to publish to should probably be something that users handle themselves by using the publishConfig field that npm looks at in package.jsons.

Maybe the changesets that are used in a prerelease could be kept in the file system still(in another directory?) so that for the final release, they could all be used to create the changelog entry for the final version.

I'm not totally sure about how specifically all of this should work and this has a bunch of open questions but I think the general idea would be good to have.

(This isn't a need to have for emotion to use it right now but would likely be necessary in the future when we do the next major version but that likely won't be for a couple months I'm guessing)

mitchellhamilton:
## [View the rendered doc](https://github.com/atlassian/changesets/blob/prerelease/docs/prereleases.md)

Closes #9


docs/prereleases.md
```diff
@@ -0,0 +1,140 @@
...
+  pkg-c@3.0.0 has no deps
+.changeset/
+  pkg-b@minor
+```
+
+```
+yarn changeset prerelease next
+```
+
+This command changes Changesets into prerelease mode which creates a `pre.json` file in the `.changeset` directory which stores Some Information Which I Haven't Totally Figured Out Yet.
```
mitchellhamilton: I went back to `pre.json` over putting it in the config because I thought having it another file would make it less likely for people to manually edit it.

Noviny:
Arghfldfdlssjbwjxyil damn GitHub on mobile. Anyway.

I fundamentally agree with the approach but want to tweak some of the user commands. The singular goal is making is impossible to do things ‘accidentally’ as opposed to being fault-tolerant.

Move exit command to:

```bash
yarn changesets prerelease —exit
```

This makes it a binary switch. Running `prerelease [TAG]` while in a prerelease should error, trying to exit should error in a non-pre mode.

The related one is I want to not have it be run `yarn changesets version` to do the versioning. This should throw an error pointing out the prerelease state and pointing to the correct command.

My weakest point is I don’t have a command to use here. Some options:

- `changesets prerelease —version`
- `changesets version —prerelease`
- something else

Basically I want to make it a unique command to switch in, to version in, and to exit.

Interestingly I don’t care about publish, but I am a bit suspicious that it’s just because I know what publish does and how it won’t really care about this internally so I don’t externally care about it appearing different. I think it’s fine though.

There’s an argument that this kind of tagging can be done for things other than pre-releases that I am having with myself but I think I don’t care. Anyone doing such custom publishes I just trust to ignore such words.

Implementation detail, but I would love if `publish` in a pre-state corrected for and avoided overwriting ‘latest’ on npm to make sure those installs happen fine. This may be related to hotfixing patches though and pre-releases might just be fine.

Docs notes for eventually:

- When to use pre
- complexities of pre with monorepos
- complexities of two publish branches in a monorepo and ways it will make you cry.

I may add more later, but overall these are really nits. Most important take aways are I agree with the direction and ideas of this and am super keen to see it fleshed out and implemented 🦋❤️

mitchellhamilton:
> Basically I want to make it a unique command to switch in, to version in, and to exit.

The reason I wanted to not do this is so that we can leverage existing release infrastrucure(i.e. the release action) so essentially we can think of the prerelease state as a really special changeset but I'm interested to hear why a separate command might be better?

> `—`

I assume this is because mobile keyboards and you mean `--`?

> Implementation detail, but I would love if publish in a pre-state corrected for and avoided overwriting ‘latest’ on npm to make sure those installs happen fine. This may be related to hotfixing patches though and pre-releases might just be fine.

Are you referring to the new package during a pre state case writing to latest because otherwise publish should make the dist tag to the tag you specify so it won't overwrite `latest`.

Noviny:
> I assume this is because mobile keyboards and you mean --?

Yes this

> publish should make the dist tag to the tag you specify 

This - looks like it should be fine.

> why a separate command might be better

My reasoning is basically pre-releases are a super-special space. Once you switch into them, things behave differently, so I want to make taking actions different too.

The main case is so you can never take actions unaware that the state you are affecting is a pre-release state.

The flip-side may be that doing accidental pre-releases isn't too bad, alongside doing accidental real releases when you meant to do a pre-release is unlikely.

So maybe it's inflicting pain for no actual benefit to fork the commands.

mitchellhamilton:
> separate command comments

What do you think of loudly saying "you're in prerelease mode, a prerelease is happening" in version + publish to mostly mitigate this?

Another concern of separate commands is, what does get-release-plan do? Does it throw? Accept an option? Does that mean you have to check if a repo is in pre mode before getting a release plan?

Noviny:
> What do you think of loudly saying

This seems mostly fine, but a) so loud we use dominik's letter package, and b) doesn't help if logs are hidden by automation.

> what does get-release-plan do?

There's a general question with that: In pre: do you run `applyPre(getReleasePlan())` or `getReleasePlan({ pre: true })`?

Since `pre` affects versioning, and likely requires another cascading iteration the way `linked` does, I think it's probably the second one, which means you need to know your pre/not-pre state before you begin it.

This is already implied by the fact you're putting pre into `assemble-release-plan`

mitchellhamilton:
> doesn't help if logs are hidden by automation.

I'm thinking we can add a warning to the PR body that the release action creates.

> There's a general question with that: In pre: do you run applyPre(getReleasePlan()) or getReleasePlan({ pre: true })?

I think the answer is `getReleasePlan()`. The reasoning being that doing a non-pre release when in pre mode would break things.

Noviny:
>I'm thinking we can add a warning to the PR body that the release action creates.

I'm happy with this mitigation.

> getReleasePlan()

`:this-tbh:`

`assembleReleasePlan` should take it as an option, but yeah, `getReleasePlan` should take in cwd at most.

packages/apply-release-plan/src/index.ts
```diff
@@ -180,33 +210,69 @@ async function getNewChangelogEntry(
...
+        try {
+          let releaseLines = await getChangelogEntry(
+            release,
+            releaseWithWorkspaces,
+            moddedChangesets,
+            getChangelogFuncs,
+            changelogOpts,
+            preState
+          );
+          if (preStatePackages) {
```
Noviny: Can we add a comment around what this is doing?

To make sure I understand, it's modifying the preStatePackages object, so that at some point later, we can write the release line info? Hence this info being returned?

packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -1,22 +1,14 @@
...
-    string,
-    {
-      name: string;
-      type: VersionType;
-      oldVersion: string;
-      changesets: string[];
-    }
-  > = new Map();
+): Map<string, InternalRelease> {
+  let releases: Map<string, InternalRelease> = new Map();
```
Noviny: releases being a map definitely makes a lot of this a lot neater 👍

packages/assemble-release-plan/src/increment.ts
```diff
@@ -0,0 +1,17 @@
...
+  release: InternalRelease,
+  preState: PreState | undefined
+) {
+  return (
+    semver.inc(release.oldVersion, release.type)! +
+    // why are we adding this ourselves rather than passing 'pre' + versionType to semver.inc?
+    // we want to make the prerelease version(the number at the end) across all packages the name
+    (preState === undefined || preState.mode === "exit"
+      ? ""
+      : `-${preState.tag}.${preState.version}`)
```
Noviny: Just checking I'm reading this right, pre will make all pre versions the same?

So if I do a pre of `button@1.1.0-rc.0` and then I cut a new pre with button and avatar, I'll get:

- `button@1.1.0-rc.1`
- `avatar@1.1.0-rc.1`

say?

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -65,18 +68,17 @@ export default function getDependents(
...
         ) {
           type = "major";
         } else {
           if (
             // TODO validate this - I don't think it's right anymore
-            !releases.some(dep => dep.name === dependent) &&
+            !releases.has(dependent) &&
             !semver.satisfies(
-              semver.inc(nextRelease.oldVersion, nextRelease.type)!,
+              incrementVersion(nextRelease, preState),
```
Noviny: 👍 abstracting this out

packages/assemble-release-plan/src/index.ts
```diff
@@ -2,31 +2,91 @@ import {
...
+  return "patch";
+}
 
 function assembleReleasePlan(
   changesets: NewChangeset[],
   workspaces: Workspace[],
   dependentsGraph: Map<string, string[]>,
-  config: Config
+  config: Config,
+  preState?: PreState
```
Noviny: Explicitly, functions that accept pre seem to also accept undefined, can we be consistent in how we signpost that?

packages/cli/src/commands/pre/index.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      "packages": Object {
+        "pkg-a": Object {
+          "highestVersionType": null,
+          "initialVersion": "1.0.0",
+          "releaseLines": Object {
+            "major": Array [],
+            "minor": Array [],
+            "patch": Array [],
+          },
+        },
```
Noviny: Okay, I think I get this but also I think it's the opposite of what I was thinking of doing.

My thought was you would store an array of the `changesets`, and not remove the changesets.

So this would look like:

```
"pkg-a": Object {
          "initialVersion": "1.0.0",
          "changesets": ["some-changeset-id"]
```

(this contextualises my - why not just write a special changeset when you exit pre btw)

Then, when running version in `pre: true`, you would filter out changesets already appearing, and don't remove the changesets.

Thinking through, this is the least deviation (storing the info here) but also means this is effectively a 'super special extra changeset-ish'?

packages/cli/src/commands/version/version.test.ts
```diff
@@ -181,3 +187,200 @@ describe("running version in a simple project", () => {
     });
   });
 });
+
+const f = fixturez(__dirname);
+
+describe("pre", () => {
+  it("should work", async () => {
```
Noviny: These are really good e2e-ish tests for version. :)

packages/pre/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+import { PreState } from "@changesets/types";
+import getWorkspaces from "get-workspaces";
+import {
+  PreExitButNotInPreModeError,
+  PreEnterButInPreModeError
+} from "@changesets/errors";
+
+export async function readPreState(cwd: string) {
+  let preStatePath = path.resolve(cwd, ".changeset", "pre.json");
+  // TODO: verify that the pre state isn't broken
```
Noviny: Is this a TODO before this ships?

packages/pre/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  await fs.writeFile(
+    preStatePath,
+    JSON.stringify({ ...preState, mode: "exit" }, null, 2) + "\n"
+  );
+}
+
+export async function enterPre(cwd: string, tag: string) {
+  let workspaces = (await getWorkspaces({
+    cwd,
+    tools: ["bolt", "root", "yarn"]
```
Noviny: Confused myself thinking the order here matters and am now wishing this was not an array. Oh well. The future.

packages/assemble-release-plan/src/increment.ts
mitchellhamilton: Yes, that’s right

packages/cli/src/commands/pre/index.test.ts
mitchellhamilton: I just did this because it was the first thing I thought of that would work. Keeping the changesets and filtering already prereleased stuff sounds like a better approach. One minor thing is that I think the changesets array shouldn’t be stored per package, it should be at the root

packages/pre/src/index.ts
mitchellhamilton: No, it’s more a "this would be good but it’s not super important

packages/apply-release-plan/src/index.ts
mitchellhamilton: Yeah, you’re understanding it right but I’m going to change it to storing the already used changesets like you proposed so this isn’t relevant anymore

docs/prereleases.md
```diff
@@ -0,0 +1,137 @@
...
+yarn changeset pre exit
+```
+
+This command will set an intent to exit prerelease mode in the `pre.json` file though it won't do any actual versioning.
+
+```
+yarn changeset version
+```
+
+The version command will apply any changesets currently in the repo and then remove the prerelease tag from the versions. The repo would now look like this:
```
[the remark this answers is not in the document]
mitchellhamilton: I like the idea of setting the initial versions back and writing a changeset to bump the necessary packages and then running version normally. I’m gonna do that.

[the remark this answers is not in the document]
mitchellhamilton: Oh wait, actually, this raises the question of what do we put in the changeset because we wouldn’t have the versions of things(unless we created a release plan to see but that feels not great) so maybe this isn’t the right thing to do.

mitchellhamilton:
One last problem/maybe it's fine:

I have `pkg-a@1.0.0`.

I enter pre with tag `next`.

I add a changeset with `pkg-a@patch`

I run version so `pkg-a` is now at `1.0.1-next.0`.

I add another changeset with `pkg-a@minor`.

I run version, should `pkg-a` be at `1.1.0-next.0` or `1.1.0-next.1`? The current implementation makes it `1.1.0-next.1`, is this the right thing to do?

Noviny:
> One last problem/maybe it's fine:

I don't want to block releasing this to niggle at this problem. I vaguely called it out above (it's the same when you do your first pre-release if there's been other pre-releases already). If we need to solve this, we can store more info, be more clever, but I'd only do that if we discover this is a genuine problem.

Noviny:
image

This error appears when you are running the version after exiting, which is a bit misleading/not ideal.

I ran some manual tests with the version command (I'm pretty comfortable with not testing changes to publish).

Current thought: I might merge and release the current `version` branch, then we merge this in immediately after and release (just making sure those changes are out in case this breaks anything). Then we can have fun from there!

ASISBusiness:
> ## [View the rendered doc](https://github.com/atlassian/changesets/blob/prerelease/docs/prereleases.md)
> Closes #9

PR #185, opened 2019-10-10
Only bump peer dependents when the peer dependency is out of range

emmatown:
Part of the motivation behind this is I want to be able to minor version Emotion. I think doing this is a good thing because people can get back nearly the exact same behavior by changing from a caret to a tilde but people also have more control over when peer dependents are bumped.

This probably isn't something we should do soon since we should probably batch other breaking changes.

### DIFF
--- a/packages/assemble-release-plan/src/index.test.ts
+++ b/packages/assemble-release-plan/src/index.test.ts
@@ -430,8 +430,8 @@ describe("bumping peerDeps", () => {
     expect(releases[0].name).toEqual("pkg-a");
     expect(releases[0].newVersion).toEqual("1.0.1");
   });
-  it("should major bump dependent when bumping caret peerDep by minor", () => {
-    setup.updatePeerDep("pkg-b", "pkg-a", "^1.0.0");
+  it("should major bump dependent when leaving range", () => {
+    setup.updatePeerDep("pkg-b", "pkg-a", "~1.0.0");
     setup.addChangeset({
       id: "anyway-the-windblows",
       releases: [{ name: "pkg-a", type: "minor" }]
@@ -471,7 +471,7 @@ describe("bumping peerDeps", () => {
     expect(releases[1].newVersion).toEqual("2.0.0");
   });
   it("should patch bump transitive dep that is only affected by peerDep bump", () => {
-    setup.updatePeerDep("pkg-b", "pkg-a", "^1.0.0");
+    setup.updatePeerDep("pkg-b", "pkg-a", "~1.0.0");
     setup.addWorkspace("pkg-c", "1.0.0");
     setup.updateDependency("pkg-c", "pkg-b", "^1.0.0");
     setup.addChangeset({
--- a/.changeset/green-flies-ring.md
+++ b/.changeset/green-flies-ring.md
@@ -1,0 +1,8 @@
+---
+"@changesets/cli": major
+"@changesets/assemble-release-plan": minor
+---
+
+Only bump dependents which depend on the given dependency as a peerDependency
+
+If you want to restore the old behaviour, you should change your peerDependencies from using a caret(`^`) to a tilde(`~`)
--- a/.changeset/green-flies-ring.md
+++ b/.changeset/green-flies-ring.md
@@ -3,6 +3,6 @@
 "@changesets/assemble-release-plan": minor
 ---
 
-Only bump dependents which depend on the given dependency as a peerDependency
+Only bump peer dependents when the peer dependency is out of range.
 
 If you want to restore the old behaviour, you should change your peerDependencies from using a caret(`^`) to a tilde(`~`)
--- a/packages/assemble-release-plan/src/determine-dependents.ts
+++ b/packages/assemble-release-plan/src/determine-dependents.ts
@@ -65,6 +65,10 @@ export default function getDependents(
         if (
           depTypes.includes("peerDependencies") &&
           nextRelease.type !== "patch" &&
+          !semver.satisfies(
+            semver.inc(nextRelease.oldVersion, nextRelease.type)!,
+            versionRange
+          ) &&
           (!releases.some(dep => dep.name === dependent) ||
             releases.some(
               dep => dep.name === dependent && dep.type !== "major"
@@ -83,6 +87,7 @@ export default function getDependents(
             type = "patch";
           }
         }
+
         return { name: dependent, type, pkgJSON: dependentPkgJSON };
       })
       .filter(({ type }) => type)


packages/assemble-release-plan/src/index.test.ts
```diff
@@ -430,8 +430,8 @@ describe("bumping peerDeps", () => {
     expect(releases[0].name).toEqual("pkg-a");
     expect(releases[0].newVersion).toEqual("1.0.1");
   });
-  it("should major bump dependent when bumping caret peerDep by minor", () => {
-    setup.updatePeerDep("pkg-b", "pkg-a", "^1.0.0");
```
Noviny: We should add a second test here which is still checking on caret, where a caret peerDep is major bumped.

mitchellhamilton: There's a test for this below this test(it was already there)

Noviny:
> This probably isn't something we should do soon since we should probably batch other breaking changes.

Agree.

Would love @lukebatchelor 's opinion before we merge this change as it impacts atlaskit stuff.

mitchellhamilton:
Also, other problem - when a package has a peer dependency with an exact version and the peer dep is patch bumped, should the peer dependent be patch or major bumped?

lukebatchelor:
Doesn't this PR lead to the issue where your peerDependency introduces some new functionality (minor change), it is still in range so you don't bump, then down the track you start consuming that functionality. At this point, your package.json is actually wrong right? Because you haven't bumped the minimum required version that would expose that feature to you, so its possible that a consumer has an older (but still satisfying) peerDep which breaks (I'll refer to this as the `inconsistent package.json problem`.

Or am I misunderstanding the change?

This does bring up a conversation that has been on our backlog for ages though, that I'll likely create a full issue for us to discuss Soon™.

The current bumping behaviour of automatically bumping the _minimum_ versions of packages when they are released at the same time was added to prevent the `inconsistent package.json problem`. But it has an unintended side effect of creating `"version fragmentation"`. 

That is: it increases the total number of combinations of our packages that are not compatible with eachother, leading to package duplication on the consumer (yarn's upgrade behaviour is very no bueno here).

An example would be:

* Button and modal are both version 1.1.1 and modal has a dep on button of version "^1.0.0"
* A consumer currently has button 1.1.0 and modal 1.1.1
* A feature is released to button (minor change) and consumed in modal at the same time (patch change)
* button 1.2.0 and modal 1.1.2 are released, but modals dep on button is now "^1.2.0" (This is good! it prevents the bug we're talking about!)
* The consumer updates their modal dep to latest (1.2.0)
* The consumers own dep on button does not satisfy modals, so it now pulls in its own copy.

At that scale, its .seriously no big problem, everything working as expected. This falls down when you have hundreds of packages depending on eachother and you need to do a sweeping change on all of them at once.

Not only would we not have hit the `inconsistent package.json problem`, but we've now created a *massive* line in the sand where any consumer upgrading one package past that line, now need to upgrade _every_ package above that line, otherwise they'll get massive duplication.

Theres a bit more nuance to this problem that we need to talk about but thought I'd mention it briefly (really Luke? Briefly?) to get you across it.

mitchellhamilton:
You're understanding this change correctly. The `inconsistent package.json problem` is certainly a concern with doing this but also, isn't it already potentially a problem because of our current behavour with non-peer dependencies since technically this case could happen:

- pkg-a@1.0.0 depends on pkg-b@^1.0.0
- pkg-b is bumped to 1.1.0
- pkg-a is bumped to 1.1.0 and starts depending on behavior added in pkg-b@1.1.0 but a package manager could technically keep pkg-b@1.0.0 installed

I feel like the trade off of "if you want Changesets to bump peer dependents on minors, use a tilde" is good because it allows both options, yes of course it increases the version fragmentation problem but idk, I feel like it's okay. ¯\\\_(ツ)_/¯

Andarist:
> Also, other problem - when a package has a peer dependency with an exact version and the peer dep is patch bumped, should the peer dependent be patch or major bumped?

IMHO, in that case, it would be OK to just patch that peer dependent - given "pragmatically correct semver" policy 😉 

> pkg-a is bumped to 1.1.0 and starts depending on behavior added in pkg-b@1.1.0 but a package manager could technically keep pkg-b@1.0.0 installed

Not sure if I understand this. When pkg-a@1.1.0 could get installed without dragging with itself also pkg-b@1.1.0?

A general note - aren't peer deps somewhat a special case in general? From things I use on a daily basis (or have used) the ones that were peers: React, Babel, TypeScript, Emotion, XState. Not really that many - usually we just deal with regular dependencies. So when one develops a monorepo with a thing used as peer dependency (such as Emotion) they should be aware of the extra risk when it comes to version management around this. Of course, this is a manual~ thing (mental overhead) and thus error-prone but I haven't yet seen many times people failing at this. However - with automated releasing, one's instinct around that might fail more often than when managing things manually.

Going back to how it works now - I don't even think this is an ultimate solution, it just maybe minimizes the risk a little bit, right? The only thing that it actually changes is that one might spot unmet peer dep warning after bumping peer-dependent package - but it's not in any way a solution that really makes things seamless (again - due to nature of peer deps). Most times in the wild when peer dep range changes it's actually a major bump for the peer-dependent package - but there is no way for changesets to automate this reliably. So maybe if it cannot be automated it just shouldn't affect (within minor ranger) other packages at all?

Feiyang1:
Is there any plan to release this change? It is a blocker for us to adopt `changeset` in [firebase-js-sdk](https://github.com/firebase/firebase-js-sdk). I opened an [issue](https://github.com/atlassian/changesets/issues/380) yesterday, and @Andarist pointed me to this PR. Thanks @Andarist !

Currently `changeset` is guessing that the required minimal version of my peerDependencies changed or will change, and bump it for me, which is not necessarily true. What makes it unusable for us is that it triggers a major version bump on the dependents, while in firebase major versions are coordinated, we can't major release something just because of changes in peerDependency version, especially when they are still in range. I agree with @mitchellhamilton that if we want to avoid `inconsistent package.json problem`, we would just use tilde version or exact version instead. 

I realized the current behavior could be desirable for some people, so perhaps we can add an option that allows peerDependencies to remain unchanged if it is in range. Something like `updatePeerDepsWhenOutofRange: true`? This way it is not a breaking change and we can release it sooner. @mitchellhamilton WDYT?

Andarist:
I would be much in favor of releasing this behavior under a flag - so people can actually use and test it, even if it comes as `unsafe*` or similar. I understand being hesitant to just change the current behavior within the current minor version so adding it as a new option (for now) would make a good compromise for both sides of the argument.

mitchellhamilton:
Yeah, :100:, this should exist as an unsafe option.

The way I think this should work is that there should be a `onlyUpdatePeerDependentsWhenOutOfRange` option in a `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH` option in the config.

Some thoughts on this:

When we add an option to this, it's us saying "There's probably a problem in here that needs to be solved but it hasn't been totally figured out yet, here's a solution to it that might be right but also might be wrong, we're not sure yet.".

This addresses the immediate problems that people have by letting people experiment with potential solutions but it also solves a much larger problem.

It doesn't restrict us in changing the implementation of Changesets but kind of more importantly, it doesn't restrict the mental model of what Changesets can do and become in the future because it doesn't force us to release solutions that are incomplete or not totally well thought out which means that instead of having to choose between two bad options in the future(1. Remove a feature which people thought was well-supported and therefore annoy a lot of people. 2. Keep a feature and annoy a lot of people/leave them with bad solutions). It's very important to note here that this does not mean we're saying "we don't care about this problem" instead it means, "we likely care about this problem but we won't restrict ourselves to this particular solution."

A practical thing: We'll likely do lots of majors to internal packages(as in the packages that aren't the cli package) when changing these options because otherwise we would have to use exact ranges everywhere but we won't do majors to the cli package when changing these options because that would create a lot of churn for consumers. Users of the experimental options should pin an exact version of `@changesets/cli`.

Feiyang1:
Thanks for the insight! That sounds reasonable to me. Do you have a sense of when you will be able to get to it (`___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH `)? It is the only blocker that prevents me from using changeset.

mitchellhamilton:
Probably not very soon. We'd be happy to accept a PR for it though. (assuming @Noviny has no objections to this which I doubt they will)

Feiyang1:
https://github.com/atlassian/changesets/pull/383 @mitchellhamilton Can you please take a look?

mitchellhamilton:
Closing this PR because having this behavior by default won't come from this PR

PR #187, opened 2019-10-11
Flag to create empty changeset

### LINKED ISSUE #145: Feature Request: add `--empty` flag to create an empty changeset
G'day.

We're looking at adding enforcement to PRs that will mean that all PRs must have a changeset. To make this cleaner/more consistent, it would be good if we could create empty changesets to signify "I have *deliberately chosen* to not release a package in this PR.

Happy to make a PR for v1.

gardnerjack:
Added the `--empty` flag to the add command, creates an empty changeset:
```
---
---
```
`yarn changeset add --empty` and `yarn changeset --empty`

Needed to change the regex for parsing changeset files because it assumed existing releases and summary. Also added tests for both creating and reading an empty changeset.

Resolves #145


packages/cli/src/types.ts
```diff
@@ -6,6 +6,7 @@ export type CliOptions = {
   verbose?: boolean;
   output?: string;
   otp?: string;
+  empty?: boolean;
```
Noviny: Minor note for learning: empty is actually always a boolean.

Why?

Because there is no 'default' option provided.

We have caught ourselves historically by effectively using 'undefined' as a distinct state for some of these. Here, we don't need a default (we will get false as the default for boolean thanks to meow).

All that said, if you need to update the PR and want to change this, go ahead, but I'm happy to merge this as is.

packages/cli/src/index.ts
```diff
@@ -91,14 +94,15 @@ const cwd = process.cwd();
   }
 
   if (input.length < 1) {
+    const { empty }: CliOptions = flags;
     // @ts-ignore if this is undefined, we have already exited
-    await add(cwd, config);
+    await add(cwd, { empty }, config);
```
Noviny: Good catch on this being the idiomatic way we were adding these config bits. I'm going to leave a second comment which I don't expect actioned in this PR.

---

@mitchellhamilton flagging your here as food-for-thought code tidy - it might be worth making these commands pass config, then... cliConfig(?) as a general pattern, allowing us to provide a default for the third argument of the functions and making tests where you just want the defaults easier. Very low priority thought though.

packages/cli/src/commands/add/__tests__/add.ts
```diff
@@ -88,7 +89,20 @@ describe("Changesets", () => {
...
+    const cwd = await copyFixtureIntoTempDir(__dirname, "simple-project");
+
+    await addChangeset(cwd, { empty: true }, defaultConfig);
+
+    const expectedChangeset = {
+      releases: [],
+      summary: ""
+    };
+    // @ts-ignore
+    const call = writeChangeset.mock.calls[0][0];
```
Noviny: Can we add a test to [writeChangeset](https://github.com/atlassian/changesets/blob/master/packages/cli/src/commands/add/__tests__/writeChangeset.ts)

that tests that when passed this, the write function behaves as intended?

packages/parse/src/index.ts
```diff
@@ -23,10 +23,14 @@ export default function parseChangesetFile(
     const yamlStuff: { [key: string]: VersionType } = yaml.safeLoad(
       roughReleases
     );
-    releases = Object.entries(yamlStuff).map(([name, type]) => ({
-      name,
-      type
-    }));
```
Noviny: I can see you added a test to the 'read' package below - it would be good to add one to 'parse', as this is the place where the change is made (the place in charge of the concern as it were).

Code otherwise looks good.

PR #190, opened 2019-10-11
Create automating-releases.md

Noviny:
I feel a bit cheeky that I mostly link out to our other tools and let their documentation carry them or not, but didn't want to duplicate documentation here.

### DIFF
--- a/docs/automating-changesets.md
+++ b/docs/automating-changesets.md
@@ -1,4 +1,4 @@
-# Automating Releases
+# Automating Changesets
 
 While changesets is designed to work with a fully manual process, it also provides tools to help automate these releases. These can be broken into two major decisions:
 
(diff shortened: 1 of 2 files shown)


docs/automating-changesets.md
```diff
@@ -0,0 +1,51 @@
...
+* A release coordinator (RC) calls to stop any merging to master
+* The RC pull down `master`, runs `changeset version`, then make a new PR with the versioning changes
+* The versioning changes are merged back into master
+* The RC pulls `master` again and runs `changeset publish`
+* The RC runs `git push --follow-tags` to push the release tags back
+* The RC unblocks merging to master
+
+This is a lot of steps, and is quite finnicky (we have to pull from master twice).
+
+We have a [github action](https://github.com/changesets/action) that replaces all these steps with merging a single automatically generated (and updated) version pull request. It has options to run `version` or `version` and `publish`.
```
mitchellhamilton: I feel like this should be above the manual steps so it's the first thing people see if they're skimming it

Noviny: Fair. I sort of realised we hadn't written down the manual steps anywhere in these docs and panicked a bit to add them. Will move them below and massage the text so that makes sense.

PR #201, opened 2019-10-24
HouseKeeping: Moved errors to one place

ajaymathur:
Moved `ExitError` and `ValidationError` out from cli and config package respectively and updated the usage.

### DIFF
--- a/packages/errors/src/index.ts
+++ b/packages/errors/src/index.ts
@@ -1,15 +1,19 @@
 import ExtendableError from "extendable-error";
 
 export class GitError extends ExtendableError {
-  constructor(public code: number, message: string) {
+  code: number;
+  constructor(code: number, message: string) {
     super(`${message}, exit code: ${code}`);
+    this.code = code;
   }
 }
 
 export class ValidationError extends ExtendableError {}
 
 export class ExitError extends ExtendableError {
-  constructor(public code: number) {
+  code: number;
+  constructor(code: number) {
     super(`The process exited with code: ${code}`);
+    this.code = code;
   }
 }
--- a/packages/errors/src/index.ts
+++ b/packages/errors/src/index.ts
@@ -3,3 +3,15 @@ export class GitError extends Error {
     super(`Error: Git - ${message}, exit code: ${exitCode}`);
   }
 }
+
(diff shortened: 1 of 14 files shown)


packages/errors/src/index.ts
```diff
@@ -3,3 +3,15 @@ export class GitError extends Error {
...
 }
+
+export class ValidationError extends Error {
+  constructor(message: string) {
+    super(`Validation Error: ${message}`);
+  }
+}
+
+export class ExitError extends Error {
+  constructor(public code: number) {
```
mitchellhamilton: TIL about this syntax, thanks!

packages/errors/src/index.ts
```diff
@@ -3,3 +3,15 @@ export class GitError extends Error {
     super(`Error: Git - ${message}, exit code: ${exitCode}`);
   }
 }
+
+export class ValidationError extends Error {
+  constructor(message: string) {
+    super(`Validation Error: ${message}`);
```
mitchellhamilton: I'd rather have it pass the message directly because either one of two things will happen with the error, the error will be handled specially and some code can make it say that it's a validation error or the error will be logged and the name of error will be shown.

```suggestion
    super(message);
```

ajaymathur: Hey @mitchellhamilton, 

If I follow your recommendation correctly it is because if we catch the error and log it, We don't want say it is validation error right?

Example:

```
try {
  throw new ValidationError('Something something')
} catch (err) {
  /**
  * Will logging `Validation Error: Something something` be useful or `Something 
  * something`
  **/
  console.log(err);
}
```

ajaymathur:
Strange the tests for failing in CI but pass in local.

packages/errors/src/index.ts
```diff
@@ -1,15 +1,19 @@
 import ExtendableError from "extendable-error";
 
 export class GitError extends ExtendableError {
-  constructor(public code: number, message: string) {
+  code: number;
+  constructor(code: number, message: string) {
```
ajaymathur: Wondering why did this happen. We can look at it later. 🙂

Issue #193, opened 2019-10-14
When setting `access` - respect package-specific config


Noviny:
I don't explicitly know this doesn't work, but we should validate it's fine, or 'fix' it.

## Problem

Currently `access` is set on all publishes, meaning a repo probably can't have both `public` and `restricted` packages (needs actual verification that this will go wrong as well, but I think it should)

## Proposed Solution

The publish command should check if `publishConfig.access` is set in each workspace's `package.json` and then respect the package config over the changeset config.

Andarist:
I'll try to investigate this in the following days and prepare a fix if needed.

PR #203, opened 2019-10-25
fix access to not use wrong names

Noviny:
Coupla things on this PR:

- this might be a breaking change to `config` and `types` - I just don't wanna. My argument on it not being is that the previous state was genuinely in error (a patch). I think this is pragmatically fine.
- Nothing actually failed to work as intended previously. If you passed `--access="private"` to npm, it ignored the flag and did the default option, which was restricted. :badpokerface:
- I added an option to make individual packages have their own `access` field. Note that npm's regular publish does not respect this field. I was unsure whether to a) just not implement this feature and only allow the global access (@Andarist I believe you had a use-case for this though), b) nest this so it's `changesets: { access: "restricted" }`.

### DIFF
--- a/.changeset/proud-rings-appear.md
+++ b/.changeset/proud-rings-appear.md
@@ -1,0 +1,17 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/cli": patch
+"@changesets/config": patch
+"get-workspaces": patch
+"@changesets/types": patch
+---
+
+Correctly handle the 'access' flag for packages
+
+This is a patch, but it's a pretty big one, and I'm pretty 😳 about it. Previously, we had access as "public" or "private", access "private" isn't valid. This was a confusino because there are three states for publishing a package:
+
+- `private: true` - the package will not be published to npm (worked)
+- `access: public` - the package will be publicly published to npm (even if it uses a scope) (worked)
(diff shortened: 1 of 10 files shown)


Noviny:
Fixes #193

emmatown (commented):
Code changes look good to me but I’m not totally sure about making this a patch since this means any project that has access: private in their config (which is the default) will throw an error which feels not good.

Maybe we should not error in config if the value is private and instead normalise it (and maybe log a warning)?

Andarist:
I was actually working on fixing #193 - just didn't have time to test it properly:
https://github.com/Andarist/changesets/commit/82c83baf019720cb3289cf460df05a4b6e150056

I was thinking about using standard publishConfig field for reading this info - wondering why this was not chosen here.

Noviny:
Oh sorry! Didn't mean to to-tread. Happy to wait for your fix. And you're right, we should use publishConfig instead of just `--access`

Andarist:
> Oh sorry! Didn't mean to to-tread.

No worries, I didn't communicate it well to indicate that I've started working on it.

> Happy to wait for your fix. And you're right, we should use publishConfig instead of just --access

Gonna combine my changes with yours and prepare a followup PR to this one later today.

Noviny:
Closing in favour of #204

PR #204, opened 2019-10-25
Respect per-package package.json publishConfig.access

### LINKED ISSUE #193: When setting `access` - respect package-specific config

Andarist:
fixes #193
closes #203 
is combination of #203 and https://github.com/Andarist/changesets/commit/82c83baf019720cb3289cf460df05a4b6e150056

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -1,6 +1,7 @@
 import * as fs from "fs-extra";
 import path from "path";
 import { ValidationError } from "@changesets/errors";
+import { warn } from "@changesets/logger";
 import { Config, WrittenConfig, Workspace } from "@changesets/types";
 import packageJson from "../package.json";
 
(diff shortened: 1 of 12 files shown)


packages/config/src/index.ts
```diff
@@ -57,15 +57,15 @@ export let parse = (
 
   if (
     json.access !== undefined &&
-    json.access !== "private" &&
+    json.access !== "restricted" &&
     json.access !== "public"
```
mitchellhamilton: We should still allow `access === "private"` so that the old default config isn't broken but let's normalize it to restricted and log a warning.

Andarist: ok, I've added handling of this

Noviny:
I think this is good to merge. I'm tempted to merge it despite failing coverage goals, but would be great not to degrade coverage. lmk what you want to do.

Andarist:
I'm afraid I won't be able to work on increasing the coverage in the following days.

PR #206, opened 2019-10-30
Initial attempt at words

Noviny:
Pls excuse me I think the words are still wrong, but wanted to get this up for discussion


docs/dictionary.md
```diff
@@ -24,6 +24,9 @@ Several of these have associated type definitions, which you can find in [our ty
...
 - **release** - The combination of versioning and publishing a package or packages which may include a build process before publishing
+- **prereleases** - A pre-release is a release that uses a tag, and is not published as `latest` on npm. This is designed for when you want to share code, but are not yet ready for it to be the main package used by everyone. Pre-releases for packages are common, but pre-releases within a monorepo presents unique problems. You can see our exhaustive pre-release documentation [in our prereleases documentation](./preleases.md). In addition, there are two different approaches to pre-releases which are defined for use separately.
+- **RC prerelease** - An RC prerelease is done before an intended important release, likely a major release. It includes semver information about what the next intended release version is, a tag, and an iterated number for the prerelease. The output of this is intended to be committed, and then further work done on the branch. RC prereleases are always published to the `next` tag on npm
```
mitchellhamilton: We should probably say that RC means release candidate somewhere

> RC prereleases are always published to the `next` tag on npm

This isn't true, that's what I've been using as an example but the tag that is in the version is what's used for the npm dist tag

mitchellhamilton:
There should probably be examples of what the versions look like for snapshot and RC

docs/dictionary.md
Noviny: I guess ambiguity, but I was assuming if you do `yarn upgrade pkgName@next` you always get the latest RC, rather than the tag is always next. Is this also false?

mitchellhamilton: If you do `yarn upgrade pkgName@next`, you'll get the latest RC on the `next` tag. If you had another pre-release line with the tag `alpha`, it would be available on the `alpha` dist tag.

Issue #211, opened 2019-11-02
get-github-info returning a single links.user


Andarist:
## Problem

I've noticed that this returns a single user - the one that has last touched the changeset file. Ideally, it would return either:
- original author
- all users responsible for a particular changeset

Noviny:
Thanks! This makes sense to me as a thing to fix. I'd probably err for original author (definitely preferable to last person to modify the file)

Probably returning all responsible people in chronological order makes the most sense, and then a person writing the release line generator can set up their desired behaviour for the info.

Andarist:
Just to give this some context - I've just changed a changeset type from minor to patch and thus the change got attributed completely to me (original author being completely lost).

Andarist:
@mitchellhamilton I see you have merged a PR addressing this, was it a full or partial fix for this? Can this issue be closed?

mitchellhamilton:
It fixed it to return the original author and we're probably not going to do the all users responsible for a changeset because that raises a lot more questions so yes, let's close this.

Andarist:
That's totally fine, a single user is most often better. 

There is still a problem with the used approach though - a changeset should be attributed to change author, not the changeset author. And from what I understand you check for the latter. I understand that retrieving a change author is tricky - but maybe something could be done to at least handle this for cases when changesets are created as part of a PR (and not when added directly on the stable branch). Additionally, a frontmatter could support specifying the author somehow.

PR #216, opened 2019-11-04
Get commit that adds a file rather than the one that last modified it

mitchellhamilton:
This addresses the main case from #211

### DIFF
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -54,12 +54,10 @@ async function tag(tagStr: string, cwd: string) {
 async function getCommitThatAddsFile(gitPath: string, cwd: string) {
   const gitCmd = await spawn(
     "git",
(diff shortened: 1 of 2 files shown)


packages/git/src/index.ts
```diff
@@ -54,12 +54,10 @@ async function tag(tagStr: string, cwd: string) {
 async function getCommitThatAddsFile(gitPath: string, cwd: string) {
   const gitCmd = await spawn(
     "git",
-    ["log", "--reverse", "--max-count=1", "--pretty=format:%h", "-p", gitPath],
+    ["log", "--diff-filter=A", "--max-count=1", "--pretty=format:%h", gitPath],
     { cwd }
   );
-  // For reasons I do not understand, passing pretty format through this is not working
-  // The slice below is aimed at achieving the same thing.
```
mitchellhamilton: If you're curious why this wasn't working, it's because we were passing `-p`

PR #221, opened 2019-11-07
--since option for status, baseBranch config option, read v1 changesets from @changesets/read and make uncommitted things show up in changed packages and with --since in status

### LINKED ISSUE #169: New packages don't show up in changed packages

### LINKED ISSUE #191: Update sinceMaster flag to allow setting branches
## Problem

If you are running releases off a release branch instead of master, you can get changesets since `my-branch`, as we explicitly only have a `--sinceMaster` flag.

## Proposed solution

- Expose the function in `@changesets/git` that takes a branch name (which the `sinceMaster` version is using under the hood)
- Add a `branch` option as a string to places which are using `sinceMaster` currently
- Add error handling around if the specified branch doesn't exist
- Add `--branch=...` to `changeset status`

## Deprecating `--sinceMaster`

I'll be honest, this flag always rubbed me the wrong way and it would be great to deprecate it. This can be two-stepped:

- Add deprecation warnings to usages of `sinceMaster` to the PR that adds `branch`
- Open a new PR with label `breaking` which removes `sinceMaster` (to be merged next time we are ready for a breaking release)

### LINKED ISSUE #192: when checking `sinceMaster`, include new changesets
## Problem

Currently, when you run `changeset status --sinceMaster` and you have uncommitted changesets, it missed them. This is a bit frowny face as a new common use-case is add a changeset, check what that releases.

## Proposed solution

Also check for unstaged + uncommitted files

### LINKED ISSUE #219: Configure a branch other than master to use as base
## Affected Packages

`@changesets/cli`

## Problem

When I run `yarn changeset` it tells me that *all* of my packages have changed and require a changeset. In reality, only some of them have changed. The root cause of this is that we use `dev` as our base branch name, rather than `master`. As such, when changesets calls `getMasterRef` it finds a ref from a very long time ago, which is where our `master` branch happens to be.

## Proposed solution

Allow the user to specify a `baseBranch` in `.changeset/config.json`, which defaults to `master`.

mitchellhamilton:
Closes #219 
Closes #191
Closes #169 
Closes #192 

I partially wanted @changesets/read to read old changesets so that status works with old changesets and partially so I can use it in the release action.

I also did some misc refactoring as I came across things.

(Sorry for the huge PR)


packages/cli/src/utils/getChangesetBase.ts
```diff
@@ -1,15 +0,0 @@
...
-  if (!projectDir) {
-    throw new Error("Could not find project directory");
-  }
-  return projectDir;
-}
-
-export default async function getChangesetBase(cwd: string) {
-  const dir = await getProjectDirectory(cwd);
-  return path.resolve(dir, ".changeset");
-}
```
mitchellhamilton: The rationale behind removing getChangesetBase is that we inconsistently use `pkg-dir` and it isn't really useful for us so we might as well get rid of it and if we don't need `pkg-dir`, we might as well get rid of getChangesetBase since `path.resolve(cwd, ".changeset")` is easier than importing this.

packages/cli/src/commands/add/index.ts
```diff
@@ -9,26 +8,14 @@ import { Config } from "@changesets/types";
...
 import printConfirmationMessage from "./messages";
 
 export default async function add(
   cwd: string,
   { empty }: { empty?: boolean },
   config: Config
 ) {
-  const changesetBase = await getChangesetBase(cwd);
-
-  if (!fs.existsSync(changesetBase)) {
```
jamiekyle-eb: I don't understand why these warnings were removed

mitchellhamilton: They weren't removed, they were moved so that they run on all commands(except init) instead of only the add command

packages/cli/src/commands/status/index.ts
```diff
@@ -15,13 +15,28 @@ export default async function getStatus(
...
-  }: { sinceMaster?: boolean; verbose?: boolean; output?: string },
+  }: {
+    sinceMaster?: boolean;
+    since?: string;
+    verbose?: boolean;
+    output?: string;
+  },
   config: Config
 ) {
-  // TODO: Check if we are no master and give a different error message if we are
```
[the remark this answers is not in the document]
mitchellhamilton: I feel like there could be a case where you want `--since=master` while on master because it will now include unstaged changes. Also, there's the problem of how do we know what kind of git ref we're on, I assume we could check if a given string is a branch but idk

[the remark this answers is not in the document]
Noviny: Since we're thoroughly in git world, I think it's fine to do something like`git rev-parse --abbrev-ref HEAD` and then `if (config.since === currentRef)` - erroring out might be too strong, but I wouldn't mind logging a warning at least.

PR #232, opened 2019-12-21
Following conventional commits

harrysolovay:
Related to [changesets/action issue 10](https://github.com/changesets/action/issues/10):

I edited the commit message to reflect the conventional commits standard (and am about to submit [a similar PR](https://github.com/changesets/action/pull/12) to [changesets/action](https://github.com/changesets/action)). I believe this will better satisfy users who already abide by the conventional commits standard. For those who don’t, it would be wonderful if there was a way to template different commit messages. If / when you are ready to think about adding that kind of pluggability, here’re my thoughts:

I’m partial to adding a `commitMessageTemplate` field to the `.changeset/config.json`, which could resemble a tagged template. I’m imagining something along these lines:

```diff
{
  "$schema": "https://unpkg.com/@changesets/config@0.3.0/schema.json",
  "changelog": "@changesets/cli/changelog",
  "commit": true,
+ “commitMessage”: “${type}(${scope}): ${description}”,
  "linked": [],
  "access": "restricted",
  "baseBranch": "master"
}
```

I don’t want to come off as pushy. Please take all of this with a grain of salt. And thanks for considering this change.


Noviny:
Hey! Thanks for patience over holidays. I think this is fine to merge (along with the change to the action). Will try and get this out later this week.

If you can update the PRs, that would be amazing, otherwise I'll take care of that.

harrysolovay:
@Noviny great, thank you! Just updated the branch :)

PR #236, opened 2020-01-02
Basic pnpm support

Andarist:
This partially addresses https://github.com/atlassian/changesets/issues/210 

This is basic support for pnpm - it seems that pnpm supports also `package.json5` and `package.yaml` (instead of package.json) and this is not covered by this PR. 

I was looking at reusing [`@pnpm/find-workspace-packages`](https://github.com/pnpm/pnpm/tree/master/packages/find-workspace-packages) but it supports those extra package.json-like files and that wouldnt work with current `pkgJsonsMissingNameField` check included in `get-workspaces` (and I didnt want to touch this right now). At the moment it also (that [`@pnpm/find-workspace-packages`](https://github.com/pnpm/pnpm/tree/master/packages/find-workspace-packages)) does some extra validations etc and those seemed a little bit unnecessary for us here. It's always possible to switch to that package in the future when you start to duplicate more & more logic from it.


.changeset/hungry-guests-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+When in pnpm-managed project use pnpm for publishing.
```
Andarist: This is needed because pnpm does some extra stuff beyond basic npm publish (it might change content of the packed tarball a little bit - for example to include root LICENSE in a published workspace package).

Andarist:
> I'm a little bit worried about using pnpm publish because it means our rule is "we use npm to do publishing except if you're using pnpm where we'll use pnpm but we don't do that for yarn or bolt" which seems not great

Personally I don't see much of a problem with this. We only chose a publish tool that makes sense in a given workspace - as yarn and bolt don't really add anything on top of the bare `npm publish` we can just use `npm` directly. But as soon as they start doing something extra we should switch to using them. If they provide a way to publish stuff through their tools (yarn does, not sure about bolt) we could switch to that right now to be more future-proof, but not sure if it's worth it.

packages/get-workspaces/src/index.ts
```diff
@@ -5,11 +5,12 @@
 import fs from "fs-extra";
 import path from "path";
 import globby from "globby";
+import readYamlFile from "read-yaml-file";
```
Noviny: Can I recommend we don't use this library?

It's a very thin wrapper around [https://www.npmjs.com/package/js-yaml](js-yaml), except it's not using `safeLoad`

I'd like to be very careful about when / how we execute user code, particularly things that might be run on our bots, so safe-loading yaml would be good.

Can you envision that causing problems in the future?

mitchellhamilton: It is using safeLoad - https://github.com/zkochan/packages/blob/master/read-yaml-file/index.js#L8

(though I don't have strong opinions on using this or js-yaml directly)

Noviny: Oh, I misread it. In this case, no worries about this, and this seems fine

Andarist: I hadn't actually chosen this particular library myself, it was used by `pnpm` itself so I've just followed suit :p when i look at it now - it seems to be authored by by `pnpm`'s author

Andarist:
> Just a quick note on this one, yarn's publish implementation has caused us a bunch of pain at Atlassian when we used it by having really subtle internal differences to npm publish that broke several builds. It's very deliberate that we use npm publish rather than yarn publish even for users we know are using yarn.

That's completely fine for me. In this particular case though `pnpm` should be preferred as it does a bunch of stuff that is expected from `pnpm`'s users.

Noviny:
> pnpm should be preferred

Oh yeah, I'm fine with that, just explaining some context around why we were doing things the way we were.

Noviny:
The only failing is coverge, which I'm tempted to let slide, unless anyone else has strong opinions about this? Will probably update branch and merge in the morning.

PR #252, opened 2020-01-20
Fix how multiline changesets of the same type are concatenated

Andarist:
I'd like to get a preliminary review of the approach I have used here. I will add a test case & changeset sometime later.

**Problem**

Multiline changesets were trimmed and `\n` was not inserted after them, so a next changeset that got appended to the result was appended at the end of the last line which messed up markdown listing etc.

### DIFF
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -16,20 +16,18 @@ import applyReleasePlan from "./";
 class FakeReleasePlan {
   changesets: NewChangeset[];
   releases: ComprehensiveRelease[];
-  baseChangeset: NewChangeset;
-  baseRelease: ComprehensiveRelease;
   config: Config;
 
   constructor(
     changesets: NewChangeset[] = [],
     releases: ComprehensiveRelease[] = []
   ) {
-    this.baseChangeset = {
+    const baseChangeset: NewChangeset = {
       id: "quick-lions-devour",
       summary: "Hey, let's have fun with testing!",
       releases: [{ name: "pkg-a", type: "minor" }]
     };
-    this.baseRelease = {
+    const baseRelease: ComprehensiveRelease = {
       name: "pkg-a",
       type: "minor",
       oldVersion: "1.0.0",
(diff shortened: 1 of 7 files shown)


Noviny:
Approach looks good to me.

packages/apply-release-plan/src/index.test.ts
```diff
@@ -252,13 +252,58 @@ describe("apply release plan", () => {
...
+            null
+          ]
+        }
+      );
+
+      let readmePath = changedFiles.find(a => a.endsWith("pkg-a/CHANGELOG.md"));
+
+      if (!readmePath) throw new Error(`could not find an updated changelog`);
+      let readme = await fs.readFile(readmePath, "utf-8");
+      expect(readme.trim()).toEqual(
```
Andarist: `outdent` wasnt really helping here, because it has trimmed some leading whitespaces incorrectly as to our expected output

packages/apply-release-plan/src/index.test.ts
```diff
@@ -16,20 +16,18 @@ import applyReleasePlan from "./";
...
-  baseChangeset: NewChangeset;
-  baseRelease: ComprehensiveRelease;
   config: Config;
 
   constructor(
     changesets: NewChangeset[] = [],
     releases: ComprehensiveRelease[] = []
   ) {
-    this.baseChangeset = {
+    const baseChangeset: NewChangeset = {
```
Andarist: this change makes sense to me because those were not manipulated directly from the outside and while reading the code I was searching for those properties being used, so this simplifies the mental model for me - i wouldnt have to look through the file because i would immediately know that they are local variables and nothing more

packages/apply-release-plan/src/index.test.ts
```diff
@@ -146,17 +144,18 @@ describe("apply release plan", () => {
...
-        ...releasePlan.releases,
-        {
-          name: "pkg-b",
-          type: "major",
-          oldVersion: "1.0.0",
-          newVersion: "2.0.0",
-          changesets: []
-        }
-      ];
+      const releasePlan = new FakeReleasePlan(
```
Andarist: just using `FakeReleasePlan` API instead of post-construction manipulation

Andarist:
@Noviny cool, I've added changesets and a test, so it should be ready to merge :)

PR #259, opened 2020-01-24
add new binary

Noviny:


emmatown (commented):
I...don't like this. I think having multiple ways to do the same thing Is Bad and it'll mean people will ask "wait, which one should I use"

Noviny:
Sure thing.

PR #263, opened 2020-01-26
Fix a bug with prereleases where a package would not be bumped with the highest bump type of all changesets from previous prereleases within the same prerelease run

### LINKED ISSUE #244: Pre mode version incorrect
## Affected Packages

https://github.com/sevenwestmedia/project-watchtower

There is only a single package currently.

## Problem

I committed pre.json onto master

pre.json
```
{
  "mode": "pre",
  "tag": "beta",
  "initialVersions": {
    "project-watchtower": "1.6.0"
  },
  "changesets": [
    "beige-islands-fry",
    "poor-dragons-act",
    "twelve-olives-burn",
    "unlucky-flowers-wave",
    "young-jokes-wink"
  ]
}
```

package.json

```
{
    "name": "project-watchtower",
    "version": "2.0.0-beta.0",
```

With the above, the changelog for the version command generated 1.6.1-beta.1, I was expecting 2.0.0-beta.1

## Proposed solution

I haven't looked too much into this project yet, have some other focuses atm, but thought I'd report the issue for now.

mitchellhamilton:
Closes #244

### DIFF
--- a/.changeset/dirty-coins-enjoy.md
+++ b/.changeset/dirty-coins-enjoy.md
@@ -1,0 +1,6 @@
+---
+"@changesets/assemble-release-plan": patch
+"@changesets/cli": patch
+---
+
+Fix a bug with prereleases where a package would not be bumped with the highest bump type of all changesets from previous prereleases within the same prerelease run
--- a/.changeset/dirty-coins-enjoy.md
+++ b/.changeset/dirty-coins-enjoy.md
@@ -3,4 +3,4 @@
 "@changesets/cli": patch
 ---
 
-Fix a bug with prereleases where a package would not be bumped with the highest bump type of all changesets from previous prereleases within the same prerelease run
+Fixed a bug where only the unreleased pre-release changesets were taken into account when calculating the new version, not previously released changesets.
--- a/packages/assemble-release-plan/src/index.ts
+++ b/packages/assemble-release-plan/src/index.ts
@@ -43,6 +43,8 @@ function assembleReleasePlan(
 
   let workspacesByName = new Map(workspaces.map(x => [x.name, x]));
 
+  let unfilteredChangesets = changesets;
+
   if (updatedPreState !== undefined) {
     for (let workspace of workspaces) {
       if (updatedPreState.initialVersions[workspace.name] === undefined) {
@@ -62,6 +64,36 @@ function assembleReleasePlan(
   // flattened down to one release per package, having a reference back to their
   // changesets, and with a calculated new versions
   let releases = flattenReleases(changesets, workspaces);
+
+  // for every release in pre mode, we want versions to be bumped to the highest bump type
+  // across all the changesets even if the package doesn't have a changeset that releases
+  // to the highest bump type in a given release in pre mode and importantly
+  // we don't want to add any new releases, we only want to update ones that will already happen
+  // because if they're not being released, the version will already have been bumped with the highest bump type
+  if (updatedPreState !== undefined && updatedPreState.mode !== "exit") {
+    let releasesFromUnfilteredChangesets = flattenReleases(
+      unfilteredChangesets,
+      workspaces
+    );
+
+    releases.forEach((value, key) => {
+      let releaseFromUnfilteredChangesets = releasesFromUnfilteredChangesets.get(
+        key
+      );
+      if (releaseFromUnfilteredChangesets === undefined) {
+        throw new InternalError("releaseFromUnfilteredChangesets is undefined");
+      }
+      releases.set(key, {
+        ...value,
+        // note that we're only setting the type, not the changesets which could be different(the name and oldVersion would be the same so they don't matter)
+        // because the changesets on a given release refer to why a given package is being released
+        // NOT why it's being released with a given bump type
+        // (the bump type could change because of this, linked or peer dependencies)
+        type: releaseFromUnfilteredChangesets.type
+      });
+    });
+  }
+
   let preVersions = new Map();
   if (updatedPreState !== undefined) {
     for (let [name, release] of releases) {
--- a/packages/assemble-release-plan/src/index.ts
+++ b/packages/assemble-release-plan/src/index.ts
@@ -45,6 +45,7 @@ function assembleReleasePlan(
 
   let unfilteredChangesets = changesets;
 
+  let preVersions = new Map();
   if (updatedPreState !== undefined) {
     for (let workspace of workspaces) {
       if (updatedPreState.initialVersions[workspace.name] === undefined) {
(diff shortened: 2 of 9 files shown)


yarn.lock
```diff
@@ -2834,6 +2834,15 @@ get-value@^2.0.3, get-value@^2.0.6:
   resolved "https://registry.yarnpkg.com/get-value/-/get-value-2.0.6.tgz#dc15ca1c672387ca76bd37ac0a395ba2042a2c28"
   integrity sha1-3BXKHGcjh8p2vTesCjlbogQqLCg=
 
+get-workspaces@^0.5.0:
```
Andarist: seems like this shouldn't land in yarn.lock?

packages/assemble-release-plan/src/index.ts
```diff
@@ -62,6 +64,36 @@ function assembleReleasePlan(
   // flattened down to one release per package, having a reference back to their
   // changesets, and with a calculated new versions
   let releases = flattenReleases(changesets, workspaces);
+
+  // for every release in pre mode, we want versions to be bumped to the highest bump type
```
Andarist: I'm not sure if I understand the problem correctly, but from what I understand this was caused by the fact that new changesets (the one not added to `pre.json` yet) could have only lower release types than what was already released before (so what was already added to `pre.json` before).

If that's the case then I'm wondering - why this had not affected Emotion? We had to have such situation there, I've often~ released only patches when in pre mode.

Also - shouldn't really `package.json#version` be accounted for as well, rather than just changesets? It's of course not advised to handle stuff manually once you start using changesets, but this could happen and this logic here could lead to unexpected results?

.changeset/dirty-coins-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/assemble-release-plan": patch
+"@changesets/cli": patch
+---
+
+Fix a bug with prereleases where a package would not be bumped with the highest bump type of all changesets from previous prereleases within the same prerelease run
```
Andarist: This message is a little bit confusing to me, I had to really focus to understand this and I'm still unsure if I got this correctly. Especially the last part is confusing for me (in relation to what is being written before) - "within the same prerelease run" is vague and hard to tell what does it mean in this context.

yarn.lock
mitchellhamilton: It's probably because of Manypkg, it's pretty much fine.

packages/assemble-release-plan/src/index.ts
mitchellhamilton: You're understanding the problem correctly.

> If that's the case then I'm wondering - why this had not affected Emotion? We had to have such situation there, I've often~ released only patches when in pre mode.

Good point, I just looked into it a bit and it looks like it's because of linked and I have a hunch that it'll cause a bug so I'll look into it more soon.

> Also - shouldn't really `package.json#version` be accounted for as well, rather than just changesets? It's of course not advised to handle stuff manually once you start using changesets, but this could happen and this logic here could lead to unexpected results?

Yeah, that could be a possibility for confusion but what would accounting for `package.json#version` mean? For example, if you had a package with `1.0.0` and someone bumped the major to `2.0.0-next.0` manually and we have a minor changeset to apply, what should the new version be? What if they bumped it to `3.0.0-next.0` instead? How do we know when to bump and when we've already bumped to a level? I think the right answer here is what we're doing where we store the version before the prereleases and apply the highest bump type to that.

yarn.lock
Andarist: interesting - I would expect no change if `@manypkg/cli` relies on `get-workspaces` (which it does), seems like on master there is no `get-workspaces` entry in yarn.lock, which seems incorrect? unless master yarn.lock got corrupted somehow?

mitchellhamilton: I would guess that this is the first time that the yarn.lock was updated since the last release so that's why there's no get-workspaces entry in the yarn.lock in master.

packages/assemble-release-plan/src/index.ts
Andarist: yeah, I guess there is really no good answer to this because we can think of weird cases for both approaches so we really shouldn't be worried much about manual changes to versions as this just has to be strongly discouraged

one additional thing that comes to my mind - technically with the given approach it's possible to "revert" the scope of changes (by publishing a major prerelease and removing that changeset later), which seems reasonable - but also somewhat dangerous for future releases? i think it really isnt much of a real-life scenario so I wouldnt be bothered with it

mitchellhamilton: > one additional thing that comes to my mind - technically with the given approach it's possible to "revert" the scope of changes (by publishing a major prerelease and removing that changeset later), which seems reasonable - but also somewhat dangerous for future releases? i think it really isnt much of a real-life scenario so I wouldnt be bothered with it

Yeah, that's true. I agree that it's not a concern.

yarn.lock
Andarist: Even with that - yarn should always use "external" version if a dependency relies on it, shouldnt it? So this entry should be there all along, it's not safe to reuse workspace version in such situation.

It ain't much of a problem, this just has surprised me and I wonder why this has happened out loud 😅

.changeset/dirty-coins-enjoy.md
JakeGinnivan: Maybe this instead?

```
Fixed a bug in version calculation where only the unreleased pre-release changesets were taken into account when calculating the semver, not previously released changesets.
```

packages/cli/src/commands/version/version.test.ts
```diff
@@ -586,4 +586,52 @@ describe("pre", () => {
       }
     ]);
   });
+  it("should use the highest bump type for between all prereleases for every prerelease", async () => {
```
JakeGinnivan: Nice tests, really easy to read 👍

packages/assemble-release-plan/src/index.ts
JakeGinnivan: > If that's the case then I'm wondering - why this had not affected Emotion? We had to have such situation there, I've often~ released only patches when in pre mode.

I actually was wondering this myself when I hit the issue. 

> Also - shouldn't really package.json#version be accounted for as well, rather than just changesets? It's of course not advised to handle stuff manually once you start using changesets, but this could happen and this logic here could lead to unexpected results?

From an explaining how version calculation happens, not taking package.json into account is much easier to explain. `Pre-release semver is calculated using both released and unreleased changesets rather than just released changesets`

JakeGinnivan: I think this comment could be simplified.

```
When calculating pre-release versions, the highest version bump of released changesets also need to be applied to the initial pre-release version otherwise calculated version number can be lower than the last released version of the package.
```

packages/assemble-release-plan/src/index.ts
```diff
@@ -62,6 +64,36 @@ function assembleReleasePlan(
...
+  // across all the changesets even if the package doesn't have a changeset that releases
+  // to the highest bump type in a given release in pre mode and importantly
+  // we don't want to add any new releases, we only want to update ones that will already happen
+  // because if they're not being released, the version will already have been bumped with the highest bump type
+  if (updatedPreState !== undefined && updatedPreState.mode !== "exit") {
+    let releasesFromUnfilteredChangesets = flattenReleases(
```
JakeGinnivan: Does this function return released AND unreleased changesets?

If it only contains released then the opposite issue could occur when the released were only a minor bump but the major contains a major?

mitchellhamilton:
I "fixed" the "bug" that meant that this bug didn't affect Emotion. I say those things in quotation marks because the bugs were mostly covered up by the slightly confusing nature of `semver.inc` though I trust the changes in 65fec0e4960a75bc9bec0619197a1ee38a6ce1dc more than relying on how semver.inc deals with pre versions even though I haven't been able to come up with a test case for it.

packages/assemble-release-plan/src/index.ts
mitchellhamilton: It contains released and unreleased.

PR #267, opened 2020-01-28
add ">=" to get-version-range-type

fzxt:
This is to workaround https://github.com/atlassian/changesets/issues/230, which is good enough for now since we can dedupe the versions within our own lockfile.

Also related:
https://github.com/atlassian/changesets/issues/99

### DIFF
--- a/packages/get-version-range-type/src/index.ts
+++ b/packages/get-version-range-type/src/index.ts
@@ -3,6 +3,6 @@ export default function getVersionRangeType(
 ): "^" | "~" | ">=" | "" {
   if (versionRange.charAt(0) === "^") return "^";
   if (versionRange.charAt(0) === "~") return "~";
-  if (versionRange.charAt(0) === ">=") return ">=";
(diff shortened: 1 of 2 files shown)


packages/get-version-range-type/src/index.ts
```diff
@@ -1,7 +1,8 @@
 export default function getVersionRangeType(
   versionRange: string
-): "^" | "~" | "" {
+): "^" | "~" | ">=" | "" {
   if (versionRange.charAt(0) === "^") return "^";
   if (versionRange.charAt(0) === "~") return "~";
+  if (versionRange.charAt(0) === ">=") return ">=";
```
Andarist: This can’t work really because you compare a single character to two characters so this will always return false

fzxt: You're right. Should be startsWith

fzxt: Thanks for taking a look @Andarist !

PR #272, opened 2020-02-02
Delete bolt-check

emmatown:
bolt-check has been replaced with Manypkg for a while so we should probably delete it here


Noviny:
I've still been recommending bolt-check to people working in `bolt` repos who want the fix, so would be good to, if we move it out of here (probably should) we move it somewhere.

@lukebatchelor do you have any idea where we could put this?

lukebatchelor:
We'll definitely be looking at using it super soon in combination with yarn 2.0.

We're not worried about where it is, as long as we can use it. 

Could put it under boltpkg org?  Or private under atlassian-fromtend?

Noviny:
@lukebatchelor oo - can we chat about yarn 2?

Also, if we got manypkg to be modular about rules it applies, would you rather use that?

Or could this be absorbed into stricter?

Back on topic, I vote move it to the bolt org.

PR #274, opened 2020-02-08
Handle missing package.json

dword-design:
Currently a missing package.json leads to a file not found error. But for tests it is useful to also be able to run get-workspaces on a folder without a package.json. This PR returns an empty array instead.

### DIFF
--- a/packages/get-workspaces/src/index.test.ts
+++ b/packages/get-workspaces/src/index.test.ts
@@ -77,7 +77,7 @@ describe("get-workspaces", () => {
   it("should return an empty array if package.json is missing", async () => {
     let cwd = await getFixturePath(__dirname, "empty");
     const workspaces = await getWorkspaces({ cwd });
-    expect(workspaces).toEqual(null);
+    expect(workspaces).toEqual([]);
   });
   it("should return an empty array if no workspaces are found", async () => {
     let cwd = await getFixturePath(__dirname, "root-only");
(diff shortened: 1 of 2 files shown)


Noviny:
I actually like that this errors, but it also masks a different problem, which is that we should call [find workspace root](https://www.npmjs.com/package/find-workspaces-root) before trying to resolve workspaces (and throw a meaningful error if there is no workspace root)

Would you be interested in tackling this change (NB: It's a breaking change as well, so would be good to bundle with your other PR)

dword-design:
@Noviny hmmm I see that it is kinda a different case if the root doesn't exist at all, but in fact the function should still return an empty array because there are no workspaces. In my intuition a missing package.json should be handled somewhere else in the client code before invoking getWorkspaces.

Noviny:
I see it more likely that getWorkspaces will cause unexpected behaviour if it returns an array here than if it errors helpfully.

It's more likely that calling `getWorkspaces` where there is no root `package.json` is a fault, rather than it is something someone deliberately wants to do. Even if there are other places it should be handled, I can't think of a situation where you should expect to call getWorkspaces in a location without a `package.json`

Do you have an example use-case where you need to be able to call it from a place without a `package.json` where you explicitly don't want it to error? You mentioned folders without a `package.json` for testing, but in that case you could mock the function and get the same effect?

dword-design:
hey @Noviny, I just moved away from workspaces. So I don't really need the package anymore. I'd say I can fix things in the existing PRs, but I don't have the time to work on new PRs.

Noviny:
We deprecated `get-workspaces` in favour of [@manypkg/get-packages](https://www.npmjs.com/package/@manypkg/get-packages) so closing this PR.

PR #275, opened 2020-02-08
Replace null by empty array

dword-design:
Currently the getWorkspaces function returns null in some cases. This makes the API much harder to use because the null case needs to be handled separately. Also, in the tests it already mentions that empty array should be returned. This PR returns empty arrays instead of null.

### DIFF
--- a/.changeset/twenty-days-relate.md
+++ b/.changeset/twenty-days-relate.md
@@ -1,0 +1,5 @@
+---
+"get-workspaces": major
+---
+
+Replace null by empty array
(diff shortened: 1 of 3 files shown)


Noviny:
I'm happy with this change, though it would be great if you added a changeset with a `major` change to `get-workspaces`.

I might see if we can batch this with another breaking change I've misplaced the conversation of, on removing `tools` as the API has you currently provide it, because the API is inherently confusing, but if that doesn't come together this week, we should just ship this.

dword-design:
@Noviny ok done!

Noviny:
We deprecated `get-workspaces` in favour of [@manypkg/get-packages](https://www.npmjs.com/package/@manypkg/get-packages) so closing this PR.

PR #281, opened 2020-03-02
Add support for other range types

### LINKED ISSUE #99: Expand on range types changesets supports
Currently changesets supports `1.0.0`, `~1.0.0`, and `^1.0.0`, which is a subset of valid semver definitions.

@mitchellhamilton found the range types that yarn supports, and suggested we should do that.

So we need to: refind that info (I have lost the link he gave me soz)
Align our range type function with it.

tarang9211:
Closes #99

### DIFF
--- a/packages/get-version-range-type/src/index.test.ts
+++ b/packages/get-version-range-type/src/index.test.ts
@@ -1,0 +1,12 @@
+import getVersionRangeType from "./";
+
+test.each([
+  ["^1.0.0", "^"],
+  ["~1.0.0", "~"],
+  [">=1.0.0", ">="],
+  ["<=1.0.0", "<="],
+  [">1.0.0", ">"],
+  ["", ""]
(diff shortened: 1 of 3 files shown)


packages/get-version-range-type/src/index.test.ts
```diff
@@ -0,0 +1,12 @@
+import getVersionRangeType from "./";
+
+test.each([
+  ["^1.0.0", "^"],
+  ["~1.0.0", "~"],
+  [">=1.0.0", ">="],
+  ["<=1.0.0", "<="],
+  [">1.0.0", ">"],
+  ["", ""]
```
mitchellhamilton:
```suggestion
  ["1.0.0", ""]
```

tarang9211: Was just about to fix this.

Noviny:
It's super-cool that we got this in! Thanks so much!

PR #282, opened 2020-03-02
Modifying display for packages

### LINKED ISSUE #96: Add 'select all' for major  and 'minor' lists, not just initial package lists
When selecting pacakge for `major` and `minor` bumps from list, also display `all changed` and `all unchanged`, in similar way to initial package select, to make mass releases easier.

@mitchellhamilton if you don't have any args against, I'll mark this is a good first issue and send it.

tarang9211:
Fixes #96


Noviny:
Thanks for this!

PR #288, opened 2020-03-05
Do not check if a 2FA token is required when on other registries

### LINKED ISSUE #247: Not compatible with private npm registries (artifactory)
## Problem

The command `npm profile get --json` only works for npmjs.org. 
When trying to run `changeset publish` with a private registry (artifactory) the command fails trying to get profile information from the registry. 

https://docs.npmjs.com/cli/profile.html

> Change your profile information on the registry. This not be available if you’re using a non-npmjs registry.

npm profile isn't supported by artifactory. Therefore changeset in its current format doesn't support artifactory :(

## Proposed solution

Move away from using `npm profile` commands when publishing in order to support private registries like artifactory.

### LINKED ISSUE #287: Possible to publish to Github Packages?
## Affected Packages

CLI publish command

## Problem

It appears that the current publish command always assumes the public registry. If there is a way to do this with the current version of `changesets` I haven't been able to figure it out.

## Proposed solution

Update publish command to check `publishConfig.registry` before publishing to the public registry.

mitchellhamilton:
Closes #247 
Closes #287 

The approach I've taken here is "if the registry isn't the npm registry or npm profile get fails, assume we don't need 2fa". Note that if the isRequired thing returns false but then a publish fails because of 2fa, we'll prompt for a token. npm profile get is purely an optimisation to make publishes faster.


ibeyene:
https://github.com/atlassian/changesets/runs/488998801?check_suite_focus=true#step:5:20

Issue #200, opened 2019-10-23
Add way to output a 'release info' thing


Noviny:
## Problem

More and more projects want access to the `release` info for a release as a whole. The `changesets` action already does this.

## Proposed solution

Expose this. The two important things are to:

- output the md as it currently is in the action version comment
- have a command to write it to a file.

I *think* it would make sense to attach config for this to the `version` command itself to output this as a file.

Noviny:
Here is the action code:

https://github.com/changesets/action

Example of it currently in action:

https://github.com/keystonejs/keystone/pull/2246

Noviny:
This basically: https://docs.npmjs.com/adding-dist-tags-to-packages

PR #290, opened 2020-03-10
Replace get-workspaces with @manypkg/get-workspaces and a bunch of other API changes for various packages

mitchellhamilton:
There's a lot of changes(especially API changes) here - please ask questions if you're not sure about any of the decisions I've made here. The changesets describe all the changes so reviewing them first would probably be good.

---

Note: I've removed bolt-check because I didn't want to migrate it since we're not using it - if someone at Atlassian wants to maintain it, they can always go back into the git history


mitchellhamilton:
From reading the coverage changes - it seems to have mainly gone down because a bunch of code was deleted

packages/apply-release-plan/src/index.ts
```diff
@@ -143,10 +142,10 @@ export default async function applyReleasePlan(
     // So we need to be careful that these iterations are properly sequential
     while (newTouchedFilesArr.length > 0) {
       let file = newTouchedFilesArr.shift();
-      await git.add(path.relative(cwd, file!), cwd);
+      await git.add(path.relative(packages.root.dir, file!), packages.root.dir);
```
Noviny: Not blocking, but a) It's weird that we're deriving this data from the packages object, but b) since we use it a bunch, can we just declare this as `dir` at the beginning of this function, so if I'm uncomfortable enough it's easy to change later?

mitchellhamilton: > It's weird that we're deriving this data from the packages object

Why is it weird to to get it from the packages object - it's _mostly_ the same as the cwd.(I say mostly because of @manypkg/find-root but because we use cwd directly in a bunch of places in Changesets they're basically the same)

> can we just declare this as dir at the beginning of this function

sure

mitchellhamilton:
> Second, just wondering why we moved getDependentsGraph? Totally fine though.

Consistency and thought it might make sense since I'm making breaking changes to the API, don't have strong options though.

Blasz:
Hi @mitchellhamilton, I'm implementing a prototype of #200 within atlassian-frontend by trying to reuse the existing changesets packages and am interested to know more about the decisions behind the API changes. The changesets currently explain 'what' but not the 'why'.

In particular, I'm curious as to why `@manypkg`'s `Package` interface is being used everywhere when only a subset of its properties are used in most places, namely the array of packages/workspaces and the cwd/root. I'm finding it difficult to construct this interface only using `bolt.getWorkspaces` since now the project package is required solely to specify the root directory and the `tool` property is required by type but not accessed at all from what I can see.

In my opinion, the functions should have a type signature containing only the data they need to make them more flexible, decoupled from `@manypkg/get-packages` and more readable.

mitchellhamilton:
> I'm finding it difficult to construct this interface only using bolt.getWorkspaces since 

Why use `bolt.getWorkspaces` when `@manypkg/get-packages` supports Bolt?

> now the project package is required solely to specify the root directory.

`@changesets/get-dependents-graph` depends on being able to get the root package. Making the root package part of the package was really nice for this reason because it reduced what assembleReleasePlan required because previously getDependentsGraph did file system operations to read the root package so the dependents graph had to be another argument to assembleReleasePlan but because it's part of the packages now, there's one less argument to assembleReleasePlan.

>  the tool property is required by type but not accessed at all from what I can see.
> 
> In my opinion, the functions should have a type signature containing only the data they need to make them more flexible, decoupled from @manypkg/get-packages and more readable.

Requiring everything is nice because it gives us as the authors of Changesets the flexibility to rely on other fields without having to do breaking changes. e.g. let's say we wanted to use the tool property in an error message or something. If it wasn't required it would be a big deal to do a major just to add a property for some error message but by being required, it makes things like that easy.

Blasz:
> Why use bolt.getWorkspaces when @manypkg/get-packages supports Bolt?

True, that might be the easiest way forward. I wanted to avoid adding another dependency to the repo but if it's being pulled in as a transitive dependency anyway then I guess it won't hurt.

> @changesets/get-dependents-graph depends on being able to get the root package. Making the root package part of the package was really nice for this reason because it reduced what assembleReleasePlan required because previously getDependentsGraph did file system operations to read the root package so the dependents graph had to be another argument to assembleReleasePlan but because it's part of the packages now, there's one less argument to assembleReleasePlan.

Sure. I guess an alternative would be to specify the root package as an argument to assembleReleasePlan instead of grouping it inside the packages argument.

I was more referring to the other places that don't access the root package such as config & applyReleasePlan.

Looking at the dependency graph code, I wonder if we really need the root package there at all? From what I can see, there isn't a need to get the dependency graph of the root package since workspaces shouldn't list it as a dependency and it doesn't look like adding a changeset for the  root package is possible.

> Requiring everything is nice because it gives us as the authors of Changesets the flexibility to rely on other fields without having to do breaking changes. e.g. let's say we wanted to use the tool property in an error message or something. If it wasn't required it would be a big deal to do a major just to add a property for some error message but by being required, it makes things like that easy.

Fair enough, although that sounds a bit YAGNI/over-engineering to me.

I'll use @manypkg/get-packages for my use case, cheers.

PR #292, opened 2020-03-11
feat: add new @changesets/write package to separate writing changesets

acheronfail:
This PR adds a new package `@changesets/write` which does the inverse of `@changesets/parse`.

It basically just extracts the `writeChangesets` function used in `@changesets/cli` so it can be re-used.


Noviny:
Hey! Thanks for this.

We initially didn't do this because writing a changeset is simple enough that reimplementing it isn't really a cost, and having it as a library implies it's doing something fancier than it is.

What's the reason for wanting this as a package as opposed to just using the code yourself?

Noviny:
PR otherwise looks fine, so happy to merge + release

acheronfail:
> We initially didn't do this because writing a changeset is simple enough that reimplementing it isn't really a cost, and having it as a library implies it's doing something fancier than it is.

Makes sense. I guess since we already have the `@changesets/parse` we've abstracted reading the changesets, why not abstract writing them too?

> What's the reason for wanting this as a package as opposed to just using the code yourself?

We need a way to programmatically create a changeset (our use case is to integrate the renovate bot into our monorepo, and there needs to be a way to generate a changeset for the diff that the renovate bot creates).

It seemed odd that we had to duplicate the same logic that `@changesets/cli` itself was doing. Plus we wanted to have the nice things that it already does automatically, like `human-id`, integration with `prettier`, etc.

acheronfail:
Looks like the changes were merged in but didn't close this PR.

I'll close it now.

PR #296, opened 2020-03-12
Replace multi-package repo with monorepo

mitchellhamilton:
This is me testing the changeset bot(I'll actually finish doing what this PR is about when I'm done testing) - PLEASE DO NOT MERGE


Noviny:
@mitchellhamilton if we're not coming back to this, can we close it?

PR #298, opened 2020-03-15
fix(changelog-github): Throw descriptive error when no options

eps1lon:
It's somewhat easy to find what you did wrong when using `"changelog": "@changesets/changelog-github"` since the offending line is included in the error but throwing the intended error instead would help even more.

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -81,7 +81,7 @@ export type ModCompWithPackage = ComprehensiveRelease & {
 export type GetReleaseLine = (
   changeset: NewChangesetWithCommit,
   type: VersionType,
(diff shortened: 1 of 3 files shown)


packages/types/src/index.ts
```diff
@@ -81,7 +81,7 @@ export type ModCompWithPackage = ComprehensiveRelease & {
 export type GetReleaseLine = (
   changeset: NewChangesetWithCommit,
   type: VersionType,
-  changelogOpts: any
+  changelogOpts: null | Record<string, any>
```
eps1lon: Refined to catch future regressions in `getReleaseLine`. I hope this is a good compromise between loose and strict types.

PR #301, opened 2020-03-26
chore: ensure we use the diverged ref when checking changed changesets

acheronfail:
I noticed that we weren't using the result of the call to `spawn` on line 73 here: https://github.com/atlassian/changesets/blob/master/packages/git/src/index.ts#L72-L77.

So, I've abstracted the common code into a function, and added a test for it.

### DIFF
--- a/packages/git/src/index.test.ts
+++ b/packages/git/src/index.test.ts
@@ -7,15 +7,61 @@ import {
   add,
   commit,
   tag,
+  getDivergedCommit,
   getChangedPackagesSinceRef,
   getChangedChangesetFilesSinceRef
 } from "./";
+
+async function getCurrentCommit(cwd: string) {
+  const cmd = await spawn("git", ["rev-parse", "HEAD"], { cwd });
+  return cmd.stdout.toString().trim();
+}
 
 describe("git", () => {
   let cwd: string;
   beforeEach(async () => {
     cwd = await copyFixtureIntoTempDir(__dirname, "with-git");
     await spawn("git", ["init"], { cwd });
+    await spawn("git", ["config", "user.email", "x@y.z"], { cwd });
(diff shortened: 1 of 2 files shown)


packages/git/src/index.test.ts
```diff
@@ -7,15 +7,61 @@ import {
...
+  return cmd.stdout.toString().trim();
+}
+
 describe("git", () => {
   let cwd: string;
   beforeEach(async () => {
     cwd = await copyFixtureIntoTempDir(__dirname, "with-git");
     await spawn("git", ["init"], { cwd });
+    await spawn("git", ["config", "--global", "user.email", "x@y.z"], { cwd });
+    await spawn("git", ["config", "--global", "user.name", "xyz"], { cwd });
```
acheronfail: I added these calls here since some users have their `git` config setup on a per-repository basis (one of those users is me :wink:).

This will make sure that the fixtures git repository will work in repositories where users have the config option `user.useConfigOnly = true`.

packages/git/src/index.test.ts
```diff
@@ -7,15 +7,61 @@ import {
...
+  return cmd.stdout.toString().trim();
+}
+
 describe("git", () => {
   let cwd: string;
   beforeEach(async () => {
     cwd = await copyFixtureIntoTempDir(__dirname, "with-git");
     await spawn("git", ["init"], { cwd });
+    await spawn("git", ["config", "user.email", "x@y.z"], { cwd });
+    await spawn("git", ["config", "user.name", "xyz"], { cwd });
```
acheronfail: I added these calls here since some users have their git config setup on a per-repository basis (one of those users is me :wink:).

This will make sure that the fixtures git repository will work in repositories where users have the config option `user.useConfigOnly` = true.

PR #311, opened 2020-04-10
feat: support workspace ranges

### LINKED ISSUE #307: Add support for the workspace protocol
## Problem

Both pnpm and Yarn v2 support workspace ranges.
* [About workspace ranges in pnpm](https://pnpm.js.org/en/workspaces#workspace-ranges-workspace)
* [About workspace ranges in Yarn v2](https://yarnpkg.com/features/workspaces#workspace-ranges-workspace)

When I ran `changeset status` on the [pnpm monorepo](https://github.com/pnpm/pnpm/), I got lots of the following errors:

```
Package "@pnpm/tarball-fetcher" must depend on the current version of "@pnpm/error": "1.2.0" vs "workspace:1.2.0"
```

Workspace ranges are really useful in monorepos, so I think it is crucial to support them. And it shouldn't be hard to support. You could probably just strip the `workspace:` part. Only `workspace:*` is a special case. Before publish, it is replaced with the current version present in the monorepo.

zkochan:
close #307


zkochan:
I am not sure how the tests are organized in this repo and where should I cover these

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -54,12 +54,16 @@ export default function getDependencyGraph(
...
       if (!match) continue;
 
       const expected = match.packageJson.version;
 
+      if (depVersion.startsWith("workspace:")) {
+        depVersion = depVersion.substr(10);
+      }
+
       // internal dependencies only need to semver satisfy, not '==='
       if (!semver.satisfies(expected, depVersion)) {
```
Andarist, line 68: From what I recall when I've checked this briefly when you have first reported this we might need to upgrade `semver` because the current version fails on `semver.satisfies('1.1.1', '*')`

.changeset/itchy-pants-obey.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": minor
+"@changesets/get-dependents-graph": minor
+---
+
+Support workspace ranges
```
Andarist, line 6: I would cover more details here - maybe something like this:
- `@changesets/apply-release-plan`: Added support for workspace ranges. They are now correctly kept and updated when applying a release plan.
- `@changesets/get-dependents-graph`: Added support for workspace ranges. Package graph validation understands them now and allows them to satisfy dependents' required ranges.

It sometimes makes sense to split changesets into multiple files, so a more fine-grained description of a change can be specified per package. Sometimes, of course, it's an overkill though.

Andarist:
> I am not sure how the tests are organized in this repo and where should I cover these

You should most likely add fixtures [here](https://github.com/atlassian/changesets/tree/492c5fa5ff3af185ed9da808df8d64cd87d4f90d/__fixtures__) and add tests in test files like [this](https://github.com/atlassian/changesets/blob/master/packages/apply-release-plan/src/index.test.ts)

zkochan:
Thanks for the review. I will try to make the appropriate changes.

.changeset/itchy-pants-obey.md
zkochan: Thanks, TBH I don't fully understand how this works. The docs don't answer all the questions I have.

packages/get-dependents-graph/src/get-dependency-graph.ts
zkochan: I think that will require a separate pull request. As far as I see, currently, if a dependent package uses `*`, it is not released. However, `workspace:*` is actually converted to an exact version during release, so all dependent packages that use `workspace:*` should be always released.

.changeset/itchy-pants-obey.md
Andarist: Just think of a single changeset file as of a release intent+changelog entry.

The release intent (frontmatter) just tells which packages are part of this intent and what kind of a bump they should get (dependents of listed packages are often also updated, but they don't necessarily get the included changelog entry, it's more like "updated dependencies: x,y,z")

And the written down changelog entry is simply added to CHANGELOG.md of the listed packages when doing a release.

This means you can mix&match those however you want within a single PR or whatever (they are actually not tied to a git flow and can be added - changed or deleted - directly on master after a change lands). So you can list packages with different version bumps, you can create many changesets with different messages, or if the message makes sense for all listed package you can just create a single changeset.

zkochan:
@Andarist too bad. You seem to be the most active contact person.

zkochan:
I'll probably look into that tomorrow.

Noviny:
Thanks for this! A good contribution.

packages/cli/src/commands/version/version.test.ts
```diff
@@ -592,7 +626,7 @@ describe("pre", () => {
       }
     ]);
   });
-  it.only("should use the highest bump type for between all prereleases for every prerelease", async () => {
```
zkochan, line 595: I don't know why, but only this test was running in the file

packages/get-dependents-graph/src/get-dependency-graph.ts
Andarist: Hm, I'm not sure why exactly I have thought `.satisfies` was not working correctly with the used `semver` version for wildcards. I've rechecked it now and it seems to be working correctly for all `semver` versions. Looking at their changelog they had issues with `intersects` and wildcards, also some regarding prereleases, that have been fixed by now. I've probably mistaken this with those issues.

Either way - I've included an integration test for this here:
https://github.com/atlassian/changesets/pull/447/files#diff-068ddbdcddd68c5cb9e2b7efda8bbc2bR348 , so we should be guarded against any potential regressions regarding this.

PR #313, opened 2020-04-13
fix: don't bump dependent if the released package is a dev dep

### LINKED ISSUE #312: Dependents are patch bumped even if the published package is a dev dependency only
## Problem

Dependent packages are patch bumped, which is good. However, dependent packages are patch bumped even if the released package is only a dev dependency.

If `foo` is in a changeset and `foo` is a dev dependency of `bar`, then `bar` should not be published because it has no changes that affect production. 

## Proposed solution

Only those dependent packages should be bumped that have foo as a non-dev dependency.

zkochan:
close #312


Noviny:
^ when two maintainers talk to each other then both go to make a comment because they didn't talk together enough 😅

mitchellhamilton:
> one way we could do this is to use the dreaded none bump type we removed, allowing us to add a release for pkg-b with a bump type of none that caused its dependency to get bumped

I was thinking that we could find the devDeps that are out of range in apply release plan and bump them there? idk how much point there is to have releases for them?

Noviny:
> find the devDeps that are out of range in apply-release-plan

This also works, it would likely just need some comments on why sometimes you change package.jsons when you aren't releasing the package, but happy to put the logic there. Easy either way.

zkochan:
The release type none seems easier to implement. Apply-release-plan doesn't have information about dependents.

zkochan:
It probably doesn't make sense to return those packages through the `releases` array, which is returned by `assemble-release-plan`. I could probably return a new array - `devDependents`.

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -95,6 +95,24 @@ describe("assemble-release-plan", () => {
     expect(releases[1].newVersion).toEqual("1.0.1");
     expect(releases[1].changesets).toEqual([]);
   });
+  it("should assemble release plan without dependent through dev dependency", () => {
+    setup.updateDevDependency("pkg-b", "pkg-a", "^1.0.0");
```
Andarist, line 99: IMHO it would be slightly easier to read/understandn this test if `pkg-a` would be the "base" package and `pkg-b` would be the one we are updating

zkochan: Unfortunately, the setup class has a changeset baked in for pkg-a already

https://github.com/atlassian/changesets/blob/c3cc23204c6cb80487aced1b37ebe8ffde0e2111/packages/assemble-release-plan/src/test-utils.ts#L61

pkg-a has a patch changeset. So I need to either rewrite that testing util or use the package that has no changesets as the dependent (which is pkg-b)

zkochan:
@Noviny I applied the solution you proposed with "none"

.changeset/moody-flies-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": patch
```
Noviny, line 2:
```suggestion
"@changesets/assemble-release-plan": major
```

.changeset/tough-lies-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
```
Noviny, line 2:
```suggestion
"@changesets/apply-release-plan": major
```

Blasz:
It doesn't look like the CLI has been major bumped as a result of this change.

PR #315, opened 2020-04-14
feat: Prettier format should be given a filepath

### LINKED ISSUE #314: Give prettier.format the filepath when formatting the package.json
## Problem

Its great to see the package.json write already runs prettier ✅. Now sometimes, the prettier config has file overrides - for example, tabWidth: 4, but with package.json tabWidth: 2 to solve that whole npm thing... 

This is a file formatted by prettier, but incorrectly formatted by the GH action: https://github.com/autoguru-au/overdrive/pull/356/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

## Proposed solution

Would be great if the `filepath` was also applied to the `.format` method:

https://github.com/atlassian/changesets/blob/492c5fa5ff3af185ed9da808df8d64cd87d4f90d/packages/apply-release-plan/src/index.ts#L105-L109

Prettier has a `filepath` option: https://prettier.io/docs/en/options.html#file-path that way formatters know its a `./package.json` file

maraisr:
Fixes #314 

Perhaps @mitchellhamilton you can help me write a unit test for this? I thought maybe not, as that is testing that prettier is doing its job.


Andarist:
You could still write an integration test for this - although it might also be overkill. [This PR](https://github.com/atlassian/changesets/pull/61/files) did not introduce any tests and I can't see any explicit tests about formatting, so I would say that a test for you change is not required.

maraisr:
I will have a go at this @Andarist - at the very least to distill some more confidence in this change

Noviny:
Happy to merge this. If you want to PR tests later that would be ❤️

maraisr:
Thanks @Noviny would be a huge help for me now! Definitely still keen to attempt the tests - so yeah, will raise a separate PR for that, increases my contribution amount too ;)

PR #318, opened 2020-04-15
Noviny/before releasing

Noviny:


Noviny:
No changeset for applyReleasePlan as it's fixes to things that haven't been released

PR #326, opened 2020-04-16
Noviny/add generate release notes

Noviny:


Noviny:
Okay, this is WIP but I made it wrong.

Goal is when this merges, 'releases' as a concept become exposed to a user so they can work on them.

packages/cli/src/index.ts
```diff
@@ -22,7 +22,7 @@ const { input, flags } = meow(
     $ changesets [command]
   Commands
     init
-    add [--empty]
+    add [--empty] [--global]
```
zkochan, line 25: It would be nice to have more info in the help output.

It is unclear what the commands do. It even a bigger mystery what the flags change.

Noviny: These need more detail and we also need better docs. Docs/help messaging are the thing I'm most interested in getting done after this global release feature gets out.

packages/read/src/index.ts
```diff
@@ -56,8 +60,63 @@ export default async function getChangesets(
...
+    console.error(
+      `We recommend looking at "aaa-global-changeset.md" or searching your changesets for a release of ""@changesets/secret-global-release""`
+    );
+    throw new Error("Multiple global changesets");
+  }
+
+  // @ts-ignore
+  return globalchangesets[0]
+    ? [globalchangesets[0], ...nonGlobalChangesets]
+    : nonGlobalChangesets;
```
Blasz, line 93: Why the `@ts-ignore`? Imo they should only be used as an absolute last resort, even casting to any is better since ts-ignore masks other errors.

packages/read/src/index.ts
```diff
@@ -56,8 +60,63 @@ export default async function getChangesets(
...
+  let [possibleGlobalChangeset, ...otherChangesets] = changesets;
+
+  if (!possibleGlobalChangeset) {
+    return {
+      changesets: [],
+      globalChangeset: undefined
+    };
+  }
+
+  if (possibleGlobalChangeset.id === "aaa-global-changeset") {
```
Blasz, line 111: Extracting this check into a type guard would mean you don't have to typecast and make the code DRYer since it's used twice in this file.

packages/read/src/legacy.ts
```diff
@@ -45,7 +45,7 @@ async function getOldChangesets(
 export default async function getOldChangesetsAndWarn(
   changesetBase: string,
   dirs: string[]
-): Promise<Array<NewChangeset>> {
+): Promise<Array<GlobalChangeset | NewChangeset>> {
```
Blasz, line 48: Is this correct? GlobalChangesets aren't handled here, I think if we fix the `MixedChangesets` type this won't need to be touched.

packages/types/src/index.ts
```diff
@@ -73,6 +97,11 @@ export type WrittenConfig = {
 
 export type NewChangesetWithCommit = NewChangeset & { commit?: string };
 
+export type MixedChangesets =
+  | [GlobalChangeset | NewChangeset, ...NewChangeset[]]
+  | [];
```
Blasz, line 102: I think the type should be `type MixedChangesets = [GlobalChangeset, ...NewChangeset[]] | NewChangeset[]`. This will probably fix issues in some other places as well such as the old changeset function.

I tested it out in the typescript playground [here](https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBA4gNgewEYEM4GEAWKB2BzCAZwmCgF4oBvKFALikOACcBLfKAXwG4BYAKFCQoAOQgB3LLgLFSFaknqNW7bv0HQAsiwAeEACaT8REoXL8oUAD5QA2vGRpD0kgBooAOk+iJ2IzJsAugHmViLiTsbAgbx8-ADGCDiMNPRauga+zsCmFDbUdFAA5ChIcYWcbvL0hXoQAGblHJVQCkV17Y0BMfGJyXGpOvoRMjm2lK01eoUcwbF8CUmkrWlDmZGjeS3VtQ0VVFttHZxd-PxAA).

Blasz (commented):
Hey, I've left a review. Comments are mostly typescript related, API and readability/maintainability.

packages/apply-release-plan/src/index.ts
```diff
@@ -89,18 +82,26 @@ export default async function applyReleasePlan(
...
+    let packageJson = updatePackageJson(
+      release,
+      versionsToUpdate,
+      pkg.packageJson
+    );
+
+    let pkgJSONPath = path.resolve(pkg.dir, "package.json");
 
-    let changelogPath = path.resolve(dir, "CHANGELOG.md");
+    let changelogPath = path.resolve(pkg.dir, "CHANGELOG.md");
```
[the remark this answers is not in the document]
Noviny: Good catch

PR #328, opened 2020-04-17
fix: pnpm detection in projects that use a shared lockfile

zkochan:


packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -20,15 +19,9 @@ function getCorrectRegistry() {
...
-    return "pnpm";
-  } catch (err) {
-    if (err.code !== "ENOENT") {
-      throw err;
-    }
-    return "npm";
-  }
+  const pm = await preferredPM(cwd);
+  if (!pm) return "npm";
+  return pm.name === "pnpm" ? "pnpm" : "npm";
```
zkochan, line 24: `preferred-pm` also detects Yarn properly. But this function was only returning npm or pnpm, so probably should stay that way.

Issue #310, opened 2020-04-10
Allow Running Custom Commands Per "Lifecycle"


jbolda:
## Affected Packages

Extent not quite known.

## Problem

We are hoping to use changesets in a polyglot monorepo. We have multiple packages with multiple languages (Rust and js at the current moment.) Obviously we are unable to version, generate changelogs and publish Rust.

## Proposed solution

I believe we could abstract out the commands that are run at the various "lifecycles" and allow them to be configured. We would set the default commands to be the existing functionality for backwards compatibility. This would solve our dilemma as we could run the Rust commands instead which mirror the current functionality. This would also allow other cases such as #171.

This would also allow changesets to expand outside of js-land. With GitHub Actions and the bot, a pure Rust package could use changesets through the automation and never need to install node. A simple script to create the markdown file could be created, or it can be done manually (or strictly through the bot).

If there is interest for this, I would be willing to tackle a PR for it.

Noviny:
This would be awesome and thanks for the interest! It's worth noting that most of the functional bits are abstracted already. The core concepts beyond are:

- write `changeset` files that contain versioning info
- generate a `release plan` out of all changesets currently on disc
- apply a release plan to a repo
- publish (go to each package and run the publish command in it)

I'm not familiar with rust, so not sure what changes you need to make to those lifecycles (beyond targeting packages that don't have a `package.json` for things).

Happy to be a sounding board for this as you experiment.

jbolda:
Great! I think by far the most work will be finding the packages. I would love some thoughts on making that flexible. I think the easy answer is a plugin system, but that seems to be a pretty hefty undertaking.

The versioning and publishing can take a command easy enough (Rust has a `cargo version` crate and standard publishing for example). I imagine the changelog generation is already pretty standard.

Looking forward to making this happen!

Noviny:
> The versioning and publishing can take a command easy enough

We already detect pnpm to run `pnpm publish` over `npm publish` so I don't think this will be too bad.

For getting rust packages, I'd look at [@manypkg/get-packages](https://github.com/Thinkmill/manypkg/tree/master/packages/get-packages) for inspiration. I'd be curious about what extending that to get rust packages looks like, rather than going full plugin just yet.

jbolda:
[...]

PR #333, opened 2020-04-18
Allow Running Custom Commands Per "Lifecycle"

jbolda:
This PR is beginning to implement changes from #310. The tests are currently broken on windows which means this is probably more psuedo-code at the moment until upstream changes fix it. I am opening a draft PR that hopefully the tests will run in CI and give me a little bit of feedback for the time being.

### DIFF
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -18,6 +18,7 @@ import prettier from "prettier";
 import versionPackage from "./version-package";
 import createVersionCommit from "./createVersionCommit";
 import getChangelogEntry from "./get-changelog-entry";
+import spawn from "spawndamnit";
 
 async function getCommitThatAddsChangeset(changesetId: string, cwd: string) {
   let commit = await git.getCommitThatAddsFile(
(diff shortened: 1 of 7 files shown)


.changeset/metal-peaches-reply.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": patch
```
ryaninvents, line 2: @jbolda Since this is a new feature, wouldn't that make it a `minor` change? Or am I missing something?

```suggestion
"@changesets/apply-release-plan": minor
```

packages/apply-release-plan/src/index.ts
```diff
@@ -89,28 +90,38 @@ export default async function applyReleasePlan(
...
-    return versionPackage(release, versionsToUpdate);
+    if (!!config.packageLifecycleCommands && !!config.packageLifecycleCommands[release.name]) {
+      customCommands.push(spawn(`${config.packageLifecycleCommands[release.name].version} ${release.type}`, { cwd: release.dir }))
+      return { ...release, packageJson: null }
+    } else {
+      return versionPackage(release, versionsToUpdate);
+    }
   });
 
+  await Promise.all(customCommands)
```
Andarist, line 104: those are not necessarily custom, the name can be misleading here, maybe just use `commands`/`versionCommands`/`finalCommands`?

.changeset/metal-peaches-reply.md
jbolda: I can certainly see it. I figured it only _needed_ patch, but it is definitely a new feature. I am fine with whatever maintains prefer.

PR #334, opened 2020-04-20
[WIP] Expose getReleasesWithChangelogs function from apply-release-plan

Blasz:
This method attaches the generated markdown changelog entries to the
release plan. This allows consumers to access the changelog entries
programatically without having to apply the release plan and write
changes to disk.

### DIFF
--- a/packages/apply-release-plan/dist/apply-release-plan.cjs.d.ts
+++ b/packages/apply-release-plan/dist/apply-release-plan.cjs.d.ts
@@ -1,0 +1,2 @@
+export * from "./declarations/src/index";
+export { default } from "./declarations/src/index";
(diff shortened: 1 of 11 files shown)


packages/apply-release-plan/dist/apply-release-plan.cjs.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./declarations/src/index";
```
Andarist, line 1: dist directories [are being gitignored](https://github.com/atlassian/changesets/blob/aa840db824c321159e3b1c66ea663b4036084bd7/.gitignore#L4), you had to somehow forcefully add them to your commits, could you remove those?

Blasz: I did it so I could install the package from this github branch but I ended up having to go through some extra trickery to get this working for a monorepo.

I don't plan on merging this PR and will decline it, as it is superseded by https://github.com/atlassian/changesets/pull/326 which exposes the functions I want to access as part of a new package.

Sorry for the confusion

PR #335, opened 2020-04-21
Use fixturez instead of jest-fixtures

zkochan:


zkochan:
@mitchellhamilton I substituted jest-fixtures with fixturez

PR #343, opened 2020-04-29
fix: release shouldn't fail on a project that references itself

zkochan:


zkochan:
yes, ignoring `file:` and `link:`  would work as well. I will update this.

Andarist:
Isnt the changeset and test description somewhat wrong now after the change has been made to ignore stuff based on file and link protocols?

zkochan:
It is not wrong, it just describes a subset of what was fixed.

Andarist:
You could possibly have "self-reference" but pointing to a version or something else, right? This is often done in tools dog-fooding their own tools (for example transpiling Babel's source with Babel). OTOH this (depending on a version from registry) probably is not supported by changesets anyway 🤔

PR #344, opened 2020-04-30
fix: bump dependent when both a dev dep and a prod dep are released

zkochan:
When both a dev dep and a prod dep of a dependent package are published, the version of the dependent package should be bumped. This fixes a regression introduced by #313.


packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -67,7 +67,8 @@ export default function getDependents(
         } else {
           if (
             // TODO validate this - I don't think it's right anymore
```
zkochan, line 69: This comment might need to be removed.

PR #355, opened 2020-05-07
fix: handle errors when checking divergent branches

acheronfail:
I just ran into the issue where my CI environment was performing a
shallow clone, and so it didn't have all the refs in its local copy.

When I used the `getReleasePlan(cwd, MY_CUSTOM_REF)` function, it
silently failed and returned an empty release plan.
It took me a while to figure out what the problem was!

This change adds error handling when we call out to `git`, so if the ref
itself doesn't actually exist then the function throws.

### DIFF
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -32,9 +32,14 @@ async function tag(tagStr: string, cwd: string) {
   return gitCmd.code === 0;
 }
 
+// Find the commit where we diverged from `ref` at using `git merge-base`
 export async function getDivergedCommit(cwd: string, ref: string) {
-  // First we need to find the commit where we diverged from `ref` at using `git merge-base`
   const cmd = await spawn("git", ["merge-base", ref, "HEAD"], { cwd });
+  if (cmd.code !== 0) {
+    throw new Error(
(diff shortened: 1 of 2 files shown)


packages/git/src/index.ts
```diff
@@ -59,6 +64,12 @@ async function getChangedFilesSince({
   const divergedAt = await getDivergedCommit(cwd, ref);
   // Now we can find which files we added
   const cmd = await spawn("git", ["diff", "--name-only", divergedAt], { cwd });
+  if (cmd.code !== 0) {
```
Andarist, line 67: when this could happen if you throw an error already in `getDivergedCommit`?

acheronfail: I agree if we've already confirmed that the given ref exists, then the likelihood that this command fails is low.

But there are some `git` configurations which can still make `git diff` return non-zero errors when it's run, if for example, they have whitespace checking enabled (https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---check) then `git diff` could return with a non-zero exit code.

I guess it's still good practice to handle the case when shelling out to `git` may unexpectedly fail, rather than silently allowing the error. :)

Andarist: U are probably right. Its a scary async world out there and better be safe than sorry 😅

Issue #189, opened 2019-10-11
Version fragmentation problem


Noviny:
Wanted to create a clear place to discuss this. Going to try and outline what the problem *is* in this description, then going to come back in comments to outline different ways of approaching it, and my idea of how changesets should or shouldn't address this problem.

> This problem exists in any instance where you release multiple interdependent packages, and scales with the interdependency of the packages.

## The problem

In the simplest form, the problem is that in an interdependent system, you may end up pulling in multiple copies of the same package.

Our first hypothetical:

I am publishing a button component and a modal component. The modal depends on button.

In essence, I get:

```
button@1.0.0
modal@1.00
    button@^1.0.0
```

If a user installs both of these packages, modern package managers will 'dedupe' the versions of button, meaning the user has one version.

If however we have:

```
button@1.1.0
modal@1.00
    button@^1.0.0
```

If the above user upgrades button, but leaves modal alone, modern package managers will keep the button modal installed 'locked' at `1.0.0` and pull in a new version of button for `button` itself.

The user now has two copies of the button code that may be nearly identical, and that the semver-ing of modal's dependency on button suggests that there should be no problem with modal using the updated button.

There are two ways this spread occurs:

- installing packages released at different times so dependencies cannot be resolved to a single version
- installing packages that get 'locked' to resolving transitive dependencies to different versions despite them being semver compatible to dedupe.

## How big of a problem is it?

This problem scales very badly with system interconnectivity, system size and release cadence.

Basically the more versions of more packages you have, which have more interdependencies, the worse it will get.

Some monorepo-based projects won't even notice this, or will be completely unaffected. Others may find it a massive drain on their ability to be installed and maintained.

## Notes

The problem you are trying to solve for is an interrelation between consumer install habits and publisher publish habits.

Most solutions will have trade-offs.

Noviny:
(I need to head home so I'm going to come back and fill this out in a bit)

## Ways to lessen the problem

### Fewer packages (less interdependency through combining single packages together)

### Fewer versions of each package (more changes per version)

###

Noviny:
(I am considering making this a PR just so I can start making threads of comments off an md doc)

Noviny:
@lukebatchelor you mentioned wanting a larger talk about version fragmentation so I made a thing.

@mitchellhamilton you probably care.

bluwy:
I believe https://changesets.dev/guide/config#updateinternaldependencies handles this problem today.

Issue #117, opened 2019-07-15
Allow people to specify dependents to bump in some way


Noviny:
Problem:

I am releasing package A and want to release package B with an updated dependency on package A.

Solution: Make a changeset bumping A, make a changeset bumping B

THIS WORKS because when we release, we also bump dependency versions of everything else that is... in the release.

This meant that if you did:

```
changeset a@patch => release => changeset b@patch => release
```

you won't update the version b depends upon.

Further, the original approach

```
changeset a@patch => changeset b@patch => release
```

produces different results.

I don't have opinions yet on how to resolve this, but the fact that to cleanly update the dependency of b on a requires knowing how changesets will be applied is bad. We need to fix this.

Problems:

- There is an implicit behaviour that I rely on to get the bumps I want
- There is an implicit behaviour in how dependents are updated that makes changesets less predictable.

mitchellhamilton:
I feel like the answer is when a package is being released, any in-monorepo dependencies that are not at the latest semver range should be bumped even if they are within semver range.

Or, we could always bump things even when they're in semver range which would make prereleases have one fewer problems though I assume that would be bad for atlaskit because it would be soooooo many releases.

One idea I've been thinking about is an `upgrade` command which bumps a dep or deps(internal or external) and adds the appropriate changeset(s). The main use case that I was thinking about was so you could bump external deps and get the same changelog entries as when internal deps are bumped.

Noviny:
> I assume that would be bad for atlaskit because it would be soooooo many releases.

Pretty much this. You end up with any publish being a publish of everything. You could make a changeset cli that did it but it would be unusable in AK.

> I feel like the answer is when a package is being released, any in-monorepo dependencies that are not at the latest semver range should be bumped even if they are within semver range.

I *think* this might be the right answer, but some further thoughts below.

The biggest problem child of this is while you end up with a bunch of things that are 'technically semver compatible', lockfiles mean that if I install a package that depends on `button@^1.0.0` then later install a package that depends on `button@^1.1.0`. Lockfiles mean I now have two versions of button.

But if keeping lockfiles happy is our actual goal, we probably want to *never* update dependencies unless we *have* to or are *asked* to (such as by an upgrade command which makes sense to me).

BUT then you make more bugs, because it's more likely someone's code doesn't work because in-repo you've accidentally/implicitly relied on a package above a version while someone else is locked below that, and resolving yarn locks is 🙃

Noviny:
[...]

PR #358, opened 2020-05-08
Make interlink dependency auto bump logic configurable

Blasz:
Relates to #117 and #189.

We'd like the ability to opt-out of the auto dependency bumping that occurs to min version ranges when two or more packages are released together, specifically in the case where the dependency has only been patch bumped.

This PR includes one way of achieving this via a `interLinkAutoBump` config option (I couldn't think of a good name, ideas welcome) which specifies the minimum bump type that the auto bumping of inter linked dependencies will apply to. By default it's set to 'patch' which should not change current behaviour of changesets.
I was originally going to add 'major' as an option as well but then decided against it since there isn't a way to specify bumping dependents (#117) and this configuration option doesn't change that.

It's unfortunate that adding a new feature behind a configuration option has resulted in two major releases. In my opinion, the package structure, function signaturres and/or versioning may need a rethink to avoid this going forward.


Noviny:
Some quick thoughts:

I'd add a changeset for the CLI so this becomes available, and tests failed, but otherwise happy with this.

This does up the need for #117
---

Re: two major versions. This makes sense for those packages, and I don't really mind doing it.

packages/types/src/index.ts
```diff
@@ -61,6 +61,8 @@ export type Config = {
   linked: Linked;
   access: AccessType;
   baseBranch: string;
+  /** Whether to auto bump inter-linked dependencies part of the same release on patch bumps and above (default) or minor bumps and above. */
+  interLinkAutoBump: "patch" | "minor";
```
Noviny, line 65: `updateInternalDependencies` is more clear to me, but I'm not super convinced.

Andarist:
I know I'm a little bit late to the party - but shouldn't this be named `updateInternalDependents`? Or have I misunderstood the purpose of this? I understand though that this has been very hard to name - even with the proposed change I'm not sure if I would know what it does by just looking at the name.

PR #359, opened 2020-05-10
Support for snapshot versioning and publish under custom tag

ajaymathur:
Resolves issue: https://github.com/atlassian/changesets/issues/327

Changes:

### 1. `version` command now supports _snapshot_ flag.

Rule:

- It will not commit the version change.

Usage:

```
$ # Creates a version as 0.0.0-[tag]-<hash>
$ # hash -> date-time combination -> YYYYMMDDHHMMSS
$ changeset version --snapshot [tag]
```

### 2. `publish` command now supports _tag_ flag

Rules:

- Throws error if project is in pre mode.

Usage:

```
$ # Publishes the packages as per changeset(s) to the provided tag.
$ changeset publish --tag <tag>
```


packages/cli/src/commands/version/index.ts
```diff
@@ -18,7 +18,11 @@ let importantEnd = chalk.red(
   "----------------------------------------------------------------------"
 );
 
-export default async function version(cwd: string, config: Config) {
+export default async function version(
+  cwd: string,
+  config: Config,
+  snapshotReleaseConfig?: SnapshotReleaseConfig
+) {
```
zkochan, line 25: is a separate config object really needed? why can't the already present config be extended?

packages/cli/src/index.ts
```diff
@@ -156,17 +156,26 @@ const cwd = process.cwd();
...
           throw new ExitError(1);
         }
         let tag = input[2];
-        if (command === "enter" && typeof tag !== "string") {
-          error("A tag must be passed when using prerelese enter");
+        if (
+          (command === "snapshot" || command === "enter") &&
+          typeof tag !== "string"
+        ) {
+          error(`A tag must be passed when using prerelese ${command}`);
```
zkochan, line 174: Why is the tag required when making a snapshot publish?

zkochan:
I don't understand what `pre` means. Why cannot it be called `changeset snapshot --tag <tag>`? I guess it would make sense to even include `publish` in there, to make it clear that a publish will happen: `changeset publish-snapshot --tag <tag>`

packages/cli/src/commands/pre/index.ts
```diff
@@ -45,5 +54,46 @@ export default async function pre(
...
+    const snapshotReleaseConfig = {
+      tag: options.tag,
+      commitHash: lastCommitHash
+    };
+
+    await version(cwd, updatedConfig, snapshotReleaseConfig);
+
+    logger.log("All the files have been updated with snapshot release.");
+
+    await publish(cwd, { otp: options.otp, tag: options.tag }, updatedConfig);
```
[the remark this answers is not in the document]
emmenko: Could be a separate improvement maybe

packages/assemble-release-plan/src/index.ts
```diff
@@ -171,7 +179,10 @@ function assembleReleasePlan(
     releases: [...releases.values()].map(incompleteRelease => {
       return {
         ...incompleteRelease,
-        newVersion: incrementVersion(incompleteRelease, preInfo)!
+        newVersion:
+          snapshotConfig === undefined
+            ? incrementVersion(incompleteRelease, preInfo)!
+            : `0.0.0-${snapshotConfig.tag}-${snapshotConfig.commitHash}`
```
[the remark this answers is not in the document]
mitchellhamilton: Using the current version or the future version(as in the current version but with bumps applied) can cause unexpected behavior when combined with other pre-releases(e.g. you have a regular pre-release at `1.0.0-beta.0` and then you had a snapshot pre-release at `1.0.0-canary-git-hash` and a consumer is using the range `^1.0.0-beta`, most people would expect that range to resolve to `1.0.0-beta.0` but it'll actually resolve to `1.0.0-canary-hash`). Using `0.0.0` solves this problem because it won't conflict with other versions.

packages/cli/src/commands/version/index.ts
mitchellhamilton: The Config is completely different from what this is so it wouldn't make sense to extend it.

packages/cli/src/index.ts
mitchellhamilton: We need a tag to know what npm dist-tag to publish to

mitchellhamilton:
`pre` means pre-release. The reason it's called `changeset pre snapshot` is that it's a snapshot **pre-release**.

Andarist:
It seems to me as well that including "publish" somewhere in the command would make it easier to understand for newcomers, even at the cost of some added verbosity.

mitchellhamilton:
Okay, let's make it `changeset publish-snapshot <tag>`

packages/cli/src/index.ts
zkochan: oh, indeed. There is no way to publish without a tag because that way it would be "latest".

In that case, makes sense. Maybe some default tag could be used. Like the branch name, or "snapshot", or "dev". But that's just a good to have.

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
zkochan: Thank you for the detailed explanation! I always forget that this is how canary versions are resolved. Might be good to add this as a comment in the code or/and an explanation in the docs about snapshot publish.

ajaymathur:
Thanks for the feedback everyone. :) 

Tasks that I am completing now:

- [ ] Update the command to `changeset publish-snapshot <tag>`. https://github.com/atlassian/changesets/pull/359#issuecomment-627022902
- [ ] Add a flag `no-publish` passing which will only make the changes in the file and skip publishing. https://github.com/atlassian/changesets/pull/359#discussion_r423251801
- [ ] Simplify the ternary statement/ code where necessary. https://github.com/atlassian/changesets/pull/359#discussion_r423252723

I am going ahead with working on change but please feel free to discuss if you think something should be added or updated.

Andarist:
Currently, regular versioning and publishing scripts are separate - IMHO it would make sense to stick to that here as well without introducing `no-publish` flag or similar because it deviates from the standard usage.

My alternative proposal would be to add `--snapshot` flag to both version and publish commands. You need to rethink if it makes sense to call `publish --snapshot tag` without previously calling version at all, but seems to me that it shouldn't create much of a problem and maybe would be useful on its own for somebody.

Q: why this has been bound to pre mode? Wouldn't it make sense to allow this outside of it? One could setup snapshot releases on each pr/commit without ever entering the pre mode - this use case seems useful.

ajaymathur:
A: Snapshot release does not have to bound to pre release, it is just the significance that pre defines a pre-release. User does not have to be in pre mode snapshot release.

---

Yes, I like the idea of `changeset publish --snapshot tag`, it read clear to me of what it does but then users will not able to use their own publish script. like mentioned in https://github.com/atlassian/changesets/pull/359#discussion_r423251801, any opinions on this?

---

Continuing on suggestion of keeping the version and publish separate I see this as:

```
$ # Gives a version as 0.0.0-<hash>, but does not commit the files no matter the config
$ changeset version --snapshot
$ # Publish command supports the tag option, this overrides config for publish - default as latest
$ changeset publish --tag <tag>
```

Opinions?

emmenko:
It makes to me, thanks.

To summarize, let me try to apply this to some examples:

**Publishing to `latest` (default)**

```
changeset version
changeset publish
```

**Publishing to `next` dist-tag instead of `latest`**

```
changeset version
changeset publish --tag next
```

**Publishing canary releases**

```
changeset version --snapshot
changeset publish --tag canary
```

**Publishing in "pre" mode**

```
changeset pre enter next
changeset version
changeset publish
```

---

Did I get this right?

Andarist:
> Yes, I like the idea of changeset publish --snapshot tag, it read clear to me of what it does but then users will not able to use their own publish script. like mentioned in #359 (comment), any opinions on this?

Isn't this problem solved by splitting this into 2 commands? @zkochan could use versioning command to adjust files and just use a custom publish command, or have I missed something?

ajaymathur:
Hey @Andarist, yes you are right. I misunderstood the second line. Sorry.

I see using snapshot flag with version and using tag option with publish looks clean. As in https://github.com/atlassian/changesets/pull/359#issuecomment-627250875

How can we decide on it?

Andarist:
Probably a good idea to wait for @mitchellhamilton’s or @Noviny’s green light

mitchellhamilton:
> ```
> $ # Gives a version as 0.0.0-<hash>, but does not commit the files no matter the config
> $ changeset version --snapshot
> $ # Publish command supports the tag option, this overrides config for publish - default as latest
> $ changeset publish --tag <tag>
> ```


This sounds good to me. One little thing: changeset publish should throw if `--tag` is passed when in pre mode (because the tag changes in a nuanced way when in pre mode). Also `changeset version --snapshot some-tag` should make the version `0.0.0-some-tag-hash`

---

Unrelated to the above - instead of using the git hash, let's use a random hash so that it's not tied to git + you should be able to do multiple releases without the last commit hash changing

zkochan:
Another good alternative to using hash would be to use date/time. I have noticed this in the typescript canary versions but they use only date.

They have versions like this: 4.0.0-dev.20200512

packages/cli/src/commands/version/index.ts
ajaymathur: Resolving this since we don't use snapshotConfig anymore.

emmenko:
FYI: React uses git SHA as well for canary/experimental releases

```
`0.0.0-experimental-${commit}`
```

https://github.com/facebook/react/blob/4469700bb60ffd62caf112e47bcfa51afdd40e47/scripts/release/utils.js#L196

https://github.com/facebook/react/blob/4469700bb60ffd62caf112e47bcfa51afdd40e47/scripts/release/utils.js#L86-L88

ajaymathur:
Hey 👋, PR is ready for review. 🙂

packages/cli/src/commands/version/index.ts
```diff
@@ -29,7 +35,11 @@ export default async function version(cwd: string, config: Config) {
   const changesets = _changesets as NonNullable<typeof _changesets>;
   const preState = _preState as NonNullable<typeof _preState>;
 
-  if (preState !== undefined && preState.mode === "pre") {
+  if (
+    options.snapshot === undefined &&
+    preState !== undefined &&
+    preState.mode === "pre"
+  ) {
```
mitchellhamilton, line 42: Could we throw here if snapshot is passed and we're in pre mode?(to give the user feedback before they're at the publish stage)

.changeset/dry-rivers-train.md
```diff
@@ -0,0 +1,8 @@
+---
+"@changesets/apply-release-plan": minor
+"@changesets/assemble-release-plan": minor
+"@changesets/cli": minor
+---
+
+- Add support for snapshot flag to version command. Usage: `changeset version --snapshot [tag]`
```
Andarist, line 7: Could you describe the motivation and the use case for this?

The second thing (`publish --tag`) is not tightly coupled to the `--snapshot` so it could be moved to a separate changeset.

packages/cli/src/commands/version/index.ts
Andarist: why pre mode should clash with snapshot releases? seems like both could be used at the same time with success

packages/apply-release-plan/src/index.ts
```diff
@@ -39,7 +39,8 @@ async function getCommitThatAddsChangeset(changesetId: string, cwd: string) {
 export default async function applyReleasePlan(
   releasePlan: ReleasePlan,
   packages: Packages,
-  config: Config = defaultConfig
+  config: Config = defaultConfig,
+  snapshot?: string | boolean
```
zkochan, line 43: Why `snapshot` is not part of `config`?

packages/cli/src/commands/version/index.ts
zkochan: seems unusual to have both a config and options. Why can't the settings from config converted to options?

.changeset/dry-rivers-train.md
```diff
@@ -4,5 +4,12 @@
 "@changesets/cli": minor
 ---
 
-- Add support for snapshot flag to version command. Usage: `changeset version --snapshot [tag]`
-- Add support for tag flag to publish command. Usage: `changeset publish --tag <tag>` 
+- Add support for snapshot flag to version command. Usage: `changeset version --snapshot [tag]`. The updated version of the packages look like `0.0.0[-tag]-YYYYMMDDHHMMSS` where YYYY, MM, DD, HH, MM and SS is the date and time of when snapshot version is created. We can use this feature with combination of the tag option in publish command to publsh packages under exprimental tags from feature branches. To publish package a snapshot version of package under experimetal we can do:
```
Andarist, line 7: This is a good quality changeset! 👍

I've just a few grammar-related nits:

```suggestion
Add support for snapshot flag to version command. Usage: `changeset version --snapshot [tag]`. The updated version of the packages looks like `0.0.0[-tag]-YYYYMMDDHHMMSS` where YYYY, MM, DD, HH, MM, and SS is the date and time of when snapshot version is created. We can use this feature with a combination of the tag option in publish command to publish packages under experimental tags from feature branches. To publish a snapshot version of a package under experimental tag we can do:
```

.changeset/violet-gorillas-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+- Add support for tag flag to publish command. Usage: `changeset publish --tag <tag>`. This will publish the packages under passed npm tag.
```
Andarist, line 5: `-` is not needed here as the changeset entry is already generated as a bullet point, this would probably cause a nested list being created unnecessarily

```suggestion
Add support for tag flag to publish command. Usage: `changeset publish --tag <tag>`. This will publish the packages under passed npm tag.
```

packages/apply-release-plan/src/index.ts
Noviny: re config: the config refers to the written config, this doesn't want to be part of that.

It could be part of the release plan, but would like to see it land and stabilise, as the release plan being relatively clean is nice.

packages/cli/src/commands/version/index.ts
mitchellhamilton: pre mode changes the npm dist tag that's used and so does the `--tag option`. Given you shouldn't commit the result of a snapshot release, it's easy to exit pre mode and then do the snapshot and you won't get any unexpected behavior. We can always change this behavior in the future if we want to in the direction of not erroring but going in the opposite direction is harder.

ajaymathur:
Hey @mitchellhamilton , thanks for a lot for accepting suggestion and merging the PR. Due to my travel arrangement this week I will not able to respond. Appreciate it. 🙂

MichaelKapustey:
Hey guys, when do you expect to publish this change? I can't wait to use it already.

PR #371, opened 2020-05-19
Add support for `ignored` config

Feiyang1:
Implements https://github.com/atlassian/changesets/issues/242

@Andarist Can you please take a look? Need to add tests, but want to confirm that this approach looks reasonable to folks first.


Andarist:
Thanks for doing this! I will get to reviewing this soon.

packages/types/src/index.ts
```diff
@@ -63,6 +63,7 @@ export type Config = {
   baseBranch: string;
   /** The minimum bump type to trigger automatic update of internal dependencies that are part of the same release */
   updateInternalDependencies: "patch" | "minor";
+  ignored: ReadonlyArray<string>;
```
Andarist, line 66: It would be great to also include support for CLI's `--ignore` flag which was also requested.

A naming consideration - seems like `--ignore` suits better for the CLI flag and `ignored` for the config option. Is it acceptable for both to differ? Would probably be quite easy to surface option being misspelled for both cases, so maybe it's not a big problem to have them use a different form of the "ignore" word?

packages/config/src/index.ts
```diff
@@ -151,12 +152,40 @@ export let parse = (json: WrittenConfig, packages: Packages): Config => {
...
+        Array.isArray(json.ignored) &&
+        json.ignored.every(pkgName => typeof pkgName === "string")
+      )
+    ) {
+      messages.push(
+        `The \`ignored\` option is set as ${JSON.stringify(
+          json.ignored,
+          null,
+          2
+        )} when the only valid values are undefined or an array package names`
```
Andarist, line 167: nit
```suggestion
        )} when the only valid values are undefined or an array of package names`
```

packages/assemble-release-plan/src/is-ignored.ts
```diff
@@ -0,0 +1,3 @@
+export default function isIgnoredPackage(packageName: string, ignored: Readonly<string[]>): boolean {
```
Andarist, line 1: you could convert ignored to a simple lookup table (`Record<string, true>`) which would make this util obsolete as it ain't doing much anyway

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -55,8 +57,12 @@ export default function getDependents(
           nextRelease.name
         );
 
-        // Firstly we check if it is a peerDependency because if it is, our dependent bump type needs to be major.
-        if (
+        // If the dependent is an ignored package, we want to bump its dependencies without a release, so setting type to "none"
+        if (isIgnoredPackage(dependent, ignoredPackages)) {
```
Andarist, line 61: 👌

packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -41,6 +43,11 @@ export default function flattenReleases(
         release.changesets.push(changeset.id);
       }
 
+      // ignored packages will not trigger a release, so set the release type to "none"
+      if (isIgnoredPackage(name, ignoredPackages)) {
+        release.type = "none";
```
Andarist, line 48: should such a release contain a list of changesets? 🤔 probably not a problem - but worth giving it a thought

packages/assemble-release-plan/src/index.ts
```diff
@@ -169,13 +174,49 @@ function assembleReleasePlan(
   return {
     changesets,
     releases: [...releases.values()].map(incompleteRelease => {
-      return {
-        ...incompleteRelease,
-        newVersion: incrementVersion(incompleteRelease, preInfo)!
-      };
+      // For ignored packages, the newVersion is the same as the old version
+      // in both prereleased and regular releases
+      if (isIgnoredPackage(incompleteRelease.name, config.ignored)) {
```
Andarist, line 179: ignored packages have already `release.type === 'none'` and this would just reuse the `oldVersion` in `incrementVersion` here:
https://github.com/atlassian/changesets/blob/5543707846a9d84e9a6fe497780309d667f57321/packages/assemble-release-plan/src/increment.ts#L10-L11
so there is no need to do this extra conditional logic here

packages/assemble-release-plan/src/index.ts
```diff
@@ -169,13 +174,49 @@ function assembleReleasePlan(
...
+        };
+      }
+
     }),
     preState: updatedPreState
   };
 }
 
+// Changesets that contains both ignored and not ignored packages are not allowed
+function validateChangesets(changesets: NewChangeset[], ignored: Readonly<string[]>): void {
```
Andarist, line 197: I don't particularly feel that mixed changesets should not be allowed, but rather that a mixed changeset should exclude listed packages (and their dependents) from the release plan.

However, that logic would be more complex and probably is disputable so throwing an error is a good first step. Different heuristic can always be implemented in another PR.

packages/assemble-release-plan/src/index.ts
```diff
@@ -169,13 +174,49 @@ function assembleReleasePlan(
...
+    for (const release of changeset.releases) {
+      if (ignored.find(ignoredPackageName => ignoredPackageName === release.name)) {
+        ignoredPackages.push(release.name);
+      } else {
+        notIgnoredPackages.push(release.name);
+      }
+    }
+
+    if (ignoredPackages.length > 0 && notIgnoredPackages.length > 0) {
+      throw new Error(`
```
Andarist, line 212: I would look for how other error/logs are being formatted in the repo but I suspect this might be not ideal as template strings preserve whitespace and we might not want to include this leading & trailing whitespace in this error message.

packages/types/src/index.ts
Feiyang1: Added `--ignore` flag to cli. I think keeping the naming consistent will provide a better UX. If I knew the cli flag and the name is the same for the config option, I will always get it right the first time when trying to add the config option, and vice versa.

For now I made it `ignore` in both CLI and the config, but I have no preference on` ignore` or `ignored`. WDYT?

packages/assemble-release-plan/src/flatten-releases.ts
Feiyang1: Changed to not add ignored packages into the `releases` map here, because they will be added in `determineDependents()` anyway. As a result, such a release will contain an empty changeset.

packages/assemble-release-plan/src/index.ts
Feiyang1: Done. Changed to use string concatenation with trailing `\n`

packages/types/src/index.ts
Andarist: yeah, that's probably best - and `ignore` works better for both contexts

packages/assemble-release-plan/src/index.ts
```diff
@@ -91,12 +94,12 @@ function assembleReleasePlan(
...
   // flattened down to one release per package, having a reference back to their
   // changesets, and with a calculated new versions
-  let releases = flattenReleases(changesets, packagesByName);
+  let releases = flattenReleases(changesets, packagesByName, config.ignored);
 
   if (updatedPreState !== undefined) {
     if (updatedPreState.mode === "exit") {
       for (let pkg of packages.packages) {
-        if (preVersions.get(pkg.packageJson.name) !== -1) {
+        if (preVersions.get(pkg.packageJson.name)) {
```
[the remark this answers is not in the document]
Andarist: I suspect this was a mistake as well - would you be able to share the repro case for the situation you have encountered? It would be great to add a test covering your fix.

packages/cli/src/commands/version/index.ts
```diff
@@ -18,7 +18,16 @@ let importantEnd = chalk.red(
   "----------------------------------------------------------------------"
 );
 
-export default async function version(cwd: string, config: Config) {
+export default async function version(
+  cwd: string, 
+  { ignore }: { ignore?: string[] },
+  config: Config) {
+  // Commandline arguments take precedence over config object,
```
Andarist, line 25: I would prefer merging both - it would be a surprise to me that one-off usage (through CLI) would override the one that is committed to the config file.

packages/cli/src/index.ts
```diff
@@ -23,7 +23,7 @@ const { input, flags } = meow(
   Commands
     init
     add [--empty]
-    version
+    version [--ignore]
```
Andarist, line 26: Interesting - I've assumed previously that it would be a publish flag, but I think this makes more sense and makes the implementation more unified 👌

packages/cli/src/index.ts
```diff
@@ -143,7 +149,27 @@ const cwd = process.cwd();
...
+          ignoreArrayFromCmd = ignore;
+        }
+
+        // Validate that items in ignoreArrayFromCmd are valid project names
+        let pkgNames = new Set(
+          packages.packages.map(({ packageJson }) => packageJson.name)
+        );
+        for (let pkgName of ignoreArrayFromCmd || []) {
+          if (!pkgNames.has(pkgName)) {
+            error(
```
Andarist, line 165: Take a look how config validation gathers all errors: https://github.com/atlassian/changesets/blob/b876789f0a3a9913db5ee86e5fc76c5560529c82/packages/config/src/index.ts#L162

This strategy is quite cool as it allows surfacing all existing problems at once rather than surfacing one, user retrying the command after fixing their problem and bumping into another one immediately.

packages/assemble-release-plan/src/index.ts
Andarist: > incrementVersion() currently keeps executing and will give release.type === 'none' packages a new version in prereleases. Is it a valid use case?

I would say that no, it's not a valid use case - at least I don't see why one would want that. Good catch! So yes - I would make this an early return in `incrementVersion`

packages/cli/src/commands/version/index.ts
Feiyang1: I think merging might be confusing. I will have to remember what is set in the config file when using `--ignore` in the cli in order to understand the result correctly. I'm not confident that I will always remember it correctly, or I will have to keep going back to the config file when using cli, which doesn't feel like a good experience.

Typescript cli flags override options set in `tsconfig.json`. For example, `tsc --lib ES5` will override the `lib` option set in `tsconfig.json`. So I actually think many people will expect the override behavior. WDYT?

packages/cli/src/index.ts
Feiyang1: Yup, the `messages` array is neat.  Do you want to apply this strategy to cli flag validation? Currently we only have a single flag for `version` command, so the `messages` strategy won't provide any benefit yet. We can always change to the `messages` strategy in the future once we add more flags.

Andarist: Right -  doing the whole messages stuff is not needed right now. I have meant that right now this validation only reports a single package name that is not part of the project, but it could actually report all of them.

packages/cli/src/commands/version/index.ts
Andarist: Maybe we should just disallow mixing those 2 right now so we don't have to make this decision right now? @mitchellhamilton @Noviny what do you think - what would be the best behavior for this?

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
Feiyang1: Added a test - https://github.com/atlassian/changesets/pull/371/files#diff-33fca048f5c1d0f8bc4f760aaaee80a9R482

packages/cli/src/index.ts
Feiyang1: Ah, that makes sense. Updated to show all missing packages.

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -348,6 +348,156 @@ describe("assemble-release-plan", () => {
...
+        undefined
+      )
+    ).toThrowErrorMatchingInlineSnapshot(`
+"Found mixed changeset big-cats-delight
+Found ignored packages: pkg-b
+Found not Ignored packages: pkg-a
+Mixed changesets that contain both ignored and not ignored packages are not allowed"
+`);
+  });
+  it("should not generate a release for package that doesn't have changeset and is not a dependent in pre exit mode", () => {
```
Andarist, line 482: could you clarify what "is not a dependent in pre exit mode" means? Im a little bit unsure, does it just mean that it had no changesets related to it generated during pre mode?

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
Andarist: ❤️

packages/assemble-release-plan/src/index.ts
```diff
@@ -178,4 +182,35 @@ function assembleReleasePlan(
...
+      } else {
+        notIgnoredPackages.push(release.name);
+      }
+    }
+
+    if (ignoredPackages.length > 0 && notIgnoredPackages.length > 0) {
+      throw new Error(
+        `Found mixed changeset ${changeset.id}\n` +
+          `Found ignored packages: ${ignoredPackages.join(" ")}\n` +
+          `Found not Ignored packages: ${notIgnoredPackages.join(" ")}\n` +
```
Andarist, line 209:
```suggestion
          `Found not ignored packages: ${notIgnoredPackages.join(" ")}\n` +
```

packages/assemble-release-plan/src/increment.ts
```diff
@@ -6,10 +6,11 @@ export function incrementVersion(
   release: InternalRelease,
   preInfo: PreInfo | undefined
 ) {
-  let version =
-    release.type === "none"
-      ? release.oldVersion
-      : semver.inc(release.oldVersion, release.type)!;
+  if (release.type === "none") {
```
Andarist, line 9: this change could use a test for a `release.type === 'none' && preInfo !== undefined && preInfo.state.mode !== "exit"` case

before the change, this would increment the version, but we don't want to do that

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -25,7 +25,8 @@ export default function getDependents(
   releases: Map<string, InternalRelease>,
   packagesByName: Map<string, Package>,
   dependencyGraph: Map<string, string[]>,
-  preInfo: PreInfo | undefined
+  preInfo: PreInfo | undefined,
+  ignoredPackages: Readonly<string[]>
 ): boolean {
```
zkochan, line 30: this function has so many arguments. Why not a single options argument?

packages/cli/src/commands/version/index.ts
mitchellhamilton: I would say we shouldn't allow it as a command line argument precisely because it raises questions like this and regardless of what we choose, people might want different behavior + having command line arguments makes it harder to reason about what Changesets/these kinds of tools in general do in a repo.

Andarist: The CLI flag is supposed to help this - https://github.com/atlassian/changesets/issues/286, for which having to update the config file, doing the publish and reverting changes to the config file seems like a bad DX. 

I'm totally fine with disallowing mixing those 2 (config option and CLI flag) ever, such decision shouldn't affect any expectations because we won't provide any default strategy for handling this - instead, we will just fail fast and hard.

mitchellhamilton: > we will just fail fast and hard.

This sounds good to me.

packages/assemble-release-plan/src/index.ts
Feiyang1: Done.

packages/cli/src/commands/version/index.ts
Feiyang1: Updated to throw when `--ignore` and ignore in the config are used at the same time.

packages/assemble-release-plan/src/increment.ts
Feiyang1: Added a test.

packages/assemble-release-plan/src/index.test.ts
Feiyang1: It means the package has no changesets and is not a dependent of any packages being released. Such package should not have a release. Originally we will generate a patch release for such package if the preState mode is `exit`.

Yeah, the message is a bit confusing. I added a describe block and updated the message. Can you please take a look if it makes sense now?

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -348,6 +348,156 @@ describe("assemble-release-plan", () => {
...
+        ignore: ["pkg-b"]
+      },
+      undefined
+    );
+
+    expect(releases.length).toEqual(1);
+    expect(releases[0].name).toEqual("pkg-a");
+    expect(releases[0].newVersion).toEqual("2.0.0");
+  });
+  it("should generate releases with 'none' release type for ignored packages through dependencies", () => {
```
Andarist, line 374: Hm, is this test allowing pkg a (which is depending on a pkg b) to be released? Which one is a dependency and which one is a dependent here? I can't check it right now myself, just want to recheck that dependent of an ignored package should not be released as we can't guarantee this to be safe (the limitation applies for both dependencies and peerDependencies I believe).

Feiyang1: It's actually `pkg-b` depending on `pkg-a` in this test. It validates that ignored packages will have release type "none" if they depend on other packages being released.

Currently I'm not doing any special handling for dependents of the ignored packages, but I think you are right that they shouldn't be released because they could depend on the new features in the ignored packages.

So I implemented it by requiring user to list all dependents of the ignored packages in the ignore array in the config file (or to the `--ignore` flag). 
The alternative is we implicitly ignore the dependents not listed in the config file, but I think it will be too confusing. Let me know if you think it is not a good approach.

Andarist: > The alternative is we implicitly ignore the dependents not listed in the config file, but I think it will be too confusing. Let me know if you think it is not a good approach.

I was thinking about implicit ignoring, but you are right that it could be confusing in some scenarios. Given that throwing an error is the most conservative and safe approach I am OK with that as it's quite easy to lift off the restrictions in the future.

packages/cli/src/index.ts
```diff
@@ -185,6 +186,19 @@ const cwd = process.cwd();
           config.ignore = ignoreArrayFromCmd;
         }
 
+        // Validate that all dependents of ignored packages are listed in the ignore list
+        const dependentsGraph = getDependentsGraph(packages);
```
Andarist, line 190: I believe this would require a new test to be added.

Feiyang1: Added tests. I moved the main function in the cli `index.ts` into a new file `run.ts`, so we can test cli flags now, which we don't/can't do before.

Feiyang1:
Added changeset

packages/assemble-release-plan/src/increment.test.ts
```diff
@@ -0,0 +1,28 @@
+import { incrementVersion } from "./increment";
+import { InternalRelease, PreInfo } from "./types";
+
+describe("incrementVersion", () => {
```
Andarist, line 4: nit: I believe we would prefer to have a more integration-like test for this. A versioning-related test has been added for example [here](https://github.com/atlassian/changesets/pull/263/files#diff-068ddbdcddd68c5cb9e2b7efda8bbc2bR589) in the past

Don't be obliged to change this now, unless you want to. We can always just refactor this later.

packages/cli/src/index.ts
```diff
@@ -233,5 +87,6 @@ ${format("", err).replace(process.cwd(), "<cwd>")}
     return process.exit(err.code);
   }
   error(err);
+  // console.log("what's going on?", err)
```
Andarist, line 90: leftover 😉

.changeset/rich-melons-press.md
```diff
@@ -0,0 +1,26 @@
+---
+"@changesets/apply-release-plan": minor
+"@changesets/assemble-release-plan": minor
+"@changesets/cli": minor
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Added support for ignoring packages in the `version` command. The version of ignored packages will not be bumped,
+but their depedencies will still be bumped normally. This is useful when you have private packages, e.g. packages under development.
```
Andarist, line 10: i don't think we need to break line manually here, much better would be to rely on renderer line-wrapping. Depending on screen width etc this break might not look naturally.

.changeset/rich-melons-press.md
```diff
@@ -0,0 +1,26 @@
+---
+"@changesets/apply-release-plan": minor
```
Andarist, line 2: Im not sure if all of those packages require the same lengthy changeset, for example `@changesets/types` could only mention what has been added to what type, but doesn't have to go in full detail about the whole added feature.

packages/assemble-release-plan/src/increment.test.ts
Feiyang1: Added integration tests.

packages/cli/src/index.ts
Feiyang1: removed.

.changeset/rich-melons-press.md
Feiyang1: Ack. removed manual line breaks.

.changeset/rich-melons-press.md
Feiyang1: Created a separate changeset with shorter description for `@changesets/types` and `@changesets/config`

Feiyang1:
Addressed all comments. I think the PR is ready to merge. Thanks @Andarist for your thorough review!

packages/cli/src/run.ts
```diff
@@ -0,0 +1,212 @@
...
+          if (!pkgNames.has(pkgName)) {
+            messages.push(
+              `The package "${pkgName}" is passed to the \`--ignore\` option but it is not found in the project. You may have misspelled the package name.`
+            );
+          }
+        }
+
+        if (config.ignore.length > 0 && ignoreArrayFromCmd) {
+          messages.push(
+            `It looks like you are trying to use the \`--ignore\` cmd flag while ignore is defined in the config file. It is currently not allowed and you can use only one of them at a time.`
```
mitchellhamilton, line 134:
```suggestion
            `It looks like you are trying to use the \`--ignore\` option while ignore is defined in the config file. This is currently not allowed, you can only use one of them at a time.`
```

.changeset/proud-tips-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Add `ignore` config option to config ignored packages. The version of ignored packages will not be bumped during a release, but their depedencies will still be bumped normally.
```
mitchellhamilton, line 6:
```suggestion
Add `ignore` config option to configure ignored packages. The versions of ignored packages will not be bumped during a release, but their dependencies will still be bumped normally. 
```

packages/cli/src/run.ts
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        // Validate that all dependents of ignored packages are listed in the ignore list
+        const dependentsGraph = getDependentsGraph(packages);
+        for (const ignoredPackage of config.ignore) {
+          const dependents = dependentsGraph.get(ignoredPackage) || [];
+          for (const dependent of dependents) {
+            if (!config.ignore.includes(dependent)) {
+              messages.push(
+                `the package "${dependent}" depends on the ignored package "${ignoredPackage}", but itself is not being ignored. Please add it to the ignore array in the config file or pass it to the \`--ignore\` flag when using cli.`
```
mitchellhamilton, line 148:
```suggestion
                `The package "${dependent}" depends on the ignored package "${ignoredPackage}", but "${dependent}"  is not being ignored. Please pass "${dependent}" to the \`--ignore\` flag.`
```

mitchellhamilton: Can you add the above check to `@changesets/config`(with different wording)?

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
mitchellhamilton: Could you move this change to another PR? I want to have a deeper look into it and it's hard to do when combined with all these other changes.

packages/cli/src/run.ts
Feiyang1: Done. Added test as well.

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
Feiyang1: Done. moved it to https://github.com/atlassian/changesets/pull/382

Andarist:
@Feiyang1 could you resolve conflicts?

@Noviny @mitchellhamilton could you do a second round of reviews? It would be great to land this soon!

Feiyang1:
@mitchellhamilton gentle ping

Andarist:
One slight problem I have found with this feature. Upgraded dependencies of ignored packages are being written to the ignored package's `package.json` - which is good, but when we don't ignore packages then "dependency release lines" are being appended at the end of patch release line. So this slight problem is that when we unignore a package those dependency release lines won't magically appear in the changelog so we'll have slightly different content than when not using this feature.

This, of course, is not a big deal - I don't quite care about those dependency release lines. It's just something I've noticed now.

PR #372, opened 2020-05-20
Fix internal update deps

Blasz:
This fixes an issue with the new `updateInternalDependencies` config setting set to `minor` where patch bumped dependencies that left the semver range of a dependent were not updated.

The fix involves adding a new `dependenciesLeavingRange` field to the release plan that lists dependencies of a released package that leave semver range. This knowledge is required by `apply-release-plan` so that it can always bump these packages regardless of the `updateInternalDependencies` setting.

I've made the field optional to not make a ton of things breaking and because it's only required by this one configuration option.

I also came across some `@ts-ignore`'s so tried to remove as many from source files as I could.

### DIFF
--- a/.changeset/old-lions-end.md
+++ b/.changeset/old-lions-end.md
@@ -1,0 +1,7 @@
+---
+"@changesets/assemble-release-plan": minor
+"@changesets/types": minor
+---
+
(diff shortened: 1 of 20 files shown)


Blasz:
Closing in favour of a new approach in #373 that does not require modifying the release plan or assemble-release-plan

PR #382, opened 2020-06-01
Fix a bug where packages that shouldn't get released get patch releases when the pre mode is exit

Feiyang1:
Fix a bug where packages that shouldn't get released get patch releases when the pre mode is exit


packages/assemble-release-plan/src/index.ts
```diff
@@ -128,7 +128,7 @@ function assembleReleasePlan(
   if (updatedPreState !== undefined) {
     if (updatedPreState.mode === "exit") {
       for (let pkg of packages.packages) {
-        if (preVersions.get(pkg.packageJson.name) !== -1) {
+        if (preVersions.get(pkg.packageJson.name)) {
```
mitchellhamilton, line 131: I know this doesn't affect the behavior but it was hard for me to understand why this fixes something and what this is actually checking and this makes that more clear. 
```suggestion
        if (preVersions.get(pkg.packageJson.name) !== 0) {
```

Andarist: I took another look at this and now I'm even more confused.

We only call `preVersions.set` within `updatedPreState.mode !== "exit"` block here:
https://github.com/atlassian/changesets/blob/64a9017aefa6b6444d83cf7802cca8e8404fb697/packages/assemble-release-plan/src/index.ts#L85
but this checks happens within `updatedPreState.mode === "exit"` block here:
https://github.com/atlassian/changesets/blob/64a9017aefa6b6444d83cf7802cca8e8404fb697/packages/assemble-release-plan/src/index.ts#L129

I suspect this might be a constant condition and this whole thing is bugged. Or have I missed something?

Feiyang1: @mitchellhamilton I don't think `!== 0` works. `preVersions.get()` returns `undefined` if there is no match. I think we only want to execute the block if there is a match, but `undefined !== 0` will evaluate to `true`, the same issue as `undefined !== -1`.

I could be wrong, because I don't really understand the purpose of this block. What @Andarist pointed out above confuses me too. @mitchellhamilton, do you have insight?

mitchellhamilton: > I suspect this might be a constant condition and this whole thing is bugged. Or have I missed something?

Yep (this is why I wanted to have a deeper look at it)

The point of this block is to make sure that any packages that were released while in pre mode will be (note that this wouldn't happen normally because when exiting pre mode, we essentially take the versions of packages before pre mode and apply the changesets to them but some packages will be bumped in pre mode but not in the normal process because semver ranges are different with prerelease versions so we need to do a patch bump to those)

I think the fix is to check `!== 0` and move the bit that sets the preVersions outside of the block that checks `updatedPreState.mode !== "exit"`

Andarist: > but some packages will be bumped in pre mode but not in the normal process because semver ranges are different with prerelease versions so we need to do a patch bump to those)

Could you elaborate on this one? It's a little bit convoluted and I don't think I understand this part.

mitchellhamilton: Let's say we have the following project:

- pkg-a@1.0.0
  - dependencies:
    - pkg-b@^1.0.0
- pkg-b@1.0.0

Then we enter pre mode.

We add a changeset for pkg-b@patch.

We release and the project should look like this:

- pkg-a@1.0.1-next.0
  - dependencies:
    - pkg-b@1.0.0-next.0 (I can't remember if we put the caret here or not)
- pkg-b@1.0.1-next.0

(note that we're bumping pkg-a here, we have to do this because `1.0.1-next.0` is not satisfied by`^1.0.0`)

Now if we exit pre mode, the project should look like this

- pkg-a@1.0.1
  - dependencies:
    - pkg-b@^1.0.1 (whether this should be updated or not is arguable but that's not the point here)
- pkg-b@1.0.1

It's important that pkg-a is bumped for two reasons:
- Saying "we're done with the prerelease now, you should go to an older version of pkg-a to get the latest stuff" is confusing for consumers of your packages
- If you then entered pre mode again without bumping pkg-a, you could get the version of pkg-a to be `1.0.1-next.0` again (this is the big issue)

Feiyang1: Ah, that makes sense. Thanks for the detailed explanation! I have updated the PR according to your request and added an integration test. PTAL.

Andarist: @mitchellhamilton This is a great visual~ explanation of this behavior. Could probably make its way to some FAQ section for pre mode.

Andarist: just a question out of my curiosity - why one of the tests has been added as an integration test and the other one is more of a unit test?

Feiyang1: @Andarist I didn't think too much about it, so there is no particular reason. Is there any convention on when should integration test be added vs unit test, or both?  I can certainly make another PR to add missing tests.

Andarist: My personal preference is to only implement integrations tests or unit tests for a particular package (treating it as a unit), so not implementing tests for particular files that have not been exposed as part of a public interface of a package. 

I'm not sure what exact convention is used here, throughout the repository, and I don't really need you to adjust any of those added tests here. I was just curious why have you chosen one over another as I don't see anyone suggesting usage one over another to you. This kind of thing just always draws my curiosity - that's why I've asked 😅 I like to understand different perspectives.

PR #385, opened 2020-06-04
Add link to the comment on GitHub with more details about one of the tests

Andarist:
This is a very valuable comment and IMHO it's worth adding it here as a quick reference for the more in-depth explanation.


Andarist:
Hm, this actually was already linked here: https://github.com/atlassian/changesets/pull/382/files#diff-d13e26984022cd1d4154e9af750e8dddR135 and I've missed it. Not sure if it's worth duplicating it in the test file as well, I will let others decide 😅

mitchellhamilton (approved):

PR #389, opened 2020-06-09
Add an experimental option `useCalculatedVersionForSnapshots`

Feiyang1:
Implements https://github.com/atlassian/changesets/issues/381#issuecomment-636549738
Currently snapshot version always starts with `0.0.0`. The experimental option allows us to use the calculated version (the next version) instead of `0.0.0`. It is useful when you want to use snapshot for canary/pre releases while still communicating the intended next version.


Feiyang1:
@mitchellhamilton When you have a moment, can you please take a look?

PR #395, opened 2020-06-17
fix: use defaults to format package.json

jonathanmorley:
prettier formatting of package.json files now defaults to using json-stringify.
Specifying `json` explicitly was causing some inconsistencies


Andarist:
Taken from Prettier docs:
> "json-stringify" (same parser as "json", but outputs like JSON.stringify) First available in v1.13.0

Do you actually know what's the actual difference? 🤔 Do you know in which version the default has changed?

jonathanmorley:
It appears that `json-stringify` was used for `package.json` files since it was introduced (See the `filenames` property under https://github.com/prettier/prettier/commit/1062f2a8898b9e1ba00de1a535797ce9b22d102a#diff-ef736ec1de130f41ef22f8e5c31df9f9)

The only difference i have seen is when I have a small objects and arrays. Specifically, having

```
{
  "files": [
    "lib"
  ]
}
```

in a package.json will be valid according to prettier defaults and prettier with the `json-stringify` parser.

However, prettier with the `json` parser (and by extension changesets) will change it to

```
{ "files": ["lib"] }
```

This was reproducible across prettier 1.18.2 (the lowest version that changesets is pinned at), and prettier 2.0.5 (latest)

Andarist:
I'm wondering if maybe we shouldn't just use a custom modifier for this, for example using this: https://github.com/microsoft/node-jsonc-parser . By converting input json and getting locations from the AST we could just manipulate the input json without any formatting changes and this is basically what we are after here as we don't care about the formatting of a file.

jonathanmorley:
I suspect that https://github.com/atlassian/changesets/commit/90f8bc95d8b47628bc07404084e0400c387dc9cc was introduced as a partial workaround to this issue.

Given:

```
09:38:23 AM ❯ cat --plain tmp/package.json
{
  "files": ["lib"],
  "files2": ["longer file"],
  "files3": ["even longer filename"]
}
```

Compare

```09:38:30 AM ❯ npx prettier tmp/package.json --parser json --print-width 20
npx: installed 1 in 1.331s
{
  "files": ["lib"],
  "files2": [
    "longer file"
  ],
  "files3": [
    "even longer filename"
  ]
}```

with 

```
09:38:52 AM ❯ npx prettier tmp/package.json --parser json-stringify
npx: installed 1 in 0.968s
{
  "files": [
    "lib"
  ],
  "files2": [
    "longer file"
  ],
  "files3": [
    "even longer filename"
  ]
}```

jonathanmorley:
I think the custom modifier would be preferable, but would you want it instead of this PR? or as a follow up?

Andarist:
Personally I would prefer repurposing this PR as the current solution only moves a problem (and not just in theory) because one can easily have custom prettier settings for JSON files (or no prettier at all) and this would in turn change formatting for those people.

This would increase the overall complexity of the solution, so I would prefer receiving a green light for this from @mitchellhamilton or @Noviny before proceeding further. It would be a shame to put a considerable effort into this to only have the work dismissed later.

jonathanmorley:
One point of contention is that for people using the default prettier settings, changesets is _already_ broken, since it overrides those settings with its own

Andarist:
> Custom prettier settings should still be respected.

Right, if possible - yes, definitely. As the changes that have to be done to the package.json are minimal though I feel like the easiest solution is to actually avoid reading any configs, or providing custom configs or things like that. Just update the JSON in place - probably could be done with RegExps, but I don't think they are ultimately best for this kind of a job, especially that they are super hard to maintain. Hence my proposal to just parse JSON to the AST, traverse it to get input locations and manipulate the input string based on this information.

maraisr:
As @Andarist has said:
> I'm wondering if maybe we shouldn't just use a custom modifier for this, for example using this: https://github.com/microsoft/node-jsonc-parser . By converting input json and getting locations from the AST we could just manipulate the input json without any formatting changes and this is basically what we are after here as we don't care about the formatting of a file.

I believe this to be a better way of doing it. This way there is no formatting bias to those that don't use prettier. I guess the only concern here is, if you are using prettier—could a version bump inline replace using the AST no longer render the file validly with prettier? I doubt it.

packages/apply-release-plan/src/version-package.ts
```diff
@@ -31,7 +36,10 @@ export default function versionPackage(
 ) {
   let { newVersion, packageJson } = release;
 
+  const edits = [];
+
   packageJson.version = newVersion;
+  edits.push(...jsonParser.modify(release.pkgJSONText, ["version"], newVersion, {}));
```
Andarist, line 42: instead of gathering "real" edits and applying them in this function, you could gather `patches` here:
```ts
type JSONPath = string[]
const patches = new Map<JSONPath, string>()
```

and later convert those to edits and apply them in one go just before writing the updated `package.json`. 

I feel that this would isolate both required changes for this more and interaction with `jsonParser` as well.

__fixtures__/simple-project-with-strange-indentation/packages/pkg-a/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "pkg-a",
+  "version": "1.0.0",
+  "dependencies": {
+      "pkg-b": "1.0.0"
```
Andarist, line 5: could you maybe make the indentation in this file even stranger? right now it looks somewhat OK at first so might be easy to miss what this is supposed to represent

packages/apply-release-plan/src/index.ts
```diff
@@ -60,9 +60,14 @@ export default async function applyReleasePlan(
...
         `Could not find matching package for release of: ${release.name}`
       );
+
+    const pkgJSONPath = path.resolve(pkg.dir, "package.json");
+    const pkgJSONText = fs.readFileSync(pkgJSONPath, { encoding: 'utf-8' });
+
     return {
       ...release,
-      ...pkg
+      ...pkg,
```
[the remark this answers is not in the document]
jonathanmorley: I was originally thinking of that, but it felt better to keep all the fs logic in here, and maintain the purity of the `versionPackage` function.

__fixtures__/simple-project-with-strange-indentation/packages/pkg-a/package.json
jonathanmorley: sure

jonathanmorley:
I wasn't able to reproduce the tabs/spaces issue from #396. I would welcome the addition of tests around that, especially if we can show that this change fixes it

mitchellhamilton:
Sorry I missed this earlier but adding jsonc-parser seems like a lot of complexity for no big gain. Let's use something like this to write the package.jsons given that package managers already use `JSON.stringify` when writing

```tsx
import * as fs from "fs-extra";
import detectIndent from "detect-indent";

let pkgRaw = await fs.readFile(packageJsonPath, "utf-8");
let indent = detectIndent(pkgRaw).indent || "  ";
await fs.writeFile(
  packageJsonPath,
  JSON.stringify(pkgJsonContent, null, indent) +
    (pkgRaw.endsWith("\n") ? "\n" : "")
);
```

Andarist:
@mitchellhamilton I'm afraid that this doesn't solve the problem at all or would only push the problem towards other cases. The proposed solution only tries to preserve the original indentation, but the problem here is not about indentation being changed, but rather formatting of particular values, mainly things like this:
```json
{
  "key": ["val1"]
}
```
This array is being sometimes kept in a single line and sometimes spread into multiple lines.

mitchellhamilton:
> This array is being sometimes kept in a single line and sometimes spread into multiple lines.

When is it kept in a single line(except if someone manually changes it which I'm not really concerned about)? Given that prettier formats package.json's with JSON.stringify now(or rather something that is functionally equivalent) and package managers use JSON.stringify, I don't see when it would happen?

Andarist:
It can be kept in a single line when somebody is using prettier's `"json"` parser - which seems like an easy "mistake" to make looking at the fact that it has been made here in changesets and I've also made it in my work's project:
```
    {
      "files": [".prettierrc", "*.json"],
      "options": {
        "parser": "json",
        "useTabs": false,
        "printWidth": 80
      }
    }
```
I had no idea there was an alternative `"json-stringify"` parser that could be used.

Given though that this setting already conflicts with the package managers I guess we could just treat Changesets doing the same thing as them and thus this shouldn't be a problem. So maybe we should just switch to using `"json-stringify"` parser?

mitchellhamilton:
> So maybe we should just switch to using "json-stringify" parser?

I don't see why we would want to use it over just JSON.stringify? From reading what it does, it handles cases where there are commas that shouldn't be there and etc. by using `@babel/parser` which doesn't really matter here?

Andarist:
No, there is no added value in that parser (that I can see). I just have meant that, for now, that would be the easiest solution for this problem as we already use prettier here. So in a sense - maybe [the original fix](https://github.com/atlassian/changesets/pull/395/commits/8a5ae4d26d5b07e96f8555a042748ee81b7fc406) by @jonathanmorley should just be restored?

emmenko:
It would be nice to fix this formatting issue with the hardcoded value of `printWidth: 20`, as it just messes things up a bit with the default prettier value of `80`.

Andarist:
@emmenko could u give more details about the issue you are experiencing?

emmenko:
Simply put, our project doesn't specify the `printWidth`, so it uses the default one of `80`. Therefore, the `package.json` files are formatted in a certain way.
Then when Changeset bumps the versions it runs prettier with the `printWidth: 20`, which introduces an unnecessary formatting change.

For example
image

Andarist:
@emmenko do u have any custom formatting specified for `.json` files? Couldn't you just reformat those however you like using precommit hook with husky + lint-staged or similar tools? The current line of my thinking aligns with @mitchellhamilton 
 - we shouldn't use Prettier here, but rather `JSON.stringify` directly which should match package managers' behavior. Given that this is their behavior and you would usually use them to bump packages such a change would not be a DX downgrade when comparing to using package managers directly.

jonathanmorley:
While updating this PR to use `JSON.stringify` directly (and not use `prettier`), I realised that `JSON.stringify` defaults to not having any whitespace (newlines, indentation, etc).

We would therefore have to make other decisions as to whether we want that whitespace, how much of it we want, and whether we want tabs or spaces.

Any of these could/would run afoul of users specific choices that don't align.

I still think that the current solution assumes the least about the package.jsons in the repos, and tries to make the minimal changes required.

emmenko:
Maybe we should just introduce some kind of "hook" configuration so that users can specify their own way of formatting things.
Then changesets simply uses `JSON.stringify` (or whatever reasonable default) and users can opt-in to "patch" the formatting as an "after" step. Conceptually similar to what `lint-staged` does.

Example option in `.changeset/config.json` (the option name is not important):

```json
"formatterCommand": "prettier --write --parser json"
```

It's assumed that the file path is provided to the command.

jonathanmorley:
I see this discussion as being how to balance 'internal complexity' and 'external complexity'

The current state of this PR (with the jsonc-parser) increases the internal complexity of the repository, especially compared to just `JSON.stringify` (and even compared to formatting with prettier)

Other suggestions appear to increase the 'external complexity' by requiring new config elements to control the formatting explicitly.

I think we probably need someone like @Andarist to weigh in on which complexity the project would want to focus more on reducing.

mitchellhamilton:
We can use detect-indent to get the indentation we should use, this is all that we need to do:

```ts
import * as fs from "fs-extra";
import { Package } from "@manypkg/get-packages";
import path from "path";
import detectIndent from "detect-indent";

async function writePackage(pkg: Package) {
  let pkgRaw = await fs.readFile(path.join(pkg.dir, "package.json"), "utf-8");
  let indent = detectIndent(pkgRaw).indent || "  ";
  return fs.writeFile(
    path.join(pkg.dir, "package.json"),
    JSON.stringify(pkg.packageJson, null, indent) +
      (pkgRaw.endsWith("\n") ? "\n" : "")
  );
}

**Comment by Andarist** [comment]
Right, and this is what package managers do. For example:
https://github.com/pnpm/pnpm/blob/cb6e6fc8aa6dd4648e3fee5e54c75ce14d24d9d1/packages/write-project-manifest/src/index.ts#L29-L30

@emmenko I'm not quite sure how a changeset-specific option is really useful here if you can do **the very** same thing by just invoking `prettier` appropriately after invoking `changeset`. This would be pretty much equivalent and doesn't require any documentation or anything whatsoever from the Changesets itself

**Comment by emmenko** [comment]
> I'm not quite sure how a changeset-specific option is really useful here if you can do the very same thing by just invoking prettier appropriately after invoking changeset.

The main problem is when using the github action, there is no way to run any custom command **before** the commit is pushed.

If you run `changeset` manually then of course you can just run `prettier` afterwards.

**Comment by Andarist** [comment]
I don't think that's entirely true. You can specify a custom `version` command:
https://github.com/changesets/action/blob/bfeb9e077e6cf393e4c4ef17e2bbc75b308b4364/action.yml#L10-L12
And that should allow you to do what you want.

**Comment by emmenko** [comment]
Could you provide an example how that would work?

**Comment by Andarist** [comment]
It would look smth like this:

**release.yml**
```yml
name: Release

on:
  push:
    branches:
      - master

jobs:
  release:
    name: Release
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Repo
        uses: actions/checkout@master
        with:
          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
          fetch-depth: 0

      - name: Setup Node.js 12.x
        uses: actions/setup-node@master
        with:
          node-version: 12.x

      - name: Install Dependencies
        run: yarn

      - name: Create Release Pull Request or Publish to npm
        id: changesets
        uses: changesets/action@master
        with:
          version: yarn version
          publish: yarn release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

**package.json**
```json
{
  "scripts": {
    "version": "changeset version && prettier '**/package.json' —write"
  }
}
```

emmenko:
Thanks I'll give it a try.

packages/apply-release-plan/src/index.test.ts
```diff
@@ -108,6 +108,35 @@ async function testSetup(
 
 describe("apply release plan", () => {
   describe("versioning", () => {
+    describe("formatting", () => {
+      it("should not reformat a package.json with a small array", async () => {
```
Andarist, line 112: I'm not super sure if this test is needed - I wouldn't say that expected results are super-known to us. We don't impose any special rules on the formatting and it's not our concern how this gets printed. Our concerns are:
- indent preservation
- inserting trailing newline conditionally

packages/apply-release-plan/src/version-package.ts
```diff
@@ -15,11 +11,7 @@ const DEPENDENCY_TYPES = [
 ] as const;
 
 export default function versionPackage(
-  release: ComprehensiveRelease & {
-    changelog: string | null;
-    packageJson: PackageJSON;
-    dir: string;
-  },
+  release: ModCompWithPackage & { changelog?: string },
```
Andarist, line 14: The type of the `changelog` property has changed here. Are we sure that the new one is correct according to the runtime? I'm surprised that this hasn't caused any type errors upstream... unless this is actually always provided as a string and null/undefined was only handled internally here 🤔 but in such a case I'm not sure why this is supported if it's not ever utilized.

Andarist: sideways q: do u have any idea what `ModCompWithPackage` stands for? 😅

packages/apply-release-plan/src/index.test.ts
Andarist: I'm actually not saying that we need tests for those "our concerns" - I'm just commenting that I feel like those would actually be more useful than the one that has been provided.

jonathanmorley: This is a regression test. It would fail if we were to return to using prettier to format the package.json. I agree that indent and trailing new line tests would also be useful

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 3
 jobs:
   test:
     docker:
-      - image: circleci/node:10.13.0
+      - image: circleci/node:12.18.2
```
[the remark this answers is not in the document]
jonathanmorley: Not sure, i'll try removing it

mitchellhamilton (approved):

Andarist:
@jonathanmorley thank you for your work and bearing with us 😅

PR #404, opened 2020-07-01
Upgrade TS and related deps

Andarist:
I was looking into if upgrading TS would resolve https://github.com/atlassian/changesets/pull/402 without explicit return type annotation and, well, it didn't. I've upgraded a bunch of deps though, so it still might be worth merging

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,11 +27,11 @@
   "author": "Ben Conolly",
   "license": "MIT",
   "dependencies": {
-    "@babel/cli": "^7.4.4",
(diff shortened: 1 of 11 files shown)


Andarist:
@mitchellhamilton any objections against this?

mitchellhamilton:
> I was looking into if upgrading TS would resolve #402 without explicit return type annotation and, well, it didn't

See https://github.com/preconstruct/preconstruct/pull/293

PR #408, opened 2020-07-07
Override commit flag for pre-releases only

hsinha:
**Current behaviour:**
If the `commit` flag in config.json is set to `true` it is still overriden to `false` when doing `changeset version` for all releases

**Fix:**
`commit` flag will be overriden only for pre-releases

### DIFF
--- a/packages/cli/src/commands/version/index.ts
+++ b/packages/cli/src/commands/version/index.ts
@@ -31,12 +31,15 @@ export default async function version(
     readPreState(cwd),
     removeEmptyFolders(path.resolve(cwd, ".changeset"))
   ]);
+  let releaseConfig = config;
 
   // temporarily needed because of TS 3.7 regression - https://github.com/microsoft/TypeScript/issues/33752
   const changesets = _changesets as NonNullable<typeof _changesets>;
   const preState = _preState as NonNullable<typeof _preState>;
 
   if (preState !== undefined && preState.mode === "pre") {
+    releaseConfig = { ...releaseConfig, commit: false };
+
     warn(importantSeparator);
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/version/index.ts
```diff
@@ -31,12 +31,15 @@ export default async function version(
...
     removeEmptyFolders(path.resolve(cwd, ".changeset"))
   ]);
+  let releaseConfig = config;
 
   // temporarily needed because of TS 3.7 regression - https://github.com/microsoft/TypeScript/issues/33752
   const changesets = _changesets as NonNullable<typeof _changesets>;
   const preState = _preState as NonNullable<typeof _preState>;
 
   if (preState !== undefined && preState.mode === "pre") {
+    releaseConfig = { ...releaseConfig, commit: false };
```
Andarist, line 41: q: why do we need this?

hsinha: To override the config **only** if its a pre-release. `changeset version` should commit changes if `commit: true` in config.json. Atleast that was the behaviour earlier. 

**Background:** This change https://github.com/atlassian/changesets/pull/359/files overrides config.json and sets `commit: false`. According to @ajaymathur the original intent was to override it only for pre-releases but right now its overriding it for all releases.

Andarist: If this was incorrectly overridden in #359 then most likely the intention was to only override it for **snapshot** releases and not when in pre-mode. Those 2 are somewhat similar but not the same.

Andarist:
This got superseded by https://github.com/atlassian/changesets/pull/437

PR #412, opened 2020-07-10
update a dependency correctly when it's declared multiple times with different range

Feiyang1:
Fixes https://github.com/atlassian/changesets/issues/411


package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "description": "A tool to help manage the versioning and changelogs for open source packages",
   "scripts": {
-    "test": "jest",
+    "test": "jest -t 'declared multiple times'",
```
Andarist, line 7:
```suggestion
    "test": "jest",
```

packages/cli/src/commands/version/version.test.ts
```diff
@@ -287,6 +287,42 @@ describe("running version in a simple project with caret dependencies", () => {
   });
 });
 
+describe.only("running version in a simple project with the same dependency declared multiple times (e.g. in both dependencies and devDependencies) with different version range", () => {
```
Andarist, line 290:
```suggestion
describe("running version in a simple project with the same dependency declared multiple times (e.g. in both dependencies and devDependencies) with different version range", () => {
```

Andarist: also - it's a `describe` containing just a single test, is it worth to wrap this into 2 hierarchical layers? I think that in here this should just be a top-level test

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -149,7 +160,7 @@ export default function getDependents({
   matched ('dependencies', 'peerDependencies', etc) and the versionRange itself ('^1.0.0')
```
Andarist, line 160: this comment has to be adjusted slightly

.changeset/nice-eels-study.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/assemble-release-plan": patch
+"@changesets/cli": patch
```
Andarist, line 4:
```suggestion
"@changesets/assemble-release-plan": patch
"@changesets/cli": patch
```

I don't think we need to bump `@changesets/apply-release-plan` because of this change. You have only implemented a logic change within `@changesets/assemble-release-plan` - listing this as a change for `@changesets/cli` makes sense though as it's the main user-facing package and people usually look into its changelog to see what has been changed.

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -59,48 +59,59 @@ export default function getDependents({
 
         const dependentPackage = packagesByName.get(dependent);
         if (!dependentPackage) throw new Error("Dependency map is incorrect");
-        const { depTypes, versionRange } = getDependencyVersionRange(
+        const dependencyVersionRanges = getDependencyVersionRanges(
```
Andarist, line 62: could you move this into `else` block in the below logic? it seems to be only relevant for it and not for the whole function body

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -59,48 +59,59 @@ export default function getDependents({
...
-              versionRange
-            )
-          ) {
+          for (const { depType, versionRange } of dependencyVersionRanges) {
+            // we check if it is a peerDependency because if it is, our dependent bump type needs to be major.
             if (
-              depTypes.includes("dependencies") ||
-              depTypes.includes("optionalDependencies") ||
-              depTypes.includes("peerDependencies")
+              depType === "peerDependencies" &&
```
Andarist, line 74: hm, I find it quite hard to read - maybe worth moving this into a separate function? that way each condition could receive a nice comment about the reasoning behind it

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -59,48 +59,59 @@ export default function getDependents({
...
+                  }
+                } else {
+                  // We don't need a version bump if the package is only in the devDependencies of the dependent package
+                  if (
+                    type !== "major" &&
+                    type !== "minor" &&
+                    type !== "patch"
+                  ) {
+                    type = "none";
+                  }
```
Andarist, line 112: Isn't this equivalent to just `type = "none";`?

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -59,48 +59,59 @@ export default function getDependents({
...
+              if (
+                // TODO validate this - I don't think it's right anymore
+                (!releases.has(dependent) ||
+                  releases.get(dependent)!.type === "none") &&
+                !semver.satisfies(
+                  incrementVersion(nextRelease, preInfo),
+                  versionRange
+                )
+              ) {
+                if (
```
Andarist, line 96: nit: I would find this one to be more readable if it would use a switch statement

packages/cli/src/commands/version/version.test.ts
Feiyang1: I added `describe` because otherwise the string becomes super long and hard to read & understand. But if you feel strongly about, I can merge them into a top level test.

packages/assemble-release-plan/src/determine-dependents.ts
Feiyang1: Done

packages/assemble-release-plan/src/determine-dependents.ts
Feiyang1: Done

packages/assemble-release-plan/src/determine-dependents.ts
Feiyang1: No. In case a package is declared as a dependency multiple times (e.g. in both peerDependencies and devDependencies), we handle them one by one, we don't want to overwrite `type` with `none` if it's set to a higher bump type.

packages/assemble-release-plan/src/determine-dependents.ts
Feiyang1: Done.

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -93,22 +91,23 @@ export default function getDependents({
...
-                ) {
-                  if (type !== "major" && type !== "minor") {
-                    type = "patch";
+                switch (depType) {
+                  case "dependencies":
+                  case "optionalDependencies":
+                  case "peerDependencies": {
+                    if (type !== "major" && type !== "minor") {
+                      type = "patch";
+                    }
```
Andarist, line 101: should this fallthrough?

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -93,22 +91,23 @@ export default function getDependents({
...
-                    type = "none";
+                  case "devDependencies": {
+                    // We don't need a version bump if the package is only in the devDependencies of the dependent package
+                    if (
+                      type !== "major" &&
+                      type !== "minor" &&
+                      type !== "patch"
+                    ) {
+                      type = "none";
+                    }
```
Andarist, line 111: could we maybe remove handling `devDependencies` at all here and just assign `"none"` upfront to the `type` instead of making it `VersionType | undefined`?

packages/cli/src/commands/version/version.test.ts
```diff
@@ -287,6 +287,42 @@ describe("running version in a simple project with caret dependencies", () => {
...
+    expect(packages.packages.map(x => x.packageJson)).toEqual([
+      {
+        dependencies: {
+          "pkg-b": "^1.0.1"
+        },
+        devDependencies: {
+          "pkg-b": "1.0.1"
+        },
+        peerDependencies: {
+          "pkg-b": "^1.0.1"
```
Andarist, line 313: I'm always a little bit when it comes to how changesets handle peer dependencies, but I have thought that updating a peer dependency with a patch release shouldn't bump this? Or it just bumps the range but doesn't bump the major of the dependent package (pkg-a here)?

packages/cli/src/commands/version/version.test.ts
```diff
@@ -287,6 +287,42 @@ describe("running version in a simple project with caret dependencies", () => {
   });
 });
 
+describe("running version in a simple project with the same dependency declared multiple times (e.g. in both dependencies and devDependencies) with different version range", () => {
```
Andarist, line 290: this is quite mouthful and when I read it I need to focus quite a bit to understand this description, maybe a single description like this:
"should update different range types across different dependency types correctly"
would be enough?

packages/assemble-release-plan/src/determine-dependents.ts
Feiyang1: IIUC, `undefined` is different from `none` and represents a valid state for packages that should be left untouched in the release, e.g. the package whose dependency's next version is still in range. So I don't think we can default `type` to `none`

packages/cli/src/commands/version/version.test.ts
Feiyang1: Correct. Patch bump to a peerDependency doesn't bump the major of the dependent package. See [here](https://github.com/atlassian/changesets/blob/master/packages/assemble-release-plan/src/determine-dependents.ts#L74).

packages/cli/src/commands/version/version.test.ts
Feiyang1: Updated.

Feiyang1:
@Andarist gentle ping

packages/cli/src/commands/version/version.test.ts
Andarist: An offtopic observation - the linked line looks suspicious. What if `nextRelease.type === 'none'`? This might be worth investigating.

__fixtures__/simple-project-same-dep-diff-range/packages/pkg-a/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "pkg-a",
+  "version": "1.0.0",
+  "dependencies": {
+    "pkg-b": "^1.0.0"
+  },
+  "peerDependencies": {
```
Andarist, line 7: was the intention to have the same range type defined in both `"dependencies"` and `"peerDependencies"`?

Feiyang1: Nop. I think the more realistic scenario is use range version in the peerDependencies, but use a fixed version in the devDependencies. I updated the test fixture to reflect that.

packages/cli/src/commands/version/version.test.ts
Feiyang1: hmm, that seems like a possibility. I might look into it when I get time and create an issue if I can reproduce. I will leave it in this PR though.

Feiyang1:
@Andarist Sorry for the extremely slow response. Can you please take another look?

Andarist (approved):

PR #413, opened 2020-07-15
Fixed mentions to @changesets/cli in the docs

Andarist:


zkochan:
TBH, I always forget whether it is plural or singular... But probably singular would make more sense.

Andarist:
Same here - I've actually noticed this because of that. I was initializing a new project and copied this header (because I wasn't 100% sure which one is correct) into `npm install` command which has failed. Also later noticed the bot's name which makes the whole thing even more confusing for users.

PR #422, opened 2020-08-02
Fix #397: Unexpected end of JSON input issue

### LINKED ISSUE #397: SyntaxError: Unexpected end of JSON input when trying to publish packages
## Affected Packages

All of the packages

## Problem

When I try to publish my packages using `pnpx changesets publish` I get this error:

```
➜ pnpx changeset publish
🦋  info npm info @tinyhttp/app
🦋  info npm info @tinyhttp/cors
🦋  info npm info @tinyhttp/etag
🦋  info npm info @tinyhttp/logger
🦋  info npm info @tinyhttp/static
🦋  info @tinyhttp/app is being published because our local version (0.1.21) has not been published on npm
🦋  info @tinyhttp/cors is being published because our local version (0.1.20) has not been published on npm
🦋  info @tinyhttp/etag is being published because our local version (0.1.20) has not been published on npm
🦋  info @tinyhttp/logger is being published because our local version (0.1.20) has not been published on npm
🦋  info @tinyhttp/static is being published because our local version (0.1.20) has not been published on npm
🦋  info Publishing "@tinyhttp/app" at "0.1.21"
🦋  info Publishing "@tinyhttp/cors" at "0.1.20"
🦋  info Publishing "@tinyhttp/etag" at "0.1.20"
🦋  info Publishing "@tinyhttp/logger" at "0.1.20"
🦋  info Publishing "@tinyhttp/static" at "0.1.20"
🦋  error SyntaxError: Unexpected end of JSON input
🦋  error     at JSON.parse (<anonymous>)
🦋  error     at internalPublish (/home/v1rtl/Coding/tinyhttp/node_modules/.pnpm/@changesets/cli@2.9.1_@types+node@14.0.13/node_modules/@changesets/cli/dist/cli.cjs.dev.js:591:19)
🦋  error     at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Same happens with `./node_modules/.bin/changeset publish`

I don't know what causes it because there's no clear error message.

## Helpful info

reproduction repository: https://github.com/talentlessguy/tinyhttp
changesets config: https://github.com/talentlessguy/tinyhttp/blob/master/.changeset/config.json
`node` version: 14.4.0
`pnpm` version: 5.0.2
`npm` version: 6.14.5

ianwalter:
Fix #397

### DIFF
--- a/.changeset/three-bananas-smile.md
+++ b/.changeset/three-bananas-smile.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Fix #397: Unexpected end of JSON input issue
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -10,17 +10,6 @@ import { TwoFactorState } from "../../utils/types";
 
 const npmRequestLimit = pLimit(40);
 
-function jsonParse(input: string) {
-  try {
-    return JSON.parse(input);
-  } catch (err) {
-    if (err instanceof SyntaxError) {
-      console.error("error parsing json:", input);
-    }
-    throw err;
-  }
-}
-
 function getCorrectRegistry() {
   let registry =
     process.env.npm_config_registry === "https://registry.yarnpkg.com"
@@ -44,11 +33,16 @@ export async function getTokenIsRequired() {
   let result = await spawn("npm", ["profile", "get", "--json"], {
     env: Object.assign({}, process.env, envOverride)
   });
-  let json = jsonParse(result.stdout.toString());
-  if (json.error || !json.tfa || !json.tfa.mode) {
-    return false;
+  let json = result.stdout.toString();
+  let profile;
+  if (json) {
+    try {
+      profile = JSON.parse(json);
+    } catch (err) {
+      // Handling the error is unnecessary since profile will be falsy.
+    }
   }
-  return json.tfa.mode === "auth-and-writes";
+  return profile && profile.tfa && profile.tfa.mode === "auth-and-writes";
 }
 
 export function getPackageInfo(pkgName: string) {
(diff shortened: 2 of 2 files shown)


ianwalter:
I don't think this actually fixes the problem. I need to push another change.

ianwalter:
Ok, I have verified that this has fixed the problem for me.

Ref: https://github.com/ianwalter/nrg/runs/938126406?check_suite_focus=true

Andarist:
Do we understand why the issue happens in the first place? I have a strong feeling that just allowing some of those paths without being able to validate stuff properly is not good as it can lead to problems down the road in certain scenarios.

ianwalter:
The first issue is that the `getTokenIsRequired` function will throw an error because stdout is not just a json string and fail the process when you aren't authenticated with `npm login` as is the case in my workflow since I'm using `NPM_TOKEN`. This is fixed by adding the `--silent` flag to the profile command but also wrapping `JSON.parse` in a try catch is better since that function should only ever return a boolean on whether 2 factor is necessary, not just throw and not allow the program to handle that situation. We should keep other registries in mind that may not even implement 2 factor and could return unexpected output. I've seen a lot of issues around this while working on other related software like `sindresorhus/np`.

Adding the `--silent` flag in general is an improvement since errors will be printed with the JSON and cause stdout to not be able to be parsed without sanitation like what is done with the `publish` logic. In the case of the `publish` logic, adding a try catch to it directly is better so that we can see the original stdout string and not just the sanitized string that is passed to `JSON.parse` when an error is thrown.

So basically, I'm not 100% sure what the exact flow is that leads to the error, but I think these changes are improvements regardless since it allows the program to maintain control, e.g. returning a boolean or `{ published: false}` where expected, instead of just erroring out.

The changes solve the issue for me, but is there an easier way to test the cli module? Usually I would just point the package to my git fork, but this doesn't work here since it's a monorepo and the preconstruct compilation aspect makes it a little trickier as well.

PR #428, opened 2020-08-04
feature(cli): add support for publishing from a subdirectory using `publishConfig.directory`

dotansimha:
The PR resolves this issue: https://github.com/atlassian/changesets/issues/423

@Andarist


Andarist:
> If a different location needs to be published, then changesets version should be executed in order to update the original package.json with the new version number, and commit it.

Is this any different from what we do now? I fail to see the difference.

> In tools like lerna, it's possible to set publishConfig.directory and this tell the publish command that the actual directory is different.

Could you link to their docs about this feature? Are there any tools handling it in the same way?

.changeset/real-spoons-raise.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/cli": patch
```
Andarist, line 2:
```suggestion
"@changesets/cli": minor
```

.changeset/real-spoons-raise.md
```diff
@@ -0,0 +1,6 @@
...
+"@changesets/types": patch
```
Andarist, line 3: As to the `@changesets/types` - I would create a separate changeset file for it, with a different description. The types package doesn't care as much about this feature as it only defines a new field to be defined in `publishConfig`

packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -103,7 +104,10 @@ async function publishAPackage(
     `Publishing ${chalk.cyan(`"${name}"`)} at ${chalk.green(`"${version}"`)}`
   );
 
-  const publishDir = pkg.dir;
+  const publishDir =
+    publishConfig && publishConfig.directory
+      ? join(pkg.dir, publishConfig.directory)
+      : pkg.dir;
```
Andarist, line 110: nit: If you rebase on top of master then you should be able to use optional chaining here 🎉 
```suggestion
  const publishDir =
    publishConfig?.directory
      ? join(pkg.dir, publishConfig.directory)
      : pkg.dir;
```

dotansimha:
@Andarist The difference is where do you run the ` npm run publish`. 
Our setup has this after `changesets version` and `yarn build`: 

```
packages/
  - a/
     - package.json
     - dist/
         - compiled.js
         - package.json
     - src/...
```

So there is a temporary `package.json` under `dist`, and this is the one we would like to publish, not the source one. 


And here's a reference for that: https://www.npmjs.com/package/@lerna/publish#publishconfigdirectory

dotansimha:
@Andarist resolved all comments :)

Andarist:
> @Andarist The difference is where do you run the npm run publish.

Right, this matter for the publish alone though - it doesn't matter for the versioning step though. Maybe I've misunderstood your original comment somehow.

.changeset/tame-pumas-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/types": patch
+---
+
+Updated signature for `publishConfig` structure of package.json
```
Andarist, line 5:
```suggestion
Updated signature of `PackageJSON['publishConfig']` to include `directory?: string`.
```

PR #437, opened 2020-08-13
Fix version command not committing when commit config option is set

### LINKED ISSUE #424: `changeset version` no longer commits
## Affected Packages

`@changesets/cli@2.9.2`

## Problem

`changeset version` does not commit when `commit: true` is set in the changesets config

Blasz:
Fixes #424 and supersedes #408.

### DIFF
--- a/packages/cli/src/commands/version/version.test.ts
+++ b/packages/cli/src/commands/version/version.test.ts
@@ -194,6 +194,41 @@ describe("running version in a simple project", () => {
     );
   });
 
+  it("should not commit the result if commit config is not set", async () => {
+    await writeChangesets([simpleChangeset2], cwd);
+    const spy = jest.spyOn(git, "commit");
+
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/version/version.test.ts
```diff
@@ -194,6 +194,41 @@ describe("running version in a simple project", () => {
...
+    const spy = jest.spyOn(git, "commit");
+
+    expect(spy).not.toHaveBeenCalled();
+
+    await versionCommand(cwd, defaultOptions, modifiedDefaultConfig);
+
+    expect(spy).not.toHaveBeenCalled();
+  });
+
+  it("should not commit the result if commit config is not set", async () => {
```
mitchellhamilton, line 208:
```suggestion
  it("should commit the result if commit config is set", async () => {
```

PR #441, opened 2020-08-14
fix: remove workspace: prefix from dependent version ranges so they can be semver evaluated

### LINKED ISSUE #440: release plan is incorrectly bumping dependent versions when workspaces protocol is used
## Affected Packages

assemble-release-plan

## Problem

When I create a `none` type changeset for a package for chore commits, dependent packages are getting patch bumped even though the actual package is not being bumped. 

I'm noticing this specifically when using the workspace protocol ranges with pnpm.

## Proposed solution

Turns out it is just the upstream semver package not being able to evaluate it correctly.

```
> semver.satisfies('1.0.0' ,'workspace:^1.0.0')
false
> semver.satisfies('1.0.0' ,'^1.0.0')
true
```

joshlartz:
Fixes #440


packages/apply-release-plan/src/index.test.ts
```diff
@@ -368,6 +369,88 @@ describe("apply release plan", () => {
         }
       });
     });
+    it("should not update dependent versions when a package has a changeset type of none", async () => {
```
Andarist, line 372: q: what for do you use changesets with a none type? 🤔

joshlartz: We automate all versioning and releases during CI. To do that, we look for changed files and enforce that a changeset must be included for that package. That means to do chore commits, like updating docs, tests, comments, and other things that wouldn’t be a releasable feature worthy of the changelog, a none type changeset is required. 

Changesets already supports this pattern, but this covers an edge case I came across when workspace versions are used in package.json it would be evaluated incorrectly and force a bump of dependents.

joshlartz: Any issues with accepting this PR @Andarist ? It definitely fixes a problem that I run into.

.changeset/chatty-eagles-look.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": none
+"@changesets/assemble-release-plan": patch
+---
+
+fix: remove workspace: prefix from dependent version ranges so they can be semver evaluated
```
Andarist, line 6:
```suggestion
Fixed an issue with determining dependent updates for dependencies using `workspace:` protocol.
```

q: didn't it affect non-none updated as well somehow? although maybe then those dependents can only get `patch` assigned so maybe that's ok? one thing that comes to my mind is checking peer dep updates, is it possible to use `workspace:` for a peer dep? 🤔

joshlartz: Right. When using a non-none changeset, it patch bumps the dependents just fine. 

I’m honestly not sure. Using peer dependencies inside a monorepo doesn’t really make sense, so I wouldn’t worry about it.

Noviny:
@joshlartz I think I'm happy to merge this - just need to resolve the conflicts and I'll give it a ✅

joshlartz:
Sorry that I totally missed your comment @Noviny! I fixed the merge conflicts and worked out another bug with how workspace protocol dependencies were handled.

joshlartz:
@Noviny @Andarist Good to merge?

.changeset/chatty-eagles-look.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": none
```
Andarist, line 2: should the `apply-release-plan` be included in this changeset? no changes were made to it

packages/apply-release-plan/src/index.test.ts
```diff
@@ -496,6 +497,88 @@ describe("apply release plan", () => {
...
```
Andarist, line 500: Note to myself: this is a new test but it was already passing before this PR.

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -167,6 +167,26 @@ describe("assemble-release-plan", () => {
     expect(releases[2].oldVersion).toEqual("1.0.0");
     expect(releases[2].newVersion).toEqual("1.0.1");
   });
+  it("should assemble release plan without dependencies when the dependent has a changeset type of none", () => {
```
Andarist, line 170: same here - new tests, but it's already passing on our base branch

packages/apply-release-plan/src/index.test.ts
```diff
@@ -496,6 +497,88 @@ describe("apply release plan", () => {
...
+      if (!pkgPathB) throw new Error(`could not find an updated package json`);
+
+      let pkgJSONB = await fs.readJSON(pkgPathB);
+
+      expect(pkgJSONB).toMatchObject({
+        name: "pkg-b",
+        version: "1.0.0"
+      });
+    });
+    it("should not update workspace dependent versions when a package has a changeset type of none", async () => {
```
Andarist, line 541: this one as well - already passing on the base branch

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -597,6 +617,65 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
...
+        undefined
+      );
+
+      expect(releases.length).toEqual(2);
+      expect(releases[0].name).toEqual("pkg-a");
+      expect(releases[1].name).toEqual("pkg-b");
+      expect(releases[1].oldVersion).toEqual("1.0.0");
+      expect(releases[1].newVersion).toEqual("1.0.0");
+    });
+    it("should assemble release plan with workspace:* dependencies", () => {
```
Andarist, line 662: and this one as well - it's passing on the base branch

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -597,6 +617,65 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
...
+      });
+
+      let { releases } = assembleReleasePlan(
+        setup.changesets,
+        setup.packages,
+        defaultConfig,
+        undefined
+      );
+
+      expect(releases.length).toEqual(2);
```
Andarist, line 656: on the base branch a third release is generated:
```json
{
  "changesets": [],
  "name": "pkg-c",
  "newVersion": "1.0.1",
  "oldVersion": "1.0.0",
  "type": "patch"
}
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -597,6 +617,65 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
...
```
Andarist, line 636: on the base branch a third release is generated:

```json
{
  "changesets": [],
  "name": "pkg-c",
  "newVersion": "1.0.1",
  "oldVersion": "1.0.0",
  "type": "patch"
}
```

.changeset/chatty-eagles-look.md
```diff
@@ -0,0 +1,6 @@
...
+"@changesets/assemble-release-plan": patch
+---
+
+Fixed an issue with determining dependent updates for dependencies using `workspace:` protocol.
```
Andarist, line 6: To be more specific - shouldn't we only mention `none` changesets here? Were other cases fixed by this? If yes - no tests were added for other cases from what I can tell.

Andarist:
@joshlartz sorry that it took so long, this PR has totally slipped me. I'm back on track so we should be able to finalize this quickly now - I've left some comments regarding the changeset and the description provided by it, gonna be waiting on your reply.

.changeset/chatty-eagles-look.md
joshlartz: I made it a none type because I added a test to the package, but you're right that there's no functional changes. It's really just a force of habit from working in a changeset enforced environment. I'm not married to it at all and can remove it if you prefer.

.changeset/chatty-eagles-look.md
joshlartz: This technically affects any changeset type when it comes across a workspace protocol because the `semver` package can't evaluate it correctly. It's just most noticeable when a none type is used because no bump should be happening. Any other patch bump or higher would obscure this.

.changeset/chatty-eagles-look.md
joshlartz: I only added tests there because of how assemble and apply work together. In case their interactions ever change in the future, the tests should help show the issue on the apply side too.

Andarist: Sure thing, the added tests are great - I've just been commenting on each test so I could build a better picture for the scope of your changes

.changeset/chatty-eagles-look.md
Andarist: q: would it be possible for it to fail in a different way? for example by bumping with a patch when the minor bump should be done? I guess not as this logic is only concerned about adding dependents to releases and they are always added with a patch bump if there is no explicit release for them.

Andarist, line 6:
```suggestion
Fixed an issue with including dependents in the release plan for dependencies using `workspace:` protocol that had a `none` changeset for them.
```

Andarist (approved):

PR #444, opened 2020-08-16
Changed the way how requests to the GitHub API were authenticated

Andarist:
fixes https://github.com/changesets/action/issues/51

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules/typescript/lib"
+}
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules/typescript/lib"
```
mitchellhamilton, line 2: Why this change?

calebeby:
Thank you!

.vscode/settings.json
Andarist: Oh, I wasn't supposed to add this to this commit - it's a local change that I'm using. I had to accidentally add it.

It is good though - by default VSCode is using its builtin TS version and this often mismatches with the one actually used by the project itself. This makes it to use a local version - which should always be preferred.

PR #446, opened 2020-08-17
Fixed an issue with dependant packages not being updated to their highest bump type in pre mode

Andarist:
fixes https://github.com/atlassian/changesets/issues/442


packages/assemble-release-plan/src/index.ts
```diff
@@ -92,9 +92,10 @@ function assembleReleasePlan(
     snapshotSuffix = getSnapshotSuffix(snapshot);
   }
 
-  let packagesByName = new Map(
+  let initialPackagesByName = new Map(
```
Andarist, line 95: Not the best name for this, but I don't know how to name this better. In general, I've spent a surprisingly long time fixing this issue (tracking it down alone was fairly quick and easy) - there is **a lot** happening here and some structures are faking data from before entering pre mode which makes this fairly hard to reason about. I would say this implementation is a good candidate for a future refactoring, but would definitely have to be done with care.

packages/cli/src/commands/version/version.test.ts
```diff
@@ -886,6 +886,114 @@ describe("pre", () => {
       }
     ]);
   });
+  it("should use the highest bump type between all prereleases when versioning a dependant package", async () => {
```
Andarist, line 889: not sure if this makes it super obvious what this test is testing for - not sure how to best call a process of versioning because of a dependency upgrade. Both of those 2 newly added tests cover what got described in a changeset - a situation when we need to upgrade a dependant package only because a dependency got upgraded but when that dependant had already a non-patch bump in the current pre mode.

Andarist:
friendly 🏓 @mitchellhamilton @Noviny , I believe this has just caused a problem with the recent Emotion prerelease as well ( https://github.com/emotion-js/emotion/pull/1982 )

mitchellhamilton (approved):

PR #448, opened 2020-08-22
Publish packages from cwd

vlad-zhukov:
Currently `<npm|yarn|pnpm> publish` command runs inside each package directory and doesn't use root `.npmrc` file. In our case we publish packages to a self-hosted registry that requires certain configs, and because of that publishing fails.


Noviny:
This change seems fine - would love to have documentation added for it though before merging.

kamleshchandnani:
Is there any update when this will be merged? 
@vlad-zhukov @Noviny I can help with the documentation if required. What are you expecting to document with respect to this issue?

Andarist:
@Noviny this isn't a configurable behavior right now, so I'm wondering if this needs any documentation. Looking at the `npm publish` docs etc this seems like a non-breaking change.

Noviny:
Yep, all good - feel free to merge and publish.

.changeset/new-mirrors-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Publish packages from cwd
```
Andarist, line 5:
```suggestion
Packages are now published from cwd (usually the root of the repository) rather than from the package directories. This respects `.npmrc` files put in the root directory.
```

Andarist (approved):

kamleshchandnani:
@vlad-zhukov Even after using this latest version my root `.npmrc` isn't respected, it just picks up the global `.npmrc` 🤔

Andarist:
@kamleshchandnani it would be great if you could prepare a runnable repro case of the problem, with clear instructions on how to reproduce this. Or even better - add a failing test case for this 😉

kamleshchandnani:
> @kamleshchandnani it would be great if you could prepare a runnable repro case of the problem, with clear instructions on how to reproduce this. Or even better - add a failing test case for this 😉

Okay so here's [the repo](https://github.com/kamleshchandnani/changesets-demo) where I was trying to reproduce the issue. Turned out that the error was something else in this case it was my registry mis typed but the reason I got confused was that my eyes just got stuck at [this line](https://github.com/kamleshchandnani/changesets-demo/runs/2493113514?check_suite_focus=true#step:5:16) which says 
> No `.npmrc` file found, creating one

and I got confused that it isn't picking up my `.npmrc` even though it is. I think the messaging can be bit more contextual if it finds my local `.npmrc`. So maybe we can change the messaging a bit? 

Also, now I think this isn't related to this issue. So do let me know if this makes sense I can open a new issue for changing the messaging :)

Andarist:
If the local `.npmrc` takes precedence over a global one then ye - definitely this could be improved, the imperfect logic is here:
https://github.com/changesets/action/blob/c2918239208f2162b9d27a87f491375c51592434/src/index.ts#L44-L53

I would appreciate it if you could report this here: https://github.com/changesets/action/issues with the context about the problem that you have encountered.

PR #453, opened 2020-09-10
Use packagesByName over packagesByNameWithVersionsFromBeforePreModeIfInPreMode - unless it is required to use the latter

Andarist:
I believe this makes more sense because it's less surprising when debugging stuff.

### DIFF
--- a/packages/assemble-release-plan/src/index.ts
+++ b/packages/assemble-release-plan/src/index.ts
@@ -129,11 +129,7 @@ function assembleReleasePlan(
       let highestPreVersion = 0;
       for (let linkedPackage of linkedGroup) {
         highestPreVersion = Math.max(
-          getPreVersion(
-            packagesByNameWithVersionsFromBeforePreModeIfInPreMode.get(
-              linkedPackage
-            )!.packageJson.version
-          ),
+          getPreVersion(packagesByName.get(linkedPackage)!.packageJson.version),
           highestPreVersion
         );
       }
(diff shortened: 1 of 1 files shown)


packages/assemble-release-plan/src/index.ts
```diff
@@ -159,11 +155,7 @@ function assembleReleasePlan(
   // releases is, at this point a list of all packages we are going to releases,
   // flattened down to one release per package, having a reference back to their
   // changesets, and with a calculated new versions
-  let releases = flattenReleases(
-    changesets,
-    packagesByNameWithVersionsFromBeforePreModeIfInPreMode,
-    config.ignore
-  );
+  let releases = flattenReleases(changesets, packagesByName, config.ignore);
```
mitchellhamilton, line 158: I don't think the change here is right and it worries me that no tests have failed.

Andarist: So this change affects `release.oldVersion` - I'm not sure if this is incorrect, I guess some code could depend on this being an "initial version" but I've tried a bunch of things and haven't found a test that would validate that the former logic was correct. It also might be really surprising that `oldVersion` points to the "initial version" rather than to the version before the current release. That "initial version" IMHO should only be relevant for this:
https://github.com/atlassian/changesets/blob/66a78a2b84785e3764fce765aa063f02c0ee222b/packages/assemble-release-plan/src/index.ts#L208-L235

Note: this logic:
https://github.com/atlassian/changesets/blob/66a78a2b84785e3764fce765aa063f02c0ee222b/packages/assemble-release-plan/src/index.ts#L202
is using the same source for `oldVersion` as the new logic here.

PR #454, opened 2020-09-11
Remove packagesByNameWithVersionsFromBeforePreModeIfInPreMode and releasesFromUnfilteredChangesets

Andarist:
cc @mitchellhamilton 😱 

~it's super weird because I would totally expect for tests added here to fail: https://github.com/atlassian/changesets/pull/263 , need to investigate why they pass~ they pass because we've removed both of those variables and it's able to compute good new version based on the given information

### DIFF
--- a/.changeset/gorgeous-weeks-begin.md
+++ b/.changeset/gorgeous-weeks-begin.md
@@ -1,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": major
+---
+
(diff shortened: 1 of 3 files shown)


Andarist:
This simplifies the code quite a bit and makes it easier to understand overall. There is only a small behavior change for `@changesets/assemble-release-plan` which after discussing with @mitchellhamilton is acceptable if we make this a major bump.

mitchellhamilton (approved):

PR #458, opened 2020-09-17
feat(config): allow to provide glob expressions for linked and ignore packages

### LINKED ISSUE #364: Allow to pass glob or regex to linked config
## Problem

When using the `linked` config option, it can get quite verbose to write and keep track of all the packages in the repo.

This becomes obvious when you want to "simulated" the "locked" mode of Lerna, where all packages follow more or less a common version (as suggested here https://github.com/atlassian/changesets/issues/356#issuecomment-625203441).

## Proposed solution

One thing that would help is to allow to pass a regex or a glob string instead of listing all packages explicitly.

For example:

```json
{
  "linked": [
    ["@changesets/*"]
  ]
}
```

Or even introduce a new option that automatically link all packages. For example:

```json
{
  "linkAllPackages": true
}
```

It would be enough to have a glob/regex, which is a bit more flexible.

What do you think?

emmenko:
Closes #364

### DIFF
--- a/docs/config-file-options.md
+++ b/docs/config-file-options.md
@@ -49,7 +49,7 @@ There are two caveats to this.
 
 These restrictions exist to ensure your repository or published code do not end up in a broken state. For a more detailed intricacies of publishing, check out our guide on [problems publishing in monorepos](./problems-publishing-in-monorepos).
 
-> NOTE: you can also provide glob expressions to match the packages, according to the [minimatch](https://www.npmjs.com/package/minimatch) format.
+> NOTE: you can also provide glob expressions to match the packages, according to the [micromatch](https://www.npmjs.com/package/micromatch) format.
 
 ## `linked` (array of package names)
 
--- a/docs/config-file-options.md
+++ b/docs/config-file-options.md
@@ -49,6 +49,8 @@ There are two caveats to this.
 
 These restrictions exist to ensure your repository or published code do not end up in a broken state. For a more detailed intricacies of publishing, check out our guide on [problems publishing in monorepos](./problems-publishing-in-monorepos).
 
+> NOTE: you can also provide glob expressions to match the packages, according to the [minimatch](https://www.npmjs.com/package/minimatch) format.
+
 ## `linked` (array of package names)
 
 This option can be used to declare that packages should 'share' a version, instead of being versioned completely independently. As an example, if you have a `@changesets/button` component and a `@changesets/theme` component and you want to make sure that when one gets bumped to `2.0.0`, the other is also bumped to `2.0.0`. To achieve this you would have the config:
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -16,7 +16,8 @@
     "@changesets/logger": "^0.0.5",
     "@changesets/types": "^3.1.0",
     "@manypkg/get-packages": "^1.0.1",
-    "fs-extra": "^7.0.1"
+    "fs-extra": "^7.0.1",
+    "minimatch": "^3.0.4"
   },
   "devDependencies": {
     "fixturez": "^1.1.0",
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -16,11 +16,11 @@
     "@changesets/logger": "^0.0.5",
     "@changesets/types": "^3.1.0",
     "@manypkg/get-packages": "^1.0.1",
-    "@types/micromatch": "^4.0.1",
     "fs-extra": "^7.0.1",
     "micromatch": "^4.0.2"
   },
   "devDependencies": {
+    "@types/micromatch": "^4.0.1",
     "fixturez": "^1.1.0",
     "jest-in-case": "^1.0.2"
   }
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -16,8 +16,9 @@
     "@changesets/logger": "^0.0.5",
     "@changesets/types": "^3.1.0",
     "@manypkg/get-packages": "^1.0.1",
+    "@types/micromatch": "^4.0.1",
     "fs-extra": "^7.0.1",
-    "minimatch": "^3.0.4"
+    "micromatch": "^4.0.2"
   },
   "devDependencies": {
     "fixturez": "^1.1.0",
--- a/packages/config/src/index.test.ts
+++ b/packages/config/src/index.test.ts
@@ -2,6 +2,7 @@ import fixturez from "fixturez";
 import { read, parse } from "./";
 import jestInCase from "jest-in-case";
 import * as logger from "@changesets/logger";
+import { Config, WrittenConfig } from "@changesets/types";
 import { Packages } from "@manypkg/get-packages";
 
 jest.mock("@changesets/logger");
@@ -16,6 +17,14 @@ let defaultPackages: Packages = {
   packages: [],
   tool: "yarn"
 };
+
+const withPackages = (pkgNames: string[]) => ({
+  ...defaultPackages,
+  packages: pkgNames.map(pkgName => ({
+    packageJson: { name: pkgName, version: "" },
+    dir: "dir"
+  }))
+});
 
 test("read reads the config", async () => {
   let dir = f.find("new-config");
@@ -49,14 +58,20 @@ let defaults = {
   }
 } as const;
 
-let correctCases = {
+let correctCases: {
+  [caseName: string]: {
+    packages?: string[];
+    input: WrittenConfig;
+    output: Config;
+  };
+} = {
   defaults: {
     input: {},
     output: defaults
   },
   "changelog string": {
     input: {
-      changelog: "some-module"
+      changelog: ["some-module", null]
     },
     output: {
       ...defaults,
@@ -126,6 +141,23 @@ let correctCases = {
       linked: [["pkg-a", "pkg-b"]]
     }
   },
+  linkedWithGlobs: {
+    packages: [
+      "pkg-a",
+      "pkg-b",
+      "@pkg/a",
+      "@pkg/b",
+      "@pkg-other/a",
+      "@pkg-other/b"
+    ],
+    input: {
+      linked: [["pkg-*", "@pkg/*"], ["@pkg-other/a"]]
+    },
+    output: {
+      ...defaults,
+      linked: [["pkg-a", "pkg-b", "@pkg/a", "@pkg/b"], ["@pkg-other/a"]]
+    }
+  },
   "update internal dependencies minor": {
     input: {
       updateInternalDependencies: "minor"
@@ -152,26 +184,27 @@ let correctCases = {
       ...defaults,
       ignore: ["pkg-a", "pkg-b"]
     }
+  },
+  ignoreWithGlobs: {
+    packages: ["pkg-a", "pkg-b", "@pkg/a", "@pkg/b"],
+    input: {
+      ignore: ["pkg-*", "@pkg/*"]
+    },
+    output: {
+      ...defaults,
+      ignore: ["pkg-a", "pkg-b", "@pkg/a", "@pkg/b"]
+    }
   }
-} as const;
+};
 
 jestInCase(
   "parse",
   testCase => {
     expect(
-      parse(testCase.input, {
-        ...defaultPackages,
-        packages: [
-          {
-            packageJson: { name: "pkg-a", version: "" },
-            dir: "dir"
-          },
-          {
-            packageJson: { name: "pkg-b", version: "" },
-            dir: "dir"
-          }
-        ]
-      })
+      parse(
+        testCase.input,
+        withPackages(testCase.packages || ["pkg-a", "pkg-b"])
+      )
     ).toEqual(testCase.output);
   },
   correctCases
@@ -182,7 +215,7 @@ let unsafeParse = parse as any;
 describe("parser errors", () => {
   test("changelog invalid value", () => {
     expect(() => {
-      unsafeParse({ changelog: {} });
+      unsafeParse({ changelog: {} }, defaultPackages);
     }).toThrowErrorMatchingInlineSnapshot(`
 "Some errors occurred when validating the changesets config:
 The \`changelog\` option is set as {} when the only valid values are undefined, a module path(e.g. \\"@changesets/cli/changelog\\" or \\"./some-module\\") or a tuple with a module path and config for the changelog generator(e.g. [\\"@changesets/cli/changelog\\", { someOption: true }])"
@@ -190,7 +223,10 @@ The \`changelog\` option is set as {} when the only valid values are undefined, 
   });
   test("changelog array with 3 values", () => {
     expect(() => {
-      unsafeParse({ changelog: ["some-module", "something", "other"] });
+      unsafeParse(
+        { changelog: ["some-module", "something", "other"] },
+        defaultPackages
+      );
     }).toThrowErrorMatchingInlineSnapshot(`
 "Some errors occurred when validating the changesets config:
 The \`changelog\` option is set as [
@@ -202,7 +238,7 @@ The \`changelog\` option is set as [
   });
   test("changelog array with first value not string", () => {
     expect(() => {
-      unsafeParse({ changelog: [false, "something"] });
+      unsafeParse({ changelog: [false, "something"] }, defaultPackages);
     }).toThrowErrorMatchingInlineSnapshot(`
 "Some errors occurred when validating the changesets config:
 The \`changelog\` option is set as [
@@ -213,7 +249,7 @@ The \`changelog\` option is set as [
   });
   test("access other string", () => {
     expect(() => {
-      unsafeParse({ access: "something" });
+      unsafeParse({ access: "something" }, defaultPackages);
     }).toThrowErrorMatchingInlineSnapshot(`
 "Some errors occurred when validating the changesets config:
 The \`access\` option is set as \\"something\\" when the only valid values are undefined, \\"public\\" or \\"restricted\\""
@@ -221,7 +257,7 @@ The \`access\` option is set as \\"something\\" when the only valid values are u
   });
   test("commit non-boolean", () => {
     expect(() => {
-      unsafeParse({ commit: "something" });
+      unsafeParse({ commit: "something" }, defaultPackages);
     }).toThrowErrorMatchingInlineSnapshot(`
 "Some errors occurred when validating the changesets config:
 The \`commit\` option is set as \\"something\\" when the only valid values are undefined or a boolean"
@@ -229,7 +265,7 @@ The \`commit\` option is set as \\"something\\" when the only valid values are u
   });
   test("linked non-array", () => {
     expect(() => {
-      unsafeParse({ linked: {} });
+      unsafeParse({ linked: {} }, defaultPackages);
     }).toThrowErrorMatchingInlineSnapshot(`
 "Some errors occurred when validating the changesets config:
 The \`linked\` option is set as {} when the only valid values are undefined or an array of arrays of package names"
@@ -237,7 +273,7 @@ The \`linked\` option is set as {} when the only valid values are undefined or a
   });
   test("linked array of non array", () => {
     expect(() => {
-      unsafeParse({ linked: [{}] });
+      unsafeParse({ linked: [{}] }, defaultPackages);
     }).toThrowErrorMatchingInlineSnapshot(`
 "Some errors occurred when validating the changesets config:
 The \`linked\` option is set as [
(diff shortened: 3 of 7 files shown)


packages/config/src/index.test.ts
```diff
@@ -17,6 +18,14 @@ let defaultPackages: Packages = {
   tool: "yarn"
 };
 
+const withPackages = (pkgNames: string[]) => ({
```
emmenko, line 21: Helper function for the tests

packages/config/src/index.ts
```diff
@@ -30,13 +31,38 @@ function getNormalisedChangelogOption(
...
+        // If there are no matching packages as a result of the minimatch filter,
+        // it is most likely the case when a package defined in the linked config
+        // does not exist.
+        // To be able to show a correct validation message, we pass the value as-is.
+        ...(matchingPackages.length > 0 ? matchingPackages : [pkgNameOrGlob])
+      ];
+    },
+    []
+  );
+}
```
emmenko, line 53: This is the main function to resolve the globs. It uses `minimatch.filter` (https://www.npmjs.com/package/minimatch#minimatchfilterpattern-options) to resolve the package name and spread it to the list.

packages/config/src/index.ts
```diff
@@ -114,16 +140,17 @@ export let parse = (json: WrittenConfig, packages: Packages): Config => {
...
       let foundPkgNames = new Set<string>();
       let duplicatedPkgNames = new Set<string>();
-      for (let linkedGroup of json.linked) {
-        for (let linkedPkgName of linkedGroup) {
-          if (!pkgNames.has(linkedPkgName)) {
+      for (let linkedGroup of json.linked as Linked) {
+        let normalizedLinkedGroup = normalizePackageNames(
+          linkedGroup,
+          pkgNames
+        );
```
emmenko, line 149: We normalize the list before doing the validation. This allows the retain the same logic as before.

packages/config/src/index.ts
```diff
@@ -168,13 +195,11 @@ export let parse = (json: WrittenConfig, packages: Packages): Config => {
         )} when the only valid values are undefined or an array of package names`
       );
     } else {
-      let pkgNames = new Set(
-        packages.packages.map(({ packageJson }) => packageJson.name)
-      );
-      for (let pkgName of json.ignore) {
-        if (!pkgNames.has(pkgName)) {
+      let normalizedIgnore = normalizePackageNames(json.ignore, pkgNames);
```
emmenko, line 198: I thought we can allow globs also for the `ignore` list.

mitchellhamilton:
@Noviny @Andarist what are your opinions on globs vs regex for this?

My thoughts:

globs:
- the inconsistency of what a "glob" is the worst
- they're kinda easier to write than regex, especially when in a string literal
- they're what people are used to writing in most tools(e.g. workspaces config)
- all valid npm packages names will only ever match that package name

regex:
- regex is consistent(or rather consistent in JS)
- The whole double escaping thing would get ANNOYING
- **v important point**:  A valid npm package name could match something that isn't the package name(e.g. you have a package named `something.another.thing`, will match `something-something.another.thing-whatever-else`, `something1another1thing` and `somethingaanotherathing` and etc. To make it only match `something.another.thing`, you would have to write `"^something\\.another\\.thing$"`). This would presumably mean we would need a syntax for "I want to do regex, not exact package name matching", probs `/regex here/`? so it would be `"/^something\\.another\\.thing$/"`
- You'd probably always want to write `^my-package$` rather than `my-package` so that it doesn't match `my-package-whatever` which is just ughhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

I feel like I'm leaning towards globs but wouldn't mind hearing other people's thoughts

packages/config/package.json
```diff
@@ -16,7 +16,8 @@
     "@changesets/logger": "^0.0.5",
     "@changesets/types": "^3.1.0",
     "@manypkg/get-packages": "^1.0.1",
-    "fs-extra": "^7.0.1"
+    "fs-extra": "^7.0.1",
+    "minimatch": "^3.0.4"
```
mitchellhamilton, line 20: Can you use `micromatch` instead? (micromatch is what fast-glob uses which globby uses which is what @manypkg/get-packages uses which is what changesets uses for getting the packages in a monorepo so it's good to use the same thing for consistency and not spending the time loading two glob implementations when people use Changesets)

(this comment is assuming we go with globs)

packages/config/src/index.test.ts
```diff
@@ -49,14 +58,20 @@ let defaults = {
...
+  };
+} = {
   defaults: {
     input: {},
     output: defaults
   },
   "changelog string": {
     input: {
-      changelog: "some-module"
+      changelog: ["some-module", null]
```
mitchellhamilton, line 74: This case is about handing the string only case, why has it been changed?

packages/config/package.json
emmenko: Damn, I thought I read "minimatch" from your comment, but it was "micromatch" instead. My bad sorry, I'll change that.

packages/config/src/index.test.ts
emmenko: Good catch, I shouldn't have commited this change. Will revert sorry

packages/config/package.json
emmenko: Done https://github.com/atlassian/changesets/pull/458/commits/2f0af89508d35db2a86d623009b49202d80f5ea8

packages/config/src/index.ts
```diff
@@ -30,13 +31,40 @@ function getNormalisedChangelogOption(
...
 
+function normalizePackageNames(
+  listOfPackageNamesOrGlob: readonly string[],
+  pkgNames: readonly string[]
+): string[] {
+  // Resolve and expand possible glob esxpressions.
+  return listOfPackageNamesOrGlob.reduce<string[]>(
+    (allResolvedPackageNames, pkgNameOrGlob) => {
+      const matchingPackages = pkgNames.filter(pkgName =>
+        micromatch.isMatch(pkgName, pkgNameOrGlob)
```
mitchellhamilton, line 42: I don't this would work for configs like `["**", "!pkg-a"]`?

Noviny:
I think I'd favour globs

Andarist:
While I hate globs and I've vented about it in another thread - I think they make more sense here, especially given serializability issues.

packages/config/src/index.ts
emmenko: I looked into that. It's a bit tricky and can become ambiguous and a bit confusing to what the expectation results are.

For example, given these packages:

```
pkg-a
pkg-b
@pkg/a
@pkg/b
@pkg-other/a
@pkg-other/b
```

and this linked config:

```js
{
  linked: [["pkg-*", "!pkg-b", "@pkg/*"]]
}
```

When I read this as a user, I can infer that I want all packages that start with `pkg-` but not `pkg-b`, plus all packages that start with `@pkg/`.
However, evaluating `!pkg-b` by itself basically means "all packages that are not pkg-b", which technically would mean to include some packages that weren't meant to be included like `@pkg-other/a` and `@pkg-other/b`.
Also the order of the linked packages might be important. For instance `["pkg-*", "!pkg-b"]` could be interpreted differently than `["!pkg-b", "pkg-*"]`.

I guess we could find and settle on a reasonable behavior but I fear that this might cause more confusion than ever.

---

FYI: with the current implementation, using a config like

```js
linked: [["pkg-*", "!pkg-b", "@pkg/*"], ["@pkg-other/a"]]
```

yields some validation errors:

```
ValidationError: Some errors occurred when validating the changesets config:
    The package "pkg-a" is defined in multiple sets of linked packages. Packages can only be defined in a single set of linked packages. If you are using glob expressions, make sure that they are valid according to https://www.npmjs.com/package/micromatch.
    The package "@pkg/a" is defined in multiple sets of linked packages. Packages can only be defined in a single set of linked packages. If you are using glob expressions, make sure that they are valid according to https://www.npmjs.com/package/micromatch.
    The package "@pkg/b" is defined in multiple sets of linked packages. Packages can only be defined in a single set of linked packages. If you are using glob expressions, make sure that they are valid according to https://www.npmjs.com/package/micromatch.
    The package "@pkg-other/a" is defined in multiple sets of linked packages. Packages can only be defined in a single set of linked packages. If you are using glob expressions, make sure that they are valid according to https://www.npmjs.com/package/micromatch.
```

Andarist: How such groups with multiple patterns are handled by other tools? Could we take a look at prior work done by others?

emmenko: I can't think of other examples of this, besides `.gitignore` and such.

Maybe we should be more strict and allow either/or and not a mix. If you provide a list, it should be a list of package names. No globs or regex.
If you want to provide a glob or regex, you provide it as a single value, not a list.

For example:

```js
linked: [["pkg-a", "pkg-c", "@pkg/a", "@pkg/b"]]
```

is equivalent of

```js
linked: [/(pkg\-[^b])|(@pkg\/\w)/]
```

image

Building regular expressions can be tricky though, especially for advanced used cases.

---

Another option I can think of is to provide a named configuration object where you explicitly define which packages should be included and which should be excluded.
Then you can use globs again.

For example:

```js
linked: [["pkg-a", "pkg-c", "@pkg/a", "@pkg/b"]]
```

is equivalent of

```js
linked: [
  {
    include: ['pkg-*', '@pkg/*'],
    exclude: ['pkg-b', '@pkg-others/*']
  }
]
```

The `exclude` list will always be resolved and applied lastly.

Andarist: One place that comes to my mind that we could look into how they have solved are workspace settings in yarn/lerna. Those are glob-based as well

emmenko: > workspace settings in yarn/lerna

Yarn and lerna (as far as I know) support globs yes but I have never seen an example for matching some packages but ignore others.

In lerna for ignoring packages they have a specific option, for example:

https://github.com/lerna/lerna/blob/c0a750e0f482c16dda2f922f235861283efbe94d/commands/list/__tests__/list-command.test.js#L353-L356

I suppose this is similar to my last proposal:

```js
linked: [
  {
    include: ['pkg-*', '@pkg/*'],
    exclude: ['pkg-b', '@pkg-others/*']
  }
]
```

Again, this is only if we want to support globs and additionally the ability to ignore certain packages.

Let me know what you think.

Andarist: > In lerna for ignoring packages they have a specific option, for example:
> https://github.com/lerna/lerna/blob/c0a750e0f482c16dda2f922f235861283efbe94d/commands/list/__tests__/list-command.test.js#L353-L356

Right, this is somewhat similar - but is for CLI args. I was asking more about how workspaces config gets interpreted. It's acquired from the config [here](https://github.com/lerna/lerna/blob/f6e7a13e60fefc523d701efddfcf0ed41a77749b/core/project/index.js#L73). This gets later resolved by [`fileFinder`](https://github.com/lerna/lerna/blob/f6e7a13e60fefc523d701efddfcf0ed41a77749b/core/project/lib/make-file-finder.js#L39) - it seems that the results of those globs are just concatenated. So I would assume that this just works like the implementation proposed by you - would be great to confirm this though.

It this is true - and you can't easily mix inclusive and exclusive patterns there then I think it's acceptable for us to do the same and, if needed, implement the proposed solution with `{ include, exclude }` object.

emmenko: Yep. Btw, Yarn also seems [to use](https://github.com/yarnpkg/yarn/search?q=micromatch) `micromatch`.

For example https://github.com/yarnpkg/yarn/blob/cbeead63564f1e1a9bdebbfec7f1fafde04f8c91/src/config.js#L785

taras:
👋 how is this going? Can I help?

emmenko:
I think we're waiting for a decision around the behavior of inclusive/exclusive patterns.

@Andarist @mitchellhamilton @Noviny any update from your side?

packages/config/src/index.ts
```diff
@@ -30,13 +31,40 @@ function getNormalisedChangelogOption(
...
+      );
+      return [
+        ...allResolvedPackageNames,
+        // If there are no matching packages as a result of the micromatch filter,
```
Andarist, line 46: if we have this knowledge here about a particular thing not being matched why do we need to fake a return value so it can be handled later? I mean - an indirection like this complicates the overall flow of the program. Couldn't we just return an error from here so it can be aggregated by the caller?

packages/config/src/index.ts
```diff
@@ -244,7 +271,11 @@ export let parse = (json: WrittenConfig, packages: Packages): Config => {
     commit:
       json.commit === undefined ? defaultWrittenConfig.commit : json.commit,
     linked:
-      json.linked === undefined ? defaultWrittenConfig.linked : json.linked,
+      json.linked === undefined
+        ? defaultWrittenConfig.linked
+        : (json.linked as Linked).map(linkedGroup =>
```
Andarist, line 276: couldnt `WrittenConfig` be adjusted to avoid casting here?

docs/config-file-options.md
```diff
@@ -49,6 +49,8 @@ There are two caveats to this.
 
 These restrictions exist to ensure your repository or published code do not end up in a broken state. For a more detailed intricacies of publishing, check out our guide on [problems publishing in monorepos](./problems-publishing-in-monorepos).
 
+> NOTE: you can also provide glob expressions to match the packages, according to the [micromatch](https://www.npmjs.com/package/micromatch) format.
```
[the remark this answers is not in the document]
emmenko: Since the `linked-packages` page is linked from the `linked` option below, there is a more prominent section there "Using glob expressions".
Wouldn't this be enough?

packages/config/package.json
```diff
@@ -16,7 +16,9 @@
     "@changesets/logger": "^0.0.5",
     "@changesets/types": "^3.1.0",
     "@manypkg/get-packages": "^1.0.1",
-    "fs-extra": "^7.0.1"
+    "@types/micromatch": "^4.0.1",
```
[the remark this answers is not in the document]
emmenko: Probably not. Should we put it as a dev dependency?

packages/config/src/index.test.ts
```diff
@@ -49,7 +58,13 @@ let defaults = {
   }
 } as const;
 
-let correctCases = {
+let correctCases: {
```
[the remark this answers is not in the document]
emmenko: You mean to extract the type object `type CorrectCases`?

packages/config/src/index.test.ts
```diff
@@ -202,7 +238,7 @@ The \`changelog\` option is set as [
   });
   test("changelog array with first value not string", () => {
     expect(() => {
-      unsafeParse({ changelog: [false, "something"] });
+      unsafeParse({ changelog: [false, "something"] }, defaultPackages);
```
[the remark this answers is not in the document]
emmenko: Because the second argument `packages` of the `parse` function is a required argument and is now used to map the package names.

https://github.com/atlassian/changesets/blob/2f0af89508d35db2a86d623009b49202d80f5ea8/packages/config/src/index.ts#L62-L66

packages/config/src/index.ts
emmenko: I didn't want to touch/change the existing validation logic, which is not really a scope of this PR.

But yeah, if you guys decide to refactor the validation flow, we can then throw here.

packages/config/src/index.ts
emmenko: Without the casting I get:

```
This expression is not callable.
  Each member of the union type '(<U>(callbackfn: (value: any, index: number, array: any[]) => U, thisArg?: any) => U[]) | (<U>(callbackfn: (value: readonly string[], index: number, array: readonly (readonly string[])[]) => U, thisArg?: any) => U[])' has signatures, but none of those signatures are compatible with each other.ts(2349)
```

image

Do you have a suggestion what we should do in this case?

packages/config/package.json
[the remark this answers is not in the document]
Andarist: that would be IMHO preferrable - unless the compiled `.d.ts` files reference `@types/micromatch` anyhow

docs/config-file-options.md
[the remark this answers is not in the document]
Andarist: that should be enough 👍

packages/config/src/index.test.ts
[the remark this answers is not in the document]
Andarist: yep, you could create `CorrectCase` and declare `correctCases` to be of type `Record<string, CorrectCase>`

packages/config/src/index.test.ts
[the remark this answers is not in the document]
Andarist: ok, gotcha - an alternative way to approach this would be to extend `unsafeParse` so that it would provide this argument if not given explicitly to the actual `parse`

 im fine either way, so do whatever you find better

packages/config/src/index.ts
Andarist: I was rather proposing that more explicit information could be returned from here - for example we could use a tuple return type to achieve that and that returned error message would be just pushed onto `messages` in the caller function. This is how all validations are handled right now and I believe we should stick to that. I understand this is sort-of doing that right now as well but in a way less explicit manner. Basically what I usually try to follow is to produce error messages as soon as they arise as most of the time you get the best context for a given error message there - even if you sometimes need to propagate such error through multiple layers up the stack so it can finally be reported somewhere.

If you do not have enough time to refactor this bit I totally understand, but if you do then I would really appreciate a change.

emmenko: Yes I agree and get your point. I would propose to tackle this as a follow up PR though, to not mix the concerns here.

At the end, what this function is doing is just extending the list of package names as-if a user entered those manually. So the validation does not need to change in this case.

packages/config/src/index.ts
Andarist: Interesting - I don't understand why this doesn't work as both of those should be arrays so how they don't have a `map` method? either way - `any[]` should probably not be included here, it somehow gets there after this condition: https://github.com/atlassian/changesets/pull/458/files#diff-679e3be50a6d82a2f7db9e0112f63992L99 . I think it might not be related to your PR, it might have just manifested itself right now.

It's some interesting TS problem - I will dig into this later.

packages/config/src/index.ts
mitchellhamilton: I'm not sure what the issue here is, the following is what micromatch does and is what we want:
```js
micromatch(
  ["pkg-a", "pkg-b", "pkg-c", "@pkg/a", "@pkg/b", "@pkg/c", "something"],
  ["pkg-*", "!pkg-b", "@pkg/*"]
)
// ["pkg-a", "pkg-c", "@pkg/a", "@pkg/b", "@pkg/c"]
```

emmenko: @mitchellhamilton thanks, yes you're right. We can simply do it like that, I think I got confused with the switch from minimatch to micromatch. My bad 🤦‍♂️

I'll push a commit with the fix.

emmenko: @mitchellhamilton @Andarist I pushed a commit with the simplified matching logic. https://github.com/atlassian/changesets/pull/458/commits/b2af53d1ac489100e92ca98aad5962e166f4fd49

Note that for keeping the current validation logic, we're returning a tuple from the `normalizePackageNames` function. The second argument is a list of packages that didn't match.
As a follow up, as suggested here https://github.com/atlassian/changesets/pull/458#discussion_r499836749, we can look into refactoring the validation logic.

packages/config/src/index.ts
Andarist: I've pushed out onto your branch a fix for this. It's a TS quirk that has been fixed in 4.1 but it seems that the fix is also causing some problems with generics: https://github.com/microsoft/TypeScript/issues/40463#issuecomment-690379467 , so we'll have to see if it even gets to 4.1

emmenko: Thanks

packages/config/src/index.ts
Andarist: From what I see in Lerna's code this doesn't conform to their logic for resolving packages, see code pointers in this post: https://github.com/atlassian/changesets/pull/458#discussion_r495765394 . Nor it does conform to the same thing in yarn: https://github.com/yarnpkg/yarn/blob/cbeead63564f1e1a9bdebbfec7f1fafde04f8c91/src/config.js#L819-L826 

Are we fine with this?

mitchellhamilton (approved):
Thanks!!

packages/config/src/index.ts
mitchellhamilton: Yeah, this is fine IMO

PR #460, opened 2020-09-22
Add `doNotThank` option to `@changeset/changelog-github`

BeeeQueue:
Takes a boolean (to disable thanking) or an array of names (to exclude them from being thanked).

Useful for if thanking people in the changelog is not necessary, e.g. thanking the main developer/maintainer over and over again, or in a private repo for work.

```
{
  "changelog": [
    {
      "repo": "org/repo",
      "doNotThank": true // Will not thank anyone
    }
  ],
}
```

```
{
  "changelog": [
    {
      "repo": "org/repo",
      "doNotThank": ["BeeeQueue"] // Will not thank BeeeQueue, but will thank anyone else (😢)
    }
  ],
}
```

---

_Was written directly on GitHub so might have style problems and has not been tested yet_

- [ ] Test locally

### DIFF
--- a/packages/changelog-github/src/index.ts
+++ b/packages/changelog-github/src/index.ts
@@ -40,12 +40,23 @@ const changelogFunctions: ChangelogFunctions = {
 
     return [changesetLink, ...updatedDepenenciesList].join("\n");
   },
-  getReleaseLine: async (changeset, type, options) => {
-    if (!options || !options.repo) {
+  getReleaseLine: async (changeset, type, options = {}) => {
+    if (!options.repo) {
       throw new Error(
         'Please provide a repo to this changelog generator like this:\n"changelog": ["@changesets/changelog-github", { "repo": "org/repo" }]'
       );
     }
+
+    const shouldThank = (user: string) => {
+      // Handle an array of names to not thank
+      if (Array.isArray(options.doNotThank)) {
+        return options.doNotThank.includes(user)
+      }
+      
+      // Should thank if it's null, true, anything except false
+      return options.doNotThank !== false
+    }
+
     const [firstLine, ...futureLines] = changeset.summary
       .split("\n")
       .map(l => l.trimRight());
@@ -56,9 +67,9 @@ const changelogFunctions: ChangelogFunctions = {
         commit: changeset.commit
       });
       return `\n\n- ${links.commit}${
-        links.pull === null ? "" : ` ${links.pull}`
+        links.pull == null ? "" : ` ${links.pull}`
       }${
-        links.user === null ? "" : ` Thanks ${links.user}!`
+        links.user == null && shouldThank(links.user) ? "" : ` Thanks ${links.user}!`
       } - ${firstLine}\n${futureLines.map(l => `  ${l}`).join("\n")}`;
     } else {
       return `\n\n- ${firstLine}\n${futureLines.map(l => `  ${l}`).join("\n")}`;


packages/changelog-github/src/index.ts
```diff
@@ -56,9 +67,9 @@ const changelogFunctions: ChangelogFunctions = {
         commit: changeset.commit
       });
       return `\n\n- ${links.commit}${
-        links.pull === null ? "" : ` ${links.pull}`
+        links.pull == null ? "" : ` ${links.pull}`
```
Andarist, line 70: IMHO would be better to keep strong comparisons here - `===`

packages/changelog-github/src/index.ts
```diff
@@ -40,12 +40,23 @@ const changelogFunctions: ChangelogFunctions = {
...
-    if (!options || !options.repo) {
+  getReleaseLine: async (changeset, type, options = {}) => {
+    if (!options.repo) {
       throw new Error(
         'Please provide a repo to this changelog generator like this:\n"changelog": ["@changesets/changelog-github", { "repo": "org/repo" }]'
       );
     }
+
+    const shouldThank = (user: string) => {
+      // Handle an array of names to not thank
```
Andarist, line 51: those comments here don't seem super useful - the code is rather simple and handles its job quite literally, so this seems repetitive for me

Andarist (commented):
I'm not sure if we'd like to proceed with this one, what are the benefits? What problems are caused by thanking all people? 🤔

packages/changelog-github/src/index.ts
BeeeQueue: Using strong comparisons on `null` is actually less safe, since using a weak one will catch both `null` and `undefined`, which you most often want to do.

packages/changelog-github/src/index.ts
BeeeQueue: I added them since there is no documentation for the option yet, so if someone goes into the code they will see it quickly without having to read the code.

Speaking of documentation, where should that be added?

packages/changelog-github/src/index.ts
Andarist: I understand the difference - can undefined be even received here though?

packages/changelog-github/src/index.ts
Andarist: Let's first discuss the merit of this option - don't want you to put too much work into this one if we decide not to proceed with this one in the end (I'm not saying that we won't, yet).

packages/changelog-github/src/index.ts
BeeeQueue: I don't know, but it's good practice to always use the weak one since it catches more potential branches with no downsides, so I did it here. 🤷

BeeeQueue:
The main issues I see occurring is for example if you're the sole developer on a project. It's very nice to get links to the PRs and commits in the changelog, but seeing `Thanks @BeeeQueue!` on all changes on every version gets a bit old... 

The other situation I've encountered is at work, where we don't really need to be thanking all the people working on the project, so it feels out of place

Andarist:
Isn't usually more information considered to be better than a lack of information? Those really seem quite harmless to me, even if a little bit excessive at times. You could rely on those to ask authors question regarding certain changes, rather than looking through `git blame` etc.

BeeeQueue:
If you like it or use it you can just not disable it, it's an option and not forced on you. 😄

Noviny:
@BeeeQueue Thanks for the PR! I'd probably actually recommend making your own functions, rather than using one of our defaults. Our goal was to make writing new generation functions easy and letting the community build lots to suit their needs, rather than having 1 or 2 canon ones that have a bunch of configuration options.

I thoroughly acknowledge that writing new changelog generators is not where we want it to be at the moment though.

I'd be more interested in this as its own package than as an option for the existing one.

BeeeQueue:
It feels very overkill to copy-paste a whole generator, publish a package, etc. just to add an option, but if that's what you prefer I'll do that instead. 😕

Andarist:
You probably could reuse our implementation and just strip away what u want with by manipulating the returned string.

We have to balance lean API and community demands, lean API doesnt only mean fewer options but also, more importantly, it means a reduced complexity of our implementation. If we add everything that is being asked for we’d end up with unmanageable code. If this issue arise again in the future we might reconsider stance on this particular option but for now i think it’s not needed as it has only been requested by you, so we cant really classify it as a popular demand.

PR #462, opened 2020-09-30
Fixed an issue with failing to parse changesets containing Windows newlines

Andarist:
I believe that it fixes https://github.com/atlassian/changesets/issues/461

### DIFF
--- a/packages/parse/src/index.test.ts
+++ b/packages/parse/src/index.test.ts
@@ -148,4 +148,56 @@ describe("parsing a changeset", () => {
       summary: ""
     });
   });
+  it("should be fine if the frontmatter is followed by a whitespace on the same line", () => {
+    const changesetMd = outdent`---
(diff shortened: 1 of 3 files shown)


packages/parse/src/index.test.ts
```diff
@@ -148,4 +148,56 @@ describe("parsing a changeset", () => {
       summary: ""
     });
   });
+  it("should be fine if the frontmatter is followed by a whitespace on the same line", () => {
```
Andarist, line 151: I've recognized this as a separate issue and fixed it first but it seems that the fix for it has also immediately resolved the Windows issue. This makes sense when I look at this regex now - can't find a situation when `\r\n` would not be covered correctly right now - please recheck though if anything pops into your mind.

mitchellhamilton (approved):

Issue #338, opened 2020-04-23
`changeset publish` fails with `Unexpected end of JSON input` after package deletion


sgoll:
## Affected Packages

`@changesets/cli@2.6.5`

## Problem

We are using `changeset publish` for packages hosted at our organization's GitHub package registry. After deleting a specific package version (the package itself still exists), `changeset publish` fails with the following error:

```
🦋  error SyntaxError: Unexpected end of JSON input
🦋  error     at JSON.parse (<anonymous>)
🦋  error     at npmRequestLimit (…/node_modules/@changesets/cli/dist/cli.cjs.dev.js:505:17)
🦋  error     at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Proposed solution

`changeset publish` should handle package deletion (as in deleting a single package version) gracefully.

Andarist:
The error seems to be coming from here:
https://github.com/atlassian/changesets/blob/d570a24740a374a9547dd7d931e6ac854e36a9fb/packages/cli/src/commands/publish/npm-utils.ts#L61

This should just check the package name (so removing a particular version shouldn't matter) and it also shouldn't crash on non-existing packages as receiving errors here is explicitly allowed and handled here:
https://github.com/atlassian/changesets/blob/d570a24740a374a9547dd7d931e6ac854e36a9fb/packages/cli/src/commands/publish/npm-utils.ts#L66-L67

I've encountered this problem in the past when one of my dependencies had old `npm` as its dependency and changesets have tried to use it (instead of the global one) but that old version had no support for `info` cmd. Could you check if that's maybe your problem as well?

sgoll:
@Andarist It seems that GitHub returns invalid data when a package version has been removed.

For any package,`npm info --json (…)` returns the correct data. But for the broken package in question (where the most recent version has been yanked from the registry), the output of `npm info` is just the empty string (no error, no contents).

Has anybody experienced this problem?

Andarist:
Huh, that's weird - maybe worth reporting to npm itself? If that's the case I guess changesets should implement a workaround for this issue.

talentlessguy:
Same issue with `pnpm`:

```
➜ pnpx changeset publish
🦋  info npm info @tinyhttp/app
🦋  info npm info @tinyhttp/cors
🦋  info npm info @tinyhttp/etag
🦋  info npm info @tinyhttp/logger
🦋  info npm info @tinyhttp/static
[...]

PR #469, opened 2020-10-09
Fix for #338 - Handle unpublished packages on github package registry

camdub:
Seeing the issue referenced in #338.

When running `npm info pkgName --json` NPM returns the 'E404' JSON response the code is expecting. If the registry target is the Github package registry, the same `npm info` command returns just an empty string.

https://github.com/atlassian/changesets/blob/d570a24740a374a9547dd7d931e6ac854e36a9fb/packages/cli/src/commands/publish/npm-utils.ts#L61

This code errors b/c it tries to parse an empty string. This PR just returns an error object similar to what NPM returns so that the publish command just sees it as an unpublished package.

I'm not super familiar with this repo, but this solution could be improved by checking the registry and only allowing empty responses to count as 404 for github packages. Unsure how to do that here.


Noviny (approved):

Noviny:
Thanks for this!

PR #482, opened 2020-11-06
None bumping peerdeps should not cause major bumps

jonathanmorley:


.changeset/blue-yaks-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": patch
+---
+
+None bumping peerdeps should not cause major bumps
```
Andarist, line 5:
```suggestion
Fixed an issue with bumping a peer dependency using a `"none"` changeset type resulting in the dependant package being major bumped.
```

Andarist (approved):

Issue #478, opened 2020-10-21
Separate publish / tag into two separate entrypoints


dfee:
## Affected Packages

pkg:@changesets/cli

## Problem

If `changeset publish` fails halfway through publishing, then no tags are created and pushed. If then, `changeset publish` succeeds, then tags are only created for packages who were published on that run. The problem, then, is then that we won't tag all releases, only the subset that were successfully published on the second run.

## Proposed solution

Split `tag` out of `publish`. `changeset version` -> `changeset publish` -> `changeset version`.

emmenko:
I have a similar issue/use case.

One of our repositories has around ~90 packages. We noticed that when all (or most of the) packages need to be published, there is a higher chance of the publishing to fail. To be more specific, packages are correctly published to NPM but the changesets command somehow fails, which then prevents tags to be created (as reported in this issue).

For example you can see in the screenshot that the version is published but the command reports an error.

<img width="1531" alt="image" src="https://user-images.githubusercontent.com/1110551/98665296-c506ab00-234b-11eb-8ab3-2e8c5d9ddbff.png">

Do you have an idea what might cause this?

However, in general it would be good to have a more resilient "error strategy". For example, retrying the publishing but do not throw in case the package is indeed published, etc.

Thanks

Andarist:
> Do you have an idea what might cause this?

We might want to limit concurrent requests to npm, even more, when publishing - I believe this has been discussed here: https://github.com/atlassian/changesets/issues/427#issuecomment-697206015

> However, in general it would be good to have a more resilient "error strategy". For example, retrying the publishing but do not throw in case the package is indeed published, etc.

That definitely would be an improvement. Additionally, to help the original issue reported here we could move creating tags at the end of the whole publish step. This should ensure that all packages are published first and only then we could attempt to create tags.

dfee:
Tags are non-destructive, right? I.e. they're not going to be overwritten on the remote. Therefore, shouldn't tagging follow the same strategy as publishing? I.e. if it's not published, attempt to publish...

Andarist:
Right, this probably could be done as well - it would probably require making an additional request to npm to check if a particular tag is published (unless we already get that info in the requests we make). Could you check what it would take to implement this?

emmenko:
Can you point out which places in the code we should be looking into?
I can also try to help out if you need some help.

dfee:
I'm generally thinking two paths:

### npm-registry path
1. enumerate current versions
[...]

PR #490, opened 2020-11-17
fix(cli): gracefully handle 403 errors when publishing

emmenko:
Ref #478

This is a (naive) attempt to gracefully handle 403 errors.

Let me know if this is an acceptable approach or if we should go into another direction.

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -1,3 +1,5 @@
+import path from "path";
+import fs from "fs-extra";
 import { ExitError } from "@changesets/errors";
 import { error, info, warn } from "@changesets/logger";
 import pLimit from "p-limit";
@@ -182,6 +184,31 @@ async function internalPublish(
       // just in case this isn't already true
       twoFactorState.isRequired = Promise.resolve(true);
       return internalPublish(pkgName, opts, twoFactorState);
+    }
+    // Sometimes NPM returns an error in case a version is already published, causing
+    // the publish script to fail and preventing changeset to finish the job.
+    // In this case, we should ignore the error:
+    // {
+    //   "error": {
+    //     "code": "E403",
+    //     "summary": "403 Forbidden - PUT https://registry.npmjs.org/<pkg_name> - You cannot publish over the previously published versions: <version>.",
+    //     "detail": "In most cases, you or one of your dependencies are requesting\na package version that is forbidden by your security policy."
+    //   }
+    // }
+    if (
+      json.error.code === "E403" &&
+      json.error.summary.includes("previously published versions")
+    ) {
+      let packageJson = await fs.readJSON(path.join(opts.cwd, "package.json"), {
+        encoding: "utf8"
+      });
+      let publishedPkgInfo = await getPackageInfo(
+        `${pkgName}@${packageJson.version}`
+      );
+      if (publishedPkgInfo.version === packageJson.version) {
+        return { published: true };
+      }
+      // The package hasn't been published apparently, let the command fail.
     }
     error(
       `an error occurred while publishing ${pkgName}: ${json.error.code}`,
--- a/.changeset/clever-zebras-roll.md
+++ b/.changeset/clever-zebras-roll.md
@@ -1,0 +1,7 @@
+---
+"@changesets/cli": patch
+---
+
+Gracefully handle 403 errors when publishing.
+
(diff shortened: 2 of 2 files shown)


emmenko:
Has anyone had a chance to look at this? Let me know in case you want to do it differently.
cc @Andarist 

It would be nice to fix this as it's still causing some issues.

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -183,6 +185,31 @@ async function internalPublish(
...
+      json.error.code === "E403" &&
+      json.error.summary.includes("previously published versions")
+    ) {
+      let packageJson = await fs.readJSON(path.join(opts.cwd, "package.json"), {
+        encoding: "utf8"
+      });
+      let publishedPkgInfo = await getPackageInfo(
+        `${pkgName}@${packageJson.version}`
+      );
+      if (publishedPkgInfo.version === packageJson.version) {
```
Andarist, line 208: what's the role of this check? shouldnt we trust the returned E403 that this particular version is already published?

Andarist:
From what I understand this can occur during a run that happens after a failing run that actually has successfully published a package? Is that statement correct?

emmenko:
Thanks for getting back to me.

The recap the issue: when publishing the packages, sometimes NPM returns `403 Forbidden` errors.

For example (see also CI run https://github.com/commercetools/ui-kit/runs/1736041131):

```
🦋  error an error occurred while publishing @commercetools-uikit/date-field: E403 403 Forbidden - PUT https://registry.npmjs.org/@commercetools-uikit%2fdate-field - You cannot publish over the previously published versions: 10.44.0. 
🦋  error a package version that is forbidden by your security policy.
🦋  error an error occurred while publishing @commercetools-uikit/async-select-input: E403 403 Forbidden - PUT https://registry.npmjs.org/@commercetools-uikit%2fasync-select-input - You cannot publish over the previously published versions: 10.44.0. 
🦋  error a package version that is forbidden by your security policy.
🦋  error an error occurred while publishing @commercetools-uikit/date-time-input: E403 403 Forbidden - PUT https://registry.npmjs.org/@commercetools-uikit%2fdate-time-input - You cannot publish over the previously published versions: 10.44.0. 
🦋  error a package version that is forbidden by your security policy.
```

This leads to changesets to not complete the publishing action, AND to not create git tags.

```
🦋  @commercetools-uikit/date-field@10.44.0
🦋  @commercetools-uikit/async-select-input@10.44.0
🦋  @commercetools-uikit/date-time-input@10.44.0
```

However, the packages are indeed published to NPM. If you check those packages listed above, the version `10.44.0` is published.

image

Therefore, the proposal in this PR is to "catch" those 403 errors and make sure that the version is indeed published. If that's the case, we can "safely" ignore the error.

PS: also worth mentioning that I have a feeling that these issues occur more frequently when lots of packages need to be published.

NMinhNguyen:
That's very interesting given that there is logic in place to check whether a package needs to be published. Perhaps there is an edge case that the check doesn't account for? https://github.com/atlassian/changesets/blob/049f0b3122f506faec857a9c36c5f4b5cc550d67/packages/cli/src/commands/publish/publishPackages.ts#L191-L193

Andarist:
Well, according to @emmenko and logs in that linked run this was the very first time that those versions were being published. So the initial check wasn't able to "discard" those from being published as they were not published at that time. 

It's super odd that a successful publish results in a 403 with such an error message. I've asked around some other people that are managing large monorepos to check if they have experienced this.

I don't see anything in particular wrong with the proposed approach - other than the fact that this shouldn't be needed, at least in theory. I would like to propose something different that I've wanted to do anyway and that might help to mitigate the issue - we should change the concurrency limit for publishing to 1.

If that won't help and this will still happen then I will feel more comfortable with using the logic proposed in this PR. WDYT?

emmenko:
> we should change the concurrency limit for publishing to 1.

Yes that would be another option. 

Would this be the place for limiting the concurrent requests?

https://github.com/atlassian/changesets/blob/12f9a433a6c3ac38f9405fcd77c9108c423d7101/packages/cli/src/commands/publish/publishPackages.ts#L92

Or did you have something else in mind? Also, should the concurrency option be configurable?

emmenko:
@Andarist I went ahead of opened a separate PR with the concurrency limit, as you suggested.
https://github.com/atlassian/changesets/pull/522

Let me know if this is what you had in mind. Thanks

Andarist:
For now let's close this one as the other PR has landed: https://github.com/atlassian/changesets/pull/522 . We can always reopen this in the future if needed.

PR #495, opened 2020-11-26
Automatically deepen clones to find correct commits

### LINKED ISSUE #494: Detect insufficient commit history for changelog in shallow repo, and deepen clone
## Affected Packages

`@changesets/changelog-git`@0.1.6
`@changesets/cli`@2.11.2

## Problem

I've been trying to diagnose why, when I run `changeset version` locally I get changelog files with correct commit history, but when my Github action or other build tools run it, all the commits show the same (latest) commit id.

This is, obviously, because on my build tools I'm running with a 1-commit shallow clone. (See also https://github.com/atlassian/changesets/issues/379.)

The recommendation at the moment seems to be "run with --depth 0", i.e. "don't use shallow clones".  That doesn't really scale, unfortunately. We have some really large (20k, 30k+ commit) repos where it's really important that we use shallow clones for builds.

Running with a deeper - but fixed - shallow clone (e.g. `--depth 50`) is also not a great option: it's entirely likely that the commits mentioned in our changelog will be older than 50 commits ago.

Would it be difficult to automatically deepen the clone if it's found that the current depth is insufficient?  We have some code that automatically does this for some of our visual-regression baseline comparisons (e.g. if our baseline details are outside of the current clone depth).

## Proposed solution

I don't yet know much about the architecture of `changesets`.  I'd be happy helping out writing some code that does the automatic deepening, as we have some experience with that, but does that require support from the changelog APIs to indicate that the correct commit was not found?

RoystonS:
This fixes #494 - where we would get the wrong commit
that added a file if we were operating in a shallow
repository that didn't contain the actual commit -
by modifying `getCommitThatAddsFile` so that it
automatically deepens a shallow clone until it finds the
_real_ commit that added a file.

### DIFF
--- a/packages/git/src/index.test.ts
+++ b/packages/git/src/index.test.ts
@@ -1,5 +1,6 @@
 import fixtures from "fixturez";
 import spawn from "spawndamnit";
+import fileUrl from "file-url";
 
 import {
   getCommitThatAddsFile,
@@ -16,6 +17,11 @@ const f = fixtures(__dirname);
 
 async function getCurrentCommit(cwd: string) {
   const cmd = await spawn("git", ["rev-parse", "HEAD"], { cwd });
+  return cmd.stdout.toString().trim();
+}
+
+async function getCurrentCommitShort(cwd: string) {
+  const cmd = await spawn("git", ["rev-parse", "--short", "HEAD"], { cwd });
   return cmd.stdout.toString().trim();
 }
 
@@ -189,16 +195,98 @@ describe("git", () => {
     it("should commit a file and get the hash of that commit", async () => {
       await add("packages/pkg-b/package.json", cwd);
       await commit("added packageB package.json", cwd);
-      const head = await spawn("git", ["rev-parse", "--short", "HEAD"], {
-        cwd
-      });
+      const headSha = await getCurrentCommitShort(cwd);
 
       const commitHash = await getCommitThatAddsFile(
         "packages/pkg-b/package.json",
         cwd
       );
 
-      expect(commitHash).toEqual(head.stdout.toString().trim());
+      expect(commitHash).toEqual(headSha);
+    });
+
+    describe("with shallow clone", () => {
+      // We will add these three well-known files
+      // over multiple commits, then test looking up
+      // the commits at which they were added.
+      const file1 = ".changeset/quick-lions-devour.md";
+      const file2 = "packages/pkg-a/index.js";
+      const file3 = "packages/pkg-b/index.js";
+
+      // The location of our cloned repo
+      let clone: string;
+      // The SHAs of the three interesting commits
+      let commit1: string, commit2: string, commit3: string;
+
+      async function createCommits(count: number, dir: string) {
+        for (let i = 0; i < count; i++) {
+          await commit("dummy commit", dir);
+        }
+      }
+
+      beforeEach(async () => {
+        // Create some test commits.
+
+        // Test commit 1.
+        // This is an important commit for testing an edge case.
+        // It adds a file and is the very first commit in the repo, which
+        // means the commit has no parent.
+        await add(file1, cwd);
+        await commit("commit1", cwd);
+        commit1 = await getCurrentCommitShort(cwd);
+
+        // Create a big history gap
+        await createCommits(60, cwd);
+
+        // Test commit 2.
+        // This commit adds a file but is early on in the history of the repo
+        // so it won't be present in a shallow clone.
+        await add(file2, cwd);
+        await commit("commit2", cwd);
+        commit2 = await getCurrentCommitShort(cwd);
+
+        // Create a big history gap
+        await createCommits(60, cwd);
+
+        // Test commit 3.
+        // This commit adds a file and will be the head commit, so it will be
+        // present in our depth-1 clone.
+        await add(file3, cwd);
+        await commit("commit3", cwd);
+        commit3 = await getCurrentCommitShort(cwd);
+
+        // Make a 1-commit-deep shallow clone of this repo
+        clone = f.temp();
+        await spawn(
+          "git",
+          // Note: a file:// URL is needed in order to make a shallow clone of
+          // a local repo
+          ["clone", "--depth", "1", fileUrl(cwd), "."],
+          {
+            cwd: clone
+          }
+        );
+      });
+
+      it("reads the SHA of a file-add if commit already included in the clone", async () => {
+        // This file was added in the head commit, so will definitely be in our
+        // 1-commit clone.
+        const commit = await getCommitThatAddsFile(file3, clone);
+        expect(commit).toEqual(commit3);
+      });
+
+      it("reads the SHA of a file-add even if not already included in the clone", async () => {
+        // Finding this commit will require deepening the clone until it appears.
+        const commit = await getCommitThatAddsFile(file2, clone);
+        expect(commit).toEqual(commit2);
+      });
+
+      it("reads the SHA of a file-add even if the first commit of a repo", async () => {
+        // Finding this commit will require deepening the clone right to the start
+        // of the repo history, and coping with a commit that has no parent.
+        const commit = await getCommitThatAddsFile(file1, clone);
+        expect(commit).toEqual(commit1);
+      });
     });
   });
 
--- a/packages/git/src/index.test.ts
+++ b/packages/git/src/index.test.ts
@@ -4,6 +4,7 @@ import fileUrl from "file-url";
 
 import {
   getCommitThatAddsFile,
+  getCommitsThatAddFiles,
   getChangedFilesSince,
   add,
   commit,
@@ -23,6 +24,11 @@ async function getCurrentCommit(cwd: string) {
 async function getCurrentCommitShort(cwd: string) {
   const cmd = await spawn("git", ["rev-parse", "--short", "HEAD"], { cwd });
   return cmd.stdout.toString().trim();
+}
+
+async function getCommitCount(cwd: string) {
+  const cmd = await spawn("git", ["rev-list", "--count", "HEAD"], { cwd });
+  return parseInt(cmd.stdout.toString(), 10);
 }
 
 describe("git", () => {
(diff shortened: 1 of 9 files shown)


RoystonS:
(CLA requires some checking. Should be done soon.)

Note that I've hardwired the auto-deepening at 50 commits at a time.  It would be possible to make this configurable, or make it increase (say, double) the number of commits deepened each time around the loop.  But I'm not sure either are worthwhile. I'm imagining that one or two rounds of 50-deepening will cover the vast majority of scenarios.

packages/git/src/index.test.ts
```diff
@@ -189,16 +195,98 @@ describe("git", () => {
...
+      // The SHAs of the three interesting commits
+      let commit1: string, commit2: string, commit3: string;
+
+      async function createCommits(count: number, dir: string) {
+        for (let i = 0; i < count; i++) {
+          await commit("dummy commit", dir);
+        }
+      }
+
+      beforeEach(async () => {
```
Andarist, line 227: I would prefer if those tests wouldn't use a shared `beforeEach` hook. It IMHO makes it harder to understand the exact flow of the test as it's not sufficient to read the test's body to understand all steps that are involved in it. With a help of helper functions, it should still be possible to write those 3 tests in a readable and somewhat DRY manner.

packages/git/src/index.ts
```diff
@@ -43,13 +43,61 @@ export async function getDivergedCommit(cwd: string, ref: string) {
...
+  async function isRepoShallow() {
+    const isShallowResult = await spawn(
+      "git",
+      ["rev-parse", "--is-shallow-repository"],
+      { cwd }
+    );
+    return isShallowResult.stdout.toString().trim() === "true";
+  }
+
+  async function deepenCloneBy(by: number) {
```
Andarist, line 98: not yet sure how to best handle this but the code that consumes this starts parallel jobs:
https://github.com/atlassian/changesets/blob/c6d38b579bdfddc22714c352c3459d04d1745326/packages/apply-release-plan/src/index.ts#L203-L208
therefore this `getCommitThatAddsFile` can potentially be executed in parallel - while operating on a shared FS resource (git). It would be best to avoid this situation.

Andarist (commented):
Overall it looks great and I'm confident that when we land this we'll improve the DX quite considerably. I've just left a comment about a single concern that I have - please take a look.

packages/git/src/index.test.ts
RoystonS: Ok.  I'll see if I can pull out common helpers instead of a full common `beforeEach`.

packages/git/src/index.ts
RoystonS: Very good call - I hadn't appreciated the parallel usage.
Ok, so we need to avoid two things:
1. Multiple parallel calls unnecessarily each deepening a clone. (5 changesets not in the current clone shouldn't cause us to run 5x50-commit-deepen calls.)
2. Multiple parallel calls doing _anything_ to a repo (specifically the `git log --diff-filter`) whilst one call is deepening it. (Don't read a repo whilst writing to it.)

I'll add a (per-repo) mutex that allows multiple readers OR a single writer.

Testing the multiple parallel job behaviour will be an "interesting challenge".

Andarist: > Testing the multiple parallel job behaviour will be an "interesting challenge".

I believe that sometimes there are just things not worth testing. Probably any test trying to test this would be rather strongly tied to the implementation and would require mocking or smth. Maybe its just such a case?

> I'll add a (per-repo) mutex that allows multiple readers OR a single writer. 

I would be interested in what do u mean by a single writer approach.

The alternative solution would be to provide a new function that would accept a list of files to check and return the list of commits. That way it could just implement a proper scheduling logic that would be local to it.

RoystonS: Yes, testing such a beast fully would require some rather evil mocking and, as you say, be rather fragile and tied to implementation.

The single-writer approach would be some sort of mutex where as many things that want to read from the repo would be able to take 'read locks' as you like, but only one thing at a time can take a 'write lock' which blocks out readers until done.  Easy to make correct, but avoiding the multiple unnecessary deepenings would require some complex code.


A new function that accepts a simple list and can take control of its own scheduling would be _far_ simpler and more self-contained rather than having this specific git function having to know that it's being accessed in a parallel way and cope with that. i.e. the loop would be "Try to find commits for all files.  Any of them need a deepening? Ok, we'll do _one_, then repeat."

The only other use seems to be in the `apply-release-plan` test-utils: https://github.com/atlassian/changesets/blob/c6d38b579bdfddc22714c352c3459d04d1745326/packages/apply-release-plan/src/test-utils/get-changelog-entry-with-git-hash.ts#L8-L16

Ok, so I'll look into changing `getCommitThatAddsFile` so it's a bulk operation.  As we would want to encourage people to use the single bulk operation rather than looping, it probably makes sense to remove the existing one-at-a-time function, but that would be a breaking change in the `git` package. Let me know if that's not what you want.

Andarist: If there is no conceptual problem with it we could make the implementation of the `getCommitThatAddsFile` as follows:
```js
const getCommitThatAddsFile = (file, cwd) => getCommitsThatAddsFiles([file], cwd)[0].commit
```

Making a breaking change to this package is not a big deal though (at least as far as I know) - it's just supposed to be used internally here and is not really up for public use.

RoystonS:
Right. After a hectic release and a Christmas break I've gotten back to this and made the changes we discussed.

Specifically:

1. I've replaced `getCommitThatAddsFile` and its multiple parallel runs (which weren't safe when needing to deepen a shallow clone) with a single `getCommitsThatAddFiles` call that retrieve the commits in parallel and safely deepen a shallow clone.
2. I've removed the big `beforeEach` from the test to make the individual tests more standalone.

Note that I've left the existing `getCommitThatAddsFile` call in there (redirected through the new call) but have marked it as 'deprecated' using the NodeJS `deprecated` call.

RoystonS:
Ook, failing tests. Will investigate; they certainly pass locally.  In the meantime, code review very welcome!

packages/git/src/index.ts
```diff
@@ -1,8 +1,10 @@
 import spawn from "spawndamnit";
+import fs from "fs";
 import path from "path";
 import { getPackages, Package } from "@manypkg/get-packages";
 import { GitError } from "@changesets/errors";
 import isSubdir from "is-subdir";
+import { deprecate } from "util";
```
Andarist, line 7: TIL

packages/git/src/index.ts
```diff
@@ -43,13 +45,126 @@ export async function getDivergedCommit(cwd: string, ref: string) {
...
+      if (info.commitSha) {
+        if (info.parentSha) {
+          // We have found the parent of the commit that added the file.
+          // Therefore we know that the commit is legitimate and isn't simply the boundary of a shallow clone.
+          map.set(info.path, info.commitSha);
+        } else {
+          commitsWithMissingParents.push(info);
+        }
+      } else {
+        // No commit for this file, which indicates it doesn't exist.
```
Andarist, line 89: could you elaborate on what does it mean? the file? the commit? seems like a file - shouldn't we throw in such a case instead of silently ignoring this?

packages/git/src/index.ts
```diff
@@ -43,13 +45,126 @@ export async function getDivergedCommit(cwd: string, ref: string) {
...
+      await deepenCloneBy(50);
+    } else {
+      // It's not a shallow clone, so all the commit SHAs we have are legitimate.
+      for (const unresolved of commitsWithMissingParents) {
+        map.set(unresolved.path, unresolved.commitSha);
+      }
+      break;
+    }
+
+    remaining = commitsWithMissingParents.map(p => p.path);
```
Andarist, line 112: this is only relevant for the `isRepoShallow()` case - so I would prefer keeping it in its block so the relation between them gets more obvious

Andarist (commented):
I have left only some very minor commits - overall an excellent job 🚀

packages/git/src/index.ts
RoystonS: This is the case where we've asked for the commit that introduced a file but the file doesn't exist.  It doesn't silently fail: it returns an `undefined` in that return slot.  So if you ask for the commits for 3 files, and only 2 exist, you'll still get the commits for the 2 that did exist, and `undefined` for the one that didn't exist. (There's a unit test for this.)

I could have made the bulk operation throw if any file didn't exist, but I think the previous implementation simply returned `undefined` for a non-existent file, so I kept that behaviour.  Happy to change it to a throw if any requested files don't exist; I slightly prefer that approach myself.

packages/git/src/index.ts
RoystonS: Nice little built-in API.  I haven't yet advertised the deprecation in the CHANGELOG entry.  I'll do that shortly.

packages/git/src/index.ts
RoystonS: Yep, you're quite right. I'd obviously got a bit of tunnel vision in there.  That `remaining = ` line path only runs if we're deepening.  Incoming commit...

packages/git/src/index.ts
RoystonS: I've added a deprecation note to the changesets file

RoystonS:
Changes made. Back to you @Andarist when you get a moment.

packages/git/src/index.ts
Andarist: > I could have made the bulk operation throw if any file didn't exist, but I think the previous implementation simply returned undefined for a non-existent file, so I kept that behaviour. 

Right, good point about this being the current behavior. Let's keep it that way for now - but I think we might change this to throwing in the future.

Andarist (approved):
@RoystonS thank you for an excellent improvement and the hard work on this PR. I will be merging this 🔜

packages/git/src/index.ts
```diff
@@ -43,13 +45,125 @@ export async function getDivergedCommit(cwd: string, ref: string) {
...
+  "Use the bulk getCommitsThatAddFiles function instead"
+);
+
+/**
+ * Get the short SHAs for the commits that added files, including automatically
+ * extending a shallow clone if necessary to determine any commits.
+ * @param gitPaths - Paths to fetch
+ * @param cwd - Location of the repository
+ */
+async function getCommitsThatAddFiles(
```
mitchellhamilton, line 61: Don't super mind if this happens in this PR but it might be nice if we batch/limit the concurrent deepens to 1 rather than having this because there could still be multiple calls to this in different places and etc.

RoystonS: I'm a _little_ confused as to what you're asking for here when you say 'batch/limit rather than having this'.  This _is_ a batched API for deepening the repo (as discussed previously: https://github.com/atlassian/changesets/pull/495#discussion_r531805043).

I _could_ add a [mutex](https://www.npmjs.com/package/async-mutex) to `getCommitsThatAddFiles` if you're worried that there'd be multiple simultaneous calls to this batched API?

mitchellhamilton: I mean keep the API getCommitThatAddsFile but batch it internally(probably just the deepen) and use a mutex or p-limit to limit simultaneous calls to deepen

RoystonS: We discussed doing that previously (https://github.com/atlassian/changesets/pull/495#discussion_r531723754). Testing it would be a little tricky, hence @Andarist and I settling on the batch call.

I could rewrite and test it again, but could I ask you and @Andarist to jointly agree on an approach before I do so?

If you want to keep `getCommitThatAddsFile` and avoid introducing `getCommitsThatAddFiles` that should be done now before merging this PR otherwise doing it later will involve an undeprecation and another deprecation. :-(

mitchellhamilton: Okay, happy to leave it how it is 👍

PR #500, opened 2020-12-02
Add release-utils package

mitchellhamilton:
This is essentially https://github.com/changesets/action but without the GitHub specific parts. I need this stuff for a project on Bitbucket so I'm gonna merge and release this now but @Andarist / @Noviny if you want to review this, that would be great. (though the main changes from the action are really just the tests which don't do mocking anymore) I'll open a PR to the action to use this at some point.


packages/release-utils/src/utils.ts
```diff
@@ -0,0 +1,131 @@
...
+    const previousVersion = previousVersions.get(pkg.dir);
+    if (previousVersion !== pkg.packageJson.version) {
+      changedPackages.add(pkg);
+    }
+  }
+
+  return [...changedPackages];
+}
+
+export function getChangelogEntry(changelog: string, version: string) {
```
mitchellhamilton, line 38: (yes, I'm aware this isn't exported/used anywhere in this PR, that will change in a future PR)

PR #504, opened 2020-12-02
Dont fail `changeset status` command if there are no changed packages

### LINKED ISSUE #496: `changeset status` errors when no publishable package has been modified
## Affected Packages

`@changesets/cli`

## Problem

Hello and thank you for changesets! It looks great and I'm excited to adopt it where I work 🙂 
I'm migrating our versioning/publishing automation from [rush](https://rushjs.io/pages/commands/rush_change/) to changesets, and changesets is a much lighter solution for my goals.

The only thing I miss from rush is their equivalent of `changeset status`, [`rush change -v`](https://rushjs.io/pages/commands/rush_change/). It automatically detects if any changes in your working branch actually affect a publishable package before erroring. That means if I have a branch that only makes changes to i.e. monorepo tooling, `rush change -v` does not error.

**TL;DR; if I modify e.g. the root `package.json` of a monorepo and run `changeset status --since=main`, it will error, even though I didn't modify any publishable packages.**

## Proposed solution

It looks like currently, `changeset status --since=branch` only checks for new changeset files in `.changeset/` to determine if it should error.

Ideally, it would also check if the diff since `branch` includes changes to files within packages folders before erroring.

vitorbal:
Fixes #496.

Add a check to the status command, so `process.exit(1)` is only called if there are no changesets _and_ a package has been changed. This prevents a few false-positives that can hurt DX. See #496 for more details.

@Andarist I wasn't sure if you wanted this to be a major bump or not.. IMO this should be a minor bump, since the new version will produce _less_ errors than before, not more (meaning, no CI checks should fail that were not already failing before). Let me know if you agree or not, and I'll add the changeset accordingly.

### DIFF
--- a/.changeset/chatty-plums-watch.md
+++ b/.changeset/chatty-plums-watch.md
@@ -2,4 +2,4 @@
 "@changesets/cli": minor
 ---
 
-`changeset status` command no longer errors when no publishable packages have been changed.
+`changeset status` command no longer errors when no packages have been changed.
--- a/.changeset/chatty-plums-watch.md
+++ b/.changeset/chatty-plums-watch.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+`changeset status` command no longer errors when no publishable packages have been changed.
--- a/packages/cli/src/commands/add/createChangeset.ts
+++ b/packages/cli/src/commands/add/createChangeset.ts
@@ -218,7 +218,6 @@ export default async function createChangeset(
       )}? (current version is ${pkg.packageJson.version})`,
       ["patch", "minor", "major"]
     );
-    console.log(type);
     if (type === "major") {
       let shouldReleaseAsMajor = await confirmMajorRelease(pkg.packageJson);
       if (!shouldReleaseAsMajor) {
--- a/packages/cli/src/commands/status/__tests__/status.ts
+++ b/packages/cli/src/commands/status/__tests__/status.ts
@@ -76,8 +76,6 @@ describe("status", () => {
   let cwd: string;
 
   beforeEach(async () => {
-    // @ts-ignore
-    jest.spyOn(process, "exit").mockImplementation(() => {});
     cwd = await f.copy("simple-project");
   });
 
@@ -97,6 +95,8 @@ describe("status", () => {
 
   it("should exit early with a non-zero error code when there are changed packages but no changesets", async () => {
     // @ts-ignore
+    jest.spyOn(process, "exit").mockImplementation(() => {});
+    // @ts-ignore
     git.getChangedPackagesSinceRef.mockImplementation(
       () => simpleChangedPackagesList
     );
@@ -107,6 +107,8 @@ describe("status", () => {
   });
 
   it("should not exit early with a non-zero error code when there are no changed packages", async () => {
+    // @ts-ignore
+    jest.spyOn(process, "exit").mockImplementation(() => {});
     // @ts-ignore
     git.getChangedPackagesSinceRef.mockImplementation(() => []);
 
@@ -121,6 +123,8 @@ describe("status", () => {
   });
 
   it("should not exit early with a non-zero code when there are changed packages and also a changeset", async () => {
+    // @ts-ignore
+    jest.spyOn(process, "exit").mockImplementation(() => {});
     // @ts-ignore
     git.getChangedPackagesSinceRef.mockImplementation(
       () => simpleChangedPackagesList
--- a/packages/cli/src/commands/status/__tests__/status.ts
+++ b/packages/cli/src/commands/status/__tests__/status.ts
@@ -66,6 +66,8 @@ describe("status", () => {
   beforeEach(async () => {
     // @ts-ignore
     git.getChangedPackagesSinceRef.mockImplementation(() => []);
+    // @ts-ignore
+    jest.spyOn(process, "exit").mockImplementation(() => {});
     cwd = await f.copy("simple-project");
   });
 
@@ -81,8 +83,6 @@ describe("status", () => {
 
   it("should exit early with a non-zero error code when there are changed packages but no changesets", async () => {
     // @ts-ignore
-    const mockExit = jest.spyOn(process, "exit").mockImplementation(() => {});
-    // @ts-ignore
     git.getChangedPackagesSinceRef.mockImplementation(() => [
       {
         name: "pkg-a",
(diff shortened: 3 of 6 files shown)


packages/cli/src/commands/status/__tests__/status.ts
```diff
@@ -76,13 +78,58 @@ describe("status", () => {
     const releaseObj = await status(cwd, {}, defaultConfig);
     expect(releaseObj).toEqual(simpleReleasePlan);
   });
-  it("should exit with a non-zero error code when there are no changesets", async () => {
+
+  it("should exit early with a non-zero error code when there are changed packages but no changesets", async () => {
```
Andarist, line 82: we'd probably need a test for no changes packages *but* with changes to the unrelated files to cover for your original issue more closely

packages/cli/src/commands/status/__tests__/status.ts
```diff
@@ -76,13 +78,58 @@ describe("status", () => {
...
     // @ts-ignore
     const mockExit = jest.spyOn(process, "exit").mockImplementation(() => {});
```
Andarist, line 84: I wonder - shouldn't this be somehow reset after each test?

Andarist (commented):
> IMO this should be a minor bump, since the new version will produce less errors than before, not more (meaning, no CI checks should fail that were not already failing before)

This sounds reasonable to me. Any objections about this one @mitchellhamilton @Noviny ?

mitchellhamilton:
Sounds reasonable to me too but I wouldn't mind getting @Noviny's thoughts on it because I've never used changeset status to enforce having changesets.

packages/cli/src/commands/add/__tests__/add.ts
```diff
@@ -19,6 +18,7 @@ const f = fixtures(__dirname);
 jest.mock("../../../utils/cli-utilities");
 jest.mock("@changesets/git");
 jest.mock("@changesets/write");
+jest.mock("@changesets/logger");
```
vitorbal, line 21: I took the liberty of refactoring the log silencing logic to rely on `jest.mock`, instead of using the `temporarilySilenceLogs`. I also silenced logs for a couple tests that were not being silenced.

I don't really know what to do with `@changesets/test-utils` though, since `temporarilySilenceLogs` is not used anywhere anymore, and that's the only export from that package. Thoughts?

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -218,7 +218,6 @@ export default async function createChangeset(
       )}? (current version is ${pkg.packageJson.version})`,
       ["patch", "minor", "major"]
     );
-    console.log(type);
```
vitorbal, line 221: Found a straggler debug log here. I checked git history and it looks like this was never intended to be logged to consumers, so I removed it. Let me know if this is not the right call!

packages/cli/src/commands/status/__tests__/status.ts
Andarist: Ok, makes sense - given that `getChangedPackagesSinceRef` is actually mocked.

packages/cli/src/commands/status/__tests__/status.ts
Andarist: TIL

packages/cli/src/commands/add/__tests__/add.ts
Andarist: Any reason why do you prefer this approach? `temporarilySilenceLogs` has the advantage of mocking the more final IO and not a package that could contain bugs. If we'd like to spy on `console.*` calls it would also make IMHO more sense to spy on them directly (potentially using test util) rather than spying on the logger package which then would pair more nicely with `temporarilySilenceLogs`.

I also believe that `temporarilySilenceLogs` is discoverable more easily and actually brings more meaning for a reader and it's easier to see what it is supposed to do. It's also easier to apply `temporarilySilenceLogs` or other utils like that on a per-file basis whereas `jest.mock` is "global" for a given test file.

packages/cli/src/commands/add/createChangeset.ts
Andarist: Yea - it's most likely a leftover that shouldn't be here, nice catch!

packages/cli/src/commands/add/__tests__/add.ts
vitorbal: Hey @Andarist thanks for the detailed reply, I appreciate it!

> Any reason why do you prefer this approach? 

Honestly the main reason was just getting rid of a util function, for simplicity's sake. It just felt simpler mocking `@changesets/logger` directly.

I also figured that any `console.*` log invocations that don't go through `@changesets/logger` were probably a mistake, so having those show up in the test run could actually be a good way to catch them (like the one in `createChangeset.ts` below, for example!).

You make good points, though, and I agree with you that `temporarilySilenceLogs` is potentially clearer for readers to understand than `jest.mock('@changesets/logger')`.

> It's also easier to apply temporarilySilenceLogs or other utils like that on a per-file basis whereas jest.mock is "global" for a given test file.

That's a good point, although I don't foresee any use for _not_ mocking logs for a specific`describe`s of a test-file, since the objective is to prevent pollution of the test run. I might be overseeing a use-case, though, so let me know.

One feasible scenario could be asserting on the log calls, but still mocking all logs "globally" for the test file. That should still be doable with the `jest.mock` version here. Let me know if that makes any sense?

Either way, say the word and I'd be happy to revert this part of the PR, I am definitely not married to this change 😄

Andarist: > One feasible scenario could be asserting on the log calls, but still mocking all logs "globally" for the test file. That should still be doable with the jest.mock version here. Let me know if that makes any sense?

Yeah, that's something we do from time to time in various projects but it usually is for warnings/errors-focused tests.

> Either way, say the word and I'd be happy to revert this part of the PR, I am definitely not married to this change 😄

I would be for reverting this part - I see some of your points as well and gonna definitely have them on my mind in the future but right now let's just wrap up this PR so we can release your awesome work :)

vitorbal: Alright, it's gone!

packages/cli/src/commands/status/index.ts
```diff
@@ -32,16 +34,22 @@ export default async function getStatus(
...
   }
-  const releasePlan = await getReleasePlan(
-    cwd,
-    since === undefined ? (sinceMaster ? "master" : undefined) : since,
-    config
-  );
-
+  const sinceBranch =
+    since === undefined ? (sinceMaster ? "master" : undefined) : since;
+  const releasePlan = await getReleasePlan(cwd, sinceBranch, config);
```
Andarist, line 39: I'm wondering - do you happen to know why this doesn't account for `config.baseBranch`? 🤔 I see that it's utilized by `@changesets/read` but not sure why. Maybe just something you have figured out while working on this PR

Andarist (approved):
Could you add a changeset for this? If not - I can do it myself later and merge it then.

packages/cli/src/commands/status/index.ts
vitorbal: Good question! I took a look [at the original PR that added it](https://github.com/atlassian/changesets/pull/221) and couldn't figure out the intent here, so I decided to follow the same logic. Maybe @mitchellhamilton remembers?

vitorbal (commented):
Sorry for the hiatus @Andarist, I was taking a few days off! I'd be glad to add the changeset. I'll add a minor bump as per our conversation earlier in this PR.

.changeset/chatty-plums-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+`changeset status` command no longer errors when no publishable packages have been changed.
```
Andarist, line 5:
```suggestion
`changeset status` command no longer errors when no packages have been changed.
```

PR #507, opened 2020-12-08
feat: bumpVersionsWithWorkspaceProtocolOnly

### LINKED ISSUE #506: Dependencies are upgraded to new major versions during release
## Affected Packages

Probably:
* assemble-release-plan
* apply-release-plan

## Problem

Let's say we have a repo with three projects: foo@2.0.0, bar@2.0.0, qar@2.0.0

foo has qar@2.0.0 in the dependencies. Which is the same as in the monorepo.

bar has an older version of qar in dependencies: 1.0.0.

Now if I release a new version of qar (v2.0.1), I only want to get it bumped inside foo. I want bar to keep using qar@1.0.0

However, seems like in some cases qar is bumped in bar as well. The reason is probably that assemble-release-plan is checking the semver ranges when finding dependents but apply-release-plan is not checking the semver ranges.

## Proposed solution

Maybe we should pass more state to apply-release-plan, to prevent it from making breaking/unwanted bumps

zkochan:
close #506


packages/cli/src/run.ts
```diff
@@ -139,7 +139,9 @@ export async function run(
         }
 
         // Validate that all dependents of ignored packages are listed in the ignore list
-        const dependentsGraph = getDependentsGraph(packages);
+        const dependentsGraph = getDependentsGraph(packages, {
+          workspaceVersionsOnly: config.workspaceVersionsOnly === true
```
Andarist, line 143: any particular reason behind this explicit comparison to `true` here? it hasn't been done the same way at other call sites, for example here: https://github.com/atlassian/changesets/pull/507/files?file-filters%5B%5D=.md&file-filters%5B%5D=.ts#diff-033b40b5770c213b81d0ae2fe63d820c6595831c6b8d8af5adb1754399409c1fR104

```suggestion
          workspaceVersionsOnly: config.workspaceVersionsOnly
```

Andarist (commented):
I've left just mostly nit-picky comments - as such I don't expect those places to be adjusted for this to get merged in if you don't feel like changing them. I've also left some minor questions - it would be lovely if you could answer those.

I'm not super fond of the `workspaceVersionsOnly` name for this but I also don't have anything in particular better than that. Maybe `manageWorkspaceVersionsOnly`?

zkochan:
maybe `bumpWorkspaceVersionsOnly`?

Andarist:
> maybe bumpWorkspaceVersionsOnly?

That would be better 👍 @mitchellhamilton any naming preferences for this one? I'll leave this open for a few days - if we don't get any further comments on this one I'll be proceeding with `bumpWorkspaceVersionsOnly`.

aspirisen:
@zkochan looks like your issue is pretty close to my one  https://github.com/atlassian/changesets/issues/512
I have a question about how does it work.

>Let's say we have a repo with three projects: foo@2.0.0, bar@2.0.0, qar@2.0.0
>foo has qar@2.0.0 in the dependencies. Which is the same as in the monorepo.
> bar has an older version of qar in dependencies: 1.0.0.

In your example, does `bar` has exact match version of `qar`? if not, how does it understand that it shouldn't bump  `qar` version in `bar`?

zkochan:
pnpm and Yarn support the `workspace:` protocol to link packages inside a monorepo:

https://pnpm.js.org/en/workspaces#workspace-protocol-workspace

So with this change, if the dependency has `"qar: "workspace:1.0.0"` in package.json, changeset will bump the version of qar. But if qar is specified without the workspace protocol: `"qar": "1.0.0"`. In this case, changeset will not bump the version of qar in dependencies.

aspirisen:
@zkochan so as I understood you need to change it manually. For example you have `foo` and `bar` and each of them are using `qar` via workspace protocol  `"qar: "workspace:1.0.0"` and therefore changesets will bump its version inside `foo` and `bar`. Then once you understand that `bar` package becomes stale and in order to avoid random breaking changes you set the version of `qar` without `workspace` and changesets stops incrementing `qar` version automatically?

zkochan:
Sounds correct. In a huge monorepo used by many teams, this scenario can happen. It did happen at my company, so until this is merged, we are using a fork of this project.

aspirisen:
Got it, thanks

mitchellhamilton:
Maybe `bumpVersionsWithWorkspaceProtocolOnly`?

zkochan:
I have renamed the setting.

zkochan:
Is there something else that needs to be done?

mitchellhamilton (approved):

PR #510, opened 2021-01-08
fix: undefined response from get-github-info call

### LINKED ISSUE #509: undefined result on get-github-info call
## Affected Packages
get-github-info

## Problem
When configuring changeset, if you pass in the repo name of the wrong format e.g `https://github.com/user/repo` instead of just `user/repo` it will cause the call to the Github API to return `undefined` and therefore causing `Cannot read property 'author' of undefined`. This error message can be misleading/confusing and so I think we need to handle that and show a nicer message.

There is an actual github action with that error of the repo that I'm working on - https://github.com/sajari/sdk-react/runs/1668436313?check_suite_focus=true

## Proposed solution
The solution is to add a more specific/stricter check for `request.repo` field to make sure that it follows the format (`userOrOrg/repoName`)

tuanddd:
Fixes #509

### DIFF
--- a/packages/get-github-info/src/index.ts
+++ b/packages/get-github-info/src/index.ts
@@ -2,6 +2,8 @@
 import fetch from "node-fetch";
 import DataLoader from "dataloader";
 
+const validRepoNameRegex = /^[A-Za-z0-9_.-]+\/[A-Za-z0-9_.-]+$/m;
+
 type RequestData = {
   commit: string;
   repo: string;
@@ -121,7 +123,7 @@ export async function getInfo(
     throw new Error("Please pass a commit SHA to getInfo");
   }
 
-  if (!request.repo) {
+  if (!request.repo || !validRepoNameRegex.test(request.repo)) {
     throw new Error(
       "Please pass a GitHub repository in the form of userOrOrg/repoName to getInfo"
     );
--- a/packages/get-github-info/src/index.ts
+++ b/packages/get-github-info/src/index.ts
@@ -2,7 +2,7 @@
 import fetch from "node-fetch";
 import DataLoader from "dataloader";
 
-const validRepoNameRegex = /^[A-Za-z0-9_.-]+\/[A-Za-z0-9_.-]+$/m;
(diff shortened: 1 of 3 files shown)


packages/get-github-info/src/index.ts
```diff
@@ -2,6 +2,8 @@
 import fetch from "node-fetch";
 import DataLoader from "dataloader";
 
+const validRepoNameRegex = /^[A-Za-z0-9_.-]+\/[A-Za-z0-9_.-]+$/m;
```
Andarist, line 5: Do you have any source which described what repo names are considered as valid by GitHub? Are dots allowed? Why `m` modifier has been used?

tuanddd: The Regex was taken from [this answer on SO](https://stackoverflow.com/a/59082561)

And yes a little oversight on my end, the `m` flag is redundant, will remove it, thanks for catching that 👍

Andarist, line 5: This should be the equivalent
```suggestion
const validRepoNameRegex = /^[\w.-]+\/[\w.-]+$/;
```

packages/get-github-info/src/index.ts
```diff
@@ -121,7 +123,7 @@ export async function getInfo(
     throw new Error("Please pass a commit SHA to getInfo");
   }
 
-  if (!request.repo) {
+  if (!request.repo || !validRepoNameRegex.test(request.repo)) {
```
Andarist, line 126: could you move the second check to a separate if and throw an error that would contain `validRepoNameRegex.source`? That way users should be able to get a little bit more information as to why the validation has failed

tuanddd:
@Andarist I've pushed the changes 🙂

tuanddd:
@Andarist What is blocking this?

Andarist (approved):

Andarist:
> @Andarist What is blocking this?

My life :P

packages/get-github-info/src/index.ts
chj-damon: why don't we support @ ? like '@ant-design/xxx'

PR #513, opened 2021-01-13
Refactor createVersionCommit

Andarist:
This is some work that I had locally for a month or so and I haven't finished it - just pushing it as a draft so I don't forget about it as I don't have time to wrap it up right now.


beeequeue:
stale

PR #516, opened 2021-01-20
Custom assemble release plan

### LINKED ISSUE #512: Specify dependency path of packages that should be bumped
## Problem

I have i.e. four packages: `app`, `page-users`, `page-settings`, `ui-kit`. 
"app" is using "page-users" and "page-settings"
"page-users" and "page-settings" are using "ui-kit"

```
                                
              app              
             /    \             
            /      \            
           /        \           
          -          -          
     page-users    page-settings
          \          /          
           \        /           
            \      /            
             -    -           
             ui-kit          
                           
```
For example if i make a breaking change in `ui-kit` and use it in `page-users` i need to bump all the packages except `page-settings`, it will use the old version of `ui-kit`, by that i will know that my changes will not break `page-settings`.

I need to specify a tree path how `changesets` will bump the versions, maybe like `linked-packages`

## Proposed solution

Probably my use case is very specific, maybe it would be possible to overload the way how `changesets` determines what packages should be bumped.

aspirisen:
Closes: https://github.com/atlassian/changesets/issues/512
Here you can override `assemble-release-plan` and create any logic for bumping packages.


Andarist:
Have you tried this in your project - could you paste in your custom assemble release plan? The split between assemble release plan and apply release plan is somewhat unintuitive for me (the apply package also does some version-related manipulations) so having a concrete example would make it a lot easier for me to review this.

That being said - I know that @mitchellhamilton had some ideas related to customizing those 2 packages, or rather maybe a way of building a custom runner that would compose our exposed pieces and custom ones. So it would be great if we could get his input regarding the approach proposed in this PR.

mitchellhamilton:
The approach I've seen taken in the past is to call assemble release plan and apply release plan yourself. Given that it's not interactive or anything, there isn't really much point to using the cli over having a script like this


```js
import assembleReleasePlan from "@changesets/assemble-release-plan";
import readChangesets from "@changesets/read";
import { read } from "@changesets/config";
import { readPreState } from "@changesets/pre";
import applyReleasePlan from "@changesets/apply-release-plan";

const preState = await readPreState(cwd);
const config = await read(cwd, packages);
const changesets = await readChangesets(cwd, sinceRef);

let releasePlan = assembleReleasePlan(changesets, packages, config, preState);

releasePlan = modifyReleasePlan(releasePlan)

await applyReleasePlan(releasePlan, packages, config);

```

aspirisen:
@Andarist something like that, so if I didn't add 'changeset' to "page-settings" or "page-users" they will not bump.
```
import assembleReleasePlanFunction from "@changesets/assemble-release-plan";

export const assembleReleasePlan: typeof assembleReleasePlanFunction = (
  changesets,
  packages,
  config,
  preState,
  snapshot
) => {
  const result = assembleReleasePlanFunction(
    changesets,
    packages,
    config,
    preState,
    snapshot
  );

  const updateOnlyIfAffected = /page-settings|page-users/;

  result.releases = result.releases.filter((r) =>
    updateOnlyIfAffected.test(r.name) ? r.changesets.length > 0 : true
  );

  return result;
};

```

aspirisen:
@mitchellhamilton do you mean that you can reproduce the behavior of `changesets` by the code that you provided?

mitchellhamilton:
Yes

with-heart:
One of the things we do with Chakra UI is publish an "aggregate package" (`@chakra-ui/react`) for user convenience that exports most of our other packages. We'd like to add a small customization to the release plan assembly that will always bump the aggregate package using the highest change of the package dependencies, so that its version accurately represents any breaking changes, new features, etc. included in those deps.

Given that we seem to have a similar use case to @aspirisen but for a different purpose, would it make sense to have a post-assemble option that allows us to specify a module with an export that takes and returns a `ReleasePlan`?

I don't mind using a custom script instead, but it would be a lot easier to be able to specify a small hook to achieve some customization here without needing to change the rest of the workflow.

Thoughts @mitchellhamilton?

Andarist:
@with-heart wouldn't running a totally separate script before calling into changesets work?

Script's logic:
1. grab existing changesets
2. duplicate them - just change the pkg name to aggregate and mention the original pkg name in the content of the changeset

After such operation, Changesets would just see those additional changeset files and I think you would pretty much land with your desired outcome.

mitchellhamilton:
> I don't mind using a custom script instead, but it would be a lot easier to be able to specify a small hook to achieve some customization here without needing to change anything else about the workflow.

You don't really need to change anything about your workflow, the only change is from running `changeset version` to the script (and you keep using the GH action by telling the action to run the script https://github.com/changesets/action#with-version-script)

beeequeue:
closing as it's stale, but we can consider it for v4

PR #522, opened 2021-01-26
fix(cli): limit publishing concurrency

emmenko:
As suggested in another PR: https://github.com/atlassian/changesets/pull/490#issuecomment-766199145

The main purpose of this is to reduce/eliminate unwanted errors and side effects caused by having too many concurrent requests for publishing packages.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -37,6 +37,7 @@
     "@changesets/types": "^3.3.0",
     "@changesets/write": "^0.1.3",
     "@manypkg/get-packages": "^1.0.1",
+    "@ricokahler/pool": "1.0.0",
     "@types/semver": "^6.0.0",
     "boxen": "^1.3.0",
     "chalk": "^2.1.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -37,7 +37,6 @@
     "@changesets/types": "^3.3.0",
     "@changesets/write": "^0.1.3",
     "@manypkg/get-packages": "^1.0.1",
-    "@ricokahler/pool": "1.0.0",
     "@types/semver": "^6.0.0",
     "boxen": "^1.3.0",
     "chalk": "^2.1.0",
--- a/packages/cli/src/commands/publish/publishPackages.ts
+++ b/packages/cli/src/commands/publish/publishPackages.ts
@@ -1,13 +1,29 @@
+import { join } from "path";
 import semver from "semver";
 import chalk from "chalk";
 import { AccessType } from "@changesets/types";
 import { Package } from "@manypkg/get-packages";
-import * as npmUtils from "./npm-utils";
 import { info, warn } from "@changesets/logger";
-import { TwoFactorState } from "../../utils/types";
 import { PreState } from "@changesets/types";
+import pool from "@ricokahler/pool";
+import * as npmUtils from "./npm-utils";
+import { TwoFactorState } from "../../utils/types";
 import isCI from "../../utils/isCI";
-import { join } from "path";
+
+type PublishedState = "never" | "published" | "only-pre";
+
+type PkgInfo = {
+  name: string;
+  localVersion: string;
+  publishedState: PublishedState;
+  publishedVersions: string[];
+};
+
+type PublishedResult = {
+  name: string;
+  newVersion: string;
+  published: boolean;
+};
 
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -71,33 +87,31 @@ export default async function publishPackages({
...
-  for (let pkgInfo of unpublishedPackagesInfo) {
-    let pkg = packagesByName.get(pkgInfo.name)!;
-    promises.push(
-      publishAPackage(
+  return pool<PkgInfo, PublishedResult>({
+    collection: unpublishedPackagesInfo,
+    // If there are many packages to be published, it's better to limit the
+    // concurrency to avoid unwanted errors, for example from NPM.
+    // TODO: make it configurable?
+    maxConcurrency: 10,
```
NMinhNguyen, line 95: From https://github.com/atlassian/changesets/pull/490#issuecomment-766199145

> we should change the concurrency limit for publishing to 1.

in which case the `for-of` loop could be refactored to something like

```js
results.push(
  await publishAPackage(
    pkg,
    access,
    twoFactorState,
    getReleaseTag(pkgInfo, preState, tag)
  )
)
```

and then outside the loop

```js
return results;
```

without needing any extra dependencies.

packages/cli/package.json
```diff
@@ -37,6 +37,7 @@
     "@changesets/types": "^3.2.0",
     "@changesets/write": "^0.1.3",
     "@manypkg/get-packages": "^1.0.1",
+    "@ricokahler/pool": "1.0.0",
```
NMinhNguyen, line 40: If you really did want to use a dependency, something more well known like https://github.com/sindresorhus/p-map might be better.

packages/cli/src/commands/publish/publishPackages.ts
emmenko: Yes I know but I think it makes sense to have a reasonable value. Unless of course there are other reasons that I don't know of.

Let's wait for @Andarist or another maintainer to give feedback first.

Thanks for your feedback btw

packages/cli/package.json
emmenko: Sure, I don't have a strong opinion about it. Whatever is preferred by the maintainers is fine.
I'll wait for feedback.

mitchellhamilton (commented):
Could you use p-limit? (it's already used elsewhere in the publishing code)

Andarist:
This could also be changed just here by using a different limit created through `p-limit`:
https://github.com/atlassian/changesets/blob/599119d1ad05e4fff307d3f8988bc1683e25cc2f/packages/cli/src/commands/publish/npm-utils.ts#L210

Both limits would work independently but this shouldn't be a problem as we don't do any parallel work when publishing.

emmenko:
Alright, I refactored the changes to use `p-limit`.

packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -1,13 +1,32 @@
...
+};
+
+type PublishedResult = {
+  name: string;
+  newVersion: string;
+  published: boolean;
+};
+
+// TODO: make it configurable?
+const npmRequestLimit = pLimit(10);
```
emmenko, line 29: Is that a reasonable concurrency value? Do you prefer something else?

Also, I'm not sure we need to make it configurable. What do you think?

packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -71,33 +90,30 @@ export default async function publishPackages({
...
-        getReleaseTag(pkgInfo, preState, tag)
-      )
-    );
-  }
-
-  return Promise.all(promises);
+  return Promise.all(
+    unpublishedPackagesInfo.map(pkgInfo =>
+      // If there are many packages to be published, it's better to limit the
+      // concurrency to avoid unwanted errors, for example from NPM.
```
NMinhNguyen, line 96:
```suggestion
      // concurrency to avoid unwanted errors, for example from npm.
```

Nit: npm is [lower case](https://css-tricks.com/start-sentence-npm/) 😅

packages/cli/src/commands/publish/publishPackages.ts
mitchellhamilton: 10 sounds reasonable. For now, let's not make it configurable

mitchellhamilton (approved):
Thanks!

PR #526, opened 2021-02-02
Fix the prerelease command in the docs.

threepointone:
I think `changset preleease` isn't a valid command ? And it doesn't match the command from the sample either.

### DIFF
--- a/docs/prereleases.md
+++ b/docs/prereleases.md
@@ -31,7 +31,7 @@ packages/
 ```
 
 ```
-yarn changeset prerelease next
+yarn changeset pre enter next
 ```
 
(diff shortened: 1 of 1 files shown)


threepointone:
I don't understand why the tests are failing, sorry :/

Andarist:
Thank you! This might be a leftover after the previous implementation or something. Don't worry about tests - the failure obviously has nothing to do with this PR, I'll have to investigate later and merge your PR then.

mitchellhamilton (approved):
Looks like CI was just a random failure, reran it and it passed. Thanks for the PR @threepointone!

PR #534, opened 2021-02-26
Fixed an issue where git.getCommitsThatAddFiles is undefined

### LINKED ISSUE #532: Changeset version not working due to wrong dependency range
## Affected Packages

`pkg:apply-release-plan`

## Problem

`changeset version` command returns `🦋  error TypeError: git.getCommitsThatAddFiles is not a function`

With the following resolution in yarn.lock:

```
"@changesets/apply-release-plan@npm:^4.2.0":
  version: 4.2.0
"@changesets/git@npm:^1.0.5":
  version: 1.0.6
```

## Proposed solution

With a clean install I don't have the issue I think this is related to [@changesets/git 1.1.0](https://github.com/atlassian/changesets/blob/bcf0f7ed0c6c2becb507fc646f1fc2c1638ba01d/packages/git/CHANGELOG.md#110) where `getCommitsThatAddFiles` was added. The version of this package was not bumped in `@changesets/apply-release-plan`  package.json even though it's using this function

If I'm right in the package.json of `@changesets/apply-release-plan` we should have `"@changesets/git": "^1.1.0",`

adriencohen:
Fixes #532

### DIFF
--- a/.changeset/perfect-days-report.md
+++ b/.changeset/perfect-days-report.md
@@ -2,4 +2,4 @@
 "@changesets/apply-release-plan": patch
 ---
 
-Fixed an issue where `git.getCommitsThatAddFiles` could be undefined due to wrong semver range for @changeset/git
+Bump the required dependency range of [`@changesets/git`](https://github.com/atlassian/changesets/tree/master/packages/changelog-github) package as the new version has added `getCommitsThatAddFiles` which is used here.
--- a/.changeset/perfect-days-report.md
+++ b/.changeset/perfect-days-report.md
@@ -1,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
+---
+
+Fixed an issue where `git.getCommitsThatAddFiles` could be undefined due to wrong semver range for @changeset/git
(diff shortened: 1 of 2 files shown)


Andarist:
I've taken another look at the changesets that lead to that broken release - I'm not entirely sure why this has happened as there was a minor bump targeting `@changesets/cli`. This might be expected as per the current logic but I'd like to investigate this - gonna handle your PR after that (should take me a couple of days max).

.changeset/perfect-days-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
+---
+
+Fixed an issue where `git.getCommitsThatAddFiles` could be undefined due to wrong semver range for @changeset/git
```
Andarist, line 5:
```suggestion
Bump the required dependency range of [`@changesets/git`](https://github.com/atlassian/changesets/tree/master/packages/changelog-github) package as the new version has added `getCommitsThatAddFiles` which is used here.
```

Andarist (approved):

PR #535, opened 2021-03-08
Allow replacing the PR/commit/author shown in the changlog line by writing `pr: some-pr-number` and etc.

mitchellhamilton:
An idea I had to solve the "oh no, I just merged a PR without a changeset and now that I add one, it'll attribute it to the wrong PR" problem.

### DIFF
--- a/packages/changelog-github/src/index.test.ts
+++ b/packages/changelog-github/src/index.test.ts
@@ -1,0 +1,101 @@
+import changelogFunctions from "./index";
+import parse from "@changesets/parse";
+
+const getReleaseLine = changelogFunctions.getReleaseLine;
+
+const data = {
+  commit: "a085003",
+  user: "Andarist",
+  pull: 1613,
+  repo: "emotion-js/emotion"
+};
+
+jest.mock(
+  "@changesets/get-github-info",
+  (): typeof import("@changesets/get-github-info") => {
+    // this is duplicated because jest.mock reordering things
+    const data = {
+      commit: "a085003",
+      user: "Andarist",
+      pull: 1613,
+      repo: "emotion-js/emotion"
+    };
+    const links = {
+      user: `[@${data.user}](https://github.com/${data.user})`,
+      pull: `[#${data.pull}](https://github.com/${data.repo}/pull/${data.pull})`,
+      commit: `[\`${data.commit}\`](https://github.com/${data.repo}/commit/${data.commit})`
+    };
+    return {
+      async getInfo({ commit, repo }) {
+        expect(commit).toBe(data.commit);
+        expect(repo).toBe(data.repo);
+        return {
+          pull: data.pull,
+          user: data.user,
+          links
+        };
+      },
+      async getInfoFromPullRequest({ pull, repo }) {
+        expect(pull).toBe(data.pull);
+        expect(repo).toBe(data.repo);
+        return {
+          commit: data.commit,
+          user: data.user,
+          links
+        };
+      }
+    };
+  }
+);
+const expected = `\n\n- [#1613](https://github.com/emotion-js/emotion/pull/1613) [\`a085003\`](https://github.com/emotion-js/emotion/commit/a085003) Thanks [@Andarist](https://github.com/Andarist)! - something\n`;
+
+describe.each([data.commit, "wrongcommit", undefined])(
+  "with commit from changeset of %s",
+  commitFromChangeset => {
+    const getChangeset = (content: string) => {
+      return [
+        {
+          ...parse(
+            `---
+      pkg: "minor"
+      ---
+      
+      something
+      ${content}
+      `
+          ),
+          id: "some-id",
+          commit: commitFromChangeset
+        },
+        "minor",
+        { repo: data.repo }
+      ] as const;
+    };
+
+    test.each(["pr", "pull request", "pull"])(
+      "override pr with %s keyword",
+      async keyword => {
+        expect(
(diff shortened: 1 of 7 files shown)


packages/changelog-github/src/index.test.ts
```diff
@@ -0,0 +1,101 @@
...
+          user: data.user,
+          links
+        };
+      }
+    };
+  }
+);
+const expected = `\n\n- [#1613](https://github.com/emotion-js/emotion/pull/1613) [\`a085003\`](https://github.com/emotion-js/emotion/commit/a085003) Thanks [@Andarist](https://github.com/Andarist)! - something\n`;
+
+describe.each([data.commit, "wrongcommit", undefined])(
```
Andarist, line 52: I get the premise of `describe.each` etc but this is somewhat hard to read right now. Could we maybe do a couple of things to improve the readability here? I would like to propose:

1. moving top-level `data` below the `jest.mock` call - i know that `jest.mock` is somewhat also using the same structure but on the other hand it already uses its copy and it would be easier to analyze the flow of this test if the test data would be closer to it
2. move `getChangeset` out of the test body
3. create a separate test block for the `commitFromChangeset === "a085003"` case 
4. duplicate the `expected` in those two tests that are using it - if all tests would use the same expected value it would imho make sense to reuse it but since one of them is using a different value it's harder to scan through this quickly to understand that

packages/changelog-github/src/index.ts
```diff
@@ -46,23 +46,71 @@ const changelogFunctions: ChangelogFunctions = {
         'Please provide a repo to this changelog generator like this:\n"changelog": ["@changesets/changelog-github", { "repo": "org/repo" }]'
       );
     }
-    const [firstLine, ...futureLines] = changeset.summary
+
+    let prFromSummary: number | undefined;
+    let commitFromSummary: string | undefined;
+    let userFromSummaryOverride: string | undefined;
```
Andarist, line 52: shouldn't all of those use either the `*Summary` or `*SummaryOverride` suffix? Is there any reason why they all don't match the same suffix?

mitchellhamilton: It was because `userFromSummaryOverride` had a markdown link rather than just the value but I just changed it so it's just the value now and they're all named `*Summary`

packages/changelog-github/src/index.test.ts
```diff
@@ -0,0 +1,107 @@
...
+          ...getChangeset(`commit: ${data.commit}`, commitFromChangeset)
+        )
+      ).toEqual(
+        `\n\n- [#1613](https://github.com/emotion-js/emotion/pull/1613) [\`a085003\`](https://github.com/emotion-js/emotion/commit/a085003) Thanks [@Andarist](https://github.com/Andarist)! - something\n`
+      );
+    });
+  }
+);
+
+test.each(["author", "user"])(
```
Andarist, line 98: would also be great to add tests for `@Andarist` and `#1613` - handling those are great DX improvements 🚀

packages/changelog-github/src/index.ts
```diff
@@ -46,23 +46,73 @@ const changelogFunctions: ChangelogFunctions = {
...
       );
     }
-    const [firstLine, ...futureLines] = changeset.summary
+
+    let prFromSummary: number | undefined;
+    let commitFromSummary: string | undefined;
+    let userFromSummary: string | undefined;
+
+    const replacedChangelog = changeset.summary
+      .replace(/(?:pr|pull|pull\s+request):\s*#?(\d+)/i, (_, pr) => {
```
Andarist, line 55: Should those be caught in the middle of the summary? I would at least require those to be at the beginning of a line

packages/get-github-info/src/index.ts
```diff
@@ -95,22 +109,37 @@ const GHDataLoader = new DataLoader(async (requests: RequestData[]) => {
...
+    let output: { commit: Record<string, any>; pull: Record<string, any> } = {
+      commit: {},
+      pull: {}
+    };
+    cleanedData[repo] = output;
+    Object.entries(data.data[`a${index}`]).forEach(([field, value]) => {
+      if (field[0] === "a") {
+        output.commit[field.substring(1)] = value;
+      } else {
+        output.pull[field.replace("pr__", "")] = value;
```
Andarist, line 126: the `pr__` prefix is somewhat self-explanatory, but the `"a"` thing is not - could u add a comment about this one? I believe it's just related to GitHub's graphql API but since I'm not familiar with it this doesn't give me much information about what's happening here (especially that we are processing `data.data` here 🤣 )

packages/changelog-github/src/index.ts
```diff
@@ -46,23 +46,73 @@ const changelogFunctions: ChangelogFunctions = {
...
       );
     }
-    const [firstLine, ...futureLines] = changeset.summary
+
+    let prFromSummary: number | undefined;
+    let commitFromSummary: string | undefined;
+    let userFromSummary: string | undefined;
+
+    const replacedChangelog = changeset.summary
+      .replace(/\n\s*(?:pr|pull|pull\s+request):\s*#?(\d+)/i, (_, pr) => {
```
Andarist, line 55: note that instead of matching `\n` you could match `^` with the `m` flag since then `^` means the start of a line instead of the start of the string
```suggestion
      .replace(/^\s*(?:pr|pull|pull\s+request):\s*#?(\d+)/mi, (_, pr) => {
```

Andarist (approved):

bjoluc:
Just looked this up again, but couldn't find it anywhere in the docs – am I missing anything?

Issue #387, opened 2020-06-07
Generating global changelog and/or GitHub release


leonardfactory:
## Affected Packages

`pkg:@changesets/cli`

## Problem

While using this awesome project, I felt the desire to have a way to generate a *global* changelog / release. The idea is to have a single changelog, where each "release" could be identified by the versioning date (or something like that). To get an idea, the [Blueprint.js changelog](https://github.com/palantir/blueprint/wiki/3.x-Changelog) is an example of what I'm talking about (implemented with lerna).

I think the problem arises in my situation since all the packages I'm publishing are meant to be relevant. There are other projects (like `pnpm`, `@changesets` itself) where only the main package is relevant (like `pnpm`, `@changesets/cli`).

## Proposed solution

I've already seen you're suggesting to implement a custom `version` command, relying on `getReleasePlan` and `assembleReleasePlan`. I'm pretty sure it is doable with a custom implementation, however I'd like to get feedback from you and maybe other folks using the project in order to know if feature could be relevant.

Furthermore, I'm not sure about your plans about custom changelog generation, is still the custom `version` command the preferred way to do this at the moment? I've seen the cli command implementation and it seems a bit complicated to implement it without basically reinventing the wheel. Even some suggestion here would be greatly appreciated!

Andarist:
Not sure what is the opinion of others - but I think some way to customize this would be helpful for people. It is very hard to account for all possible use cases, so probably accepting custom "generator" of this stuff with documented input & output would be the way to go - similarly to how you already can customize changelog generator. Maybe it would be just a matter of a different output from those changelog generators? Could you list the exact things that you'd like to customize? cc @zkochan - this issue might be of interest to you as well.

Of course, if we could just document how to implement your own `version` command this could work too, but I don't believe this is trivial right now.

leonardfactory:
In my opinion, current changelog entries generator is fine. I was thinking about a "collector", gathering changes grouped by package, e.g.:

> ### Version packages 09.06.2020
>
> #### `@scope/pkg-a` v2.0.0
> 
> Major Changes:
> - A breaking change
> 
> #### `@scope/pkg-b` v1.1.0
>
> Minor Changes:
> - A new feature

Main difference here is that this `CHANGELOG.md` "collected" file could be placed in project root folder. This could be especially useful for projects using a simulated locked mode with linked packages.

In this case it would be necessary to customize:
- Output file (even if `CHANGELOG.md` could be considered a sensible default)
[...]

Issue #264, opened 2020-01-26
[proposal] Aggregated changelog in root of monorepo?


KingScooty:
Since moving over to changesets from lerna, one of the things i miss is not having a root aggregated changelog log of all changes in the root of the monorepo project.

I'm curious what the reasoning for this is?

The root monorepo changelog (that lerna generated) was a great way for me to give stakeholders and customers a super quick overview of all the things that have changed in one place.

Is it a possible feature to be added going forward?

Andarist:
Would this be an alternative for per-package changelogs or an addition to them?

KingScooty:
Addition i think. I imagine there may be people that want one or the other, but i definitely got value out of having both. Perhaps i have a dated workflow and need to change it (willing to be challenged if people disagree with the idea).

I've just found that there's some interesting maintenance that comes with having a monorepo. The tooling, scripts etc. Essentially monorepo _features_ that can change frequently as new ways to maintain your mono repo/new best practices come about. They may not be a feature to product, but they are to developers. It's nice to be able to celebrate/advertise those changes somewhere. Perhaps the root monorepo changelog doesn't aggregate anything. Perhaps it's its own thing that summarises what's going on around the monorepo codebase?

In a world where business pays the bills, it's nice to be able to show off these changes to stakeholders. After all monorepo tooling is supposed to help speed up output and improve consistency. I'll take anything that helps me prove to stakeholders that codebase maintenance is just as important as feature development.

Keen to know people's thoughts.

Noviny:
I think this would be a good thing to be able to get. Issue #200 is a potential way to implement this, with this being a step beyond that around having opinions about where and how to store such an aggregate.

dotansimha:
My two cents that having a single `CHANGELOG` in the root will probably break things like Renovate/npm from detecting the actual changes in a specific package release.

chathu-novade:
This would be great to have "changelog per package and a combined changelog in the root"

Jym77:
I'm also maintaining a monorepo.
I used to have one (manual 🙈 ) changelog at the root, and switched to changesets for ease of use.
I'm still pushing some notes to the root changelog, but now it's only the list of changed packages in that release (with links to their changelog) which I hack around at release time (essentially running `grep` on all changesets). 

From my user point of view, I think a nice way would be to have the changelog generator functions (`ChangelogFunctions`) have a third sibling to add to that global changelog. These functions already seem to be receiving the set of changesets in a nice format…
[...]

PR #537, opened 2021-03-11
feat: enable output of global changelog

with-heart:
Hi all! I'd been meaning to work on functionality mentioned in #200, #264, and #387 for awhile now, and finally had a chance to take a look. I wanted to open this as a draft to get some feedback on this approach, as something feels off to me about this.

As of right now, this functionality is attached to the `version` command by means of the `--globalChangelog` flag, which naively places `CHANGELOG.md` at the root of `cwd`. I've tested this with the [`chakra-ui`](https://github.com/chakra-ui/chakra-ui) repo, so I can verify that it works correctly. You can see the output of that here: https://gist.github.com/with-heart/ebdb9e9e74b3f84e221111d12c3723ca

# Changes
- Modified `generateMarkdown` to return `{title: string, body: string}` rather than `string`. This was needed because `title` is normally `## X.Y.Z` in the individual packages, but we'd want this to be `<package-name>@X.Y.Z` in a global changelog similar to how it works with "Version Packages" PRs.
- Modified existing individual package changelog generation within `applyReleasePlan` to utilize above change in data shape
- Added a new section to `applyReleasePlan` that combines the individual changelogs into a single `CHANGELOG.md`

# Questions/Issues
- Does it still make sense to start by collecting changelogs as markdown?

#200 mentions outputting a "'release info' thing", but it seems weird to me that the output of this would be a markdown file, which is a fairly static piece of output instead of something like a json file. In my mind, a nice opportunity for refactoring the releases->changelogs process would be to come up with a metadata object that those changelogs are built from, with the additional benefit that the output artifact could be a json file, which would allow anyone to consume it to produce whatever they wanted, whether that's a global changelog or something else.

- How do we envision configuration of the global changelog? With the Chakra global changelog, I think I'd want to sort the packages based on the highest version (packages with `major` changes come before packages with only `minor` changes come before packages with only `patch` changes), but I'm not sure how to easily allow that type of customization.

### DIFF
--- a/packages/apply-release-plan/src/get-changelog-entry.ts
+++ b/packages/apply-release-plan/src/get-changelog-entry.ts
@@ -2,6 +2,7 @@ import { ChangelogFunctions, NewChangesetWithCommit } from "@changesets/types";
 
 import { ModCompWithPackage } from "@changesets/types";
 import startCase from "lodash.startcase";
+import { ChangelogEntry } from "./types";
 import { shouldUpdateDependencyBasedOnConfig } from "./utils";
 
 type ChangelogLines = {
@@ -35,7 +36,7 @@ export default async function generateMarkdown(
     updateInternalDependencies: "patch" | "minor";
     onlyUpdatePeerDependentsWhenOutOfRange: boolean;
   }
-) {
+): Promise<ChangelogEntry | null> {
   if (release.type === "none") return null;
 
   const releaseObj: ChangelogLines = {
@@ -102,12 +103,22 @@ export default async function generateMarkdown(
     )
   );
 
-  return [
-    `## ${release.newVersion}`,
-    await generateChangesForVersionTypeMarkdown(releaseObj, "major"),
-    await generateChangesForVersionTypeMarkdown(releaseObj, "minor"),
-    await generateChangesForVersionTypeMarkdown(releaseObj, "patch")
-  ]
-    .filter(line => line)
-    .join("\n");
+  const title = `## ${release.newVersion}`;
+  const major = await generateChangesForVersionTypeMarkdown(
+    releaseObj,
+    "major"
+  );
+  const minor = await generateChangesForVersionTypeMarkdown(
+    releaseObj,
+    "minor"
+  );
+  const patch = await generateChangesForVersionTypeMarkdown(
+    releaseObj,
+    "patch"
+  );
+
+  return {
+    title,
+    body: [major, minor, patch].filter(line => line).join("\n")
+  };
 }
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -61,7 +61,10 @@ export default async function applyReleasePlan(
   releasePlan: ReleasePlan,
   packages: Packages,
   config: Config = defaultConfig,
-  snapshot?: string | boolean
+  options: {
+    snapshot?: string | boolean;
+    output?: boolean;
+  } = {}
 ) {
   let cwd = packages.root.dir;
 
@@ -95,7 +98,7 @@ export default async function applyReleasePlan(
     cwd
   );
 
-  if (releasePlan.preState !== undefined && snapshot === undefined) {
+  if (releasePlan.preState !== undefined && options.snapshot === undefined) {
     if (releasePlan.preState.mode === "exit") {
       await fs.remove(path.join(cwd, ".changeset", "pre.json"));
     } else {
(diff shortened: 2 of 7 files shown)


Andarist:
What's the intention of the global changelog? Or rather - what kind of things people could want from it? Are there any other use cases beyond just gathering all generated per-package changelogs in a single document?

with-heart:
> What's the intention of the global changelog? Or rather - what kind of things people could want from it? Are there any other use cases beyond just gathering all generated per-package changelogs in a single document?

That's a good question and after thinking about it for two weeks 😅 I haven't been able to think of any other use cases besides our own.

For Chakra, global changelog is part of a move we'd like to make to abstract our monorepo a little bit. Right now, we have a ton of individual changelog files being generated by `changesets` and just as many individual GitHub releases containing the same information. It's great to have that information available, but for most of our users, it's serious information overload and makes it difficult to navigate.

One of the packages we publish is `@chakra-ui/react`, which is like an aggregate of most of our other packages and offers a single entrypoint for most of our React users. So ideally we'd like to abstract the monorepo behind that. We already refer to the Chakra UI version as the version of the aggregate package, and we'd like to go a bit further by offering a global changelog that is organized around that package's version. 

In my mind, that'd look like one of these (leaning towards 1, but really not sure which would be more useful tbh):

```md
# `@chakra-ui/react@1.4.0`

## Minor Changes

- `@chakra-ui/package-a@1.2.0` Line for change! Thanks user1!
- `@chakra-ui/package-b@1.3.0` Line for change! Thanks user2!

## Patch Changes

- `@chakra-ui/package-c@1.1.3` Line for fix! Thanks user3!

# `@chakra-ui/react@1.3.4`

## Patch Changes

- `@chakra-ui/package-c@1.1.2` Line for fix! Thanks user4!
```

```md
# `@chakra-ui/react@1.4.0`

## `@chakra-ui/package-a@1.2.0`

### Minor Changes

- Line for change! Thanks user1!

## `@chakra-ui/package-b@1.3.0`

### Minor Changes

- Line for change! Thanks user2!

## `@chakra-ui/package-c@1.1.3`

### Patch Changes

- Line for fix! Thanks users3!

# `@chakra-ui/react@1.3.4`

## `@chakra-ui/package-c@1.1.2`

### Patch Changes

- Line for fix! Thanks user4!
```

joeldenning:
fwiw we also would find this helpful for the create-single-spa project.

NMinhNguyen:
I think this feature would be amazing for projects like Babel (or a design system) where there are a lot of packages, and having to navigate to each package's changelog becomes burdensome. It's useful to see at a glance what packages were updated with each release of such a project.

DelysidBot (approved):

NMinhNguyen:
Having used https://github.com/changesets/action for releases and the Version Packages PRs sort of as a global changelog, I think that something like https://github.com/babel/babel/releases/tag/v7.16.3 would probably be the best option. Currently, if you add a changeset that touches multiple packages, the Version Packages PR will contain "duplicate" entries for this change in every package's changelog. It'd be useful to deduplicate them the way Babel does (see how `babel-plugin-transform-parameters`, `babel-traverse` are on the same line):

```markdown
## v7.16.3 (2021-11-09)

#### :bug: Bug Fix
* `babel-helpers`
  * [#13862](https://github.com/babel/babel/pull/13862) fix(helpers): match `Reflect.get` behaviour ([@lightmare](https://github.com/lightmare))
* `babel-plugin-transform-parameters`, `babel-traverse`
  * [#13941](https://github.com/babel/babel/pull/13941) Support transforming params of arrow functions in class fields ([@nicolo-ribaudo](https://github.com/nicolo-ribaudo))
* `babel-parser`
  * [#13928](https://github.com/babel/babel/pull/13928) fix: incorrect await rejection following arrow function in parameters (Closes [#13872](https://github.com/babel/babel/issues/13872)) ([@The-x-Theorist](https://github.com/The-x-Theorist))
* Other
  * [#13918](https://github.com/babel/babel/pull/13918) Fix `parserOverride` support in `@babel/eslint-parser` ([@nicolo-ribaudo](https://github.com/nicolo-ribaudo))
```

Being able to see all the major changes in 1 place is probably very useful too, and Babel's changelog enables you to do that.

CxRes:
Does the Changesets team plan on accepting/implementing the global change log feature?

Andarist:
@CxRes I would be open to discussing and accepting a PR for this: https://github.com/changesets/changesets/issues/683 . I think this would be a more flexible approach and it could be used to achieve the "global changelog". Let me know if you are interested in helping out with this.

CxRes:
@Andarist I have only recently started playing with Changesets and do not know my way around the code to make a helpful contribution. It seemed to me though that this PR mostly resolves the issue; again, I do not understand the ways in which it falls short wrt #683. I would be happy to discuss but unless I understand the issue/code better I am in no position to offer any help.

All that said, this is definitely a useful feature I would like to see implemented! It is more consistent for my use case of a monolithic project that I am migrating to a monorepo to keep a single change log going.

emmenko:
Hi 👋,

What the current status with this feature? Is there anything we can do to help getting this implemented?

PR #558, opened 2021-04-01
adds a release plan flag to the version command

jcreamer898:
Hey y'all! Long time listener first time caller.

Was thinking it'd be nice to do `changesets version --release-plan` and dump said release plan out to the consumer.

This is a first pass so totally up for any other ideas, additions here, but visibility into the release plan is pretty cool to have. At my last gig (Eventbrite) we dumped the release plan out to a file by manually creating it, and then we could run builds on multiple CI machines using the plan.

Now, I'm at Microsoft and am hoping to leverage changesets here too, and thought I'd go ahead and contribute this idea.

Thanks!

### DIFF
--- a/packages/cli/src/commands/version/index.ts
+++ b/packages/cli/src/commands/version/index.ts
@@ -23,9 +23,11 @@ export default async function version(
   cwd: string,
   options: {
     snapshot?: string | boolean;
+    releasePlan?: string | boolean;
   },
   config: Config
 ) {
+
   const releaseConfig = {
     ...config,
     // Disable committing when in snapshot mode
@@ -70,6 +72,10 @@ export default async function version(
     preState,
     options.snapshot
   );
+
+  if (options.releasePlan) {
+    return releasePlan;
+  }
 
   await applyReleasePlan(
     releasePlan,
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/version/index.ts
```diff
@@ -71,6 +73,10 @@ export default async function version(
     options.snapshot
   );
 
+  if (options.releasePlan) {
```
Andarist, line 76: is your intention to skip the `applyReleasePlan` altogether? It actually still resolves some version/dependency-related stuff so I feel like it would be safer to include its logic in what is being returned

packages/cli/src/index.ts
```diff
@@ -42,6 +42,9 @@ const { input, flags } = meow(
         type: "string",
         isMultiple: true
       },
+      releasePlan: {
```
Andarist, line 45: it would make sense to me to change this to a `--json` flag, see https://github.com/atlassian/changesets/issues/391

jcreamer898: Yeah, I thought about that, good idea.

PR #559, opened 2021-04-02
Added new `--open` flag to the `add` command

### LINKED ISSUE #554: include --open flag on add command?
## Affected Packages

`pkg:cli`

## Problem

Sometimes you want to just edit the changeset by hand instead of using the prompts, so you use `--empty`. But then you have to manually go find the right filename that was just created. It would be nice to be able to open it automatically with the default editor.

## Proposed solution

Potentially as an `--open` flag.

Andarist:
closes #554 

cc @ianstormtaylor

### DIFF
--- a/.changeset/strange-news-rhyme.md
+++ b/.changeset/strange-news-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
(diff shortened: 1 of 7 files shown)


mitchellhamilton (approved):

ianstormtaylor:
@Andarist thank you! Looking forward to using it.

PR #569, opened 2021-04-13
feat: allow publishing private packages

gakonst:
Supersedes https://github.com/atlassian/changesets/pull/420.

We get all packages (instead of just the public ones) and we only publish the public ones, while returning `published = true` for everything else, so that they can be used in the tagging step after publishing.

Context: https://github.com/atlassian/changesets/issues/399, https://github.com/atlassian/changesets/issues/478#issuecomment-780951976

cc @mitchellhamilton

### DIFF
--- a/packages/cli/src/commands/publish/index.ts
+++ b/packages/cli/src/commands/publish/index.ts
@@ -79,8 +79,13 @@ export default async function run(
     if (tool !== "root") {
       for (const pkg of successful) {
         const tag = `${pkg.name}@${pkg.newVersion}`;
-        log("New tag: ", tag);
-        await git.tag(tag, cwd);
+        const isMissingTag = !(await git.tagExists(tag));
+        if (isMissingTag) {
+          log("New tag: ", tag);
+          await git.tag(tag, cwd);
+        } else {
+          log("Skipping existing tag: ", tag);
+        }
       }
     } else {
       const tag = `v${successful[0].newVersion}`;
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -34,6 +34,13 @@ async function tag(tagStr: string, cwd: string) {
   return gitCmd.code === 0;
 }
 
+async function tagExists(tagStr: string) {
+  const gitCmd = await spawn("git", ["tag", "-l", tagStr]);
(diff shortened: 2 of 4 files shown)


gakonst:
Hi @mitchellhamilton is there any feedback for this PR?

packages/cli/src/commands/publish/index.ts
```diff
@@ -79,8 +79,13 @@ export default async function run(
...
       for (const pkg of successful) {
         const tag = `${pkg.name}@${pkg.newVersion}`;
-        log("New tag: ", tag);
-        await git.tag(tag, cwd);
+        const isMissingTag = !(await git.tagExists(tag));
+        if (isMissingTag) {
+          log("New tag: ", tag);
+          await git.tag(tag, cwd);
+        } else {
+          log("Skipping existing tag: ", tag);
```
Andarist, line 87: When this branch could be chosen? Tag conflict seems like a problem and not something that should be silently silenced

packages/git/src/index.ts
```diff
@@ -34,6 +34,13 @@ async function tag(tagStr: string, cwd: string) {
   return gitCmd.code === 0;
 }
 
+async function tagExists(tagStr: string) {
```
Andarist, line 37: this is a new export from the `@changesets/git` package so a changeset should be created to account for that

andrewabest:
This change would be super useful for us - we want to use changesets to manage release notes and versioning for node packages that are not published to npm.

If `changeset publish` creates and outputs tags for private packages, the GitHub Action will proceed to [create a GitHub release](https://github.com/changesets/action/blob/master/src/run.ts#L125) for the change with the appropriate release notes, which is our goal.

JakeGinnivan:
Hey @mitchellhamilton @Andarist just giving this a friendly bump, i'm just setting up another project and would love to move away from my patch-package approach :)

JakeGinnivan:
Just playing with this, it still checks NPM for the existence of the private packages through.

With my PR for public NPM packages I used that status, otherwise I used tags. I think it should clearly separate the logic for each.

Andarist:
This got superseded by https://github.com/changesets/changesets/pull/662

PR #572, opened 2021-04-13
Document `bumpVersionsWithWorkspaceProtocolOnly` option

Andarist:
Related to https://github.com/atlassian/changesets/pull/507

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -4,20 +4,36 @@ import micromatch from "micromatch";
 import { ValidationError } from "@changesets/errors";
 import { warn } from "@changesets/logger";
 import { Packages } from "@manypkg/get-packages";
-import { Config, WrittenConfig, Linked } from "@changesets/types";
+import { Config, WrittenConfig } from "@changesets/types";
 import packageJson from "../package.json";
 import { getDependentsGraph } from "@changesets/get-dependents-graph";
 
-export let defaultWrittenConfig = {
+type MakeSureThatSchemaDescribesAllProperties<
+  T extends WrittenConfig,
+  Schema = typeof import("../schema.json")["properties"]
+> = {
+  [K in keyof T]: K extends keyof Schema
+    ? T[K]
+    : "This property is missing a definition in the schema";
+};
+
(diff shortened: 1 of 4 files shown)


packages/config/src/index.ts
```diff
@@ -4,20 +4,36 @@ import micromatch from "micromatch";
...
 
-export let defaultWrittenConfig = {
+type MakeSureThatSchemaDescribesAllProperties<
+  T extends WrittenConfig,
+  Schema = typeof import("../schema.json")["properties"]
+> = {
+  [K in keyof T]: K extends keyof Schema
+    ? T[K]
+    : "This property is missing a definition in the schema";
+};
```
Andarist, line 18: It's obviously somewhat quirky but also somewhat useful 🤷‍♂️ I'm OK with removing this if you feel like it's too much

mitchellhamilton: lol, i like it. we might want to move it to a test file so it doesn't run for consumers?

packages/config/src/index.ts
```diff
@@ -4,20 +4,36 @@ import micromatch from "micromatch";
...
   changelog: "@changesets/cli/changelog",
   commit: false,
-  linked: [] as Linked,
+  linked: [],
   access: "restricted",
   baseBranch: "master",
   updateInternalDependencies: "patch",
-  ignore: [] as ReadonlyArray<string>
-} as const;
+  bumpVersionsWithWorkspaceProtocolOnly: false,
```
mitchellhamilton, line 34: tbh, I think we should be more intentional about what we include in the default written, for example `updateInternalDependencies` is something that very few people should use so it probably shouldn't have been included there and I feel like `bumpVersionsWithWorkspaceProtocolOnly` fits in the same category.

andrewsellers:
Hi, the `bumpVersionsWithWorkspaceProtocolOnly` config option / feature is great, thanks for including this 🙌 

It took a little tracking down though as wasn't surfaced in the current docs - is this PR still active / due to be merged?

beeequeue:
this is now documented in the new website: https://changesets.dev/guide/config#bumpversionswithworkspaceprotocolonly

PR #576, opened 2021-04-23
Add verdaccio to the cool list

juanpicado:
We have been using `changesets` for quite a while and decided to use it definitely, cool project. I'd like also to be on the list.

https://github.com/verdaccio/verdaccio


mitchellhamilton (approved):
Thanks!

PR #584, opened 2021-05-02
Avoid an infinite loop when git commands fail to execute when Changesets try to retrieve commits that added files

### LINKED ISSUE #571: Infinite loop makes changeset version command hang
## Affected Packages

`@changesets/git`

## Problem

a loop that never ends
https://github.com/atlassian/changesets/blob/18d3da3b707fe834e0bb1cddce1186bb24b75ef2/packages/git/src/index.ts#L112

Andarist:
This addresses the direct problem in https://github.com/atlassian/changesets/issues/571 . We need to investigate the issue more to address the core of the problem though.

fix #571


Andarist:
I will address the failing tests later - I don't think this should block the potential PR approval though (unless you think otherwise).

emmatown (commented):
Could you add the command that was run to the error that is thrown?

mitchellhamilton:
> I will address the failing tests later - I don't think this should block the potential PR approval though (unless you think otherwise).

We've explicitly intended to support running Changesets without git and the tests failing seem like they would break that?

Andarist:
> Could you add the command that was run to the error that is thrown?

Sure thing.

> We've explicitly intended to support running Changesets without git and the tests failing seem like they would break that?

That's interesting - because the original report was stating that the infinite loop has been caused by the lack of git in the container but that would mean that our tests would not be able to complete before this change here. So there has to be some additional thing about the setup of the reporter.

Gonna try to investigate this later.

kenrick95:
Hi @Andarist, are there any blockers on this PR? Seems like would be helpful to solve #571

yonran:
> Could you add the command that was run to the error that is thrown?

I ran into this infinite loop with no error printed in a GitHub Workflow when I ran `changeset version` in a shallow checkout without git credentials:

```
on:
    workflow_dispatch:
jobs:
    release:
        name: Create Release
        runs-on: ubuntu-latest
        permissions:
            contents: write
            packages: write
        steps:
            - uses: actions/checkout@v4
              with:
                  # don’t save GITHUB_TOKEN to .git/config; use credential helper instead
                  # because we don't want postinstall scripts
                  # to have access to write credentials
                  persist-credentials: false
            - name: Install node
              uses: actions/setup-node@v4
            - name: configure git to use $GIT_PASSWORD for github credential
              shell: bash
              run: |
                  set -eux
                  git config --global credential.https://github.com.helper '!f(){ test "$1" = get && printf "username=x-access-token\npassword=%s\n" "$GIT_PASSWORD"; }; f'
            - run: npm ci --ignore-scripts
            # - run: npm reinstall dependencies with postinstall scripts
            - run: npx changeset version
              env:
                # I didn’t realize that I need this:
                # GIT_PASSWORD: ${{ github.token }}
```

Steps to reproduce locally:

First, do a shallow clone

```
git clone --depth=1 --branch=<branch> https://github.com/<org>/<repo>.git
```

Then, remove github credentials
* e.g. if using git-credential-osxkeychain, then in Keychain Access, for the github.com Internet Password, remove git-credential-osxkeychain from authorized programs in Access Control
* e.g. if using `gh auth authorize-git`, remove the `[credential "https://gist.github.com "]` `helper = "!gh auth git-credential"` from `~/.gitconfig` or `~/.config/git/config/`

Then run `changeset version`

```
npm ci --ignore-scripts
npx changeset version
```

bluwy:
> We need to investigate the issue more to address the core of the problem though.

I believe #1045 may be the solution to the core problem

bluwy (approved):
I went ahead and revert some non-0 handling since it's impossible for them to return non-0. @Andarist can you check the code again?

Honestly, not really happy with the git functions, some doesn't error and returns a boolean, some throws an error, some does a `console.log`. Argument format is also a bit sporadic. But I don't really want to change those for now.

Andarist:
I agree this package isn't particularly consistent - it woul be nice to clean it up some time later.

PR #585, opened 2021-05-04
Add support for workspace version alias

javier-garcia-meteologica:
Adds support for `workspace:^` and `workspace:~`.


Andarist:
Could you point to any documentation about those two?

javier-garcia-meteologica:
This feature was recently merged into pnpm and at yarn they are working to implement it.

https://github.com/pnpm/pnpm/pull/3368
https://github.com/yarnpkg/berry/issues/2658

There is no documentation yet about this feature, but I should probably write about it.

javier-garcia-meteologica:
I was wrong, it's already documented in the [workspaces section of pnpm](https://pnpm.io/workspaces)

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -72,7 +72,11 @@ export default function getDependencyGraph(
       const expected = match.packageJson.version;
 
       if (depVersion.startsWith("workspace:")) {
-        depVersion = depVersion.substr(10);
+        const workspaceDepVersion = depVersion.substr(10);
+        depVersion =
+          workspaceDepVersion === "^" || workspaceDepVersion === "~"
```
Andarist, line 77: It kinda feels like we should fill in the true current versions here. I'm not super sure yet if that's the correct way of approaching that though - gonna need to give this a deeper thought to check how this influences both `@changesets/assemble-release-plan` and `@changesets/apply-release-plan`

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -72,7 +72,11 @@ export default function getDependencyGraph(
...
+            ? "*"
+            : workspaceDepVersion;
```
Andarist, line 79: Note to myself - we need to recheck how this works with prereleases. From what I recall `*` isn't always satisfied by prerelease versions.

javier-garcia-meteologica: I didn't know about that rule, you are right.

```js
> semver.satisfies('1.0.0-alpha.1', '*')                                  
false
```

So `workspace:*`, `workspace:^` and `workspace:~` should skip semver checks altogether and go directly to `dependencies.push(depName)`.

javier-garcia-meteologica: I added a test for prerelease versions. I've never done a prerelease, so if something doesn't make sense tell me.

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -70,15 +70,23 @@ export default function getDependencyGraph(
...
+        // Resolve workspace version aliases
+        depVersion =
+          workspaceDepVersion === "^" || workspaceDepVersion === "~"
+            ? "*"
+            : workspaceDepVersion;
       } else if (opts?.bumpVersionsWithWorkspaceProtocolOnly === true) {
         continue;
       }
 
+      const matchesAnyVersion = isWorkspaceProtocol && depVersion === "*";
```
Andarist, line 86: I would go for what you have suggested earlier - just call `dependencies.push(depName);` and `continue` the loop instead of faking `*` and handling it later

packages/get-dependents-graph/src/get-dependency-graph.ts
Andarist: > I added a test for prerelease versions.

Thank you, that is awesome and is of a great help ❤️ 

--- 

To reflect the actual usage you would have to enter the pre mode, like in those tests here:
https://github.com/atlassian/changesets/blob/8d0b27ddb3038156c741c741f706dfc47094740a/packages/cli/src/commands/version/version.test.ts#L583

packages/apply-release-plan/src/version-package.ts
```diff
@@ -59,7 +59,11 @@ export default function versionPackage(
           continue;
         const usesWorkspaceRange = depCurrentVersion.startsWith("workspace:");
         if (usesWorkspaceRange) {
-          depCurrentVersion = depCurrentVersion.substr(10);
+          const workspaceDepVersion = depCurrentVersion.substr(10);
+          depCurrentVersion =
+            workspaceDepVersion === "^" || workspaceDepVersion === "~"
```
Andarist, line 64: We are kinda losing the range type here so it will get computed to `""` few lines below it:
https://github.com/atlassian/changesets/blob/8d0b27ddb3038156c741c741f706dfc47094740a/packages/apply-release-plan/src/version-package.ts#L77
which will lead to `newNewRange` being set incorrectly:
https://github.com/atlassian/changesets/blob/8d0b27ddb3038156c741c741f706dfc47094740a/packages/apply-release-plan/src/version-package.ts#L79

And this might lead to updating the package json file incorrectly here:
https://github.com/atlassian/changesets/blob/8d0b27ddb3038156c741c741f706dfc47094740a/packages/apply-release-plan/src/index.ts#L133

However, I see that you have added an integration test for this as well:
https://github.com/atlassian/changesets/blob/8d0b27ddb3038156c741c741f706dfc47094740a/packages/cli/src/commands/version/version.test.ts#L358-L361

So I wonder - where my reasoning has failed? is `version` set here to a modifier? that would explain why `newNewRange` gets set to a correct value

Andarist (commented):
This looks great - I still need to review added tests carefully but I don't expect that there is anything wrong with them. It might take me a couple of days to merge this in so please be patient. In the meantime, you still need to sign the CLA so I could merge this.

packages/apply-release-plan/src/version-package.ts
javier-garcia-meteologica: I think the confusion comes because we are trying to solve the same problem from different perspectives. The problem is how to specify minor `^` and patch `~` dependencies pointing to the latest version during a release, but match any workspace version during development.

The previous solution was to specify workspace version ranges `workspace:^1.0.0`. When a dependency has a version bump, also bump all workspace version ranges of dependents.

The alternative is to create version aliases `workspace:^` and `workspace:~`. These aliases don't need to be bumped, the always point the latest version, they will be substituted with `^1.0.0` and `~1.0.0` during release.

Regarding the code, `workspace:*`, `workspace:^` and `workspace:~` will be substituted with `*`. Then the following condition will be false and versionPackage will do nothing. 

https://github.com/atlassian/changesets/blob/8d0b27ddb3038156c741c741f706dfc47094740a/packages/apply-release-plan/src/version-package.ts#L70-L76 

I added a new test for apply-release-plan to make it clear.

NMinhNguyen:
Sorry for being annoying, but wanted to ask if there's still plans to merge this?

Andarist (approved):

Andarist:
@javier-garcia-meteologica thank you for your contribution. This was well implemented and you have followed up on my suggestions and questions promptly. I'm very sorry I couldn't get to this sooner - I've only found some new energy to work on Changesets recently and I hope to get to any future PRs way quicker.

belgattitude:
@Andarist, just to say: huge thanks for your work. It's really appreciated.

Andarist:
Thank you - it means a lot. :)

packages/cli/src/commands/version/version.test.ts
```diff
@@ -362,6 +362,39 @@ describe("workspace range", () => {
       }
     ]);
   });
+
+  it("should bump dependant package when bumping a `workspace:^` dependency", async () => {
```
Andarist, line 366: I've found out that perhaps this test is incorrect. I would appreciate some second eyes on my reasoning though. 

`workspace:^` is basically equivalent to the `^${currentVersionOfThatDependency}`. So in this case this would be an equivalent of `^1.0.0`. Therefore if we just bump the `pkg-b` with a patch bump then... it shouldn't update the dependent pkg (`pkg-a`) as the new version fo the `pkg-b` (`1.0.1`) doesn't leave the current dependency range.

There are options for always bumping dependent packages etc but that's not the default.

F3n67u: I agree that we should not bump in this test. I tried to switch from `^1.0.0` to `workspace:^` to reference packages inside my monorepo and found the behavior is different.

PR #603, opened 2021-06-17
Fix the unhandled error from @changesets/git that is causing a infini…

### LINKED ISSUE #602: Unhandled error cause infinite loop in @changesets/git
## Affected Packages

* @changesets/git

## Problem

In some git solutions like Bitbucket, `git fetch --deepen [Number]` does not work.

When the command `changeset version` is runned in a step from `bitbucket-pipeline`, that is using a shallow clone, the step locks.

This happens because the `changeset version` try to get more commits from history using the `git fetch --deepen` command

## Proposed solution

Throw an error if the returning code from `git fetch --deepen [Number]` is different from 0.

https://github.com/atlassian/changesets/blob/main/packages/git/src/index.ts#L165

leometzger:
Resolves #602 

Throw an error when the command has not executed sucessfully.

### DIFF
--- a/.changeset/afraid-geckos-visit.md
+++ b/.changeset/afraid-geckos-visit.md
@@ -1,0 +1,5 @@
+---
+"@changesets/git": patch
+---
+
+Fix the unhandled error from @changesets/git that is causing infinite loop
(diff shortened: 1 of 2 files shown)


Andarist:
Could you fork from [this PR](https://github.com/atlassian/changesets/pull/584), apply suggestions from the PR, and make tests passing? I've started working on fixing the issue that you are facing there but didn't have time to wrap up this work. It would be lovely if you could pick it up from that point.

leometzger:
I have implemented a possible solution in https://github.com/atlassian/changesets/pull/611

PR #617, opened 2021-07-22
Update usage of since in changeset status example code

elisechant:


bluwy:
Looks like this has been updated in https://github.com/changesets/changesets/pull/734, but thanks for the contribution!

PR #634, opened 2021-09-08
Implement changeset tag and getAllTags. Resolves #612.

### LINKED ISSUE #612: Create git tags during version instead of publish
## Affected Packages

`@changesets/assemble-release-plan`
`@changesets/apply-release-plan`

## Problem

For the https://github.com/single-spa/create-single-spa monorepo, I use `changeset version` but use `pnpm publish -r` rather than `changeset publish`, as encouraged by https://pnpm.io/using-changesets.

The problem is that no git tags are created with those two commands:

| command | creates git tags? |
| --------- | ----------------- |
| `changeset version` | ❌  |
| `changeset publish` | ✅  |
| `pnpm version` | ✅  |
| `pnpm publish` | ❌  |

My ideal would be creating a separate git tag for each package in the monorepo - for example `webpack-config-single-spa@3.1.0` and `webpack-config-single-spa-react@2.1.0`

I first proposed a change to pnpm publish, but they thought the best place to implement the tagging would be during changeset version (see https://github.com/pnpm/pnpm/issues/3586).

### Option 1

Add a new flag `changeset version --git-tag` that commits and creates git tags. This would more closely match npm, yarn, and pnpm, all of which create git tags during their respective `version` commands.

### Option 2

Add a new command `changeset tag` that can be executed after `changeset version` and commit are completed, as an alternative to `changeset publish` for those using pnpm to publish rather than changeset

joeldenning:
See #612. This is my first pull request to this repository - let me know if I missed any steps for contributing.


packages/git/src/index.ts
```diff
@@ -26,6 +26,16 @@ async function commit(message: string, cwd: string) {
   return gitCmd.code === 0;
 }
 
+async function getAllTags(cwd: string): Promise<string[]> {
+  const gitCmd = await spawn("git", ["tag"], { cwd });
+  if (gitCmd.code === 0) {
+    const tags = gitCmd.stdout.toString().split("\n");
+    return tags;
+  } else {
+    throw Error("Unable to retrieve git tags");
```
joeldenning, line 35: I wasn't sure if this was the best way to report an error or not - note that the stdout and stderr of the git command are being swallowed which could be tricky for people trying to debug.

packages/cli/src/commands/tag/index.ts
```diff
@@ -0,0 +1,19 @@
+import * as git from "@changesets/git";
+import { getPackages } from "@manypkg/get-packages";
+import { log } from "@changesets/logger";
+
+export default async function run(cwd: string) {
+  const { packages } = await getPackages(cwd);
+
+  const allExistingTags = await git.getAllTags(cwd);
```
joeldenning, line 8: Wasn't sure if I should add a check to verify a clean git status before tagging? As far as I know, git tag will ignore any uncommitted changes, but the user might not have realized they haven't committed everything yet. Thoughts?

packages/git/src/index.ts
```diff
@@ -26,6 +26,16 @@ async function commit(message: string, cwd: string) {
   return gitCmd.code === 0;
 }
 
+async function getAllTags(cwd: string): Promise<string[]> {
+  const gitCmd = await spawn("git", ["tag"], { cwd });
+  if (gitCmd.code === 0) {
+    const tags = gitCmd.stdout.toString().split("\n");
```
Andarist, line 32: With my git version/config the `stdout` contains a trailing newline - so to be more precise we'd need this:
```suggestion
    const tags = gitCmd.stdout.toString().trim().split("\n");
```

Andarist: I'm also wondering if this would work on WIndows - is the line separator unified there in this context?

packages/git/src/index.ts
```diff
@@ -26,6 +26,16 @@ async function commit(message: string, cwd: string) {
   return gitCmd.code === 0;
 }
 
+async function getAllTags(cwd: string): Promise<string[]> {
+  const gitCmd = await spawn("git", ["tag"], { cwd });
```
Andarist, line 30: I'm slightly worried that tags might not always be available on CI when working with shallow gi clones and similar setups. I'm no git expert so this would require some research on the topic.

packages/git/src/index.ts
Andarist: In another PR of mine, I've just been throwing smth like `throw new Error(gitCmd.stderr.toString());`. This seems to address the issue that you have raised with stuff being swallowed. It's probably ain't perfect as that stderr won't be nicely formatted, and the failure code will still be somewhat lost - but it still seems like it provides more information to the user and as such, I would prefer something along those lines.

In the future, we might create a special `GitError` or something to handle this better.

docs/command-line-options.md
```diff
@@ -139,3 +140,15 @@ pre [exit|enter {tag}]
...
+
+```
+tag
+```
+
+The tag command creates git tags for the current version of all packages. The tags created are equivalent to those created by [`changeset publish`](#publish), but the `tag` command does not publish anything to npm.
+
+This is helpful in situations where a different tool, such as `pnpm publish -r`, is used to publish packages instead of changeset. For situations where `changeset publish` is executed, running `changeset tag` is unneeded. See [issue 612](https://github.com/atlassian/changesets/issues/612) for the original inspiration behind this command.
+
+The git tags are created in the format `pkg-name@version-number` and are based on the current version number of the package.json for each package. It is expected that [`changeset version`](#version) is run before `changeset tag`, so that the package.json versions are updated before the git tags are created.
```
Andarist, line 154: > The git tags are created in the format `pkg-name@version-number` and are based on the current version number of the package.json for each package.

This might not be entirely correct. The mentioned format is true for monorepos with defined workspaces, but Changesets also support single-package repositories and to make things more idiomatic in those scenarios the pattern is `v1.0.0`. 

> It is expected that [`changeset version`](#version) is run before `changeset tag`, so that the package.json versions are updated before the git tags are created.

Should we maybe validate this in the `tag` command? We could probably use `@changesets/read` and ensure that there are no pending changesets. WDYT?

.changeset/nervous-actors-protect.md
```diff
@@ -0,0 +1,10 @@
+---
+"@changesets/cli": minor
+"@changesets/git": minor
+---
+
+Add `changeset tag` and `getAllTags`
+
+The `changeset tag` command creates git tags for each package.
+
+The `getAllTags` function within @changesets/git retrieves all git tags for a repository.
```
Andarist, line 10: I think I would split this changeset file into 2. The `getAllTags` part is not relevant for people reading about the new feature in`@changesets/cli`.

Don't be afraid of describing a single PR from different angles :) Each touched package might often need a different summary of changes (the change should be - IMHO - described primarily from the perspective of that package).

Andarist:
@joeldenning no worries. Sorry for my late reply anyway - i was lately busy and didnt have time to work on Changesets much but i think i got some fresh energy for that (still not a lot of time though 😜). Hopefully, i should be more responsive in the coming weeks.

If you dont end up addressing the issues I will eventually pick this up and bring this through a finish line.

I would appreciate an answer on the question asked here though: https://github.com/atlassian/changesets/pull/634#discussion_r725699020 . Im rethinking how to reimplement publishing and this would help me to consider your use case.

packages/cli/src/commands/tag/index.ts
joeldenning: As far as I can tell, changesets publish does not verify a clean git status before publishing.

.changeset/nervous-actors-protect.md
joeldenning: 👍🏿  will update

packages/git/src/index.ts
joeldenning: I read through https://git-scm.com/docs/git-clone and tested locally with `git clone --depth=1`. I saw that the tags are included in a shallow clone unless the `--no-tags` flag is passed to git clone.

Different CI/CD tools run git clone with a wide variety of flags, so there may be some tools that pass the `--no-tags` flag. I looked at github's actions/checkout and it allows you to configure the `--depth` flag but I see no mention of the `--no-tags` flag. [This documentation](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches) indicates that it's quite easy to get all the git tags and branches, if needed. I didn't look at travis, circle, or any other CI/CD tools.

packages/git/src/index.ts
joeldenning: 👍🏿  i'll update my code

packages/git/src/index.ts
joeldenning: I booted up my windows laptop (despite a bsod!) and tested this out. Everything worked, although I ended up adding a `.filter(Boolean)` because it was extracting an empty string as a tag, even after I added the `.trim()`.

Here's the code I ran on my windows machine as a test:

```js
const spawn = require('spawndamnit');

spawn("git", ["tag"], { cwd: process.cwd() }).then(gitCmd => {
  const tags = gitCmd.stdout.toString().trim().split("\n").filter(Boolean);
  console.log("tags", tags);
})
```

I ran the test code in a repo with no tags, with one tag, and with two tags. It worked in all cases. Regarding windows new line characters, I'm not sure why the `"\n"` works there but I think it's because a lot of Windows programs use `\n\r` in their code, so that code looking for normal new lines or carriage returns all work. It also might be possible that NodeJS's spawn command does some cross-platform stuff when converting the stdout to a node string. Either way, it seems to work well on Windows.

docs/command-line-options.md
joeldenning: Re-reading the code I wrote, I don't see how the `v` will be present in the git tag. Were you saying that you'd like it to do that, to match `changeset publish`? If so, how can I detect whether the code is running in a single-package repo, so that I can insert the `v` in that situation during `changeset tag`?

joeldenning:
I pushed a couple new commits addressing the feedback, and responded to your question above.

docs/command-line-options.md
Andarist: > So I won't plan on updating changesets tag to do that check, to match publish

Good call 👍 

> If so, how can I detect whether the code is running in a single-package repo, so that I can insert the v in that situation during changeset tag?

This is what we currently do when publishing:
https://github.com/atlassian/changesets/blob/c1decd7b162e560b86ded277772fd8c2a89240b1/packages/cli/src/commands/publish/index.ts#L79-L89

docs/command-line-options.md
```diff
@@ -139,3 +140,15 @@ pre [exit|enter {tag}]
...
+
+## tag
+
+```
+tag
+```
+
+The tag command creates git tags for the current version of all packages. The tags created are equivalent to those created by [`changeset publish`](#publish), but the `tag` command does not publish anything to npm.
+
+This is helpful in situations where a different tool, such as `pnpm publish -r`, is used to publish packages instead of changeset. For situations where `changeset publish` is executed, running `changeset tag` is unneeded. See [issue 612](https://github.com/atlassian/changesets/issues/612) for the original inspiration behind this command.
```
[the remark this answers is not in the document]
Andarist: Both of the things mentioned could be supported by the `changeset publish`. The latter already is, since we literally call `npm publish` in the given directory. My plan for the former is to invoke the appropriate `pack` command and publish that (so replacing would happen within `pnpm`), this would - most likely - require us to also implement running some lifecycle scripts within changeset (instead of relying on them being executed within `npm publish`) but as far as I remember this (running the scripts) is exactly what `pnpm` does under the hood. So it looks like a valid strategy for us too.

packages/git/src/index.ts
joeldenning: I think it's reasonable that the `changeset tag` command would require that you've cloned the tags along with the repo. It's in the name :)

packages/git/src/index.ts
joeldenning: Here's the resulting output:

```sh
chars [
  118, 49, 46, 48, 46, 48,
  10, 118, 49, 46, 48, 46,
  49, 10
]
```

joeldenning:
I've responded to all comments and think I've addressed all of the review feedback - let me know if there's anything I looked over or other things to change.

docs/command-line-options.md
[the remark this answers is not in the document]
Andarist: @zkochan using smth like `pnpm pack && npm publish lib.tar.gz` would work though, right?

packages/git/src/index.ts
Andarist: That array is an equivalent of `'v1.0.0\nv1.0.1\n'`, with no hidden character of anything like that. So a single `.trim()` should already handle that:
```js
[118, 49, 46, 48, 46, 48, 10, 118, 49, 46, 48, 46, 49, 10 ].map(s => String.fromCharCode(s)).join('').trim().split("\n") // ['v1.0.0', 'v1.0.1']
```

I would prefer to avoid that extra `.filter(Boolean)` unless we can write a test that would fail without it.

Andarist:
@joeldenning thank you for your contribution and for bearing with me. I've made a TODO item for this week to land this - it's very close to that, just some nits and one last look at this. I can take care of all of that myself so we don't end up 🏓 each other any longer.

docs/command-line-options.md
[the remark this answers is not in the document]
zkochan: right

Andarist (approved):

Andarist:
Ufff, managed to land this before the end of the week 😉 

Thank you very much for your contribution ❤️

PR #638, opened 2021-09-16
Make master branch rename suggestion more clear

brianespinosa:
I was confused looking at the docs about the main branch name and I thought for a minute that it was telling me we should try to rename the main branch to master. I added the suggestion of an alternative.


brianespinosa:
I don't think a changeset is needed, but if someone wants me to add one, let me know.

Andarist (approved):

PR #640, opened 2021-09-20
Update globby dependency to fix vulnerability

Benjamin1333:
globby 8.0.0-9.2.0 depends on fast-glob <=2.2.7 which causes a oderate vulnerability. Maybe you wanna update the globby dep?

### DIFF
--- a/packages/get-workspaces/package.json
+++ b/packages/get-workspaces/package.json
@@ -9,7 +9,7 @@
   "dependencies": {
     "@changesets/types": "^4.0.1",
     "fs-extra": "^7.0.1",
-    "globby": "^9.2.0",
+    "globby": "^12.0.2",
     "read-yaml-file": "^1.1.0"
   },
   "devDependencies": {}


kossahl (approved):

packages/get-workspaces/package.json
```diff
@@ -9,7 +9,7 @@
   "dependencies": {
     "@changesets/types": "^4.0.1",
     "fs-extra": "^7.0.1",
-    "globby": "^9.2.0",
+    "globby": "^12.0.2",
```
Andarist, line 12: this can't be updated like this because this version is ESM-only and we still support CJS consumers

bluwy:
It doesn't look like we rely on the affected `globby` anymore if you update `@changesets/cli` to latest:

```
$ npm ls globby 
changesets-ls@ /Users/bjorn/Work/repros/changesets-ls
└─┬ @changesets/cli@2.27.9
  └─┬ @manypkg/get-packages@1.1.3
    └── globby@11.1.0
```

PR #646, opened 2021-10-04
Add Astro to projects using Changesets!

jasikpark:


Andarist (approved):
Astro is a cool project ❤️ if you find any problems with Changesets or in a need of some feature that we don't provide then don't hesitate and reach out to me.

PR #668, opened 2021-11-10
Add resolutions to package.json type definition

marcodejongh:
Technically workspace packages can have resolutions too, not just root level package.json's.
Ran into this missing when trying to use `@manypkg/get-packages` to clean up workspace resolutions

### DIFF
--- a/.changeset/hot-weeks-dance.md
+++ b/.changeset/hot-weeks-dance.md
@@ -2,4 +2,4 @@
 "@changesets/types": patch
 ---
 
-Add resolutions to packagejson type definition
+Add `resolutions` to the `PackageJSON` type.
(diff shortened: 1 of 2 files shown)


Andarist:
> Ran into this missing when trying to use @manypkg/get-packages to clean up workspace resolutions

Does manypkg depend on this type from the `@changesets/type` package?

marcodejongh:
@Andarist Correct, so if I do 
```
const { tool, root, packages } = await getPackages(process.cwd());
packages.map(pkg => pkg.packageJson.resolutions);
```

I get a typecheck error since resolutions is not on the interface, this PR addresses that

.changeset/hot-weeks-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/types": patch
+---
+
+Add resolutions to packagejson type definition
```
Andarist, line 5:
```suggestion
Add `resolutions` to the `PackageJSON` type.
```

Andarist (approved):

PR #674, opened 2021-11-20
Rework publishing to work with all popular package managers

### LINKED ISSUE #432: changeset publish does not replace workspace:* with current version in workspace before publishing
## Affected Packages

@changesets/apply-release-plan

and maybe others...

## Problem

@zkochan has already raised an issue with support of workspace ranges in dependencies, and he already proposed a solution for his immediate need of replacing workspace:x.y.z with x.y.z and it's working. However the case of "workspace:*" is not taken care of yet. The documented behavior in all package managers is for it to be replaced with the current version in the workspace.

changeset publish does not do that resulting in the NPM package being published with 'workspace:*' which obviously does not work.

## Proposed solution

Not sure, but @zkochan already had a comment regarding this at: https://github.com/atlassian/changesets/pull/311/files#r407058482 and to my knowledge yarn2 properly publishes such packages when invoked with 'yarn npm publish'. Not sure about pnpm.
Maybe the solution is to let the package manager handle this?

**Workaround so far:**
I am using yarn v2, so the workaround for me is ugly but produces consistent results. 
- Just stop at changeset version
- Use yarn npm publish to publish packages
- manually create git tags
- push to git

### LINKED ISSUE #598: [Feature Request] Support for Yarn 2 "workspace ranges"
## Affected Packages

`@changesets/cli`

## Problem

First off, thanks for the great tool!

Yarn 2 introduces a feature called [workspace ranges](https://yarnpkg.com/features/workspaces#workspace-ranges-workspace).

With workspace ranges, you can specify the intent of "depend on the current version of local workspace package, whatever the actual published version it may be". If you think about the workflow you're simutaneously modifying multiple packages, this can be extremely useful.

The transformation of `workspace:...` version to actual versions that npm can understand is handled by yarn at publish time.

> When a workspace is packed into an archive (whether it's through `yarn pack` or one of the publish commands like `yarn npm publish`), we dynamically replace any workspace: dependency by:
> 
> * The corresponding version in the target workspace (if you use * or a project-relative path)
> * The associated semver range (for any other range type)

However, [as current `changeset publish` command only supports either npm or pnpm as publish client](https://github.com/atlassian/changesets/blob/c426035565cfac518238c8bf32f3c496c66c0657/packages/cli/src/commands/publish/npm-utils.ts#L36-L56), this transformation cannot take place, and `changeset publish` aborts with:

```
🦋  error TypeError: Invalid comparator: packages/my-package
🦋  error     at Comparator.parse (/home/runner/work/npm-packages/npm-packages/.yarn/cache/semver-npm-5.7.1-40bcea106b-06ff0ed753.zip/node_modules/semver/semver.js:754:11)
🦋  error     at new Comparator (/home/runner/work/npm-packages/npm-packages/.yarn/cache/semver-npm-5.7.1-40bcea106b-06ff0ed753.zip/node_modules/semver/semver.js:737:8)
🦋  error     at Range.<anonymous> (/home/runner/work/npm-packages/npm-packages/.yarn/cache/semver-npm-5.7.1-40bcea106b-06ff0ed753.zip/node_modules/semver/semver.js:925:12)
🦋  error     at Array.map (<anonymous>)
🦋  error     at Range.parseRange (/home/runner/work/npm-packages/npm-packages/.yarn/cache/semver-npm-5.7.1-40bcea106b-06ff0ed753.zip/node_modules/semver/semver.js:924:13)
🦋  error     at Range.<anonymous> (/home/runner/work/npm-packages/npm-packages/.yarn/cache/semver-npm-5.7.1-40bcea106b-06ff0ed753.zip/node_modules/semver/semver.js:867:17)
🦋  error     at Array.map (<anonymous>)
🦋  error     at new Range (/home/runner/work/npm-packages/npm-packages/.yarn/cache/semver-npm-5.7.1-40bcea106b-06ff0ed753.zip/node_modules/semver/semver.js:866:40)
🦋  error     at versionPackage (/home/runner/work/npm-packages/npm-packages/.yarn/cache/@changesets-apply-release-plan-npm-5.0.0-19ff54c5d1-d43561e831.zip/node_modules/@changesets/apply-release-plan/dist/apply-release-plan.cjs.dev.js:160:9)
🦋  error     at /home/runner/work/npm-packages/npm-packages/.yarn/cache/@changesets-apply-release-plan-npm-5.0.0-19ff54c5d1-d43561e831.zip/node_modules/@changesets/apply-release-plan/dist/apply-release-plan.cjs.dev.js:312:12
```

## Proposed solution

If either

- `@changesets/cli` allows the configuration of publish command (Related issue: #580) or,
- it supports Yarn 2 out of the box

It would be awesome!

### LINKED ISSUE #813: Yarn not used for publishing
## Affected Packages
N/A

## Problem

Changesets doesn't seem to use yarn for publishing (only npm or pnpm) : 
https://github.com/changesets/changesets/blob/34933f041ab87d0fd1dc0250b894e011cf12da00/packages/cli/src/commands/publish/npm-utils.ts#L39

This forces users to maintain yarn and npm configurations in order for example to publish to private repos.
What's the reason behind this choice ?

## Proposed solution
Support yarn for publishing.

Thanks in advance.

### LINKED ISSUE #1454: Publishing with yarn doesn't work correctly
## Affected Packages

@changesets/cli

## Problem

Publishing with yarn should be done via `yarn npm publish` but `@changesets/cli` is using `npm publish` instead. This means some workspaces features that are yarn specific do not work as expected. For instance, when publishing with `yarn npm publish` yarn replaces inner packages with version `workspace:^` with the correct version. While when publishing with `npm publish` npm ignores packages with version `workspace:^`. This results in incorrectly published packages because when installing it package managers won't be able to resolve `workspace:^`.

I believe the above is also the reason why `.yarnrc.yml` is ignored when publishing via changesets.

This is simple to reproduce:
1. In a monorepo managed by yarn create packages "@example/a" and "@example/b".
2. The package.json of @example/a contains @example/b as a dependency:
```
"dependencies": {
  "@example/b": "workspace:^"
}
```
3. Create a changeset for @example/a.
4. Generate the new versions via `changeset version`.
5. Then publish via `changeset publish`.
6. Now install the new version in a new project and check the package.json in the `node_modules` and also the new info added to `yarn.lock`. You'll see `workspace:^` hasn't been replaced.

Related issues:
* https://github.com/changesets/changesets/issues/1411
* https://github.com/changesets/changesets/issues/650

## Proposed solution

The proposed solution is to use `yarn npm publish` instead of `npm publish`. I tested this out in a private monorepo and it works. I also opened a PR for this change [here](https://github.com/changesets/changesets/pull/1449) and would like help getting it reviewed and approved.

This issue is pretty urgent for us at my company. Any guidance on how to move this forward would be very much appreciated.

Andarist:
fixes https://github.com/atlassian/changesets/issues/598
fixes https://github.com/atlassian/changesets/issues/432
fixes https://github.com/changesets/changesets/issues/813
fixes https://github.com/changesets/changesets/issues/1454
(maybe https://github.com/changesets/changesets/issues/1411)

at the moment this is being opened as a draft, to gather initial feedback

The goal of this PR is to delegate as much as possible to yarn berry / pnpm so they can do their magic with rewriting package manifests and stuff

TODO:
- [ ] implement improved authentication setup for yarn berry in the https://github.com/changesets/action since it doesn't rely on `.npmrc`

cc @schickling @zkochan (might be interesting to you)

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -174,7 +174,11 @@ async function internalPublish(
     !isCI &&
     // yarn berry doesn't accept `--otp` and it asks for it on its own
     publishTool.name !== "yarn";
-  let publishFlags = opts.access ? ["--access", opts.access] : [];
+  let publishFlags = publishTool.name !== "yarn" ? ["--json"] : [];
+
+  if (opts.access) {
+    publishFlags.push("--access", opts.access);
+  }
   publishFlags.push("--tag", opts.tag);
 
   if (shouldHandleOtp && (await twoFactorState.isRequired)) {
@@ -189,7 +193,7 @@ async function internalPublish(
   };
   let { code, stdout, stderr } = await spawn(
     publishTool.name,
-    [...publishTool.args, "--json", ...publishFlags, ...publishTool.flags],
+    [...publishTool.args, ...publishFlags, ...publishTool.flags],
     {
       cwd: opts.cwd,
       env: Object.assign({}, process.env, envOverride)
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -36,23 +36,33 @@ function getCorrectRegistry(packageJson?: PackageJSON): string {
 
 async function getPublishTool(
   cwd: string
-): Promise<{ name: "npm" } | { name: "pnpm"; shouldAddNoGitChecks: boolean }> {
+): Promise<{ name: "npm" | "pnpm" | "yarn"; args: string[]; flags: string[] }> {
   const pm = await preferredPM(cwd);
-  if (!pm || pm.name !== "pnpm") return { name: "npm" };
-  try {
-    let result = await spawn("pnpm", ["--version"], { cwd });
-    let version = result.stdout.toString().trim();
-    let parsed = semver.parse(version);
-    return {
-      name: "pnpm",
-      shouldAddNoGitChecks:
-        parsed?.major === undefined ? false : parsed.major >= 5
-    };
-  } catch (e) {
-    return {
-      name: "pnpm",
-      shouldAddNoGitChecks: false
-    };
+  if (!pm) {
+    return { name: "npm", args: ["publish"], flags: [] };
+  }
+  const version = (await spawn(pm.name, ["--version"], { cwd })).stdout
+    .toString()
+    .trim();
+
+  const parsed = semver.parse(version)!;
+
+  switch (pm.name) {
+    case "npm":
+      return { name: "npm", args: ["publish"], flags: [] };
+    case "pnpm":
+      if (parsed.major < 5) {
+        return { name: "pnpm", args: ["publish"], flags: [] };
+      }
+      return { name: "pnpm", args: ["publish"], flags: ["--no-git-checks"] };
+    case "yarn":
+      // classic yarn doesn't do anything special when publishing, let's stick to the npm client in such a case
+      if (parsed.major < 2) {
+        return { name: "npm", args: ["publish"], flags: [] };
+      }
+      return { name: "yarn", args: ["npm", "publish"], flags: [] };
+    default:
+      return { name: "npm", args: ["publish"], flags: [] };
   }
 }
 
@@ -144,6 +154,14 @@ export let getOtpCode = async (twoFactorState: TwoFactorState) => {
   return askForOtpCode(twoFactorState);
 };
 
+const isOtpError = (error: any) => {
+  // The first case is no 2fa provided, the second is when the 2fa is wrong (timeout or wrong words)
+  return (
+    error.code === "EOTP" ||
+    (error.code === "E401" && error.detail.includes("--otp=<code>"))
+  );
+};
+
 // we have this so that we can do try a publish again after a publish without
 // the call being wrapped in the npm request limit and causing the publishes to potentially never run
 async function internalPublish(
@@ -152,14 +170,16 @@ async function internalPublish(
   twoFactorState: TwoFactorState
 ): Promise<{ published: boolean }> {
   let publishTool = await getPublishTool(opts.cwd);
+  let shouldHandleOtp =
+    !isCI &&
+    // yarn berry doesn't accept `--otp` and it asks for it on its own
+    publishTool.name !== "yarn";
   let publishFlags = opts.access ? ["--access", opts.access] : [];
   publishFlags.push("--tag", opts.tag);
-  if ((await twoFactorState.isRequired) && !isCI) {
+
+  if (shouldHandleOtp && (await twoFactorState.isRequired)) {
     let otpCode = await getOtpCode(twoFactorState);
     publishFlags.push("--otp", otpCode);
-  }
-  if (publishTool.name === "pnpm" && publishTool.shouldAddNoGitChecks) {
-    publishFlags.push("--no-git-checks");
   }
 
   // Due to a super annoying issue in yarn, we have to manually override this env variable
@@ -169,12 +189,34 @@ async function internalPublish(
   };
   let { code, stdout, stderr } = await spawn(
     publishTool.name,
-    ["publish", opts.cwd, "--json", ...publishFlags],
+    [
+      ...publishTool.args,
+      opts.cwd,
+      "--json",
+      ...publishFlags,
+      ...publishTool.flags
+    ],
     {
       env: Object.assign({}, process.env, envOverride)
     }
   );
+
   if (code !== 0) {
+    // yarn berry doesn't support --json and we don't attempt to parse its output to a machine-readable format
+    if (publishTool.name === "yarn") {
+      const output = stdout
+        .toString()
+        .trim()
+        .split("\n")
+        // this filters out "unnamed" logs: https://yarnpkg.com/advanced/error-codes/#yn0000---unnamed
+        // this includes a list of packed files and the "summary output" like: "Failed with errors in 0s 75ms"
+        // those are not that interesting so we reduce the noise by dropping them
+        .filter(line => !/YN0000:/.test(line))
+        .join("\n");
+      error(`an error occurred while publishing ${pkgName}:`, `\n${output}`);
+      return { published: false };
+    }
+
     // NPM's --json output is included alongside the `prepublish` and `postpublish` output in terminal
     // We want to handle this as best we can but it has some struggles:
     // - output of those lifecycle scripts can contain JSON
@@ -185,13 +227,7 @@ async function internalPublish(
       getLastJsonObjectFromString(stdout.toString());
 
     if (json?.error) {
-      // The first case is no 2fa provided, the second is when the 2fa is wrong (timeout or wrong words)
-      if (
-        (json.error.code === "EOTP" ||
-          (json.error.code === "E401" &&
-            json.error.detail.includes("--otp=<code>"))) &&
-        !isCI
-      ) {
+      if (shouldHandleOtp && isOtpError(json.error)) {
         if (twoFactorState.token !== null) {
           // the current otp code must be invalid since it errored
           twoFactorState.token = null;
@@ -206,10 +242,10 @@ async function internalPublish(
         json.error.detail ? "\n" + json.error.detail : ""
       );
     }
-
     error(stderr);
     return { published: false };
   }
+
   return { published: true };
 }
 
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -189,14 +189,9 @@ async function internalPublish(
   };
   let { code, stdout, stderr } = await spawn(
     publishTool.name,
-    [
-      ...publishTool.args,
-      opts.cwd,
-      "--json",
-      ...publishFlags,
-      ...publishTool.flags
-    ],
+    [...publishTool.args, "--json", ...publishFlags, ...publishTool.flags],
     {
+      cwd: opts.cwd,
       env: Object.assign({}, process.env, envOverride)
     }
   );
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -56,7 +56,7 @@ async function getPublishTool(
       }
       return { name: "pnpm", args: ["publish"], flags: ["--no-git-checks"] };
     case "yarn":
-      // classic yarn doesn't do anything special when publishing, let's stick to the npm client in such a case
+      // Yarn Classic doesn't do anything special when publishing, let's stick to the npm client in such a case
       if (parsed.major < 2) {
         return { name: "npm", args: ["publish"], flags: [] };
       }
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -34,36 +34,25 @@ function getCorrectRegistry(packageJson?: PackageJSON): string {
     : registry;
 }
 
+const getPublishToolVersion = async (name: string, cwd: string) =>
+  (await spawn(name, ["--version"], { cwd })).stdout.toString().trim();
+
 async function getPublishTool(
   cwd: string
-): Promise<{ name: "npm" | "pnpm" | "yarn"; args: string[]; flags: string[] }> {
-  const pm = await preferredPM(cwd);
-  if (!pm) {
-    return { name: "npm", args: ["publish"], flags: [] };
-  }
-  const version = (await spawn(pm.name, ["--version"], { cwd })).stdout
-    .toString()
-    .trim();
+): Promise<{ name: "npm" | "pnpm" | "yarn"; version: string }> {
+  const name = (await preferredPM(cwd))?.name || "npm";
 
-  const parsed = semver.parse(version)!;
+  const version = await getPublishToolVersion(name, cwd);
 
-  switch (pm.name) {
-    case "npm":
-      return { name: "npm", args: ["publish"], flags: [] };
-    case "pnpm":
-      if (parsed.major < 5) {
-        return { name: "pnpm", args: ["publish"], flags: [] };
-      }
-      return { name: "pnpm", args: ["publish"], flags: ["--no-git-checks"] };
-    case "yarn":
-      // Yarn Classic doesn't do anything special when publishing, let's stick to the npm client in such a case
-      if (parsed.major < 2) {
-        return { name: "npm", args: ["publish"], flags: [] };
-      }
-      return { name: "yarn", args: ["npm", "publish"], flags: [] };
-    default:
-      return { name: "npm", args: ["publish"], flags: [] };
+  if (name === "yarn" && semver.lt(version, "2.0.0")) {
+    // Yarn Classic doesn't do anything special when publishing, let's stick to the npm client in such a case
+    return {
+      name: "npm",
+      version: await getPublishToolVersion("npm", cwd)
+    };
   }
+
+  return { name, version };
 }
 
 export async function getTokenIsRequired() {
@@ -172,13 +161,15 @@ async function internalPublish(
   let publishTool = await getPublishTool(opts.cwd);
   let shouldHandleOtp =
     !isCI &&
-    // yarn berry doesn't accept `--otp` and it asks for it on its own
-    publishTool.name !== "yarn";
+    (publishTool.name === "yarn"
+      ? semver.gte(publishTool.version, "3.2.0-rc.8")
+      : true);
   let publishFlags = publishTool.name !== "yarn" ? ["--json"] : [];
 
   if (opts.access) {
     publishFlags.push("--access", opts.access);
   }
+
   publishFlags.push("--tag", opts.tag);
 
   if (shouldHandleOtp && (await twoFactorState.isRequired)) {
@@ -186,19 +177,24 @@ async function internalPublish(
     publishFlags.push("--otp", otpCode);
   }
 
+  if (publishTool.name === "pnpm" && semver.gte(publishTool.version, "5.0.0")) {
+    publishFlags.push("--no-git-checks");
+  }
+
   // Due to a super annoying issue in yarn, we have to manually override this env variable
   // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
   const envOverride = {
     npm_config_registry: getCorrectRegistry()
   };
-  let { code, stdout, stderr } = await spawn(
-    publishTool.name,
-    [...publishTool.args, ...publishFlags, ...publishTool.flags],
-    {
-      cwd: opts.cwd,
-      env: Object.assign({}, process.env, envOverride)
-    }
-  );
+  let { code, stdout, stderr } =
+    publishTool.name === "yarn"
+      ? await spawn("yarn", ["npm", "publish", ...publishFlags], {
+          cwd: opts.cwd,
+          env: Object.assign({}, process.env, envOverride)
+        })
+      : await spawn(publishTool.name, ["publish", opts.cwd, ...publishFlags], {
+          env: Object.assign({}, process.env, envOverride)
+        });
 
   if (code !== 0) {
     // yarn berry doesn't support --json and we don't attempt to parse its output to a machine-readable format
--- a/.changeset/dry-cows-wait.md
+++ b/.changeset/dry-cows-wait.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+Refactored the publishing part of the code to invoke Yarn Berry instead of npm CLI for publishing purposes. This ensures that packages are correctly packed, so for instance workspace ranges can be correctly substituted now.
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -48,7 +48,7 @@
     "meow": "^6.0.0",
     "outdent": "^0.5.0",
     "p-limit": "^2.2.0",
-    "preferred-pm": "^3.0.0",
+    "preferred-pm": "^3.0.3",
     "semver": "^5.4.1",
     "spawndamnit": "^2.0.0",
     "term-size": "^2.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3523,10 +3523,18 @@ find-up@^4.0.0, find-up@^4.1.0:
     locate-path "^5.0.0"
     path-exists "^4.0.0"
 
-find-yarn-workspace-root2@^1.2.11:
-  version "1.2.11"
-  resolved "https://registry.yarnpkg.com/find-yarn-workspace-root2/-/find-yarn-workspace-root2-1.2.11.tgz#b14a8be43503ec8c0cdf7f5923a3534860bdd92e"
-  integrity sha512-TkwH3kDD1Z8X9LnrU3gbyYw8KtLZZjsym6c2YIRFCDIMJD1JL9+Blkzpm07db+Q6ZVLISqrW9cN0CvXN4rr5ag==
+find-up@^5.0.0:
+  version "5.0.0"
+  resolved "https://registry.yarnpkg.com/find-up/-/find-up-5.0.0.tgz#4c92819ecb7083561e4f4a240a86be5198f536fc"
+  integrity sha512-78/PXT1wlLLDgTzDs7sjq9hzz0vXD+zn+7wypEe4fXQxCmdmqfGsEPQxmiCSQI3ajFV91bVSsvNtrJRiW6nGng==
+  dependencies:
+    locate-path "^6.0.0"
+    path-exists "^4.0.0"
+
+find-yarn-workspace-root2@1.2.16:
+  version "1.2.16"
+  resolved "https://registry.yarnpkg.com/find-yarn-workspace-root2/-/find-yarn-workspace-root2-1.2.16.tgz#60287009dd2f324f59646bdb4b7610a6b301c2a9"
+  integrity sha512-hr6hb1w8ePMpPVUK39S4RlwJzi+xPLuVuG8XlwXU3KD5Yn3qgBWVfy3AzNlDhWvE1EORCE65/Qm26rFQt3VLVA==
   dependencies:
     micromatch "^4.0.2"
     pkg-dir "^4.2.0"
@@ -5198,6 +5206,13 @@ locate-path@^5.0.0:
   dependencies:
     p-locate "^4.1.0"
 
+locate-path@^6.0.0:
+  version "6.0.0"
+  resolved "https://registry.yarnpkg.com/locate-path/-/locate-path-6.0.0.tgz#55321eb309febbc59c4801d931a72452a681d286"
+  integrity sha512-iPZK6eYjbxRu3uB4/WZ3EsEIMJFMqAoopl3R+zuq0UjcAm/MO6KCweDgPfP3elTztoKP3KtnVHxTn2NHBSDVUw==
+  dependencies:
+    p-locate "^5.0.0"
+
 lodash.sortby@^4.7.0:
   version "4.7.0"
   resolved "https://registry.yarnpkg.com/lodash.sortby/-/lodash.sortby-4.7.0.tgz#edd14c824e2cc9c1e0b0a1b42bb5210516a42438"
@@ -5897,6 +5912,13 @@ p-locate@^4.1.0:
   dependencies:
     p-limit "^2.2.0"
 
+p-locate@^5.0.0:
+  version "5.0.0"
+  resolved "https://registry.yarnpkg.com/p-locate/-/p-locate-5.0.0.tgz#83c8315c6785005e3bd021839411c9e110e6d834"
+  integrity sha512-LaNjtRWUBY++zB5nE/NwcaoMylSPk+S+ZHNB1TzdbMJMny6dynpAGt7X/tl/QYq3TIeE6nxHppbo2LGymrG5Pw==
+  dependencies:
+    p-limit "^3.0.2"
+
 p-map@^2.0.0:
   version "2.1.0"
   resolved "https://registry.yarnpkg.com/p-map/-/p-map-2.1.0.tgz#310928feef9c9ecc65b68b17693018a665cea175"
@@ -6113,13 +6135,13 @@ posix-character-classes@^0.1.0:
   resolved "https://registry.yarnpkg.com/posix-character-classes/-/posix-character-classes-0.1.1.tgz#01eac0fe3b5af71a2a6c02feabb8c1fef7e00eab"
   integrity sha1-AerA/jta9xoqbAL+q7jB/vfgDqs=
 
-preferred-pm@^3.0.0:
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/preferred-pm/-/preferred-pm-3.0.0.tgz#26cc7fbbfc1815e4cb8272d5fdde15dc96220fb8"
-  integrity sha512-NbN+2UuqjakJpyHamsuIWyeFdQcFUQHF9nkw16hpFE++z3px+/KDsj+AF1h0BlnsBJi1Z5U4EKBW7XnHriny8g==
+preferred-pm@^3.0.3:
+  version "3.0.3"
+  resolved "https://registry.yarnpkg.com/preferred-pm/-/preferred-pm-3.0.3.tgz#1b6338000371e3edbce52ef2e4f65eb2e73586d6"
+  integrity sha512-+wZgbxNES/KlJs9q40F/1sfOd/j7f1O9JaHcW5Dsn3aUUOZg3L2bjpVUcKV2jvtElYfoTuQiNeMfQJ4kwUAhCQ==
   dependencies:
-    find-up "^4.1.0"
-    find-yarn-workspace-root2 "^1.2.11"
+    find-up "^5.0.0"
+    find-yarn-workspace-root2 "1.2.16"
     path-exists "^4.0.0"
     which-pm "2.0.0"
 


emmatown (commented):
Maybe we should log stdout and stderr when publish fails so people can debug when something goes wrong like we fail to parse the json? (and maybe even land that and the `getLastJsonObjectFromString` stuff separately to package manager changes?)

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -35,23 +36,33 @@ function getCorrectRegistry(packageJson?: PackageJSON): string {
...
+  switch (pm.name) {
+    case "npm":
+      return { name: "npm", args: ["publish"], flags: [] };
+    case "pnpm":
+      if (parsed.major < 5) {
+        return { name: "pnpm", args: ["publish"], flags: [] };
+      }
+      return { name: "pnpm", args: ["publish"], flags: ["--no-git-checks"] };
+    case "yarn":
+      // classic yarn doesn't do anything special when publishing, let's stick to the npm client in such a case
```
belgattitude, line 59: "classic yarn"  might be confusing as yarn 2+ use classic install mode (nmLinker: module). 

Suggested: "yarn v1 doesn't do..."

Andarist: Yarn Classic is the official "name" of the Yarn v1 - where the new one is called Yarn Berry. So I think this is using the right name here 🤔

belgattitude: Looks fine then, thx @Andarist

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -151,57 +170,81 @@ async function internalPublish(
...
     npm_config_registry: getCorrectRegistry()
   };
-  let { stdout } = await spawn(
+  let { code, stdout, stderr } = await spawn(
     publishTool.name,
-    ["publish", opts.cwd, "--json", ...publishFlags],
+    [
+      ...publishTool.args,
+      opts.cwd,
+      "--json",
```
belgattitude, line 195: Looks like if yarn does not support --json, the argument should be conditionally added

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -151,57 +170,81 @@ async function internalPublish(
...
   const envOverride = {
     npm_config_registry: getCorrectRegistry()
   };
-  let { stdout } = await spawn(
+  let { code, stdout, stderr } = await spawn(
     publishTool.name,
-    ["publish", opts.cwd, "--json", ...publishFlags],
+    [
+      ...publishTool.args,
+      opts.cwd,
```
belgattitude, line 194: `opts.cwd`

I don't think it will work with yarn 2+, not sure about the other package managers (https://yarnpkg.com/cli/npm/publish has no reference to path)

I think this should work

```
 const args = [
    ...publishTool.args,
   // "--json", 
    ...publishFlags,
    ...publishTool.flags
  ];
  let { code, stdout, stderr } = await spawn(
    publishTool.name,
    args,
    {
      cwd: opts.cwd,
      env: Object.assign({}, process.env, envOverride)
    }
  );
``` 

Just tested it and it works on yarn 3. But I'm wondering if it would be cleaner to extract the command and add tests on it. I'll have a look tomorrow to see if I can help

packages/cli/src/commands/publish/npm-utils.ts
Andarist: yes, thank you for spotting that - this is an obvious oversight on my part, I knew that this ain't supported as it makes me handle yarn berry in a different way just a few lines later

just out of curiosity - is using an unknown flag makes yarn throw an error?

belgattitude: Unfortunately yes, just tested with your current P/R to be sure:

https://github.com/soluble-io/cache-interop/pull/222#issuecomment-979372099

BTW I love the butterfly in the yarn output, made me smile :)

```
🦋  error Unknown Syntax Error: Extraneous positional argument ("/home/sebastien/github/cache-interop/packages/cache-redis").
```

packages/cli/src/commands/publish/npm-utils.ts
Andarist: Ah, yes - good catch. I got `Unknown Syntax Error: Extraneous positional argument ("packages/test").` with this.

I think it's safe to just use `{ cwd: opts.cwd }` for all package managers.

packages/cli/src/commands/publish/npm-utils.ts
Andarist: I think you have tested the other case 😅 but this one doesn't work either:
> Unknown Syntax Error: Unsupported option name ("--json").

Andarist:
@belgattitude thank you very much for testing this ❤️ I will make necessary changes here soon, I would only want to land this first https://github.com/atlassian/changesets/pull/676 so I could rebase this one on top of it.

packages/cli/src/commands/publish/npm-utils.ts
Andarist: > Just tested it and it works on yarn 3. But I'm wondering if it would be cleaner to extract the command and add tests on it. I'll have a look tomorrow to see if I can help

Tests for this whole thing would be great. I'm just unsure how to test this best - we really should have some e2e tests here because there are so many little details hidden within those publish commands that we really should invoke those and verify what can be observed somehow.

packages/cli/src/commands/publish/npm-utils.ts
belgattitude: Yep :sweat_smile:, I've been struggling a bit with how to build your P/R to use it...

But that definitely happened and it was nice to see.

Andarist: I will make sure to setup snapshot releases for PRs before we go into re-testing this after I apply the suggested changes

packages/cli/src/commands/publish/npm-utils.ts
belgattitude: Totally, me I would be lazy and just unit test something in the lines of

```typescript
type Params= {
   pm: 'pnpm' | 'yarn' | 'npm',
   pmVersion: number,
   tag: string,
   access: 'public' | 'private'
}
const getPublishSpawnParams = (params): {cmd: string, args: string[]} => {
 ///....
}

describe('when pnpm', () => {
   it('should return a pnpm..', () => {
      expect(getPublishSpawnParams({pm: 'pnpm', pmVersion: '5', tag: 'main', access: 'public', otp: 'xxx'}).toStrictEqual({
          cmd: 'pnpm',
          args: ['publish', '--access', 'public', '--tag', 'main', '--json']
      })
   }); 
}

describe('when yarn < 2', () => {
   it('should return a npm publish..', () => {
      expect(getPublishSpawnParams({pm: 'yarn', pmVersion: '1', tag: 'main', access: 'public', otp: 'xxx'}).toStrictEqual({
          cmd: 'npm',
          args: ['publish', '--access', 'public', '--tag', 'main', '--otp', 'xxx']
      })
   }); 
}

describe('when yarn 2+', () => {
   it('should return a yarn..', () => {
      expect(getPublishSpawnParams({pm: 'yarn', pmVersion: '2', tag: 'main', access: 'public', otp: 'xxx'}).toStrictEqual({
          cmd: 'yarn',
          args: ['publish', '--access', 'public', '--tag', 'main']
      })
   }); 
}
```

Not so sure about how to mock spawn, so that might make sense

I'm bad with the names hey :)

belgattitude: I'm just checking and you're totally right so much small details, that's insane :)

So definitely I would live well with unit tests cause e2e looks a huge effort, especially with minor versions.

I would start with some specs for the most common cases (npm, yarn 1/2+, pnpm 6) and live with a "dirty" function that can be improved later (with things like https://www.npmjs.com/package/compare-versions to adapt the result)

Andarist:
@belgattitude I've rebased this and applied your suggestions. I've also created a PR to add snapshot releases ([here](https://github.com/atlassian/changesets/pull/681)). This can only be merged in when we confirm that it has been done in a secure way, so I don't want to rush it. If you could retest this PR "manually" (like you have done before) - I would highly appreciate it.

belgattitude:
Hey @Andarist thanks a lot. Here it is

<details>
  <summary>yarn changeset publish</summary>

```
sebastien@seb-thinkpad-x1:~/github/cache-interop$ yarn changeset publish
🦋  warn ===============================IMPORTANT!===============================
🦋  warn You are in prerelease mode so packages will be published to the canary
🦋  warn         dist tag except for packages that have not had normal releases which will be published to latest
🦋  warn ----------------------------------------------------------------------
🦋  info npm info @soluble/cache-interop
🦋  info npm info @soluble/cache-ioredis
🦋  info npm info @soluble/cache-redis
🦋  info npm info @soluble/dsn-parser
🦋  info @soluble/cache-interop is being published because our local version (0.8.2-canary.0) has not been published on npm
🦋  info @soluble/cache-ioredis is being published because our local version (0.8.4-canary.0) has not been published on npm
🦋  info @soluble/cache-redis is being published because our local version (0.8.4-canary.0) has not been published on npm
🦋  info @soluble/dsn-parser is being published because our local version (1.3.3-canary.0) has not been published on npm
🦋  info Publishing "@soluble/cache-interop" at "0.8.2-canary.0"
🦋  info Publishing "@soluble/cache-ioredis" at "0.8.4-canary.0"
🦋  info Publishing "@soluble/cache-redis" at "0.8.4-canary.0"
🦋  info Publishing "@soluble/dsn-parser" at "1.3.3-canary.0"
🦋  info will publish with 'yarn npm publish --access public --tag canary'
🦋  info will publish with 'yarn npm publish --access public --tag canary'
🦋  info will publish with 'yarn npm publish --access public --tag canary'
🦋  info will publish with 'yarn npm publish --access public --tag canary'
🦋  error an error occurred while publishing @soluble/dsn-parser: 
🦋  error ➤ YN0033: No authentication configured for request
🦋  error an error occurred while publishing @soluble/cache-ioredis: 
🦋  error ➤ YN0033: No authentication configured for request
🦋  error an error occurred while publishing @soluble/cache-redis: 
🦋  error ➤ YN0033: No authentication configured for request
🦋  error an error occurred while publishing @soluble/cache-interop: 
🦋  error ➤ YN0033: No authentication configured for request
🦋  @soluble/cache-interop@0.8.2-canary.0
🦋  @soluble/cache-ioredis@0.8.4-canary.0
🦋  @soluble/cache-redis@0.8.4-canary.0
🦋  @soluble/dsn-parser@1.3.3-canary.0

```
</details>

The command is not complaining anymore with yarn 3.2.0-rc.6.

> 🦋  info will publish with 'yarn npm publish --access public --tag canary'

PS: I've added the log here: https://github.com/atlassian/changesets/pull/682/files just to be sure.

That said it's not a full publish, I'll need --otp for that and [can only publish my repo](https://github.com/soluble-io/cache-interop/pull/222) from a gh action for now.

So all look fine. My concerns are related to

- Logging: As it look probable we run into bugs, I would add some logging about the command used.
- Prevent regressions: I feel we should be still have some kind of unit tests for the command (and pm versions). To ensure future change does not break things.

Of course all is up to you, I come from different ecosystem so my reflexes often lack perspective and love to learn and share in general.

Let me know I still have few free days, and I'll gladly help.

Andarist:
> Logging: As it look probable we run into bugs, I would add some logging about the command used.

That's probably a good call - anything that you would like to see logged beside the information added in your PR [here](https://github.com/atlassian/changesets/pull/682/files#diff-55663caf0b01c5bac3a7d6c3322f12c33fca10b13e116cfc05ef2e16d7582a5fR197)?

> Prevent regressions: I feel we should still have some kind of unit tests for the command (and pm versions). To ensure future changes does not break things.

I agree that we need some tests here - I'm just on the fence when it comes to unit tests here. The logic itself is pretty straightforward - the only problem is that I'm not always sure what exact combination of options has to be passed in a specific situation. And when it comes to that I think that comments can be more valuable than unit tests. 

I just kinda feel that I would be that more confident about this logic if I would have those unit tests. I'm not a testing expert though and probably many people would say that this would be worthwhile - I'm just not feeling that. Maybe @mitchellhamilton would have some additional opinion about this.

> The command is not complaining anymore with yarn 3.2.0-rc.6.

Just to make sure that we are on the same page - everything looks OK now from your end, it's just that publish fails because OTP is required and you are not using the automation token because you are running the publish script locally and not on the CI. Right?

mitchellhamilton:
> testing

tbh, what I'd really like is tests that run with all the different package managers against a local registry (probably https://verdaccio.org) with a way to run the tests against actual registries.

Andarist:
Let's move the tests-related discussion to [the dedicated thread](https://github.com/atlassian/changesets/issues/107). I don't think this is sort of out of the scope of this PR and should be addressed separately.

belgattitude:
> anything that you would like to see logged beside the information added in your PR here?

New to the project so I don't know but I would follow the concerns from https://github.com/atlassian/changesets/pull/674#pullrequestreview-811980332 @mitchellhamilton 

> Just to make sure that we are on the same page - everything looks OK now from your end, it's just that publish fails because OTP is required

Yep. Don't want to [mess](https://github.com/soluble-io/cache-interop/pull/222/files#diff-3f2bf3353c2a9d4bbdd723e6c7e58361082234fba5cf6fdc3595a11ff729effc) too much my repo :smile:. A canary of changeset can help for me to test it out fully, but as the "yarn command" is okay, I feel I can be confident enough the tool will do it's job. 

> About tests

Just want to express something about unit.  I'm certainly not trying to create a debate, just sharing and if it's not useful, please forget.

1. Don't test what you don't own.

    I like the idea of e2e, but in this specific case I would be confident that:

    - given the package manager cli receive the correct params it will do the job or it's a bug on their side. They have an e2e suite that runs on many os... spawn (which I don't know about) has it's own tests too, I guess. So for me what is to test is the command we run, less the fact that npm, pnpm, yarn 1.2.3 will actually publish what what we are feeding to it. Command line is somehow a contract. Same idea when I test a server, do I need to test if my express server works or just my handler/use-case (business...) .... There's not big difference when you write the tests, but running them is much easier when you don't need to start the actual express server (time, flacky...).

2. What kind of unit tests

   For example we could list the commands that are expected by each tool (from their doc). To illustrate (I haven't checked just taken from what I read in this P/R)

    | Given                |  Expected         |
    | -------------------- | -------------------- |
    | Yarn 1  --otp     | `npm publish --access public --tag main --otp xxx --json` |
    | Yarn 1              | `npm publish --access public --tag main --json`  |
    | Yarn 2              | `yarn npm publish --access public --tag main`   |
    | Yarn 2  --otp     | *should throw 'Unsupported --otp, upgrade to yarn 3.2+'*  |
    | Yarn 3.2 --otp   | `yarn npm publish --access public --tag main --otp xxx`   |
    | Pnpm 6            | `yarn npm publish --access public --tag main --json --no-git-checks`   |
    | Npm 6  --otp    | `npm publish --access public --tag main --otp xxx --json`  |
    | Npm 8  --otp    | `npm publish --access public --tag main --otp xxx --json`  |

    That could be written in many ways... In this scenario looks a fit for kind of acceptance / BDD / cucumber style. either given/when/expect like [here](https://github.com/atlassian/changesets/pull/674#discussion_r757142693) or even with a simple it.each and an array of scenarios. 

    That should suffice to prevent regressions, as you said the code is trivial but how much variations will come in the future. Personally as a contributor (If I become, I'll become :smile: ) I would be more confident to make changes.

If you like I could add an example test file in a separate P/R... The only thing is that I don't really know node, so a reviewer/guide will be needed :smile: 

Hey thanks again

belgattitude:
@Andarist a little addition. I'm not sure you're aware of it. When I faced the issue with yarn/pnpm workspace: protocol, I just did a quick hack for my own use case. Rather than change npm for publication, I intercepted the pack step (yarn pack, pnpm pack) to generate the archive. see here https://github.com/atlassian/changesets/pull/585#issuecomment-940756397.

I'm thinking of it now, I guess it's a bit late to tell and I don't want to add confusion here. So feel free to ignore, but it might help to harmonize and have one tool to test for publishing.

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -169,12 +193,29 @@ async function internalPublish(
   };
   let { code, stdout, stderr } = await spawn(
     publishTool.name,
-    ["publish", opts.cwd, "--json", ...publishFlags],
+    [...publishTool.args, ...publishFlags, ...publishTool.flags],
     {
+      cwd: opts.cwd,
```
Andarist, line 198: I think that `opts.cwd` might have been passed as an argument before so that `.npmrc` would be taken from the root of the monorepo - since older versions of npm did not understand workspaces the support for them was implemented in other tools (like Changesets). I would still have to confirm that but this is my worry.

ragingwind:
@mitchellhamilton  any updates?

Andarist:
@ragingwind I still have some minor things to adjust here and to figure out the authentication story in `changesets/action`. I didn't have much time to spend on this lately - but I will be working on landing this PR as soon as I can.

ragingwind:
@Andarist Thank you. It would be great I could use it with prerelease. We use yarn@3.1.0 now

ilyaGurevich:
@Andarist definitely looking forward to this PR, we are attempting to upgrade our monorepo from yarn v1 to yarn v3 and would like to also retain our conventional-commits/semantic releases that lerna used to give but is no longer compatible with yarn berry

Andarist:
@ragingwind I would say that at the moment using this packed version from CodeSandbox CI is the easiest:
```
@changesets/cli@https://pkg.csb.dev/changesets/changesets/commit/98401c27/@changesets/cli/_pkg.tgz
```
Since you are using Yarn 3 you'd have to use [this Yarn plugin](https://github.com/yarnpkg/berry/issues/3327#issuecomment-995886724) to resolve requested dependencies correctly. 

You could also try installing this version from npm:
```
@changesets/cli@2.19.0-temp.0
```
but it's a little behind the one available on CodeSandbox CI - and any future prereleases like that will only be available through CodeSandbox CI so that's what I recommend now.

Note that if you are also using `changesets/actions` then you should authenticate correctly before calling our action, otherwise Yarn won't be able to authenticate publish requests. I'm a little bit worried that `npm info` calls might fail either way - but if you could confirm that then I would have an actionable item to fix as part of this PR.

 > and would like to also retain our conventional-commits/semantic releases that lerna used to give

@ilyaGurevich note that how Changesets work is slightly different from that - the generated changelog updates etc are not based on commit messages but on the committed changeset `.md` files

unional:
Also notice that `changeset` is reading registry settings from `.npmrc` instead of `.yarnrc.yml` even the repo is using `yarn` instead of `npm`.

7rulnik:
@Andarist do you need some additional help to finish it?

gamertike:
hw is this pr going and when will i be able to use this? atm moment i really need this because my publishing automation for my project is broken due to yarn not being supported.

terenceodonoghue:
Just checking on this PR, keen to see it finished. @Andarist if you need any help (or sponsorship) to see it through, I'd be happy to help, it's an awesome tool otherwise. 🦋

scottdickerson:
are there any updates on this, we really miss the great capabilities of `yarn changeset publish`

tianyingchun:
how about this issue progress? really need this feature without any hack.

tianyingchun:
@Andarist using `@changesets/cli@2.22.0-temp.0` publish failed.

➤ YN0000: Done in 7s 668ms
🦋  info npm info @armit/cli
🦋  info npm info @armit/common
🦋  info npm info @armit/eslint-config-bases
🦋  info @armit/cli is being published because our local version (0.0.5) has not been published on npm
🦋  info @armit/common is being published because our local version (0.0.5) has not been published on npm
🦋  info @armit/eslint-config-bases is being published because our local version (0.0.4) has not been published on npm
🦋  info Publishing "@armit/cli" at "0.0.5"
🦋  info Publishing "@armit/common" at "0.0.5"
🦋  info Publishing "@armit/eslint-config-bases" at "0.0.4"
🦋  error an error occurred while publishing @armit/eslint-config-bases: 
🦋  error ➤ YN0000: CHANGELOG.md
🦋  error ➤ YN0000: README.md
🦋  error ➤ YN0000: package.json
🦋  error ➤ YN0000: src/bases/index.js
🦋  error ➤ YN0000: src/bases/jest.js
🦋  error ➤ YN0000: src/bases/playwright.js
🦋  error ➤ YN0000: src/bases/prettier.js
🦋  error ➤ YN0000: src/bases/react.js
🦋  error ➤ YN0000: src/bases/regexp.js
🦋  error ➤ YN0000: src/bases/rtl.js
🦋  error ➤ YN0000: src/bases/sonar.js
🦋  error ➤ YN0000: src/bases/storybook.js
🦋  error ➤ YN0000: src/bases/tailwind.js
🦋  error ➤ YN0000: src/bases/typescript.js
🦋  error ➤ YN0000: src/bases/vitest.js
🦋  error ➤ YN0000: src/helpers/getDefaultIgnorePatterns.js
🦋  error ➤ YN0000: src/helpers/getPrettierConfig.js
🦋  error ➤ YN0000: src/helpers/index.js
🦋  error ➤ YN0000: src/index.js
🦋  error ➤ YN0000: src/patch/modern-module-resolution.js
🦋  error ➤ YN0000: src/prettier.base.config.js
🦋  error ➤ YN0033: No authentication configured for request
🦋  error an error occurred while publishing @armit/cli: 
🦋  error ➤ YN0000: CHANGELOG.md
🦋  error ➤ YN0000: LICENSE
🦋  error ➤ YN0000: README.md
🦋  error ➤ YN0000: bin/armit.mjs
🦋  error ➤ YN0000: dist/index.d.ts
🦋  error ➤ YN0000: dist/index.js
🦋  error ➤ YN0000: dist/index.js.map
🦋  error ➤ YN0000: index.js
🦋  error ➤ YN0000: package.json
🦋  error ➤ YN0033: No authentication configured for request
🦋  error an error occurred while publishing @armit/common: 
🦋  error ➤ YN0000: CHANGELOG.md
🦋  error ➤ YN0000: LICENSE
🦋  error ➤ YN0000: README.md
🦋  error ➤ YN0000: dist/index.d.ts
🦋  error ➤ YN0000: dist/index.js
🦋  error ➤ YN0000: dist/index.js.map
🦋  error ➤ YN0000: index.js
🦋  error ➤ YN0000: package.json
🦋  error ➤ YN0033: No authentication configured for request
🦋  @armit/cli@0.0.5
🦋  @armit/common@0.0.5
🦋  @armit/eslint-config-bases@0.0.4
Error: The process '/usr/local/bin/yarn' failed with exit code 1
    at m._setResult (/home/runner/work/_actions/changesets/action/v1/dist/index.js:102:7258)
Error: The process '/usr/local/bin/yarn' failed with exit code 1
    at m.CheckComplete (/home/runner/work/_actions/changesets/action/v1/dist/index.js:102:[66](https://github.com/armitjs/armit/actions/runs/3301463560/jobs/5446958342#step:5:67)86)
    at ChildProcess.<anonymous> (/home/runner/work/_actions/changesets/action/v1/dist/index.js:102:5[72](https://github.com/armitjs/armit/actions/runs/3301463560/jobs/5446958342#step:5:73)3)
    at ChildProcess.emit (events.js:314:20)
    at maybeClose (internal/child_process.js:1022:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:2[87](https://github.com/armitjs/armit/actions/runs/3301463560/jobs/5446958342#step:5:88):5)

HosseinAgha:
@Andarist could you please rebase with the main branch? I need the latest main branch changes in the csb releases.

jmatsushita:
I've been trying out this PR with @qnighy's patch and with yarn `4.0.0-rc.45`

I get the following error message when running `yarn changeset publish`:
```
🦋  error error while checking if token is required npm ERR! code E401
🦋  error npm ERR! 401 Unauthorized - GET https://registry.npmjs.org/-/npm/v1/user
🦋  error {
🦋  error   "error": {
🦋  error     "code": "E401",
🦋  error     "summary": "401 Unauthorized - GET https://registry.npmjs.org/-/npm/v1/user",
🦋  error     "detail": ""
🦋  error   }
🦋  error }
🦋  error
🦋  error npm ERR! A complete log of this run can be found in:
🦋  error npm ERR!     /Users/jun/.npm/_logs/2023-06-09T10_56_34_917Z-debug-0.log
```

I traced this to the fact that `getTokenIsRequired()` is still in the hotpath, but that would require `npm profile get --json` to work, whereas it's not configured (since `yarn` uses `.yarnrc.yml` for registry configuration.

If I simply short circuit this function and `return false;` then it works. A temporary fix of course, but I hope this would help and maybe give back a tiny bit of momentum to get this over the line. Happy to test new patches!

jmatsushita:
Just bumped into the fact that for `changesets` to check if the package is published in a registry, [it uses `getPackageInfo` which is still using `npm info`](https://github.com/changesets/changesets/blob/760821164b4f144e7aca2bd4d410dce8c5ce641d/packages/cli/src/commands/publish/npm-utils.ts#L84-L90) instead of `yarn npm info`. In the case where registry access is configured only through a `.yarnrc.yml` file this will fail with a 404 (since it doesn't use the proper registry for that `@scope`) and will try to republish the package which is already in the registry, and fail if the package was already published (since the `yarn npm publish` part works correctly).
```
🦋  warn Received 404 for npm info "@scope/package"
🦋  info @scope/package is being published because our local version (0.0.0) has not been published on npm
🦋  info Publishing "@scope/package" at "0.0.0"
🦋  error an error occurred while publishing @scope/package: 
🦋  error ➤ YN0000: output/Main/index.js
🦋  error ➤ YN0000: package.json
🦋  error ➤ YN0035: The remote server failed to provide the requested resource
🦋  error ➤ YN0035:   Response Code: [40](https://github.com/org/repo/actions/runs/.../jobs/...#step:12:41)3 (Forbidden)
🦋  error ➤ YN0035:   Request Method: PUT
🦋  error ➤ YN0035:   Request URL: https://gitlab.com/api/v4/projects/.../packages/npm/@scope%2fpackage
🦋  @scope/package@0.0.0
```

A similar logic than `internalPublish` but for `yarn npm info` works https://github.com/changesets/changesets/blob/760821164b4f144e7aca2bd4d410dce8c5ce641d/packages/cli/src/commands/publish/npm-utils.ts#L189-L197 

Here's an updated version of the above patch:

<details>
<summary>.yarn/patches/@changesets-cli-npm-2.26.0-49d5c5f72d.patch</summary>
diff --git a/dist/cli.cjs.dev.js b/dist/cli.cjs.dev.js
index b1582198d3d2631569ff59f27170354a932d3ad0..2826a10919ca861e82efa105b6604fa52eda9847 100644
--- a/dist/cli.cjs.dev.js
+++ b/dist/cli.cjs.dev.js
@@ -1,5 +1,4 @@
 'use strict';
-
 var meow = require('meow');
 var errors = require('@changesets/errors');
 var logger = require('@changesets/logger');
@@ -705,31 +704,33 @@ function getCorrectRegistry(packageJson) {
   return !registry || registry === "https://registry.yarnpkg.com" ? "https://registry.npmjs.org" : registry;
 }
 
+const getPublishToolVersion = async (name, cwd) => (await spawn__default['default'](name, ["--version"], {
+  cwd
+})).stdout.toString().trim();
+
 async function getPublishTool(cwd) {
-  const pm = await preferredPM__default['default'](cwd);
-  if (!pm || pm.name !== "pnpm") return {
-    name: "npm"
-  };
+  var _await$preferredPM;
 
-  try {
-    let result = await spawn__default['default']("pnpm", ["--version"], {
-      cwd
-    });
-    let version = result.stdout.toString().trim();
-    let parsed = semver__default['default'].parse(version);
-    return {
-      name: "pnpm",
-      shouldAddNoGitChecks: (parsed === null || parsed === void 0 ? void 0 : parsed.major) === undefined ? false : parsed.major >= 5
-    };
-  } catch (e) {
+  const name = ((_await$preferredPM = await preferredPM__default['default'](cwd)) === null || _await$preferredPM === void 0 ? void 0 : _await$preferredPM.name) || "npm";
+  const version = await getPublishToolVersion(name, cwd);
+
+  if (name === "yarn" && semver__default['default'].lt(version, "2.0.0")) {
+    // Yarn Classic doesn't do anything special when publishing, let's stick to the npm client in such a case
     return {
-      name: "pnpm",
-      shouldAddNoGitChecks: false
+      name: "npm",
+      version: await getPublishToolVersion("npm", cwd)
     };
   }
+
+  return {
+    name,
+    version
+  };
 }
 
 async function getTokenIsRequired() {
+  let publishTool = await getPublishTool(process.cwd());
+  if (publishTool.name === "yarn") return false;
   // Due to a super annoying issue in yarn, we have to manually override this env variable
   // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
   const envOverride = {
@@ -754,6 +755,7 @@ async function getTokenIsRequired() {
 }
 function getPackageInfo(packageJson) {
   return npmRequestLimit(async () => {
+    let publishTool = await getPublishTool(process.cwd());
     logger.info(`npm info ${packageJson.name}`); // Due to a couple of issues with yarnpkg, we also want to override the npm registry when doing
     // npm info.
     // Issues: We sometimes get back cached responses, i.e old data about packages which causes
@@ -761,7 +763,9 @@ function getPackageInfo(packageJson) {
     // as they will always give a 404, which will tell `publish` to always try to publish.
     // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
 
-    let result = await spawn__default['default']("npm", ["info", packageJson.name, "--registry", getCorrectRegistry(packageJson), "--json"]); // Github package registry returns empty string when calling npm info
+    let result = publishTool.name === "yarn" ? await spawn__default['default']("yarn", ["npm", "info", packageJson.name, "--json"], {
+      env: Object.assign({}, process.env)
+    }) : await spawn__default['default']("npm", ["info", packageJson.name, "--registry", getCorrectRegistry(packageJson), "--json"]); // Github package registry returns empty string when calling npm info
     // for a non-existent package instead of a E404
 
     if (result.stdout.toString() === "") {
@@ -816,20 +820,32 @@ let getOtpCode = async twoFactorState => {
   }
 
   return askForOtpCode(twoFactorState);
+};
+
+const isOtpError = error => {
+  // The first case is no 2fa provided, the second is when the 2fa is wrong (timeout or wrong words)
+  return error.code === "EOTP" || error.code === "E401" && error.detail.includes("--otp=<code>");
 }; // we have this so that we can do try a publish again after a publish without
 // the call being wrapped in the npm request limit and causing the publishes to potentially never run
 
+
 async function internalPublish(pkgName, opts, twoFactorState) {
   let publishTool = await getPublishTool(opts.cwd);
-  let publishFlags = opts.access ? ["--access", opts.access] : [];
+  let shouldHandleOtp = !isCI__default['default'] && (publishTool.name === "yarn" ? semver__default['default'].gte(publishTool.version, "3.2.0-rc.8") : true);
+  let publishFlags = publishTool.name !== "yarn" ? ["--json"] : [];
+
+  if (opts.access) {
+    publishFlags.push("--access", opts.access);
+  }
+
   publishFlags.push("--tag", opts.tag);
 
-  if ((await twoFactorState.isRequired) && !isCI__default['default']) {
+  if (shouldHandleOtp && (await twoFactorState.isRequired)) {
     let otpCode = await getOtpCode(twoFactorState);
     publishFlags.push("--otp", otpCode);
   }
 
-  if (publishTool.name === "pnpm" && publishTool.shouldAddNoGitChecks) {
+  if (publishTool.name === "pnpm" && semver__default['default'].gte(publishTool.version, "5.0.0")) {
     publishFlags.push("--no-git-checks");
   } // Due to a super annoying issue in yarn, we have to manually override this env variable
   // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
@@ -842,21 +858,35 @@ async function internalPublish(pkgName, opts, twoFactorState) {
     code,
     stdout,
     stderr
-  } = await spawn__default['default'](publishTool.name, ["publish", opts.cwd, "--json", ...publishFlags], {
+  } = publishTool.name === "yarn" ? await spawn__default['default']("yarn", ["npm", "publish", ...publishFlags], {
+    cwd: opts.cwd,
+    env: Object.assign({}, process.env, envOverride)
+  }) : await spawn__default['default'](publishTool.name, ["publish", opts.cwd, ...publishFlags], {
     env: Object.assign({}, process.env, envOverride)
   });
 
   if (code !== 0) {
-    // NPM's --json output is included alongside the `prepublish` and `postpublish` output in terminal
+    // yarn berry doesn't support --json and we don't attempt to parse its output to a machine-readable format
+    if (publishTool.name === "yarn") {
+      const output = stdout.toString().trim().split("\n") // this filters out "unnamed" logs: https://yarnpkg.com/advanced/error-codes/#yn0000---unnamed
+      // this includes a list of packed files and the "summary output" like: "Failed with errors in 0s 75ms"
+      // those are not that interesting so we reduce the noise by dropping them
+      .filter(line => !/YN0000:/.test(line)).join("\n");
+      logger.error(`an error occurred while publishing ${pkgName}:`, `\n${output}`);
+      return {
+        published: false
+      };
+    } // NPM's --json output is included alongside the `prepublish` and `postpublish` output in terminal
     // We want to handle this as best we can but it has some struggles:
     // - output of those lifecycle scripts can contain JSON
     // - npm7 has switched to printing `--json` errors to stderr (https://github.com/npm/cli/commit/1dbf0f9bb26ba70f4c6d0a807701d7652c31d7d4)
     // Note that the `--json` output is always printed at the end so this should work
+
+
     let json = getLastJsonObjectFromString(stderr.toString()) || getLastJsonObjectFromString(stdout.toString());
 
     if (json !== null && json !== void 0 && json.error) {
-      // The first case is no 2fa provided, the second is when the 2fa is wrong (timeout or wrong words)
-      if ((json.error.code === "EOTP" || json.error.code === "E401" && json.error.detail.includes("--otp=<code>")) && !isCI__default['default']) {
+      if (shouldHandleOtp && isOtpError(json.error)) {
         if (twoFactorState.token !== null) {
           // the current otp code must be invalid since it errored
           twoFactorState.token = null;
diff --git a/dist/cli.cjs.prod.js b/dist/cli.cjs.prod.js
index 5b1b7dd6439be4f0b721a17e683b1435c254f5b8..803762c16b9c7f484948171e42e1d8b4d95cbc7c 100644
--- a/dist/cli.cjs.prod.js
+++ b/dist/cli.cjs.prod.js
@@ -1,5 +1,5 @@
 "use strict";
-
+console.log()
 var meow = require("meow"), errors = require("@changesets/errors"), logger = require("@changesets/logger"), util = require("util"), fs = require("fs-extra"), path = require("path"), getPackages = require("@manypkg/get-packages"), getDependentsGraph = require("@changesets/get-dependents-graph"), config = require("@changesets/config"), chalk = require("chalk"), child_process = require("child_process"), termSize = require("term-size"), enquirer = require("enquirer"), externalEditor = require("external-editor"), ansiColors = require("ansi-colors"), git = require("@changesets/git"), writeChangeset = require("@changesets/write"), resolveFrom = require("resolve-from"), semver = require("semver"), outdent = require("outdent"), applyReleasePlan = require("@changesets/apply-release-plan"), readChangesets = require("@changesets/read"), assembleReleasePlan = require("@changesets/assemble-release-plan"), pre$1 = require("@changesets/pre"), pLimit = require("p-limit"), preferredPM = require("preferred-pm"), spawn = require("spawndamnit"), isCI = require("is-ci"), table = require("tty-table"), getReleasePlan = require("@changesets/get-release-plan");
 
 function _interopDefault(e) {
@@ -396,28 +396,25 @@ function getCorrectRegistry(packageJson) {
   return registry && "https://registry.yarnpkg.com" !== registry ? registry : "https://registry.npmjs.org";
 }
 
+const getPublishToolVersion = async (name, cwd) => (await spawn__default.default(name, [ "--version" ], {
+  cwd: cwd
+})).stdout.toString().trim();
+
 async function getPublishTool(cwd) {
-  const pm = await preferredPM__default.default(cwd);
-  if (!pm || "pnpm" !== pm.name) return {
-    name: "npm"
+  var _await$preferredPM;
+  const name = (null === (_await$preferredPM = await preferredPM__default.default(cwd)) || void 0 === _await$preferredPM ? void 0 : _await$preferredPM.name) || "npm", version = await getPublishToolVersion(name, cwd);
+  return "yarn" === name && semver__default.default.lt(version, "2.0.0") ? {
+    name: "npm",
+    version: await getPublishToolVersion("npm", cwd)
+  } : {
+    name: name,
+    version: version
   };
-  try {
-    let version = (await spawn__default.default("pnpm", [ "--version" ], {
-      cwd: cwd
-    })).stdout.toString().trim(), parsed = semver__default.default.parse(version);
-    return {
-      name: "pnpm",
-      shouldAddNoGitChecks: void 0 !== (null == parsed ? void 0 : parsed.major) && parsed.major >= 5
-    };
-  } catch (e) {
-    return {
-      name: "pnpm",
-      shouldAddNoGitChecks: !1
-    };
-  }
 }
 
 async function getTokenIsRequired() {
+  let publishTool = await getPublishTool(process.cwd());
+  if (publishTool.name === "yarn") return false;
   const envOverride = {
     npm_config_registry: getCorrectRegistry()
   };
@@ -432,8 +429,11 @@ async function getTokenIsRequired() {
 
 function getPackageInfo(packageJson) {
   return npmRequestLimit((async () => {
+    let publishTool = await getPublishTool(process.cwd());
     logger.info("npm info " + packageJson.name);
-    let result = await spawn__default.default("npm", [ "info", packageJson.name, "--registry", getCorrectRegistry(packageJson), "--json" ]);
+    let result = "yarn" === publishTool.name ? await spawn__default.default("yarn", [ "npm", "info", packageJson.name, "--json" ], {
+    env: Object.assign({}, process.env)
+  }) : await spawn__default.default("npm", [ "info", packageJson.name, "--registry", getCorrectRegistry(packageJson), "--json" ]);
     return "" === result.stdout.toString() ? {
       error: {
         code: "E404"
@@ -466,23 +466,36 @@ let otpAskLimit = pLimit__default.default(1), askForOtpCode = twoFactorState =>
   return twoFactorState.token = val, val;
 })), getOtpCode = async twoFactorState => null !== twoFactorState.token ? twoFactorState.token : askForOtpCode(twoFactorState);
 
+const isOtpError = error => "EOTP" === error.code || "E401" === error.code && error.detail.includes("--otp=<code>");
+
 async function internalPublish(pkgName, opts, twoFactorState) {
-  let publishTool = await getPublishTool(opts.cwd), publishFlags = opts.access ? [ "--access", opts.access ] : [];
-  if (publishFlags.push("--tag", opts.tag), await twoFactorState.isRequired && !isCI__default.default) {
+  let publishTool = await getPublishTool(opts.cwd), shouldHandleOtp = !isCI__default.default && ("yarn" !== publishTool.name || semver__default.default.gte(publishTool.version, "3.2.0-rc.8")), publishFlags = "yarn" !== publishTool.name ? [ "--json" ] : [];
+  if (opts.access && publishFlags.push("--access", opts.access), publishFlags.push("--tag", opts.tag), 
+  shouldHandleOtp && await twoFactorState.isRequired) {
     let otpCode = await getOtpCode(twoFactorState);
     publishFlags.push("--otp", otpCode);
   }
-  "pnpm" === publishTool.name && publishTool.shouldAddNoGitChecks && publishFlags.push("--no-git-checks");
+  "pnpm" === publishTool.name && semver__default.default.gte(publishTool.version, "5.0.0") && publishFlags.push("--no-git-checks");
   const envOverride = {
     npm_config_registry: getCorrectRegistry()
   };
-  let {code: code, stdout: stdout, stderr: stderr} = await spawn__default.default(publishTool.name, [ "publish", opts.cwd, "--json", ...publishFlags ], {
+  let {code: code, stdout: stdout, stderr: stderr} = "yarn" === publishTool.name ? await spawn__default.default("yarn", [ "npm", "publish", ...publishFlags ], {
+    cwd: opts.cwd,
+    env: Object.assign({}, process.env, envOverride)
+  }) : await spawn__default.default(publishTool.name, [ "publish", opts.cwd, ...publishFlags ], {
     env: Object.assign({}, process.env, envOverride)
   });
   if (0 !== code) {
+    if ("yarn" === publishTool.name) {
+      const output = stdout.toString().trim().split("\n").filter((line => !/YN0000:/.test(line))).join("\n");
+      return logger.error(`an error occurred while publishing ${pkgName}:`, "\n" + output), 
+      {
+        published: !1
+      };
+    }
     let json = getLastJsonObjectFromString(stderr.toString()) || getLastJsonObjectFromString(stdout.toString());
     if (null != json && json.error) {
-      if (("EOTP" === json.error.code || "E401" === json.error.code && json.error.detail.includes("--otp=<code>")) && !isCI__default.default) return null !== twoFactorState.token && (twoFactorState.token = null), 
+      if (shouldHandleOtp && ("EOTP" === (error = json.error).code || "E401" === error.code && error.detail.includes("--otp=<code>"))) return null !== twoFactorState.token && (twoFactorState.token = null), 
       twoFactorState.isRequired = Promise.resolve(!0), internalPublish(pkgName, opts, twoFactorState);
       logger.error(`an error occurred while publishing ${pkgName}: ${json.error.code}`, json.error.summary, json.error.detail ? "\n" + json.error.detail : "");
     }
@@ -490,6 +503,7 @@ async function internalPublish(pkgName, opts, twoFactorState) {
       published: !1
     };
   }
+  var error;
   return {
     published: !0
   };
diff --git a/dist/cli.esm.js b/dist/cli.esm.js
index ced46d2520d4bff71b94e4023b1841373a55712b..044d375056d5007859ee1a2e9c3dc7f0c1b4fcd2 100644
--- a/dist/cli.esm.js
+++ b/dist/cli.esm.js
@@ -682,31 +682,33 @@ function getCorrectRegistry(packageJson) {
   return !registry || registry === "https://registry.yarnpkg.com" ? "https://registry.npmjs.org" : registry;
 }
 
+const getPublishToolVersion = async (name, cwd) => (await spawn$1(name, ["--version"], {
+  cwd
+})).stdout.toString().trim();
+
 async function getPublishTool(cwd) {
-  const pm = await preferredPM(cwd);
-  if (!pm || pm.name !== "pnpm") return {
-    name: "npm"
-  };
+  var _await$preferredPM;
 
-  try {
-    let result = await spawn$1("pnpm", ["--version"], {
-      cwd
-    });
-    let version = result.stdout.toString().trim();
-    let parsed = semver.parse(version);
-    return {
-      name: "pnpm",
-      shouldAddNoGitChecks: (parsed === null || parsed === void 0 ? void 0 : parsed.major) === undefined ? false : parsed.major >= 5
-    };
-  } catch (e) {
+  const name = ((_await$preferredPM = await preferredPM(cwd)) === null || _await$preferredPM === void 0 ? void 0 : _await$preferredPM.name) || "npm";
+  const version = await getPublishToolVersion(name, cwd);
+
+  if (name === "yarn" && semver.lt(version, "2.0.0")) {
+    // Yarn Classic doesn't do anything special when publishing, let's stick to the npm client in such a case
     return {
-      name: "pnpm",
-      shouldAddNoGitChecks: false
+      name: "npm",
+      version: await getPublishToolVersion("npm", cwd)
     };
   }
+
+  return {
+    name,
+    version
+  };
 }
 
 async function getTokenIsRequired() {
+  let publishTool = await getPublishTool(process.cwd());
+  if (publishTool.name === "yarn") return false;
   // Due to a super annoying issue in yarn, we have to manually override this env variable
   // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
   const envOverride = {
@@ -731,6 +733,7 @@ async function getTokenIsRequired() {
 }
 function getPackageInfo(packageJson) {
   return npmRequestLimit(async () => {
+    let publishTool = await getPublishTool(process.cwd());
     info(`npm info ${packageJson.name}`); // Due to a couple of issues with yarnpkg, we also want to override the npm registry when doing
     // npm info.
     // Issues: We sometimes get back cached responses, i.e old data about packages which causes
@@ -738,7 +741,9 @@ function getPackageInfo(packageJson) {
     // as they will always give a 404, which will tell `publish` to always try to publish.
     // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
 
-    let result = await spawn$1("npm", ["info", packageJson.name, "--registry", getCorrectRegistry(packageJson), "--json"]); // Github package registry returns empty string when calling npm info
+    let result = publishTool.name === "yarn" ? await spawn$1("yarn", ["npm", "info", packageJson.name, "--json"], {
+    env: Object.assign({}, process.env)
+  }) : await spawn$1("npm", ["info", packageJson.name, "--registry", getCorrectRegistry(packageJson), "--json"]); // Github package registry returns empty string when calling npm info
     // for a non-existent package instead of a E404
 
     if (result.stdout.toString() === "") {
@@ -793,20 +798,32 @@ let getOtpCode = async twoFactorState => {
   }
 
   return askForOtpCode(twoFactorState);
+};
+
+const isOtpError = error => {
+  // The first case is no 2fa provided, the second is when the 2fa is wrong (timeout or wrong words)
+  return error.code === "EOTP" || error.code === "E401" && error.detail.includes("--otp=<code>");
 }; // we have this so that we can do try a publish again after a publish without
 // the call being wrapped in the npm request limit and causing the publishes to potentially never run
 
+
 async function internalPublish(pkgName, opts, twoFactorState) {
   let publishTool = await getPublishTool(opts.cwd);
-  let publishFlags = opts.access ? ["--access", opts.access] : [];
+  let shouldHandleOtp = !isCI && (publishTool.name === "yarn" ? semver.gte(publishTool.version, "3.2.0-rc.8") : true);
+  let publishFlags = publishTool.name !== "yarn" ? ["--json"] : [];
+
+  if (opts.access) {
+    publishFlags.push("--access", opts.access);
+  }
+
   publishFlags.push("--tag", opts.tag);
 
-  if ((await twoFactorState.isRequired) && !isCI) {
+  if (shouldHandleOtp && (await twoFactorState.isRequired)) {
     let otpCode = await getOtpCode(twoFactorState);
     publishFlags.push("--otp", otpCode);
   }
 
-  if (publishTool.name === "pnpm" && publishTool.shouldAddNoGitChecks) {
+  if (publishTool.name === "pnpm" && semver.gte(publishTool.version, "5.0.0")) {
     publishFlags.push("--no-git-checks");
   } // Due to a super annoying issue in yarn, we have to manually override this env variable
   // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
@@ -819,21 +836,35 @@ async function internalPublish(pkgName, opts, twoFactorState) {
     code,
     stdout,
     stderr
-  } = await spawn$1(publishTool.name, ["publish", opts.cwd, "--json", ...publishFlags], {
+  } = publishTool.name === "yarn" ? await spawn$1("yarn", ["npm", "publish", ...publishFlags], {
+    cwd: opts.cwd,
+    env: Object.assign({}, process.env, envOverride)
+  }) : await spawn$1(publishTool.name, ["publish", opts.cwd, ...publishFlags], {
     env: Object.assign({}, process.env, envOverride)
   });
 
   if (code !== 0) {
-    // NPM's --json output is included alongside the `prepublish` and `postpublish` output in terminal
+    // yarn berry doesn't support --json and we don't attempt to parse its output to a machine-readable format
+    if (publishTool.name === "yarn") {
+      const output = stdout.toString().trim().split("\n") // this filters out "unnamed" logs: https://yarnpkg.com/advanced/error-codes/#yn0000---unnamed
+      // this includes a list of packed files and the "summary output" like: "Failed with errors in 0s 75ms"
+      // those are not that interesting so we reduce the noise by dropping them
+      .filter(line => !/YN0000:/.test(line)).join("\n");
+      error(`an error occurred while publishing ${pkgName}:`, `\n${output}`);
+      return {
+        published: false
+      };
+    } // NPM's --json output is included alongside the `prepublish` and `postpublish` output in terminal
     // We want to handle this as best we can but it has some struggles:
     // - output of those lifecycle scripts can contain JSON
     // - npm7 has switched to printing `--json` errors to stderr (https://github.com/npm/cli/commit/1dbf0f9bb26ba70f4c6d0a807701d7652c31d7d4)
     // Note that the `--json` output is always printed at the end so this should work
+
+
     let json = getLastJsonObjectFromString(stderr.toString()) || getLastJsonObjectFromString(stdout.toString());
 
     if (json !== null && json !== void 0 && json.error) {
-      // The first case is no 2fa provided, the second is when the 2fa is wrong (timeout or wrong words)
-      if ((json.error.code === "EOTP" || json.error.code === "E401" && json.error.detail.includes("--otp=<code>")) && !isCI) {
+      if (shouldHandleOtp && isOtpError(json.error)) {
         if (twoFactorState.token !== null) {
           // the current otp code must be invalid since it errored
           twoFactorState.token = null;
diff --git a/package.json b/package.json
index 0912d0f8fbf6f88c623ba4de0c87345947cde035..4ea6809033e063422e98cc3370685eb9b46847a2 100644
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "meow": "^6.0.0",
     "outdent": "^0.5.0",
     "p-limit": "^2.2.0",
-    "preferred-pm": "^3.0.0",
+    "preferred-pm": "^3.0.3",
     "resolve-from": "^5.0.0",
     "semver": "^5.4.1",
     "spawndamnit": "^2.0.0",
</details>

Which works in yarn `4.0.0-rc.45` with these incantations in the `package.json`:
```json
{
  "name": "package",
  "version": "0.0.0",
  "devDependencies": {
    "@changesets/cli": "patch:@changesets/cli@npm%3A2.26.0#~/.yarn/patches/@changesets-cli-npm-2.26.0-49d5c5f72d.patch"
  },
  "packageManager": "yarn@4.0.0-rc.45",
  "resolutions": {
    "@changesets/cli@^2.26.0": "patch:@changesets/cli@npm%3A2.26.0#./.yarn/patches/@changesets-cli-npm-2.26.0-49d5c5f72d.patch"
  },
}
```

castastrophe:
Any interest in picking this work back up? We love changesets at the Adobe Spectrum CSS project but we're leveraging yarn 4.1.1 and the workspaces:^ syntax and would love to see even a small update to using `yarn npm publish` instead of `npm publish` so that the workspace identifiers are correctly replaced in the released code.

tianyingchun:
any plan of this?

fmal:
I would highly recommend switching to https://github.com/monoweave/monoweave for Yarn-based projects. I had a great experience with this tool in my monorepos.

tianyingchun:
waiting changesets support `workspace:*`

tien:
Hey @Andarist sorry for the ping, I got surprised by this pretty bad today 😅

Are you planning on picking this PR up again? Else, one option I can think of is just to remove `changeset publish` entirely, and delegate publishing logic to the package manager (could actually be the better way to go).

HaakonSvane:
+1

bensampaio:
Hey @Andarist, at my company we stumbled upon this issue a few weeks ago and it cost us quite some outages and bugs due to duplicate versions or incorrect resolutions in yarn.lock. I'm surprised to find out that this issue has been there since 2021. It hasn't been fixed and it is also not in the docs anywhere. It could have saved us a lot of headache to know that yarn publish isn't supported. In my opinion, this deserves a warning on your main README file.

Fortunately, we managed to fix this issue with the changes in this [PR](https://github.com/changesets/changesets/pull/1449). So far I didn't get your feedback there. Could you let me know why isn't it an option to merge it?

Most importantly, if you are not fixing this issue, do you have at least a recommendation for people stuck in this problem? At my company, we chose changesets and are willing to help if the effort is manageable. Please let us know if there's anything we can do. And please document this because others should know before they choose this tool.

kvz:
Would love to see this get landed. Could @transloadit sponsor this effort perhaps?

kvz:
Is this an ai response? I actually offered to sponsor this. And we allocate 80% of revenue to open source at Transloadit

HaakonSvane:
> Is this an ai response? I actually offered to sponsor this. And we allocate 80% of revenue to open source at Transloadit

There are obviously more people visiting this PR than just yourself. While it is great to hear about how you help the open source community, this is not very common for companies, especially those where IT is not considered part of their core business, to do. I posted my message in the hopes that other frustrated developers that might come across this PR have a look at the presentation and reflect while we wait

artyomSultanov:
+1

imarabinda:
+1

paulbalaji:
+1

Andarist:
This is going to be superseded by https://github.com/changesets/changesets/pull/2097

PR #676, opened 2021-11-22
Fixed an issue with parsing `--json` output when publishing

Andarist:
This work has been moved out of https://github.com/atlassian/changesets/pull/674

closes https://github.com/atlassian/changesets/issues/567
closes https://github.com/atlassian/changesets/issues/397

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -184,7 +184,7 @@ async function internalPublish(
       getLastJsonObjectFromString(stderr.toString()) ||
       getLastJsonObjectFromString(stdout.toString());
 
-    if (json.error) {
+    if (json?.error) {
       // The first case is no 2fa provided, the second is when the 2fa is wrong (timeout or wrong words)
       if (
         (json.error.code === "EOTP" ||
@@ -205,9 +205,10 @@ async function internalPublish(
         json.error.summary,
         json.error.detail ? "\n" + json.error.detail : ""
       );
-      error(stderr);
-      return { published: false };
     }
+
+    error(stderr);
+    return { published: false };
   }
   return { published: true };
 }
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -9,6 +9,7 @@ import semver from "semver";
 import { askQuestion } from "../../utils/cli-utilities";
 import isCI from "../../utils/isCI";
 import { TwoFactorState } from "../../utils/types";
+import { getLastJsonObjectFromString } from "../../utils/getLastJsonObjectFromString";
 
 const npmRequestLimit = pLimit(40);
 const npmPublishLimit = pLimit(10);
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -166,41 +167,47 @@ async function internalPublish(
...
       }
-      // just in case this isn't already true
-      twoFactorState.isRequired = Promise.resolve(true);
-      return internalPublish(pkgName, opts, twoFactorState);
+      error(
+        `an error occurred while publishing ${pkgName}: ${json.error.code}`,
+        json.error.summary,
+        json.error.detail ? "\n" + json.error.detail : ""
+      );
+      error(stderr);
```
Andarist, line 208: Added this as suggested [here](https://github.com/atlassian/changesets/pull/674#pullrequestreview-811980332)

For the moment I've refrained from printing `stdout` too because it contains the output of all lifecycle scripts and it feels like this could get really, really, verbose.

A somewhat weird effect is also that with npm7+ this will contain at the end the JSON mentioned by the `error` above this line - but I guess it's not worth trying to remove that.

packages/cli/src/utils/getLastJsonObjectFromString.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    try {
+      return JSON.parse(str);
+    } catch (err) {
+      // move past the potentially leading `{` so the regexp in the loop can try to match for the next `{`
+      str = str.slice(1);
+    }
+  }
+  return null;
+};
```
mitchellhamilton, line 15: How would this handle the thing below and similar things with `{` and `}` in strings?

```json
{"blah": "{"}
```

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -166,41 +167,47 @@ async function internalPublish(
...
+    // NPM's --json output is included alongside the `prepublish` and `postpublish` output in terminal
+    // We want to handle this as best we can but it has some struggles:
+    // - output of those lifecycle scripts can contain JSON
+    // - npm7 has switched to printing `--json` errors to stderr (https://github.com/npm/cli/commit/1dbf0f9bb26ba70f4c6d0a807701d7652c31d7d4)
+    // Note that the `--json` output is always printed at the end so this should work
+    let json =
+      getLastJsonObjectFromString(stderr.toString()) ||
+      getLastJsonObjectFromString(stdout.toString());
+
+    if (json.error) {
```
mitchellhamilton, line 187:
```suggestion
    if (json?.error) {
```

mitchellhamilton (changes requested):
Could do the suggestion below and log stderr regardless of whether `json?.error` is truthy or not so that users can see stderr just in case we fail to find a json object in the output?

packages/cli/src/commands/publish/npm-utils.ts
Andarist: Do we have any scenario in which this would be currently needed? Or is it intended to be as an extra precaution?

packages/cli/src/utils/getLastJsonObjectFromString.ts
Andarist: the given example is not even that problematic - we try to parse from `{` till the end, and since this is a valid JSON it just works

if we consider an input like:
```
"{" {"blah": "{"}
```
then it also works because it will fail on the quoted one and retry on the second one (and succeed there)

The idea behind this helper is a little bit indirect but the gist is that even if we catch `{` somewhere, wherever really, be it in a quoted string, be it orphaned, be it part of some other JSON, then we try to parse from it and we simply fail because all of those cases can't be parsed as a single JSON. So we continue and we only parse successfully when we get to the last JSON in the string.

There is also an assumption that we can't have some "invalid" characters after that last JSON - this helper isn't working correctly if we consider such scenario (so `{"a": "b"}randomtext` won't result in a "last json object from string"). Such a scenario has no application for us - because the JSON that we are after should really always be the last thing in the `stdout`/`stderr`

packages/cli/src/commands/publish/npm-utils.ts
mitchellhamilton: Don't see anything that would require it, just an extra precaution because getting `Uncaught TypeError: Cannot read properties of null (reading 'error')` would be pretty annoying.

packages/cli/src/utils/getLastJsonObjectFromString.ts
mitchellhamilton: > There is also an assumption that we can't have some "invalid" characters after that last JSON - this helper isn't working correctly if we consider such scenario (so {"a": "b"}randomtext won't result in a "last json object from string"). Such a scenario has no application for us - because the JSON that we are after should really always be the last thing in the stdout/stderr

Running `npm publish`(`npm@8.1.0`) in a non-tty env, I get what's shown below. How is it working if there won't be "invalid" characters after the last json given there are "invalid" characters after the json

```
npm ERR! code EOTP
npm ERR! This operation requires a one-time password from your authenticator.
npm ERR! You can provide a one-time password by passing --otp=<code> to the command you ran.
npm ERR! If you already provided a one-time password then it is likely that you either typoed
npm ERR! it, or it timed out. Please try again.
{
  "error": {
    "code": "EOTP",
    "summary": "This operation requires a one-time password from your authenticator.",
    "detail": "You can provide a one-time password by passing --otp=<code> to the command you ran.\nIf you already provided a one-time password then it is likely that you either typoed\nit, or it timed out. Please try again."
  }
}

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/mitchell/.npm/_logs/2021-11-25T22_31_42_856Z-debug.log
```

Andarist: Sorry, I forgot that I've actually put this at the beginning of this function: `str = str.replace(/[^}]*$/, "");`.

So some invalid characters after the last JSON are allowed, but it would break on stuff like:
```
{"some": "json"}

npm ERR!     /some/weird/path/file}name.log
```

This seems acceptable as it's a really bizarre case - we can always refine this logic later when the need arises. Unless you have any idea how to improve this logic in an easy and quick way to handle more scenarios?

packages/cli/src/commands/publish/npm-utils.ts
Andarist: Ok, I've adjusted the code as suggested and made sure that even if it's missing then we still `return { published: false }` (as this is still within the `if (code !== 0)` block)

mitchellhamilton (approved):
Given that we'll always log `stderr` now, I'm fine with `getLastJsonObjectFromString` possibly breaking for edge cases since the worse case is someone has to provide a 2fa code with `--otp` rather than the prompt.

peschee:
👏🏼

Andarist:
@peschee Sorry, I've missed your comment here - what do you mean? Changesets Action is just using the installed Changesets CLI - and AFAIR it doesn't make direct calls to the npm CLI. So this shouldn't quite affect it directly, unless I'm missing something.

PR #677, opened 2021-11-22
defaultWrittenConfig change baseBranch from master to main

### LINKED ISSUE #675: Change default base branch: master -> main
Since GitHub [changed the name of the default branch on new repos to `main`](https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/) on Oct 1, 2020, `changesets` might consider updating the default `baseBranch` to match:

```diff
$ yarn changeset init
$ cat .changeset/config.json
{
  "$schema": "https://unpkg.com/@changesets/config@1.6.2/schema.json",
  "changelog": "@changesets/cli/changelog",
  "commit": false,
  "linked": [],
  "access": "restricted",
- "baseBranch": "master",
+ "baseBranch": "main",
  "updateInternalDependencies": "patch",
  "ignore": []
}
```

janosh:
Closes #675.

Should I add a changelog entry?

### DIFF
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -33,7 +33,9 @@ export default async function init(cwd: string) {
       }
       await fs.writeFile(
         path.resolve(changesetBase, "config.json"),
-        JSON.stringify(defaultWrittenConfig, null, 2)
+        // modify base branch to "main" without changing defaultWrittenConfig since it also serves as a fallback
+        // for config files that don't specify a base branch so changing that to main would be a breaking change
+        JSON.stringify({ ...defaultWrittenConfig, baseBranch: "main" }, null, 2)
       );
     } else {
       warn(
@@ -44,7 +46,7 @@ export default async function init(cwd: string) {
     await fs.copy(path.resolve(pkgPath, "./default-files"), changesetBase);
     await fs.writeFile(
       path.resolve(changesetBase, "config.json"),
-      JSON.stringify(defaultWrittenConfig, null, 2)
+      JSON.stringify({ ...defaultWrittenConfig, baseBranch: "main" }, null, 2)
     );
 
     log(
--- a/packages/cli/src/commands/init/__tests__/command.ts
+++ b/packages/cli/src/commands/init/__tests__/command.ts
@@ -33,9 +33,9 @@ describe("init", () => {
       true
     );
     await initializeCommand(cwd);
-    expect(await fs.readJson(path.join(cwd, ".changeset/config.json"))).toEqual(
-      defaultWrittenConfig
-    );
+    expect(
+      await fs.readJson(path.join(cwd, ".changeset/config.json"))
+    ).toEqual({ ...defaultWrittenConfig, baseBranch: "main" });
   });
   it("shouldn't overwrite a config if it does exist", async () => {
     const cwd = await f.copy("simple-project");


packages/config/src/index.test.ts
```diff
@@ -40,7 +40,7 @@ test("read reads the config", async () => {
     changelog: false,
     commit: true,
     access: "restricted",
-    baseBranch: "master",
+    baseBranch: "main",
```
Andarist, line 43: Correct me if I'm wrong but it looks like this is actually a breaking change. I've looked into the `new-config` fixture and it only has such `.changeset/config.json`:
```json
{
  "changelog": false,
  "commit": true
}
```

I think the point of this PR was only to initialize the `.changeset/config.json` with the new value but not to change what it defaults to when it is not explicitly given etc.

janosh: > I think the point of this PR was only to initialize the .changeset/config.json with the new value but not to change what it defaults to when it is not explicitly given etc.

What's the difference there?

janosh: @Andarist Happy to make changes here as necessary but I didn't quite understand your point.

Andarist: The problem with the current solution is that if somebody already has a config like:
```json
{
  "changelog": false,
  "commit": true
}
```
then with this change Changesets would start using `main` as their `baseBranch` and thus it would change behavior, by potentially crashing because a `main` branch might not even exist in a given repository.

When initializing Changesets we write `defaultBranch` into the generated `.changeset/config.json`, even though it's not strictly required because that property has a default. But since we are outputting this property it can be changed somewhat safely because people who are already using Changesets are not running `changesets init`, they have already done that in the past.

janosh: I see, makes sense. So we need two separate objects, one like `defaultWrittenConfig` which sets the default values in a new config file and another `defaultConfigValues` that sets the values for configs that don't specify optional keys??

Andarist: Ye, smth like that would work. Please only that because this is a monorepo we might have to implement this in a non-intuitive way. We don't want to break people who don't upgrade all of the transitive dependencies at once. 

I think that the safest solution is to not touch `defautlWrittenConfig` at all (as weirdly as it might be) - and to just override this value before we actually write it in the config file.

janosh: > I think that the safest solution is to not touch defautlWrittenConfig at all (as weirdly as it might be) - and to just override this value before we actually write it in the config file.

I agree, seems like the best solution here.

janosh:
@Andarist Anything left to do here?

janosh:
@Andarist Anything left to do here?

janosh:
@Andarist Could you let me know if this change is still acceptable?

packages/cli/src/commands/init/index.ts
```diff
@@ -44,7 +46,7 @@ export default async function init(cwd: string) {
     await fs.copy(path.resolve(pkgPath, "./default-files"), changesetBase);
     await fs.writeFile(
       path.resolve(changesetBase, "config.json"),
-      JSON.stringify(defaultWrittenConfig, null, 2)
+      JSON.stringify({ ...defaultWrittenConfig, baseBranch: "main" }, null, 2)
```
Andarist, line 49: I would extract this to a variable on the top of the file, so the added comment could be added to that variable so it would be visible~ in a common place for both places that have been touched here.

Andarist:
@janosh sorry for the delay, it's just sometimes hard to get to everything on time. The PR looks good - I just left a single nit-level comment. Could you also add a changeset to this PR?

janosh:
@Andarist I had deleted my fork so couldn't make any further changes here. Closing in favor of #776 where I addressed your comments.

Issue #581, opened 2021-04-29
Mixed versions across workspaces


raphaelokon:
Hi there.

We are evaluating to replace Lerna with Yarn workspaces + Changesets and need input if the following concept is possible with changesets:

```js
  "workspaces": [
    "libs/*",  // All packages in `libs` should be versioned independenly
    "main/*",  // All packages in `main` should have one fixed/locked versioning
  ]
```

We would ideally have an independent versioning (like all packages can get a custom version) for all packages inside the `libs/` workspace and a fixed version across packages (like in LErna’s fixed/locked mode) under `main/`. 

Is this possible with changesets?
 
Cheers 😊

Andarist:
It's possible - you could use `linked` config option for all packages in the `main` directory. Note that this works slightly in a different way than Lerna's fixed/locked mode but I think that it should suffice for the majority of use cases.

There have been some requests for the "true" fixed/locked mode so we could consider adding support for this somehow but at the moment we don't have manpower to design such a feature.

raphaelokon:
Hi @Andarist 

So grateful for your feedback on this. I already saw the `linked` config option, but the example made me puzzle as how to always force each package inside the `linked` array to update to one and the same version.

> There have been some requests for the "true" fixed/locked mode so we could consider adding support for this somehow but at the moment we don't have manpower to design such a feature.

Sure – we could look into helping out. 

How would you describe the difference between Lerna’s fixed/locked mode and Changesets fixed mode?

Thanks again and have a good one.

Andarist:
> How would you describe the difference between Lerna’s fixed/locked mode and Changesets fixed mode?

If by "Changesets fixed mode" you are referring to our linked mode then:
- if you have a set of linked packages and you publish multiple packages from that set then all of those releases will have the very same version
- but if you only release some subset of packages from that set of linked packages then those packages that don't have to be released won't be released. Whereas with Lerna's fixed mode all packages are always released together, even if some of them have not been touched

Andarist:
The fixed mode is being implemented in https://github.com/changesets/changesets/pull/690 . Please check out if it matches your expectations

Issue #683, opened 2021-12-01
Aggregate releases


Andarist:
It's a very popular request to aggregate all of the changelogs stuff in a single GitHub release. A lot of open-source monorepos revolve around a single project and thus creating dozens of releases with each `changeset publish` gets quite noisy - this is especially true for linked packages or fixed packages (FYI I'm working on introducing the fixed package groups).

I propose to introduce support for this:
```td
type LinkedPatterns = string[];
type Linked = Array<
  | { name?: string; packages: LinkedPatterns; aggregate?: boolean }
  | LinkedPatterns
>;
```

Since the goal of this feature is to combine things into a single GitHub release we need something to put in the title of the release. The `name` option would be used for that and the created release would have such title `${name} v${version}`. It IMHO makes sense to infer the name from the used scope as that would work for the majority of use cases - but that can also be left out from the MVP. 

`aggregate` (or `aggregateRelease`) would be an opt-in for this feature - `@changesets/action` could pick this up and decide how to handle the releasing part of the things. Some people have also suggested that a single CHANGELOG could be created (that would be a copy of the aggregated release) - I think there is a value in this feature, but I'm not sure if this should automatically be enabled with just `aggregate` or if this requires an additional option. I'm afraid of introducing an option creep though so I would really prefer to keep this simple. 

The big question is - how do we concatenate particular changelogs? How do we sort packages? My ideas:
- allow for stuff defined in `packages` to decide about this, for instance: `packages: ['@chakra-ui/react', '@chakra-ui/*']` would tell Changesets to use the current CHANGELOG update from `@chakra-ui/react` first and to put everything else after that
- we could try to sort the packages based on the dependency graph, so basically try to infer the order from the graph - the packages with the most descendants would take precedence over packages with less of those. This is kinda magical and probably just a possible enhancement over the previous idea. I would keep this one out of the MVP

There is also an additional problem - some projects, like Chakra, have an "aggregate" package like `@chakra-ui/react`. The purpose of this package is to just reexport stuff from everything else and nothing beyond that. I think they currently just add `@chakra-ui/react` to each created changeset because of that. This would mean that every change in the aggregate release would get kinda duplicated because of that - as it would be described both by `@chakra-ui/react` and some other, inner, package like `@chakra-ui/button` or something. Could their workflow be adjusted with the `aggregate` in place in a way that they wouldn't suffer from this problem?

[...]

PR #690, opened 2021-12-04
Add support for `fixed` packages

Andarist:
This introduces an alternative to the linked mode. Linked mode is being constantly misunderstood and confuses people while the fixed mode is known from other tools (like Lerna).

TODO:

- [x] add changesets per affected package
- [ ] verify if there are any missing essential tests
- [x] verify how this works with generating changelogs
- [ ] docs - I think this might come in a follow up PR

closes https://github.com/changesets/changesets/issues/581
closes https://github.com/changesets/changesets/issues/570


packages/assemble-release-plan/src/index.test.ts
```diff
@@ -95,6 +95,44 @@ describe("assemble-release-plan", () => {
     expect(releases[1].newVersion).toEqual("1.0.1");
     expect(releases[1].changesets).toEqual([]);
   });
+  it("should update multiple dependents of a single package", () => {
```
Andarist, line 98: those two tests previously had been configuring `linked` but that wasn't relevant for them so I've moved them to the top block~

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -254,189 +292,7 @@ describe("assemble-release-plan", () => {
     expect(releases[0].name).toEqual("pkg-a");
     expect(releases[0].newVersion).toEqual("2.0.0");
   });
-  it("should assemble release plan for linked packages", () => {
```
Andarist, line 257: I've grouped tests related to the linked packages in a describe block - no significant changes were made to those moved tests

packages/cli/src/commands/version/version.test.ts
```diff
@@ -144,36 +145,6 @@ describe("running version in a simple project", () => {
     });
   });
 
-  it("should bump packages to the correct versions when packages are linked", async () => {
```
Andarist, line 147: similarly here - I've moved tests related to linked packages to their own block

__fixtures__/fixed-packages/.changeset/config.js
```diff
@@ -0,0 +1,87 @@
+/*
```
mitchellhamilton, line 1: This config doesn't actually work, right? Since it's a js file

mitchellhamilton (commented):
The option seems fine to me. Def curious to see what the changelog looks like. We might even want changelog generators to have a separate export to define what we say in the case of "this package hasn't changed at all but it's being released so all the versions are the same"

__fixtures__/fixed-packages/.changeset/config.js
Andarist: I have no idea - I've just copied the `linked-packages` fixtures and modified it slightly. I will verify tomorrow if this can be removed.

Andarist:
> The option seems fine to me. Def curious to see what the changelog looks like. We might even want changelog generators to have a separate export to define what we say in the case of "this package hasn't changed at all but it's being released so all the versions are the same"

Definitely lots to discuss when it comes to the changelog generation stuff - I will reach out to you in the coming days to talk this through further, but in the meantime you could take a look at this: https://github.com/changesets/changesets/issues/683 . I have some new ideas/thoughts on this issue - but the use cases mentioned there are still relevant.

fivethreeo:
What is the status of this pull? Would it be possible to try this now if I install changesets from the branch?

Andarist:
@fivethreeo I'm waiting for review and need to recheck the todo list added in the thread. 

You can try it out by installing this build:
```
yarn add --dev @changesets/cli@https://pkg.csb.dev/changesets/changesets/commit/98401c27/@changesets/cli/_pkg.tgz
```

packages/assemble-release-plan/src/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  let highestReleaseType = releases[0].type;
+
+  for (let release of releases) {
+    switch (release.type) {
+      case "major":
+        return "major";
+      case "minor":
+        highestReleaseType = "minor";
+        break;
+      // patch never needs to override another value
```
mitchellhamilton, line 24: Wouldn't patch need to override a `none`?

mitchellhamilton:
A test that generates a changelog would be great (I imagine for something being bumped that didn't have any changesets it would just have the version and nothing else?)

Andarist:
@mitchellhamilton just added such a test. I think that an "empty" entry is acceptable for now - we should gather feedback from users before introducing more options. It is very likely that most people who want the fixed mode might want aggregate changelogs (https://github.com/changesets/changesets/issues/683) and this would "solve" this problem without adding new functions to changelog generators.

mitchellhamilton (approved):

PR #693, opened 2021-12-06
Fix bug in 'flatten releases' function where 'none' release types don't play nicely with other release types

luciaquirke:
Patch/minor/major releases can be overridden by none releases during the 'assemble release plan' stage of `changeset version`, making package versions fail to be bumped.

The fix consists of handling the 'none' release type in the `flatten-releases` function, and slightly simplifying the logic.


packages/assemble-release-plan/src/index.test.ts
```diff
@@ -74,6 +74,53 @@ describe("assemble-release-plan", () => {
...
+        { name: "pkg-c", type: "major" }
+      ]
+    });
+    setup.addChangeset({
+      id: "big-cats-yelp",
+      releases: [
+        { name: "pkg-a", type: "none" },
+        { name: "pkg-b", type: "none" },
+        { name: "pkg-c", type: "none" }
+      ]
```
Andarist, line 100: in a way, this part is not really required for this test, but I guess it doesn't hurt it

.changeset/wild-schools-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": patch
```
Andarist, line 2: q: is only this package affected? or is the CLI affected too? or maybe is this problem "fixed up" in some other part of the code when CLI runs end to end?

luciaquirke: I don't think the bug is fixed up anywhere because we found some packages in production that hadn't been bumped as a result of this bug. The CLI is affected because it calls @changesets/assemble-release-plan.

Andarist, line 2:
```suggestion
"@changesets/assemble-release-plan": patch
"@changesets/cli": patch
```

.changeset/wild-schools-brake.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fix flatten-releases so 'none' release types don't override other release types
```
Andarist, line 5:
```suggestion
Fixed an issue with `none` release type sometimes overriding other release types and thus preventing a release from happening.
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -102,6 +102,53 @@ describe("assemble-release-plan", () => {
     expect(releases[0].type).toEqual("major");
     expect(releases[0].newVersion).toEqual("2.0.0");
   });
+  it("should handle 'none' changeset types with other changeset types", () => {
```
Andarist, line 105:
```suggestion
  it("`none` changeset should not override other release types", () => {
```

Andarist (approved):

PR #695, opened 2021-12-07
Setup CodeSandbox CI for PRs

Andarist:
This replaces and closes https://github.com/changesets/changesets/pull/681

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,0 +1,4 @@
+{
+  "packages": ["packages/*"],
+  "node": "14"
+}


package.json
```diff
@@ -16,7 +16,8 @@
     "changeset": "packages/cli/bin.js",
     "check-all": "yarn test && yarn lint && yarn types:check",
     "version-packages": "changeset version && yarn prettier",
-    "release": "yarn build && changeset publish"
+    "release": "yarn build && changeset publish",
+    "codesandbox:build": "yarn build && yarn version-packages"
```
mitchellhamilton, line 20: I think `version-packages` would fail if there were changesets on CodeSandbox CI since it won't have a GitHub token for the changelog generation. We should just be able to remove it.
```suggestion
    "codesandbox:build": "yarn build"
```

Andarist: Good catch, I've fixed that. The PR is ready for re-review

mitchellhamilton (approved):

PR #701, opened 2021-12-08
Added support for a `--no-git-tag` CLI flag that can be used with `changeset publish`

Andarist:
This should be mostly used together with snapshot releases. They are usually created in the CI environment with custom scripts. So even though the tags were always created there it didn't matter that much because nobody has been pushing them on their own.

However, it feels like for snapshot releases the git tags are rarely what you actually want - and this matters especially if you want to prepare a snapshot machine locally, on your own machine. So to make this use case easier I'm proposing to introduce this flag.

It somewhat feels like a stopgap solution - but since `--snapshot` is not used with `changeset publish` I don't currently have a better idea to do this. Unless we'd infer the release type from the released package versions - this kinda seems magical though and I'm worried about introducing such a solution prematurely.

### DIFF
--- a/packages/cli/src/commands/publish/index.ts
+++ b/packages/cli/src/commands/publish/index.ts
@@ -39,7 +39,7 @@ function showNonLatestTagWarning(tag?: string, preState?: PreState) {
 
 export default async function run(
   cwd: string,
-  { otp, tag }: { otp?: string; tag?: string },
+  { otp, tag, gitTag = true }: { otp?: string; tag?: string; gitTag?: boolean },
   config: Config
 ) {
   const releaseTag = tag && tag.length > 0 ? tag : undefined;
@@ -72,20 +72,23 @@ export default async function run(
   if (successful.length > 0) {
     success("packages published successfully:");
     logReleases(successful);
-    // We create the tags after the push above so that we know that HEAD wont change and that pushing
-    // wont suffer from a race condition if another merge happens in the mean time (pushing tags wont
-    // fail if we are behind the base branch).
-    log(`Creating git tag${successful.length > 1 ? "s" : ""}...`);
-    if (tool !== "root") {
-      for (const pkg of successful) {
-        const tag = `${pkg.name}@${pkg.newVersion}`;
+
+    if (gitTag) {
+      // We create the tags after the push above so that we know that HEAD wont change and that pushing
+      // wont suffer from a race condition if another merge happens in the mean time (pushing tags wont
+      // fail if we are behind the base branch).
+      log(`Creating git tag${successful.length > 1 ? "s" : ""}...`);
+      if (tool !== "root") {
+        for (const pkg of successful) {
+          const tag = `${pkg.name}@${pkg.newVersion}`;
+          log("New tag: ", tag);
+          await git.tag(tag, cwd);
+        }
+      } else {
+        const tag = `v${successful[0].newVersion}`;
         log("New tag: ", tag);
         await git.tag(tag, cwd);
       }
-    } else {
-      const tag = `v${successful[0].newVersion}`;
-      log("New tag: ", tag);
-      await git.tag(tag, cwd);
     }
   }
 
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -7,13 +7,13 @@ import { run } from "./run";
 const { input, flags } = meow(
   `
   Usage
-    $ changesets [command]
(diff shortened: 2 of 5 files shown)


packages/cli/src/commands/publish/index.ts
```diff
@@ -72,20 +72,23 @@ export default async function run(
...
-    // fail if we are behind the base branch).
-    log(`Creating git tag${successful.length > 1 ? "s" : ""}...`);
-    if (tool !== "root") {
-      for (const pkg of successful) {
-        const tag = `${pkg.name}@${pkg.newVersion}`;
+
+    if (gitTag) {
+      // We create the tags after the push above so that we know that HEAD wont change and that pushing
+      // wont suffer from a race condition if another merge happens in the mean time (pushing tags wont
+      // fail if we are behind the base branch).
```
Andarist, line 79: quite frankly - i have no idea what that comment is about, it might refer to some older implementation

packages/cli/src/index.ts
```diff
@@ -7,13 +7,13 @@ import { run } from "./run";
 const { input, flags } = meow(
   `
   Usage
-    $ changesets [command]
+    $ changeset [command]
```
Andarist, line 10: this kinda showcase how confusing the name of the script is - should we create two executables: changeset and changesets?

packages/cli/src/index.ts
```diff
@@ -7,13 +7,13 @@ import { run } from "./run";
...
   Commands
     init
     add [--empty] [--open]
-    version [--ignore]
-    publish [--otp=code]
-    status [--since <branch>] [--verbose] [--output=JSON_FILE.json]
+    version [--ignore] [--snapshot <?name>]
+    publish [--tag <name>] [--otp <code>] [--no-git-tag]
+    status [--since <branch>] [--verbose] [--output JSON_FILE.json]
```
Andarist, line 16: I've just unified this a little bit and added stuff that was missing from here

packages/cli/src/index.ts
```diff
@@ -48,7 +48,14 @@ const { input, flags } = meow(
...
       open: {
         type: "boolean"
+      },
+      gitTag: {
+        type: "boolean",
+        default: true
       }
+      // mixed type like this is not supported by `meow`
+      // if it gets passed explicitly then it's still available on the flags with an inferred type though
+      // snapshot: { type: "boolean" | "string" },
```
Andarist, line 58: i've left this out as a comment since I believe that every supported option should be listed here.

packages/cli/src/types.ts
```diff
@@ -1,9 +1,4 @@
-import { AccessType } from "@changesets/types";
-
 export type CliOptions = {
-  commit?: boolean;
-  changelog?: string;
-  access?: AccessType;
```
Andarist, line 6: those were unused and unsupported - probably leftovers from the previous implementation

packages/cli/src/types.ts
```diff
@@ -13,6 +8,7 @@ export type CliOptions = {
   ignore?: string | string[];
   snapshot?: string | boolean;
   tag?: string;
+  gitTag?: boolean;
```
Andarist, line 11: I'm not sure about the name of this option. Maybe it should be something like `--skip-git-tags`? 

One thing I've noticed just now is that `changeset tag` creates git tags, while `--tag` (accepted by the publish command) is used for npm/release tagging. Can we do better? Should we rename `changeset tag` to be something else (like `changeset git-tag`)?

packages/cli/src/commands/publish/index.ts
mitchellhamilton: I think what it's referring to is if you're publishing on every git push, you don't want to create the tags then publish to npm because whether it's published or not is what determines whether it needs to do a publish.  (basically replace the `the push above` with `the publish above` in the comment)

packages/cli/src/types.ts
mitchellhamilton: `--no-git-tag` seems fine to me

> Should we rename changeset tag to be something else (like changeset git-tag)?

I don't think it's _that bad_ but maybe.

mitchellhamilton (approved):

packages/cli/src/commands/publish/index.ts
Andarist: Hm, i still dont get it - we query npm info to check if something should be published or not, so creating git tags should affect this at all

PR #703, opened 2021-12-10
Fixed issues with `*` ranges not being replaced in premode

Andarist:


packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -183,12 +182,24 @@ function getDependencyVersionRanges(
...
+
+    if (versionRange.startsWith("workspace:")) {
       dependencyVersionRanges.push({
         depType: type,
-        versionRange: deps[dependencyName].replace("workspace:", "")
+        versionRange:
+          // intentionally keep other workspace ranges untouched
+          // this has to be fixed but this should only be done when adding appropriate tests
+          versionRange === "workspace:*"
+            ? // workspace:* actually means the current exact version, and not a wildcard similar to a reguler * range
```
mitchellhamilton, line 195: 😶

Andarist: Hehe, ye - this is weird. Basically `workspace:*` means "replace with the current version (exact)", `workspace:^` means "replace with the current version, modified with ^" (and the same applies to `workspace:~`).

mitchellhamilton (approved):

PR #704, opened 2021-12-12
Accept tag dependencies as valid

Andarist:


packages/cli/src/commands/version/version.test.ts
```diff
@@ -267,6 +267,37 @@ describe("running version in a simple project", () => {
     expect(changesetDir).toContain(".ignored-temporarily.md");
   });
 
+  it("should not update a dependant that uses a tag as a dependency rage for a package that could otherwise be local", async () => {
```
Andarist, line 270: This test is already passing on the main branch but I have thought that it might be worth it to add it here nevertheless. The only difference is that this will no longer print errors in the console - this is being checked in a unit test for `getDependencyGraph` so I've skipped rechecking this here.

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -65,35 +79,40 @@ export default function getDependencyGraph(
...
-      if (!semver.satisfies(expected, depVersion)) {
+      const range = getValidRange(depRange);
+
+      // `depRange` could have been a tag and if a tag has been used there might have been a reason for that
+      // we should not count this as a local monorepro dependant
+      if (!range && !isProtocolRange(depRange)) {
+        continue;
+      }
+
+      if (!range || !range.test(expected)) {
```
Andarist, line 108: this is kinda `if (!isProtocolRange(depRange) || !range.test(expected))` but that wouldn't quite satisfy TS as `range` could still be null

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -65,39 +79,44 @@ export default function getDependencyGraph(
...
         }
       } else if (opts?.bumpVersionsWithWorkspaceProtocolOnly === true) {
         continue;
       }
 
-      // internal dependencies only need to semver satisfy, not '==='
-      if (!semver.satisfies(expected, depVersion)) {
+      const range = getValidRange(depRange);
+
+      if ((range && !range.test(expected)) || isProtocolRange(depRange)) {
```
mitchellhamilton, line 102: Isn't `isProtocolRange(depRange)` doing nothing here(and also the other place it's used)? Since if it's a "protocol range" then it's not a valid semver range?

mitchellhamilton (commented):
By, tag, you mean a dep with anything other than just a semver range, right?

Andarist:
> By, tag, you mean a dep with anything other than just a semver range, right?

Somewhat, kinda - there is not an obvious way to distinguish a tag from other things. If we take a look at [this list](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies) we can distinguish such things:
- semver ranges/versions
- HTTP URLs, git (so with a protocol)
- */''/x/X (wildcards)
- user/repo (special case for GitHub)
- tags
- path/path/path (local paths)

We know though that we also have other protocols, like `workspace:`, `link:` and even `patch:` (available at least in Yarn Berry). 

So the code in this PR doesn't target tags specifically as I don't ensure that some other categories stay unmatched. However, when I look at this - those categories should probably also stay untouched in the same way. We could add some tests for those - but I'm not sure if it's worth it right now (given my limited time to work on this)

packages/get-dependents-graph/src/get-dependency-graph.ts
Andarist: Not quite - for instance `link:` would result in `range === null` and we might not want to accept it for production dependencies. In general - this code might not be complete, but it's more complete than the previous version. I want to understand the use cases better before allowing for more things.

Some more information can be deduced from https://github.com/changesets/changesets/pull/704#issuecomment-993406755

mitchellhamilton: Oh sorry, I was reading this wrong

mitchellhamilton:
Seems strange that we wouldn't error for tags and GitHub repos but would for things that have explicit protocols? Feels like we should just only error when it's a range but it's out of range of the actual version?

Andarist:
> Seems strange that we wouldn't error for tags and GitHub repos but would for things that have explicit protocols? Feels like we should just only error when it's a range but it's out of range of the actual version?

We have an explicit test that disallows `link:` protocol to be used for production dependencies:
https://github.com/changesets/changesets/blob/3281faa736407bc927f2e8e1ac26f619a1a94918/packages/get-dependents-graph/src/get-dependency-graph.test.ts#L78

Like I've mentioned in [other comment of mine](https://github.com/changesets/changesets/pull/704#discussion_r768535594) here:
> In general - this code might not be complete, but it's more complete than the previous version. I want to understand the use cases better before allowing for more things.

I'm unsure how all of the protocol types work with packing - and I assume that this might be different per package manager (I hope that this is not pm-dependent but I assume the worst :P). And I don't have time to investigate this right now - so I've decided to leave the existing behavior untouched and fix what I know that has to be fixed.

mitchellhamilton (approved):

PR #705, opened 2021-12-12
Fix issue with bumping packages without `package.json#version`

Andarist:


packages/cli/src/commands/version/version.test.ts
```diff
@@ -267,6 +267,83 @@ describe("running version in a simple project", () => {
     expect(changesetDir).toContain(".ignored-temporarily.md");
   });
 
+  it("should throw an error if a package.json is missing the name field", async () => {
```
Andarist, line 270: This is not a new test - I've only found an old fixture that was currently unused and I've thought that it would be nice to reintroduce a test for this (as the original test for it was moved to `manypkg`)

emmatown (commented):
Making it `0.0.0` if there's no version feels like weird behavior to me. I feel like I'd rather have an error?

Andarist:
> Making it 0.0.0 if there's no version feels like weird behavior to me. I feel like I'd rather have an error?

That is also an option I've considered - it would be more of a breaking change though, especially for snapshot releases as previously those would be successfully and "correctly" published as `0.0.0-timestamp`.

I'm OK with throwing though - if we face any backlash we can always reconsider.

Crunchyman-ralph:
I'm running into this exact issue, can we ping someone who can merge this ? @emmatown ?

bluwy:
I'll close this as this is a bit outdated. Right now the behaviour if a package without version is added to the changeset, it'd error with `Mixed changesets that contain both ignored and not ignored packages are not allowed`, or simply ignore the changeset if it only contain the package without version

PR #706, opened 2021-12-13
Fixed an issue with none releases ending up as snapshot releases

Andarist:


mitchellhamilton (commented):
This is confusing me, the tests seem to be about ignored packages but the title sounds like it's about changesets with `none` releases?

Andarist:
> This is confusing me, the tests seem to be about ignored packages but the title sounds like it's about changesets with none releases?

That is cause updating a dependency of an ignored package ends up being a none release for that ignored package - so it doesn't get released but so we can adjust its dependency ranges. This reflects the actual use case which got me to fix this.

There might be also a more basic test case to be written here though - I didn't evaluate it. The fixed bug is about a situation in which Changesets itself creates a none release for a package on behalf of the user. Can this happen in any other situation? I agree though that maybe the changeset description should be adjusted here

Andarist:
@mitchellhamilton I've tweaked the changeset and the test title, I've also added a new test case for a situation with explicit none release type. Could you re-review it?

mitchellhamilton (approved):

PR #708, opened 2021-12-13
Upgrade `is-ci`

Andarist:


packages/cli/src/utils/isCI.ts
```diff
@@ -1,4 +0,0 @@
-// @ts-ignore
-import isCI from "is-ci";
-
-export default !!(isCI || process.env.GITHUB_ACTIONS);
```
Andarist, line 4: this is no longer needed since `is-ci` detects GitHub Actions correctly now: https://github.com/watson/ci-info/pull/42#issuecomment-780935392

mitchellhamilton (approved):

Issue #707, opened 2021-12-13
npm Automation tokens not supported


b3nThomas:
## Affected Packages
`@changesets/cli@2.18.1`

## Problem
`changeset publish` attempts to use the `npm profile get --json` command to work out if 2fa is enabled etc.
When using an Automation level npm token, this throws a 403 error as these tokens literally only have access to read or publish, and I think `npm whoami` too.:
```
13:11:02                      echo //registry.npmjs.org/:_authToken=**** > .npmrc &&
13:11:02                      npm profile get --json
13:11:02                  '
13:11:03  []                   / profile:
[]                   / profile:
[]                   - profile:
[]                   - profile:
[]                   - profile:
[]                   \ profile:
[]                   \ profile:

npm ERR! code E403
13:11:03  npm ERR! 403 403 Forbidden - GET https://registry.npmjs.org/-/npm/v1/user
13:11:03  npm ERR! 403 In most cases, you or one of your dependencies are requesting
13:11:03  npm ERR! 403 a package version that is forbidden by your security policy, or
13:11:03  npm ERR! 403 on a server you do not have access to.
13:11:03  {
13:11:03    "error": {
13:11:03      "code": "E403",
13:11:03      "summary": "403 Forbidden - GET https://registry.npmjs.org/-/npm/v1/user",
13:11:03      "detail": "In most cases, you or one of your dependencies are requesting\na package version that is forbidden by your security policy, or\non a server you do not have access to."
13:11:03    }
13:11:03  }
13:11:03  
13:11:03  npm ERR! A complete log of this run can be found in:
13:11:03  npm ERR!     /root/.npm/_logs/2021-12-13T13_11_03_532Z-debug.log
13:11:03  
[Pipeline] }
```

I had to work backwards to find this command as you actually only get this little error message:
```
13:06:31   1 file changed, 1 insertion(+), 1 deletion(-)
13:06:32  78🦋  info npm info @******/*****
13:06:33  error parsing json: 
13:06:33  SyntaxError: Unexpected end of JSON input
13:06:33      at JSON.parse (<anonymous>)
13:06:33      at jsonParse (/usr/src/app/node_modules/@changesets/cli/dist/cli.cjs.dev.js:601:17)
13:06:33      at getTokenIsRequired (/usr/src/app/node_modules/@changesets/cli/dist/cli.cjs.dev.js:651:14)
13:06:33      at processTicksAndRejections (node:internal/process/task_queues:96:5)
```
[...]

PR #710, opened 2021-12-15
Errors received from `npm profile get --json` will now always be printed

Andarist:
closes https://github.com/changesets/changesets/issues/707

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -65,6 +65,13 @@ export async function getTokenIsRequired() {
   let result = await spawn("npm", ["profile", "get", "--json"], {
     env: Object.assign({}, process.env, envOverride)
   });
+  if (result.code !== 0) {
+    error(
+      "error while checking if token is required",
+      result.stderr.toString().trim() || result.stdout.toString().trim()
+    );
+    return false;
+  }
   let json = jsonParse(result.stdout.toString());
   if (json.error || !json.tfa || !json.tfa.mode) {
     return false;
--- a/.changeset/friendly-mugs-drive.md
+++ b/.changeset/friendly-mugs-drive.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Errors received from `npm profile get --json`, that is used to check if 2FA is required, will now always be printed.


packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -65,6 +65,13 @@ export async function getTokenIsRequired() {
   let result = await spawn("npm", ["profile", "get", "--json"], {
     env: Object.assign({}, process.env, envOverride)
   });
+  if (result.code !== 0) {
+    error(
+      "error while checking if token is required",
+      result.stderr.toString().trim() || result.stdout.toString().trim()
+    );
+    return false;
```
Andarist, line 73: I'm not super fond of this but this is what is already being returned below `if (json.error)`. So I've figured out that this should just behave the same now.

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -65,6 +65,13 @@ export async function getTokenIsRequired() {
   let result = await spawn("npm", ["profile", "get", "--json"], {
     env: Object.assign({}, process.env, envOverride)
   });
+  if (result.code !== 0) {
+    error(
+      "error while checking if token is required",
+      result.stderr.toString().trim() || result.stdout.toString().trim()
+    );
```
Andarist, line 72: It would be great to establish a cohesive logging format, but I kinda don't have the mental energy for that right now. 

As to the `stderr || stdout` stuff - this is the same stuff that got fixed [here](https://github.com/changesets/changesets/pull/676). npm 7 has changed the behavior and they have started to print json errors to stderr

mitchellhamilton (approved):

packages/cli/src/commands/publish/npm-utils.ts
mitchellhamilton: The whole checking whether 2fa is required before trying to publish is really just an optimization, if it turns out that it is required but this returned false, the publishing will handle that fine and prompt so returning false here is kinda totally fine tbh.

Andarist: In such a case - maybe it’s not worth to print the error at all?

mitchellhamilton: I think it's still nice to see the error otherwise I'd be afraid that this would silently break at some point in the future and publish would just be slower but there wouldn't be a clear reason.

PR #713, opened 2021-12-15
Fixed an issue with pre mode needs to be re-enterable after exiting

### LINKED ISSUE #712: pre mode needs to be re-enterable after exiting
## Affected Packages

pkg:@changesets/pre

## Problem

Below is the scene that how I used and the problem

```bash
# do something, and want to bump a prerelease alpha version to test

$ changeset pre enter alpha
🦋  success Entered pre mode with tag alpha
🦋  info Run `changeset version` to version packages with prerelease versions

# do something, like commit and version

$ changeset pre exit
🦋  success Exited pre mode
🦋  info Run `changeset version` to version packages with normal versions

# do something ...
# I don't want to bump a patch version or publish now,
# and want to continue bump alpha version,
#
# so I need enter pre mode again,
# in this time, it throw an Error, even though I have really exited pre mode above
$ changeset pre enter alpha
🦋  error `changeset pre enter` cannot be run when in pre mode
🦋  info If you're trying to exit pre mode, run `changeset pre exit`

$ cat .changeset/pre.json
{
  "mode": "exit",
  "tag": "alpha",
  "initialVersions": {
    ...
  },
  "changesets": [
    "xxx"
  ]
}
```

I know I can edit `pre.json` manually, but I think it needs to be improved in changesets.

## Proposed solution

[Here](https://github.com/changesets/changesets/blob/%40changesets/pre%401.0.8/packages/pre/src/index.ts#L52-L54) the code of function [`enterPre`](https://github.com/changesets/changesets/blob/%40changesets/pre%401.0.8/packages/pre/src/index.ts#L52-L54), and I noticed there is a TODO that say `verify that the pre state isn't broken`, I think it means like this:

```typescript
  export async function enterPre(cwd: string, tag: string) {
   let packages = await getPackages(cwd);
   let preStatePath = path.resolve(packages.root.dir, ".changeset", "pre.json");
-  // TODO: verify that the pre state isn't broken
   let preState: PreState | undefined = await readPreState(packages.root.dir);
-  if (preState !== undefined) {
+  if (preState !== undefined && preState.mode === "pre") {
     throw new PreEnterButInPreModeError();
   }
   let newPreState: PreState = {
     mode: "pre",
     tag,
     initialVersions: {},
-    changesets: []
+    changesets: preState?.changesets ?? []
   };
     newPreState.initialVersions[pkg.packageJson.name] = pkg.packageJson.version;
   }
   await fs.writeFile(preStatePath, JSON.stringify(newPreState, null, 2) + "\n");
 }
```

zthxxx:
closes #712

### DIFF
--- a/.changeset/afraid-dots-know.md
+++ b/.changeset/afraid-dots-know.md
@@ -3,4 +3,4 @@
 "@changesets/pre": patch
 ---
 
-Fixed an issue with not being to reenter exited pre mode.
+Fixed an issue with not being able to reenter exited pre mode.
--- a/.changeset/afraid-dots-know.md
+++ b/.changeset/afraid-dots-know.md
@@ -1,5 +1,6 @@
 ---
 "@changesets/cli": patch
+"@changesets/pre": patch
 ---
 
-support re-enter pre mode if already exited
+Fixed an issue with not being to reenter exited pre mode.
--- a/.changeset/afraid-dots-know.md
+++ b/.changeset/afraid-dots-know.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+support re-enter pre mode if already exited
--- a/.changeset/thirty-panthers-film.md
+++ b/.changeset/thirty-panthers-film.md
@@ -1,0 +1,5 @@
+---
+"@changesets/pre": patch
+---
+
+pre mode needs to be re-enterable after exiting, close #712
--- a/.changeset/thirty-panthers-film.md
+++ b/.changeset/thirty-panthers-film.md
@@ -1,5 +1,2 @@
 ---
-"@changesets/pre": patch
 ---
-
-pre mode needs to be re-enterable after exiting, close #712
--- a/packages/pre/src/index.test.ts
+++ b/packages/pre/src/index.test.ts
@@ -20,6 +20,16 @@ let preStateForSimpleProject: PreState = {
   tag: "next"
 };
 
+let preStateForExited: PreState = {
+  changesets: ["slimy-dingos-whisper"],
+  initialVersions: {
+    "pkg-a": "1.0.0",
+    "pkg-b": "1.0.0"
+  },
+  mode: "exit",
+  tag: "beta"
+};
+
 describe("enterPre", () => {
   it("should enter", async () => {
     let cwd = f.copy("simple-project");
@@ -37,6 +47,21 @@ describe("enterPre", () => {
     );
     await expect(enterPre(cwd, "some-tag")).rejects.toBeInstanceOf(
       PreEnterButInPreModeError
+    );
+  });
+  it("should enter if already exited pre mode", async () => {
+    let cwd = f.copy("simple-project");
+    await fs.writeJSON(
+      path.join(cwd, ".changeset", "pre.json"),
+      preStateForExited
+    );
+    await enterPre(cwd, "next");
+    expect(await fs.readJson(path.join(cwd, ".changeset", "pre.json"))).toEqual(
+      {
+        ...preStateForExited,
+        mode: "pre",
+        tag: "next"
+      }
     );
   });
 });
--- a/packages/pre/src/index.ts
+++ b/packages/pre/src/index.ts
@@ -47,16 +47,16 @@ export async function exitPre(cwd: string) {
 export async function enterPre(cwd: string, tag: string) {
   let packages = await getPackages(cwd);
   let preStatePath = path.resolve(packages.root.dir, ".changeset", "pre.json");
-  // TODO: verify that the pre state isn't broken
-  let preState = await readPreState(packages.root.dir);
-  if (preState !== undefined) {
+  let preState: PreState | undefined = await readPreState(packages.root.dir);
+  // can't reenter if pre mode still exists, but enterable if already exited
+  if (preState?.mode === "pre") {
     throw new PreEnterButInPreModeError();
   }
   let newPreState: PreState = {
     mode: "pre",
     tag,
     initialVersions: {},
-    changesets: []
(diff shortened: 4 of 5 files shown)


packages/pre/src/index.ts
```diff
@@ -47,16 +47,16 @@ export async function exitPre(cwd: string) {
 export async function enterPre(cwd: string, tag: string) {
   let packages = await getPackages(cwd);
   let preStatePath = path.resolve(packages.root.dir, ".changeset", "pre.json");
-  // TODO: verify that the pre state isn't broken
-  let preState = await readPreState(packages.root.dir);
-  if (preState !== undefined) {
+  let preState: PreState | undefined = await readPreState(packages.root.dir);
+  // can't reenter if pre mode still exists, but enterable if already exited
+  if (preState?.mode === "pre") {
```
Andarist, line 52: IMHO this makes sense

packages/pre/src/index.ts
```diff
@@ -47,16 +47,16 @@ export async function exitPre(cwd: string) {
...
     throw new PreEnterButInPreModeError();
   }
   let newPreState: PreState = {
     mode: "pre",
     tag,
     initialVersions: {},
```
Andarist, line 58: should `initialVersions` be "reset" though? My first instinct would be that the `newPreState` should just have `tag` and `mode` updated

packages/pre/src/index.test.ts
```diff
@@ -39,6 +49,21 @@ describe("enterPre", () => {
       PreEnterButInPreModeError
     );
   });
+  it("should enter if already exited pre mode", async () => {
```
Andarist, line 52: could you also add a test at the CLI level too? it would go [here](https://github.com/changesets/changesets/blob/ab8f2b29d5cf921d97987a665051b67a7ec55ed3/packages/cli/src/commands/pre/index.test.ts#L27) and I imagine it would look smth like this:
1. `f.copy`
2. `changeset pre enter beta`
3. do stuff
4. `changeset pre exit`
5. `changeset pre enter beta`

packages/pre/src/index.ts
Andarist: ah, this is only initializing the value, and versions are filled in in the loop below this. So the question is - should the current versions be put as `initialVersions`? or should the existing `initialVersions` be preserved?

zthxxx: It doesn't make any difference, the `initialVersions` not really be used anywhere yet.

packages/pre/src/index.test.ts
zthxxx: Done.

Due to both `exit` `enter` are independent operator, and the connection between them is only `pre.json`,

so I only test the "re-enter" step, rather than the flow 2 to 4 in your given

packages/pre/src/index.ts
Andarist: You are right, `initialVersions` are not actually used anywhere - I'm not sure why do we even keep this info around 🤔 I've asked Mitchell a question about this.

packages/pre/src/index.ts
zthxxx: I'm not sure what need I do or not?  👀

Remove the `// comment`, or remove types, or just click resolve this conversation?

Andarist: i was just confirming that this makes sense to me 😉 nothing to change here :)

packages/pre/src/index.ts
zthxxx: Any update about this question?

Or It's exactly another issue and not about this PR "re-enterable pre mod", so can I just click "resolve conversation" then open an issue to discuss?

packages/pre/src/index.ts
```diff
@@ -47,16 +47,16 @@ export async function exitPre(cwd: string) {
 export async function enterPre(cwd: string, tag: string) {
   let packages = await getPackages(cwd);
   let preStatePath = path.resolve(packages.root.dir, ".changeset", "pre.json");
-  // TODO: verify that the pre state isn't broken
-  let preState = await readPreState(packages.root.dir);
-  if (preState !== undefined) {
+  let preState: PreState | undefined = await readPreState(packages.root.dir);
+  // can't reenter if pre mode still exists, but enterable if already exited
```
Andarist, line 51:
```suggestion
  // can't reenter if pre mode still exists, but we should allow exited pre mode to be reentered
```

.changeset/thirty-panthers-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/pre": patch
+---
+
+pre mode needs to be re-enterable after exiting, close #712
```
Andarist, line 5:
```suggestion
---
---
```

.changeset/afraid-dots-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
```
Andarist, line 3:
```suggestion
---
"@changesets/cli": patch
"@changesets/pre": patch
---
```

.changeset/afraid-dots-know.md
```diff
@@ -0,0 +1,5 @@
...
+
+support re-enter pre mode if already exited
```
Andarist, line 5:
```suggestion
Fixed an issue with not being to reenter exited pre mode.
```

.changeset/afraid-dots-know.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/cli": patch
+"@changesets/pre": patch
+---
+
+Fixed an issue with not being to reenter exited pre mode.
```
Andarist, line 6:
```suggestion
Fixed an issue with not being able to reenter exited pre mode.
```

PR #715, opened 2021-12-16
Cleanup test output

Andarist:


packages/release-utils/src/run.test.ts
```diff
@@ -22,6 +23,7 @@ const writeChangesets = (changesets: Changeset[], cwd: string) => {
 };
 
 jest.setTimeout(10000);
+silenceLogsInBlock();
```
Andarist, line 26: the logs created by this test were especially noisy - cause they were logging all the executed git commands and stuff like that and made it harder to quickly check which tests failed etc

packages/release-utils/src/utils.ts
```diff
@@ -96,12 +97,12 @@ export async function execWithOutput(
...
 ) {
-  console.log(`Running ${command} ${args.join(" ")}`);
+  process.stdout.write(`Running: ${command} ${args.join(" ")}` + os.EOL);
   let childProcess = spawn(command, args, {
     cwd: options.cwd
   });
-  childProcess.on("stdout", data => console.log(data.toString()));
-  childProcess.on("stderr", data => console.error(data.toString()));
+  childProcess.on("stdout", data => process.stdout.write(data));
+  childProcess.on("stderr", data => process.stderr.write(data));
```
Andarist, line 105: this is more correct because the received data doesn't have to end up with a new line

packages/test-utils/src/index.ts
```diff
@@ -8,24 +8,77 @@
...
+  const originalLoggerWarn = logger.warn;
+  const originalLoggerSuccess = logger.success;
+
+  const originalConsoleError = console.error;
+  const originalConsoleInfo = console.info;
+  const originalConsoleLog = console.log;
+  const originalConsoleWarn = console.warn;
+
+  const originalStdoutWrite = process.stdout.write;
+  const originalStderrWrite = process.stderr.write;
```
Andarist, line 24: This has a drawback - it might be harder to use `console.log` in the "silenced" tests. I can live with that - WDYT?

types/spawndamnit.d.ts
```diff
@@ -1,9 +1,10 @@
...
-  import { SpawnOptions, ChildProcess } from "child_process";
+  import { SpawnOptions } from "child_process";
+  import { EventEmitter } from "events";
 
   export default function spawn(
     cmd: string,
     args: Array<string>,
     opts?: SpawnOptions
-  ): Promise<{ stdout: Buffer; code: number; stderr: Buffer }> & ChildProcess;
+  ): Promise<{ stdout: Buffer; code: number; stderr: Buffer }> & EventEmitter;
```
Andarist, line 9: the previous version was incorrect:
https://github.com/jamiebuilds/spawndamnit/blob/9fec44510749481dd11906f3a979ed9ebdda4fed/promise.js#L18

mitchellhamilton (approved):

PR #720, opened 2021-12-20
fix: make cli skip private packages without version field

### LINKED ISSUE #620: Omit private packages from lists of changed and unchanged packages offered by the CLI when running `yarn changeset`
## Affected Packages

- @changesets/cli

## Problem

The list of packages offered by `yarn changeset` includes [private packages](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#private) (i.e. packages with `private: true` in package.json). In projects with broad contribution, contributors may not be aware that only packages that get published should be included in changesets, and include those private packages. This creates noise for maintainers, who have to explain the issue and ask the contributor to fix the changeset.

## Proposed solution

If a package is marked as private, it shouldn't be included in the lists of changed and unchanged packages offered by the CLI. This behavior could be behind an optional flag if there are workflows where including private packages in this list makes sense.

bhovhannes:
This closes #620 

Per conversation there, it was decided to skip packages without "version" field in CLI. I limited that to _private_ packages without "version" field to limit the blast radius of this behavior change.


bhovhannes:
@Andarist I'll be grateful for you review. I believe this does not contradict with #705 .

I am not sure which type of tests to write to this change and where, as the change affects what CLI is showing, which is mocked in existing tests I've looked at. I can spend more time on tests once we'll be settled with implementation.

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -58,6 +58,14 @@ async function getPackagesToRelease(
     );
   }
 
+  const pkgJsonsByName = getPkgJsonByName(allPackages);
+
+  // filter out private packages without 'version' field
+  allPackages = allPackages.filter(pkg => !isInternalPackage(pkg.packageJson));
+  changedPackages = changedPackages.filter(
+    pkgName => !isInternalPackage(pkgJsonsByName.get(pkgName)!)
+  );
```
Andarist, line 67: Hm, I guess it makes sense to filter this here - as we might need to keep those packages in the "true" `allPackages` so we can still treat them as internal packages so we can update their dependency ranges etc. The only valid changeset type for them should be `'none'`

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -96,6 +104,16 @@ function formatPkgNameAndVersion(pkgName: string, version: string) {
   return `${bold(pkgName)}@${bold(version)}`;
 }
 
+function getPkgJsonByName(packages: Package[]) {
+  return new Map(
+    packages.map(({ packageJson }) => [packageJson.name, packageJson])
+  );
+}
+
+function isInternalPackage(packageJson: PackageJSON) {
```
Andarist, line 113: We sometimes refer to in-monorepo packages as "internal" so this function name might be a little bit confusing. Maybe this should be changed to smth like `isNonVersionablePackage`?

bhovhannes: Renamed to `isNonVersionablePackage`. It is longer but reflects better what function does. Thanks!

packages/cli/src/commands/add/createChangeset.ts
bhovhannes: I am not sure I understand you correctly here. Do you mean that we could exclude private packages without `version` field from `allPackages` even before passing `allPackages` to this function, somewhere in `commands/add/index.ts` ?

Andarist: I say that we could exclude them at a different stage - like when getting the list of packages. But that would prevent us from treating those packages as "internal" and thus we couldn't update dependency ranges in them etc. So I think it makes sense to handle this here.

bhovhannes: I see. I was also thinking about that and at the end decided to filter them here, right before showing list of packages to the user. That way filtering affects only the list which cli presents to the user and does not affect other logic.

bhovhannes:
@Andarist do you have any comments here or this can be merged?

bhovhannes:
@Andarist friendly ping. Can this be merged or something else needs to be done first ?

Andarist:
@bhovhannes thanks for the ping and sorry for going MIA on this one. I have a couple of things to check in relation to private packages and I would like to land this after I investigate all of them, so my decision can take all of the gathered information into account. For the time being, I think this is OK to go but IIRC somebody has raised concerns that such packages might not be "linkable" in some monorepo setups and I need to check this out.

bhovhannes:
> @bhovhannes thanks for the ping and sorry for going MIA on this one. I have a couple of things to check in relation to private packages and I would like to land this after I investigate all of them, so my decision can take all of the gathered information into account. For the time being, I think this is OK to go but IIRC somebody has raised concerns that such packages might not be "linkable" in some monorepo setups and I need to check this out.

I see. Let me know if something will be needed from my side. I plan to migrate one of my monorepos to changesets (pnpm + changesets + turborepo combo) and my monorepo contains private packages without version field, so I'll pay attention to that.

> It seems though that you have rebased this PR in a way that I see now some stuff that shouldn't be here. If you could clean this up - I would highly appreciate it.

rebased again and squashed all my commits into one

PR #730, opened 2022-01-04
Update changesets/action version in workflow

zthxxx:
Due to the master branch has deprecated in https://github.com/changesets/action/pull/120

Also update actions/checkout fetch depth to 1, due to it's out of date after changesets/changesets#495

### DIFF
--- a/.github/workflows/changeset-version.yml
+++ b/.github/workflows/changeset-version.yml
@@ -13,9 +13,6 @@ jobs:
       - name: Checkout Repo
         # https://github.com/actions/checkout
         uses: actions/checkout@v2
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/changeset-version.yml
```diff
@@ -13,8 +13,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@master
         with:
-          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
-          fetch-depth: 0
+          fetch-depth: 1
```
Andarist, line 16: This is the default now so this can be freely dropped.

zthxxx: okay, updated.

zthxxx:
thanks for review

PR #731, opened 2022-01-05
Mention Airnode in changelogs users

Siegrift:
Hi, thanks for this awesome library. We are using it in [Airnode project](https://github.com/api3dao/airnode) and we are really happy how it formalized and simplified our release process.


Andarist:
Thank you for the PR - this list is getting pretty lengthy and I think how to actually downsize it, or replace it with something else (like a list of logos?). There needs to be some popularity threshold (or some other, arbitrary reason to include a project on the list) - otherwise, this would just be totally unmanageable. Given that your project has only 100+ stars I don't think it's best for us to include it on the list right now. I hope that you don't feel offended by it - it's not personal. The project itself looks cool.

PR #732, opened 2022-01-05
Fixed an issue with `changeset version` will bump unrelated packages prerelease version

### LINKED ISSUE #729: `changeset version` with `pre mode: exit` will incorrect bump all pre-release version to normal version, even without any changesets files
## Affected Packages

`@changeset/cli`

## Problem

In my expectation, `changeset version` should only consume "changeset files" and bump version relate to those packages (and any dependency on them),

`changeset version` should NOT bump version of unrelated packages, and most of the time it does (includes no pre.json and `mode: pre` in pre.json).

However, `changeset version` with `pre.json` and `mode: exit` will incorrect bump all pre-release version to normal version, even without any changesets files.

### Reproduce

1. expected, normally changeset version

```bash
ls .changeset
README.md   config.json


head packages/create/package.json
{
  "name": "@leetsolve/create",
  "version": "0.1.1-alpha.1",
  "description": "",


changeset version


# version will not changed, that's expect
head packages/create/package.json
{
  "name": "@leetsolve/create",
  "version": "0.1.1-alpha.1",
  "description": "",
```

2. expected, changeset version with pre mode

```bash
ls .changeset
README.md   config.json   pre.json


# here we are in 'pre mode'
head pre.json
{
  "mode": "pre",
  "tag": "test",
  "initialVersions": {


head packages/create/package.json
{
  "name": "@leetsolve/create",
  "version": "0.1.1-alpha.1",
  "description": "",


changeset version


# version will not changed, that's expect
head packages/create/package.json
{
  "name": "@leetsolve/create",
  "version": "0.1.1-alpha.1",
  "description": "",
```

3. INCORRECT, changeset version with `mode: exit`

```bash
ls .changeset
README.md   config.json   pre.json


# here we exited 'pre mode'
head pre.json
{
  "mode": "exit",
  "tag": "test",
  "initialVersions": {


head packages/create/package.json
{
  "name": "@leetsolve/create",
  "version": "0.1.1-alpha.1",
  "description": "",


changeset version


# INCORRECT, version bumped, but in fact there is no changeset files used in `changeset version`
head packages/create/package.json
{
  "name": "@leetsolve/create",
  "version": "0.1.1",
  "description": "",
```

## Proposed solution

Do not bump version of unrelated packages when exited 'pre mode'

zthxxx:
This closes #729


zthxxx:
@Andarist I'll be grateful for you review. 

I known you are busy with maintaining so many awesome projects, that’s admirable, but I’m not sure another who I should ask for review in changesets.

packages/assemble-release-plan/src/index.ts
```diff
@@ -118,24 +117,6 @@ function assembleReleasePlan(
     releasesValidated = !linksUpdated && !dependentAdded;
   }
 
-  if (preInfo?.state.mode === "exit") {
-    for (let pkg of packages.packages) {
-      // If a package had a prerelease, but didn't trigger a version bump in the regular release,
-      // we want to give it a patch release.
-      // Detailed explanation at https://github.com/changesets/changesets/pull/382#discussion_r434434182
-      if (preInfo.preVersions.get(pkg.packageJson.name) !== 0) {
```
Andarist, line 126: So basically, in a way, this got moved `determineDependents`, right?

packages/cli/src/commands/version/version.test.ts
```diff
@@ -117,14 +116,27 @@ describe("running version in a simple project", () => {
 
   describe("when there are no changeset commits", () => {
```
Andarist, line 117: q: why this test had to be adjusted? how this PR has affected it?

Andarist:
@zthxxx sorry for the lack of a response sooner. This stuff is complicated and delicate and requires me to really focus on it to review this. I hope to get to this in the following week - but no hard promises.

If you could write down how this was behaving before your patch and how does it behave after your change I would highly appreciate it. I understand that this is somewhat covered by tests but I'm more interested in the implementation details now - what logic was behind the previous solution and how it was adjusted to cover for your use case **without** breaking existing use cases.

packages/assemble-release-plan/src/index.ts
zthxxx: yes

zthxxx:
@Andarist issue and reproduce steps here: https://github.com/changesets/changesets/issues/729
(this may answer you about how this was behaving before)

---

Explanation for details (and what I thought):

First when we execute `changeset version`, we expect some result, **version bumped** or **do nothing**, but what conditions need bump version?

Beginning with the most simple project(condition), a package with version `1.0.0` and a normal changesets config file, and no other files. So simple right?

```
./
├── .changeset/
│    └── config.json
└── packages/
     └── package-a/
          └── package.json    // version: 1.0.0
```

When we run `changeset version` at here, nothing will change, no doubt.
But if there is a changeset file with wrote `package-a: patch`, after `changeset version` package-a will get `version: 1.0.1`, easy to understand, right?

---

Let's go next, if package initial state is in prerelease version, and also no changeset file, but have a `pre.json` (with `mode: pre` or `mode: exit` are same):

```
./
├── .changeset/
│    ├── config.json
│    └── pre.json             // mode: pre / exit
└── packages/
     └── package-a/
          └── package.json    // version: 1.0.0-alpha.0
```

What we expect after run `changeset version`？ ->  `pre.json` should be removed, but version of package-a should still be `1.0.0-alpha.0`, nothing changed.

---

**Why we should have expectation like this?**

One principle of the `changesets` is **"all the version bumped come from changeset files being consumed"**.
So if no changeset files consumed, no package's version should be changed.

However, as #729 said, changesets has a bug here that `changeset version` with `pre.json` and `mode: exit` will bump all prerelease version even there isn't any changeset file.

I think this due the code here [`assemble-release-plan/src/index.ts#L121-L127`](https://github.com/changesets/changesets/blob/a0ebd5c8faf5cf16db2240ad15584912bc3bcba7/packages/assemble-release-plan/src/index.ts#L121-L127),
this code will make all package in prerelease will get a default release-plan to bump version.

So I move the logic (create default release-plan) to `function determineDependents()`, which will it created exactly **decided by changeset file of packages and their dependent-graph**.

---

Aslo I added more test for "Combination" cases which I mentioned above, let's talk about whole cases which are no longer simple:

(run `changesets version`)


|      | changeset files<br />(no / have, related / have, unrelated) | package version<br />(normal / prerelease) | pre.json<br />(no / `mode: pre` / `mode:exit`) | expect: version changed?                             |
| ---- | ------------------------------------------------------------ | ------------------------------------------- | ----------------------------------------------- | ---------------------------------------------------- |
| 1    | no                                                           | normal / prerelease                         | no / `mode: pre` / `mode: exit`                 | nothing change                                       |
| 2    | have, related                                                | normal                                      | no                                              | normal bumped                                        |
| 3    | have, related                                                | normal                                      | `mode: pre`                                     | bump to prerelease                                   |
| 4    | have, related                                                | prerelease                                  | `mode: pre`                                     | still bump prerelease number                         |
| 5    | have, related                                                | prerelease                                  | `mode: exit`                                    | bump to normal                                       |
| 6    | have, related                                                | normal                                      | `mode: exit`                                    | normal bumped                                        |
| 7    | have, unrelated                                              | normal / prerelease                         | no / `mode: pre` / `mode: exit`                 | nothing change<br />(that's like no changeset files) |
| 9    | have, unrelated, <br />but package in dependent-graph        |                                             |                                                 | behavior same as cases 2 / 3 / 4 / 5 / 6             |

Andarist:
Thank you for the detailed response - I, unfortunately, didn't yet have time to read through it and analyze this properly but your response should help me to do so when I get to it. I know that this doesn't help you all that much but I've just wanted to say that I remember this issue and I'm grateful for your effort here. I hope to get to review this later this week or in the following week.

> (PS: I known this is complicated and delicate, so we hope peoples please redesign prerelease method in changesets@v3 🙏)

Heh, ye - the prerelease mode can get hairy at times. However, I feel like it gets complicated mostly when you work with mergebacks, and if you don't want to have all of your packages participating in the prerelease mode. If you have any ideas on how this could be redesigned holistically - I'm all ears (a separate issue should be created though). There are so many expectations about it that it's quite hard to satisfy all the possible workflows.

zthxxx:
@Andarist Your reply is much appreciated, I was just wondering if you had any questions about this issue in last week?

Andarist:
I think the existing behavior is caused by the fact that such a scenario is possible:
1. enter pre mode
2. create a changeset like this `pkg-a: patch`
3. release a prerelease of that package (resulting in `pkg-a@1.0.0-alpha.1`~)
4. remove that changeset file
5. exit pre mode

It's hard to determine if the existing prerelease version of `pkg-a` was created "in response" to a changeset or if it has been created in any other way. We can't roll this version back (that would create problems for the next prerelease "cycle" like mentioned by @Rugvip) and we potentially also can't always leave it "as is". Some other packages might depend on `pkg-a` and if they exit pre mode then I'm pretty sure that it would be a bad thing if a "stable" version of a package `pkg-b` would still depend on a prerelease version of `pkg-a`.

I guess **maybe** we could leave the version as is if no "stable" package (after we calculate final versions for exiting the pre mode) would depend on this package `pkg-a`. This raises the complexity though and I'm not sure if the return of the investment is worth it.

I see how this might look weirdly from the theoretical point of view (and I like to build on fundamentally good principles, so I recognize this as a potential problem) and I see how in turn this might cause some unexpected results in the "real world". However, I'm wondering how your full end-to-end workflow looks like that this matters. It would be great if you could describe a real-life use case that would advocate for this proposed change.

I'm also interested in hearing your opinion about those two mentioned problems (1. removing a changeset file during prerelease mode 2. other stable packages depending on that package after exiting)

zthxxx:
@Andarist   
1. This is a typical "Edge Case", `changesets` always will not cause the edge case if it used correctly, but package maybe had a prerelease version before use `changesets` that like some project migrating to use `changesets`.
2. But fix the edge case will make `changesets` logic self-consistent by the principle **"all the version bumped come from changeset files being consumed"**, with not breaking anything.
3. Nothing about "roll back", that's Rugvip's wrong understanding.
4. If you're worried about that "dependentcy" with keep version, see this PR code and unit tests,
    I had do and test it, all packages versions will updated by their dependent-graph.
5. If you're worried about that "complexity", see this PR code, as you said,
    that's just **one line** judging condition move from move from the wrong place to the right place (`assembleReleasePlan` -> `determineDependents`).

Andarist:
> This is a typical "Edge Case", changesets always will not cause the edge case if it used correctly, but package maybe had a prerelease version before use changesets that like some project migrating to use changesets.


Well, I've personally removed changeset files when being in premode - it's because those changesets are later used for constructing the "final" changelog (when we exit the premode). So I would consider adding/removing/editing changesets in premode as something normal. After all, I could try to implement something but I could discover, while in premode, that the implementation is wrong (or the feature itself is wrong).

I need to be able to "roll back" on my own from such work and a typical workflow for this would include:
- reverting the changes (either with `git revert` or manually)
- removing associated changesets

This scenario, which I consider to be valid, ends up being the same as the situation that you have faced. It's impossible to differentiate them though (at least with how we are currently tracking things).

So I wonder if your issue shouldn't be addressed by:
- better docs
- added validation when entering premode (or when exiting)

zthxxx:
I prefer to agree that the "premode" feature itself is wrong 😄, but we still used it now at v2, so I try to fix even the edge case.

Back to this PR, I want the fix can be merge, so I hope to known what needs to be adjusted in this PR.

zthxxx:
@Andarist  Sorry to bother you, I have resolve the conflicts. Is there any update for you? I think I need more advice to known how to adjust this PR.

Andarist:
No worries.

Quite frankly, I'm leaning towards "closing" this issue as a design limitation or something. There are those cases when (IMHO) it's better to leave the prerelease mode for a package than to leave it accidentally in a prerelease mode. Please also check [this recent PR](https://github.com/changesets/changesets/pull/751) that fixes a bug that has caused some packages to stay in the prerelease mode incorrectly. This makes me believe that the concerns that I've raised here were not artificial - such situations happen from time to time.

I see your point of view and I partially agree with you. However, I don't see how we could satisfy both scenarios at the same time automatically and I think that the other one is somewhat more important (and how it currently works, so in a way, your fix could be seen as a breaking change).

You could also read through [this issue](https://github.com/changesets/changesets/issues/756#issuecomment-1039997529) and see what I've mentioned about how publishing with bare yarn and pnpm works. This is about something slightly different (publishing, and not versioning) but I feel like the problem space is at least partially related. In other words, this is how many tools work.

I agree though that perhaps this could be explicitly called out in the docs to avoid confusion etc.

zthxxx:
@Andarist  Thanks for your reply and review, although I'm sorry to this, I accept close the issue.

Andarist:
@zthxxx please note that i very much appreciate this PR, the discussion and you bearing with my slow responses ❤️

PR #734, opened 2022-01-12
update automate changesets documentation

eranshapira:
```🦋  warn --sinceMaster is deprecated and will be removed in a future major version, use --since=master instead```

### CODE REVIEW

File: docs/automating-changesets.md [hunk 1]
```diff
@@ -29,7 +29,7 @@ In some cases, you may want to make CI fail if not changeset is present, to ensu
 In your CI process add a step that runs:
 
 ```bash
-changeset status --sinceMaster
+changeset status --since=master
```
Andarist, line 32:
```suggestion
changeset status --since=main
```

Issue #688, opened 2021-12-02
Ability to set `commit` to different values in add and version commands.


tido64:
## Affected Packages

@changesets/cli

## Problem

We recently migrated to Changesets and are seeing some issues related to the `commit` option being used by both `add` and `version` commands. We set it to `true`, thinking that it would be convenient if didn't have to run `git add/commit` immediately after having run `changeset add`. But when the `action/changesets` picks up the changes and runs `changeset version`, it adds `[skip ci]` to the commit message and submits the PR. When we merge the PR, nothing happens because CI is skipped.

I looked through the docs and even looked into the code to see if I had missed a configuration somewhere, but it turns out it was removed back in 2.0, with the message below:

>   1. If you use `commit` and `skipCI` in `versionOptions` or `publishOptions`, set commit to `true`, all commits will include a skip ci message. if you have a use case for only using commit on one command or not including a skip ci message by default, contact us and we will talk about re-implementing these features.

https://github.com/atlassian/changesets/blob/678b16a6e49976bd3884919303bc22ed6690d151/packages/cli/CHANGELOG.md?plain=1#L738

So I guess what I'm asking is if we can re-implement this feature 😄 

## Proposed solution

Ability to override `commit` for one command, or at least split up the `commit` option so we can set different values for add and version commands.

Andarist:
q: is the problem more about `commit` option affecting both commands or about `[skip ci]` being added to the commit message? It seems that you potentially want to disable the latter but using `commit: true` for both commands would be OK for you?

I'm slightly unsure why the `[skip ci]` is there - it kinda seems dangerous to me, especially as the default for the `commit` option. At least my normal workflow is that I very often include changesets together with code-oriented commits or add them quickly on top of my changes (in a separate commit) but before pushing my branch out. In both cases I definitely want the CI to run.

tido64:
> q: is the problem more about `commit` option affecting both commands or about `[skip ci]` being added to the commit message? It seems that you potentially want to disable the latter but using `commit: true` for both commands would be OK for you?

Yes, just removing `[skip ci]` from the commit message would be enough for us. The `commit` option affecting both commands would've been fine if it wasn't for that.

> I'm slightly unsure why the [skip ci] is there

I'm also confused why it was added, tbh. I dunno if there is a use case for it.

Andarist:
[...]

PR #735, opened 2022-01-13
Update Commit config to separate add/version and add back skipCi

rohit-gohri:
PoC for a possible solution for #688. Made commit option configurable separately for `add`/`version` commands. Adds back `skipCI` option

```
{
  "commit": true
}
// above evaluates to below
{
  "commit": {
    "add": {
      "skipCI": false
    },
    "version": {
      "skipCI": true
    }
  }
}
```

Other ways:
```
// only add
{
  "commit": {
    "add": {
      "skipCI": false
    },
    "version":false
  }
}
// only version
{
  "commit": {
    "add": false,
    "version": {
      "skipCI": false
    }
  }
}
```


Andarist:
Thank you for the PR - I'll try to do a review as soon as possible but I have lots on my plate right now and can't get to everything right away (even though I would really like to 😛 )

Andarist:
I see how this solution is more robust but I'm not yet sure if we want to allow people to hook into such stuff. I think that **for now** I would prefer if we'd go with a simpler, builtin, solution.

I imagine that we could support this:
```json
{
  "commit": {
    "version": {
      "skipCI": true
    },
    "publish": {
      "skipCI": true
    }
  }
}
```

We'd have to make sure that we use the same default values as the current, implicit, ones. A config like this should only enable auto-commits for the version command:
```json
{
  "commit": {
    "version": true
  }
}
```

rohit-gohri:
@Andarist Have updated the PR to an approach similar to what you described

Andarist:
Thank you for the PR - I will try to review it after the weekend

packages/types/src/index.ts
```diff
@@ -58,9 +58,18 @@ export type PackageJSON = {
...
+
+export type Commit = {
+  add: false | CommitOptions;
+  version: false | CommitOptions;
+};
+
 export type Config = {
   changelog: false | readonly [string, any];
-  commit: boolean;
+  commit: Commit;
```
Andarist, line 72: This is a breaking change for all packages that receive config from the outside, such as `@changesets/apply-release-plan`. Previously they could assume that this is a boolean. I think it's not worth a breaking change and perhaps we don't have to normalize this (at least not for the time being) to this object shape.

packages/config/src/index.ts
```diff
@@ -359,3 +397,5 @@ export let defaultConfig = parse(defaultWrittenConfig, {
   tool: "root",
   packages: [fakePackage]
 });
+
+export { getAddLine, getVersionLine };
```
Andarist, line 401: I think that those don't belong in this package, at least for now - moving those should be reverted, they should stay declared closer to where they are actually called.

packages/config/src/index.ts
```diff
@@ -113,14 +118,45 @@ export let parse = (json: WrittenConfig, packages: Packages): Config => {
...
+  ) {
     messages.push(
       `The \`commit\` option is set as ${JSON.stringify(
         json.commit,
         null,
         2
-      )} when the only valid values are undefined or a boolean`
+      )} when the only valid values are undefined or a boolean or an object`
     );
+  } else if (typeof json.commit === "object") {
```
Andarist, line 133: we could also add a check if `skipCI` is of the correct type within this block

Andarist:
Actually… while taking a shower i was thinking about this more 😅and i think i’ve figured out an alternative, more robust, design. Gonna post it later today so maybe let’s hold working on this PR until i write it down

Andarist:
My proposal is to accept a string for the commit option that would point to a module responsible for generating those commit messages. We already do a similar thing with "changelog generators":
https://github.com/changesets/changesets/blob/c2b9d44886740f850704b06a4d15beca2ca3a1d8/packages/apply-release-plan/src/index.ts#L201-L223

I feel like this would be the most robust option that wouldn't require us to add a somewhat weird shape of the `commit` config option. This way the default module for this (that probably should be just an additional entry in the `@changesets/cli`) could not add `[skip ci]` at all but it would be super easy to create a module that would call the default one and just adjust the return value slightly. I do a similar thing with the default changelog generator here:
https://github.com/contentlayerdev/contentlayer/blob/7c06f352e73e265ad57e0ef12c88b892bc85837d/scripts/changeset-changelog-generator.js

Note that when using the default "commit generator" we'd have to add `[skip ci]` to the message associated with one of the commands to maintain backward compatibility. This could happen in the code and outside of this default module though - unless this would be too cumbersome to pull off in the codebase, then it's just fine to include that in the default module.

Open questions:
- what arguments function(s) exposed by this module should receive?
- should the module expose a single function or two? it's important to know the content of which command we are generating but this information could be given in an argument, so using a single function is an option. I didn't think through which one should be preferred here though.

rohit-gohri:
> My proposal is to accept a string for the commit option that would point to a module responsible for generating those commit messages.
Ha, that was the orignal PR's approach. It's still in this old branch: https://github.com/changesets/changesets/compare/main...rohit-gohri:old
I should have just opened a new PR 🤦  but I force-pushed instead to keep the discussion in single place.

Though I didn't move it into it's own module, I'll open a new PR with the changes.

> what arguments function(s) exposed by this module should receive?
the changesets for the `add` and the release info for the `publish`, and any config?

> should the module expose a single function or two?
I think 2 makes more sense as the input arguments would be completely different.

Andarist:
> Ha, that was the orignal PR's approach. It's still in this old branch: https://github.com/changesets/changesets/compare/main...rohit-gohri:old

Heh, ye - I've reconsidered my previous opinion after getting a better look at what the alternative actually looks like 😅 

> Though I didn't move it into it's own module

It doesn't have to be a new module - an additional exntrypoint in the `@changesets/cli` should be sufficient. Take a look at how the one for the changelog is being defined here:
https://github.com/changesets/changesets/blob/0d1debd8f5d364073d627a017316a25554dec4d5/packages/cli/package.json#L26

> I'll open a new PR with the changes.

Feel free to force-push to this one if you prefer that.

> the changesets for the add and the release info for the publish, and any config?

Yep, sounds alright.

> I think 2 makes more sense as the input arguments would be completely different.

Agreed. Such functions could actually be typed somewhat OK with the upcoming [dependent parameters](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6-rc/#dependent-parameters-cfa) but ye, it's probably more straightforward to use separate functions.

rohit-gohri:
Opened https://github.com/changesets/changesets/pull/768 based on the discussion here, didn't force push so that we can compare the approaches. Closing this.

PR #740, opened 2022-01-23
fix changeset parser fail when summary is empty

akphi:
Fix https://github.com/changesets/action/issues/138

@Andarist Could you do me a favor to upgrade `changeset-action` to use the latest parser when we merge this PR? Thank you so much!

### DIFF
--- a/.changeset/light-cats-flow.md
+++ b/.changeset/light-cats-flow.md
@@ -1,4 +1,5 @@
 ---
+"@changesets/cli": patch
 "@changesets/parse": patch
 ---
 
--- a/.changeset/light-cats-flow.md
+++ b/.changeset/light-cats-flow.md
@@ -1,0 +1,5 @@
+---
+"@changesets/parse": patch
+---
+
+Fixed an issue with failing to parse changesets containing a completely empty summary.
--- a/packages/parse/src/index.test.ts
+++ b/packages/parse/src/index.test.ts
@@ -136,6 +136,28 @@ describe("parsing a changeset", () => {
       summary: expectedSummary
     });
   });
+  it("should be fine if the summary body is empty", () => {
+    const changesetMd = outdent`---
+    "cool-package": minor
+    ---`;
+
+    const changeset = parse(changesetMd);
+    expect(changeset).toEqual({
+      releases: [{ name: "cool-package", type: "minor" }],
+      summary: ""
+    });
+  });
+  it("should be fine if the summary body contains only an empty space", () => {
+    const changesetMd = outdent`---
+    "cool-package": minor
+    --- `;
+
+    const changeset = parse(changesetMd);
+    expect(changeset).toEqual({
+      releases: [{ name: "cool-package", type: "minor" }],
+      summary: ""
+    });
+  });
   it("should be fine if the changeset is empty", () => {
     const changesetMd = outdent`---
     ---
@@ -149,11 +171,7 @@ describe("parsing a changeset", () => {
     });
   });
   it("should be fine if the changeset only contains frontmatter", () => {
-    const changesetMd = outdent`---
-    ---
-    `;
-
-    const changeset = parse(changesetMd);
+    const changeset = parse(`---\n---`);
     expect(changeset).toEqual({
       releases: [],
       summary: ""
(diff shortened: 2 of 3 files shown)


packages/parse/src/index.test.ts
```diff
@@ -148,6 +148,17 @@ describe("parsing a changeset", () => {
       summary: ""
     });
   });
+  it("should be fine if the changeset only contains frontmatter", () => {
```
Andarist, line 151: This is a somewhat different test case that was originally reported. The reported case was about an empty summary but it had some releases declared in the frontmatter. I believe that it's worth adding tests for both of those cases.

packages/parse/src/index.test.ts
```diff
@@ -148,6 +148,17 @@ describe("parsing a changeset", () => {
...
+    const changesetMd = outdent`---
+    ---
+    `;
+
+    const changeset = parse(changesetMd);
```
Andarist, line 156: `outdent` removes a trailing newline by default and all but it feels a little bit implicit here. I would prefer if we could do this explicitly here:
```suggestion
    const changeset = parse('---\n---');
```

packages/parse/src/index.ts
```diff
@@ -1,7 +1,7 @@
 import yaml from "js-yaml";
 import { Release, VersionType } from "@changesets/types";
 
-const mdRegex = /\s*---([^]*?)\n\s*---\s*\n([^]*)/;
+const mdRegex = /\s*---([^]*?)\n\s*---(?:\s*\n([^]*))?$/;
```
Andarist, line 4: This version of the regexp doesn't allow for a valid~ input like `"---\n--- "`

We can fix this with:
```suggestion
const mdRegex = /\s*---([^]*?)\n\s*---(\s*(?:\n|$)[^]*)/;
```

By using this suggestion we can also revert the change in the `summary` declaration as this regexp doesn't use an optional capturing group so it will always end up using an empty string for a `roughSummary` match.

akphi: Very nice catch! This is updated now

packages/parse/src/index.test.ts
akphi: Done!

packages/parse/src/index.test.ts
akphi: I totally overlooked that you have `outdent`. Updated!

packages/parse/src/index.test.ts
```diff
@@ -148,6 +170,13 @@ describe("parsing a changeset", () => {
       summary: ""
     });
   });
+  it("should be fine if the changeset only contains frontmatter", () => {
```
Andarist, line 173:
```suggestion
  it("should be fine if the changeset is empty and without any trailing whitespace", () => {
```

packages/parse/src/index.test.ts
```diff
@@ -136,6 +136,28 @@ describe("parsing a changeset", () => {
...
+    "cool-package": minor
+    ---`;
+
+    const changeset = parse(changesetMd);
+    expect(changeset).toEqual({
+      releases: [{ name: "cool-package", type: "minor" }],
+      summary: ""
+    });
+  });
+  it("should be fine if the summary body contains only an empty space", () => {
```
Andarist, line 150:
```suggestion
  it("should be fine if there is no summary body and the frontmatter has some trailing whitespace", () => {
```

packages/parse/src/index.test.ts
```diff
@@ -136,6 +136,28 @@ describe("parsing a changeset", () => {
       summary: expectedSummary
     });
   });
+  it("should be fine if the summary body is empty", () => {
```
Andarist, line 139:
```suggestion
  it("should be fine if the summary body is completely empty and there is no trailing whitespace", () => {
```

Andarist:
@akphi could you also add a changeset for this change?

> @Andarist Could you do me a favor to upgrade changeset-action to use the latest parser when we merge this PR? Thank you so much!

Sure thing.

akphi:
> @akphi could you also add a changeset for this change?

Done!

.changeset/light-cats-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/parse": patch
```
Andarist, line 2:
```suggestion
"@changesets/cli": patch
"@changesets/parse": patch
```

PR #747, opened 2022-02-09
Fix peer dep changelogs

### LINKED ISSUE #746: Peer dependency changelog lines are added under patch instead of major
## Affected Packages
`pkg:apply-release-plan`

## Problem

Dependency release lines are unconditionally added under the `patch` heading instead of peer dependency lines being added under `major`.

Blasz:
Fixes #746.

This also adds a new `dependencyType` argument to `getDependencyReleaseLine` to distinguish peer vs regular dependencies. We want to distinguish them because we currently output a shortened `Updated dependencies` message to prevent changelog noise from the sheer number of dependencies that our atlassian-frontend repo has. Peer dependency upgrades are more important as they trigger a major release and so we want to output the specific dependencies and their versions for peer upgrades and only peer upgrades.


packages/apply-release-plan/src/get-changelog-entry.ts
```diff
@@ -56,45 +70,56 @@ export default async function getChangelogEntry(
...
   );
-
   changelogLines.patch.push(
     changelogFuncs.getDependencyReleaseLine(
-      relevantChangesets,
+      getRelevantChangesets(dependentReleases, changesets),
       dependentReleases,
-      changelogOpts
+      changelogOpts,
+      "dependencies"
```
Andarist, line 122: If I read the code correctly this could actually be also caused by an optional dependency.

q: do you plan to utilize this new argument in your code?

packages/apply-release-plan/src/index.test.ts
```diff
@@ -1900,6 +1900,82 @@ describe("apply release plan", () => {
...
+            null
+          ],
+          commit: false,
+          linked: [],
+          access: "restricted",
+          baseBranch: "main",
+          updateInternalDependencies: "patch",
+          ignore: [],
+          ___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH: {
+            onlyUpdatePeerDependentsWhenOutOfRange: false,
```
Andarist, line 1945: Out of curiosity - so I assume that you are still using this default setting, doesn't it create major releases for you way too often?

packages/apply-release-plan/src/get-changelog-entry.ts
Blasz: Releases only contain packages specified in the `dependencies` and `peerDependencies` fields of package.json so I don't think optional dependencies would be included here unless they were also specified in `dependencies` which isn't recommended behaviour.

We're planning to use it to distinguish peer dep upgrade lines from non-peer dep ones so that we output a more detailed message for major peer upgrades.

This distinction could also be made using a `bumpType` or `isPeer` argument instead.

packages/apply-release-plan/src/index.test.ts
Blasz: Yeah we are. It does for packages that have peer dependencies that upgrade frequently yes. The alternative though is that the minimum versions of peer dependency version ranges become stale which can break things, especially with aggressive de-duplication.

It probably is worth re-exploring though.

n1ru4l:
I have another question that is potentially related to this PR.

I also noticed that changesets planned MAJOR releases, in which only PATCH updates were listed. After some research and finding this issue, I now understand that peerDependency updates are currently treated as MAJOR updates, but incorrectly listed as PATCH dependencies within the generated CHANGELOG.md file.

My first thought is that a `peerDependency` update should not automatically result in a MAJOR bump. In my perception the package that lists a package as a peerDependency should be bumped according to the version bump of the peerDependency 🤔 .

```
peer dependency major -> major bump of dependent
peer dependency minor -> minor bump of dependent
peer dependency patch -> patch bump of dependent
```

What is the current reason for treating each update of a peerDependency as a MAJOR change?

GaboFDC:
> Two packages `core` and `cli`. When `core` is bumped from 2.0.0 to 2.1.0 and `cli` depends on it in form of a version range `^2.0.0`, it may or may not require a bump in version range.
> 
> a) changesets should try to be smart and safely bump the version range (major -> major, minor -> minor, patch -> patch) b) or leave it entirely to the developer to decide
> 
> My 32 cents.

This is exactly what #827 proposed as well

dotansimha:
Regarding this PR, @Blasz @n1ru4l @Andarist is there anything we can do to resolve the issues and get this merged? I can try to help :)

beeequeue:
closing as stale, and because we are updating the bump type of peer dependents in #2090

PR #751, opened 2022-02-11
Fix for dependent packages not being bumped when exiting prerelease

Rugvip:
We ran into this issue the other day, where if the release plan logic explicitly determines that a packages doesn't needs a bump, it would be ignored when exiting pre-release too.

### DIFF
--- a/.changeset/strong-crews-film.md
+++ b/.changeset/strong-crews-film.md
@@ -1,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": patch
+---
+
+Fixed an issue where dependent packages would not get bumped properly when exiting prerelease mode.
--- a/.changeset/strong-crews-film.md
+++ b/.changeset/strong-crews-film.md
@@ -1,5 +1,6 @@
 ---
 "@changesets/assemble-release-plan": patch
+"@changesets/cli": patch
 ---
 
 Fixed an issue where dependent packages would sometimes not get bumped properly when exiting prerelease mode.
--- a/.changeset/strong-crews-film.md
+++ b/.changeset/strong-crews-film.md
@@ -2,4 +2,4 @@
 "@changesets/assemble-release-plan": patch
 ---
 
-Fixed an issue where dependent packages would not get bumped properly when exiting prerelease mode.
+Fixed an issue where dependent packages would sometimes not get bumped properly when exiting prerelease mode.
--- a/packages/assemble-release-plan/src/index.test.ts
+++ b/packages/assemble-release-plan/src/index.test.ts
@@ -869,7 +869,7 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
       expect(releases[0].newVersion).toEqual("1.0.1");
     });
 
-    it("should always bump dependents when exiting pre-release mode", () => {
+    it("should bump dev dependents when exiting pre-release mode", () => {
       setup.updatePackage("pkg-a", "1.0.1-next.0");
       setup.updatePackage("pkg-b", "1.0.1-next.0");
       setup.updateDevDependency("pkg-b", "pkg-a", "1.0.1-next.0");
@@ -898,7 +898,7 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
       expect(releases[1].newVersion).toEqual("1.0.1");
     });
 
-    it("should not bump dependents when exiting pre-release mode if ignored", () => {
+    it("should not bump ignored dev dependents when exiting pre-release mode", () => {
       setup.updatePackage("pkg-a", "1.0.1-next.0");
       setup.updatePackage("pkg-b", "1.0.1-next.0");
       setup.updateDevDependency("pkg-b", "pkg-a", "1.0.1-next.0");
--- a/packages/assemble-release-plan/src/index.test.ts
+++ b/packages/assemble-release-plan/src/index.test.ts
@@ -869,6 +869,65 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
       expect(releases[0].newVersion).toEqual("1.0.1");
     });
 
+    it("should always bump dependents when exiting pre-release mode", () => {
+      setup.updatePackage("pkg-a", "1.0.1-next.0");
+      setup.updatePackage("pkg-b", "1.0.1-next.0");
+      setup.updateDevDependency("pkg-b", "pkg-a", "1.0.1-next.0");
+
+      const { releases } = assembleReleasePlan(
+        setup.changesets,
+        setup.packages,
+        {
+          ...defaultConfig
+        },
+        {
+          changesets: ["strange-words-combine"],
+          tag: "next",
+          initialVersions: {
+            "pkg-a": "1.0.0",
+            "pkg-b": "1.0.0"
+          },
+          mode: "exit"
+        }
+      );
+
+      expect(releases.length).toEqual(2);
+      expect(releases[0].name).toEqual("pkg-a");
+      expect(releases[0].newVersion).toEqual("1.0.1");
+      expect(releases[1].name).toEqual("pkg-b");
+      expect(releases[1].newVersion).toEqual("1.0.1");
+    });
+
+    it("should not bump dependents when exiting pre-release mode if ignored", () => {
+      setup.updatePackage("pkg-a", "1.0.1-next.0");
+      setup.updatePackage("pkg-b", "1.0.1-next.0");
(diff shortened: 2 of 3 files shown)


packages/assemble-release-plan/src/index.ts
```diff
@@ -124,7 +124,8 @@ function assembleReleasePlan(
       // we want to give it a patch release.
       // Detailed explanation at https://github.com/changesets/changesets/pull/382#discussion_r434434182
       if (preInfo.preVersions.get(pkg.packageJson.name) !== 0) {
-        if (!releases.has(pkg.packageJson.name)) {
+        const existingRelease = releases.get(pkg.packageJson.name);
+        if (!existingRelease || existingRelease.type === "none") {
```
Andarist, line 128: this might create a second release for the same package and I would consider that to be a problem because each package should have at most one release. It just seems like a reasonable invariant to maintain.

I would have to think more to figure out if there is some e2e test to be written for this situation though - I think this might create some minor unwanted side-effect, just ain't sure at the spot what that is.

Rugvip: We're overriding the existing release though right? And that release is guaranteed to have no effect since it's `none`?

I was considering setting the release type to `undefined` for dev dependencies instead, but that feels a bit more risky. Overall I assume the release logic is basically a `max(allReasonsToRelease)`  for each package right? Which this would seem to follow.

Is there some situation where you'd actually want to override a release `type` downwards? As in some piece of logic determining we're doing a `minor` release, but that being overridden by something else that says it should just be a `patch`?

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -685,6 +685,35 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
       expect(releases[0].newVersion).toEqual("1.0.1");
     });
 
+    it("should always bump dependents when exiting pre-release mode", () => {
+      setup.updatePackage("pkg-a", "1.0.1-next.0");
+      setup.updatePackage("pkg-b", "1.0.1-next.0");
+      setup.updateDevDependency("pkg-b", "pkg-a", "1.0.1-next.0");
```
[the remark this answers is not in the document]
Rugvip: I should add that this is not the only way to produce this issue, we've also seen it with a combination of transitive devDependencies and version ranges, although I wasn't able to reproduce that case.

packages/assemble-release-plan/src/index.ts
Andarist: > We're overriding the existing release though right? And that release is guaranteed to have no effect since it's none?

You are right, I've reviewed this in a hurry and I've assumed that we are pushing here to an array or something but this is actually kept in a map.

> Is there some situation where you'd actually want to override a release type downwards?

I don't think so but can't say that I'm completely sure about it. I think that in general assemble-release-plan is used to gather the releases and compute the highest bump type for each. The new version should always be derivable from this information. In apply-release-plan we, well, apply this information and adjust dependency ranges in dependent packages for all releases 

> As in some piece of logic determining we're doing a minor release, but that being overridden by something else that says it should just be a patch?

This should never happen - we should always choose the highest release type based on all applied constraints. "Downgrading" a release would have undetermined and bad consequences.

---

So the actual concern that I have after reviewing this is that for a none release we'd lose `release.chagesets` with the proposed logic. So I would propose to branch this logic:
- when there are no releases then create a patch release for it here
- when there is already a none release then just override it with a patch type

Rugvip: Pushed an update. I felt this conveyed the intent most clearly, but I can switch it over to for example one of these too:

```ts
} else if (existingRelease.type === "none") {
  releases.set(pkg.packageJson.name, {
    ...existingRelease,
    type: "patch"
  });
}
```

```ts
} else if (existingRelease.type === "none") {
  existingRelease.type = "patch";
}
```

packages/assemble-release-plan/src/index.test.ts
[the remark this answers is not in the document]
Andarist: Just a follow-up question that helps me understand how you are using Changesets etc. How does this package get itself into a prerelease stage without having any explicit/transitive release for itself (that would be different that none)? Did you remove some changeset files **after** publishing this package in a prerelease mode?

packages/assemble-release-plan/src/index.ts
Andarist: I think that the second one is the most straightforward one - it's closer to "preserve everything while changing the release type". The first one is OK too but no strong reason to have this updated in an immutable way.

Andarist: There is also an issue here that... we might end up bumping an ignored package because their implementation also "piggybacks" on none releases. It feels wrong, right? 🤔 I need to think more about how to best handle both scenarios.

packages/assemble-release-plan/src/index.test.ts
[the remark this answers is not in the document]
Andarist: 👋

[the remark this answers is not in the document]
Rugvip: Tbh I don't know. It happened to a package that has been around for a while. I think what might've happened was that we changed the dependency type or range in the middle of a prerelease. For example switching something from a regular dep to a dev dep I assume would be able to trigger this behavior.

packages/assemble-release-plan/src/index.ts
Andarist: Have you tried running this patch against the whole test suite? IIRC this would prevent dev dependency ranges to be updated within `@changesets/apply-release-plan`.

Andarist: Hm, our postinstall hook should allow for all packages to be available in node_modules and for the test suite to be fully runnable. Either way - I've confirmed that those tests would fail with the proposed patch:
```
assemble-release-plan › should assemble release plan without dependent through dev dependency
same package in different dependency types › should update different range types correctly
pre › should not bump packages through devDependencies
pre › linked › should not bump a linked package if its linked devDep gets released
```

And by looking at snapshot diffs those failures are related exactly to what I've mentioned - dev dep ranges not being updated correctly.

I very much agree that the current implementation with none changesets is a little bit quirky, so I'm not at all surprised that it raises some eyebrows. However, it somewhat neatly allows us to just handle all in a similar fashion - without introducing more concepts, special cases etc. Maybe there is a better design to be found here, but kinda whatever works, right? 😉

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -869,6 +869,65 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
       expect(releases[0].newVersion).toEqual("1.0.1");
     });
 
+    it("should always bump dependents when exiting pre-release mode", () => {
```
Andarist, line 872:
```suggestion
    it("should bump dev dependents when exiting pre-release mode", () => {
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -869,6 +869,65 @@ Mixed changesets that contain both ignored and not ignored packages are not allo
...
+      );
+
+      expect(releases.length).toEqual(2);
+      expect(releases[0].name).toEqual("pkg-a");
+      expect(releases[0].newVersion).toEqual("1.0.1");
+      expect(releases[1].name).toEqual("pkg-b");
+      expect(releases[1].newVersion).toEqual("1.0.1");
+    });
+
+    it("should not bump dependents when exiting pre-release mode if ignored", () => {
```
Andarist, line 901:
```suggestion
    it("should not bump ignored dev dependents when exiting pre-release mode", () => {
```

.changeset/strong-crews-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": patch
+---
+
+Fixed an issue where dependent packages would not get bumped properly when exiting prerelease mode.
```
Andarist, line 5:
```suggestion
Fixed an issue where dependent packages would sometimes not get bumped properly when exiting prerelease mode.
```

.changeset/strong-crews-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": patch
```
Andarist, line 2:
```suggestion
"@changesets/assemble-release-plan": patch
"@changesets/cli": patch
```

Issue #737, opened 2022-01-16
Error messages are not human readable


BlackGlory:
@changesets/cli 2.19.0
```
🦋  info Publishing "@middleware/stream" at "0.11.2"
🦋  error an error occurred while publishing @middleware/stream: 1 command failed 
🦋  error sh -c yarn run-s clean build
🦋  error <Buffer 45 52 52 4f 52 3a 20 22 62 75 69 6c 64 3a 63 6f 6d 70 69 6c 65 22 20 65 78 69 74 65 64 20 77 69 74 68 20 32 2e 0a 45 52 52 4f 52 3a 20 22 62 75 69 6c ... 488 more bytes>
🦋  error packages failed to publish:
🦋  @middleware/stream@0.11.2
error Command failed with exit code 1
```

Andarist:
@BlackGlory could you create a repro case for this problem?

BlackGlory:
I'm afraid I can't create it, because this error only appears on publish, and disappears after multiple calls.

Andarist:
I've managed to find the place where that Buffer could indeed be logged "as is" and I've prepared a fix for this: https://github.com/changesets/changesets/pull/760

Hopefully, with this patch released soon-ish you could get more info about your particular problem. Note that recently we go a similar report here: https://github.com/changesets/changesets/issues/759. Perhaps your problem is the same?

BlackGlory:
I think it is the same.

Andarist:
So you are using snap too?

BlackGlory:
My Node.js v16.13.1 is installed via nvm, so this may not be relevant to snap.

JakeGinnivan:
FWIW I am using GitHub actions with

```
            - name: Setup Node.js 16.x
              uses: actions/setup-node@v2
              with:
                  node-version: 16.x
                  registry-url: 'https://npm.pkg.github.com'
```

After this PR I am getting

```
🦋  error 
🦋  error 
🦋  error packages failed to publish:
```

The packages are not published yet so this is the initial publish. Will update when get to the bottom of why it's failing (assume it's a token/scoping/npmrc issue or similar)

JakeGinnivan:
Ok, mine was interesting.

PNPM wasn't returning an error when `--json` was passed to the publish command. The issue was I was setting `publishConfig.directory` because NX builds to a dist folder, but it's not removing that from the copied package.json so it then tried to follow the publishConfig.directory again which resulted in a path which didn't exist

[...]

PR #760, opened 2022-02-23
Fixed an issue with logging non-stringified Buffers in case of publish errors

Andarist:
fixes https://github.com/changesets/changesets/issues/737

The problem is that all functions in `@changesets/logger` accept `any`s and I think that should be fixed cause it can lead to mistakes like this.

This could be fixed with this patch:
<details>
<summary>patch</summary>

```diff
diff --git a/packages/logger/src/index.ts b/packages/logger/src/index.ts
index d0d3f51..8992673 100644
--- a/packages/logger/src/index.ts
+++ b/packages/logger/src/index.ts
@@ -15,22 +15,22 @@ function format(args: Array<any>, customPrefix?: string) {
   );
 }
 
-export function error(...args: Array<any>) {
+export function error(...args: Array<string>) {
   console.error(format(args, chalk.red("error")));
 }
 
-export function info(...args: Array<any>) {
+export function info(...args: Array<string>) {
   console.info(format(args, chalk.cyan("info")));
 }
 
-export function log(...args: Array<any>) {
+export function log(...args: Array<string>) {
   console.log(format(args));
 }
 
-export function success(...args: Array<any>) {
+export function success(...args: Array<string>) {
   console.log(format(args, chalk.green("success")));
 }
 
-export function warn(...args: Array<any>) {
+export function warn(...args: Array<string>) {
   console.warn(format(args, chalk.yellow("warn")));
 }

```
</details>

However, this would require further changes in the codebase and I didn't want to make them right now. Mainly all `.catch`ed errors would become harder to pass to `logger.error`:
https://github.com/changesets/changesets/blob/75c30fc58b236392832fcffc2d64a16ba5ebed02/packages/cli/src/utils/cli-utilities.ts#L84-L86

On the other hand, this is somewhat desirable, because at least we'd have to make the explicit conversion at those call sites. Thoughts?

### DIFF
--- a/.changeset/chilly-swans-approve.md
+++ b/.changeset/chilly-swans-approve.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Fixed an issue with logging non-stringified Buffers in case of publish errors.
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -214,7 +214,7 @@ async function internalPublish(
       );
     }
 
-    error(stderr);
+    error(stderr.toString());
     return { published: false };
   }
   return { published: true };


mitchellhamilton (approved):

mitchellhamilton:
> On the other hand, this is somewhat desirable, because at least we'd have to make the explicit conversion at those call sites. Thoughts?

Yeah, I think being explicit about it would be nice

Issue #756, opened 2022-02-15
Unexpected first time publish without adding a changeset


melanieseltzer:
## Affected Packages

I'm unsure.

## Problem

I'm not sure if this is a problem or if this is me misunderstanding something.

I [have this PR](https://github.com/melanieseltzer/toolkit/pull/6) that creates a brand new package (`@mels/eslint-config-base`). It hadn't been published to npm yet before creating the PR. **I hadn't even added a changeset yet** to the PR because I'm still working on it, and the version is `0.0.0` in the package.json.

I have a [GitHub Actions workflow](https://github.com/melanieseltzer/toolkit/blob/main/.github/workflows/prerelease.yml) in the repo that creates snapshots on pull requests, so that I can test PRs by installing at each commit if needed.

Despite not even adding a changeset yet to my PR, [it's gone and published `0.0.0` to npm](https://www.npmjs.com/package/@mels/eslint-config-base?activeTab=versions). Is that intended behavior? 🤔 I thought that it was supposed to wait until there's a changeset before publishing something? Or if it needed to publish something first (just to get something up in the registry since it didn't exist previously), shouldn't it have published `0.0.0-prerelease-96f1f5f` in this case?

[Full link to logs](https://github.com/melanieseltzer/toolkit/runs/5194225946?check_suite_focus=true).

```
6s
Run COMMIT_SHA=653789aa0bef7e6da3408f2cc26[1](https://github.com/melanieseltzer/toolkit/runs/5194225946?check_suite_focus=true#step:5:1)0bc808bf6a0b
yarn run v1.22.[17](https://github.com/melanieseltzer/toolkit/runs/5194225946?check_suite_focus=true#step:5:17)
$ changeset version --snapshot prerelease-653789a
🦋  warn No unreleased changesets found, exiting.
Done in 0.51s.
yarn run v1.[22](https://github.com/melanieseltzer/toolkit/runs/5194225946?check_suite_focus=true#step:5:22).17
$ changeset publish --tag prerelease
🦋  warn ===============================IMPORTANT!===============================
🦋  warn Packages will be released under the prerelease tag
🦋  warn ----------------------------------------------------------------------
🦋  info npm info @mels/eslint-config-base
🦋  info npm info @mels/eslint-patch
🦋  warn Received 404 for npm info "@mels/eslint-config-base"
🦋  info @mels/eslint-config-base is being published because our local version (0.0.0) has not been published on npm
[...]

PR #762, opened 2022-02-23
Add some clarification around accidental publishing

melanieseltzer:
Document the possibility of publishing unwanted versions to npm when doing automatic snapshots on PRs.

Ref: https://github.com/changesets/changesets/issues/756

### DIFF
--- a/docs/command-line-options.md
+++ b/docs/command-line-options.md
@@ -103,13 +103,23 @@ changeset publish [--otp={token}]
 
 This publishes changes to npm, and creates git tags. This works by going into each package, checking if the version it has in its `package.json` is published on npm, and if it is not, running the `npm publish`.
 
-**Important:** There is an unintended consequence when doing automatic snapshots on pull requests, please see [snapshot releases](./snapshot-releases.md#automatic-snapshots-on-prs) for more info.
-
 Because this command assumes that last commit is the release commit you should not commit any changes between calling version and publish. These commands are separate to enable you to check if the releases changes are acurate.
 
 `--otp={token}` - allows you to provide an npm one-time password if you have auth and writes enabled on npm. The CLI also prompts for the OTP if it's not provided with the --otp option.
 
 `--tag TAGNAME` - for packages that are published, the chosen tag will be used instead of `latest`, allowing you to publish changes intended for testing and validation, not main consumption. This will most likely be used with [snapshot releases](./snapshot-releases.md).
+
+### Unintended first time publish
+
+You might have a GitHub Action set up that does [automatic snapshots on pull requests](./snapshot-releases.md#automatic-snapshots-on-prs). There is an unintended consequence to this with brand new packages that have never been published to npm before.
+
+`changeset version` and `changeset publish` are decoupled in that no information is shared between them.
+
+`changeset publish` actually doesn't care if a changeset is even generated, it will still publish the version that doesn't exist on npm even if `changeset version` exits because no unreleased changesets were found.
+
+So, if you have automatic snapshots on PRs set up in the repo, and then create a PR adding a brand new package with a version of `0.0.0` in package.json (common scenario when just starting out, a throwaway version number when still in development), the GitHub Action will run on the PR and `changeset publish` will see that `0.0.0` has not been published on npm for this package, and **it will publish that version.**
+
+You can get around this by temporarily marking `private: true` in package.json while setting up the new package, and removing it when you're ready to generate a changeset.
 
 ### Git Tags
 
--- a/docs/command-line-options.md
+++ b/docs/command-line-options.md
@@ -101,25 +101,25 @@ Snapshot is used for a special kind of publishing for testing - it creates tempo
 changeset publish [--otp={token}]
 ```
 
-This publishes changes to npm, and creates git tags. This works by going into each package, checking if the version it has in its `package.json` is published on npm, and if it is not, running the `npm publish`.
+This publishes changes to npm, and creates git tags. The command works by:
+
+- querying current versions in the repository
+- comparing the version it has in its `package.json` with published versions
+- trying to publish what has not been published yet by running `npm publish`
+
+Importantly, `changeset publish` doesn't care about changeset files at all. If invoked, it will still run through the above actions whether a changeset file has been generated or not.
+
+This means an accidental publish can happen if you have an unpublished package (or even an unpublished version of a package), start using Changesets (but don't generate a changeset yet), and use our [changesets/action](https://github.com/changesets/action). If you don't immediately add a changeset file then the first run of the action just calls `changeset publish` and it publishes what has not been published yet. Unfortunately, this means something marked with a throwaway version of say `0.0.0` in package.json would still get published.
+
+You are more likely to run into this if you have a GitHub Action set up that does [automatic snapshots on pull requests](./snapshot-releases.md#automatic-snapshots-on-prs), because running `changeset version && changeset publish` sequentially in CI means that `changeset publish` will run regardless if `changeset version` did anything or exited.
+
+You can get around this behavior by temporarily marking `private: true` in package.json while in development, and removing it when you're ready to generate a changeset and publish the package.
 
 Because this command assumes that last commit is the release commit you should not commit any changes between calling version and publish. These commands are separate to enable you to check if the releases changes are acurate.
 
 `--otp={token}` - allows you to provide an npm one-time password if you have auth and writes enabled on npm. The CLI also prompts for the OTP if it's not provided with the --otp option.
 
 `--tag TAGNAME` - for packages that are published, the chosen tag will be used instead of `latest`, allowing you to publish changes intended for testing and validation, not main consumption. This will most likely be used with [snapshot releases](./snapshot-releases.md).
-
-### Unintended first time publish
-
-You might have a GitHub Action set up that does [automatic snapshots on pull requests](./snapshot-releases.md#automatic-snapshots-on-prs). There is an unintended consequence to this with brand new packages that have never been published to npm before.
-
-`changeset version` and `changeset publish` are decoupled in that no information is shared between them.
-
-`changeset publish` actually doesn't care if a changeset is even generated, it will still publish the version that doesn't exist on npm even if `changeset version` exits because no unreleased changesets were found.
-
-So, if you have automatic snapshots on PRs set up in the repo, and then create a PR adding a brand new package with a version of `0.0.0` in package.json (common scenario when just starting out, a throwaway version number when still in development), the GitHub Action will run on the PR and `changeset publish` will see that `0.0.0` has not been published on npm for this package, and **it will publish that version.**
-
-You can get around this by temporarily marking `private: true` in package.json while setting up the new package, and removing it when you're ready to generate a changeset.
 
 ### Git Tags
 
--- a/docs/command-line-options.md
+++ b/docs/command-line-options.md
@@ -103,6 +103,8 @@ changeset publish [--otp={token}]
 
 This publishes changes to npm, and creates git tags. This works by going into each package, checking if the version it has in its `package.json` is published on npm, and if it is not, running the `npm publish`.
 
+**Important:** There is an unintended consequence when doing automatic snapshots on pull requests, please see [snapshot releases](./snapshot-releases.md#automatic-snapshots-on-prs) for more info.
+
 Because this command assumes that last commit is the release commit you should not commit any changes between calling version and publish. These commands are separate to enable you to check if the releases changes are acurate.
 
 `--otp={token}` - allows you to provide an npm one-time password if you have auth and writes enabled on npm. The CLI also prompts for the OTP if it's not provided with the --otp option.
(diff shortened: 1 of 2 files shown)


docs/snapshot-releases.md
```diff
@@ -51,3 +51,58 @@ yarn add your-package-name@bulbasaur
...
+
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v2
+        with:
+          node-version: "14"
+
+      - name: Install Dependencies
+        run: yarn
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
Andarist, line 81: This is only needed when you are managing private packages, right? Perhaps this could be left out for brevity of this example

docs/snapshot-releases.md
```diff
@@ -51,3 +51,58 @@ yarn add your-package-name@bulbasaur
...
+
+      - name: Publish Pre-release
+        run: |
+          COMMIT_SHA=${{ github.event.pull_request.head.sha }}
+          COMMIT_SHA_SHORT=$(git rev-parse --short "$COMMIT_SHA")
+          yarn changeset version --snapshot prerelease-$COMMIT_SHA_SHORT
```
Andarist, line 87: Any particular reason why we are simply not doing this here?
```suggestion
          COMMIT_SHA_SHORT=$(git rev-parse --short HEAD)
          yarn changeset version --snapshot prerelease-$COMMIT_SHA_SHORT
```

docs/snapshot-releases.md
```diff
@@ -51,3 +51,58 @@ yarn add your-package-name@bulbasaur
...
+
+## Automatic snapshots on PRs
+
+You can automatically generate snapshots on pull requests with GitHub Actions:
+
+```yaml
+name: Pre-release
+
+on:
+  pull_request:
```
Andarist, line 63:
```suggestion
on: pull_request
```

docs/snapshot-releases.md
```diff
@@ -51,3 +51,58 @@ yarn add your-package-name@bulbasaur
...
+
+jobs:
+  publish_prerelease:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
Andarist, line 71: This was needed in the past but it shouldn't be needed for already quite some time (`@changesets/cli` tries to auto-deepen the git history, so you don't have to fetch *all* of it upfront)
```suggestion
```

docs/snapshot-releases.md
```diff
@@ -51,3 +51,58 @@ yarn add your-package-name@bulbasaur
 ## What to do with the snapshot branch
 
 In almost all circumstances, we recommend that the changes after you have run `version` get merged back into your main branch. With snapshots, this is not the case. We recommend that you do not push the changes from this running of `version` to any branch. This is because the snapshot is intended for installation only, not to represent the correct published state of the repo. Save the generated version, and the tag you used, but do not push this to a branch you are planning to merge into the main branch, or merge it into the main branch.
+
+## Automatic snapshots on PRs
+
+You can automatically generate snapshots on pull requests with GitHub Actions:
```
Andarist, line 57: I wonder if we should mention that this strategy is only safe and valid when working in a closed environment without forks. The problem with forks is that we can't really give our secrets, that are required for publishing, to the fork as that would be a security concern.

docs/snapshot-releases.md
```diff
@@ -51,3 +51,58 @@ yarn add your-package-name@bulbasaur
...
+
+`changeset publish` actually doesn't care if a changeset is even generated, it will still publish the version that doesn't exist on npm even if `changeset version` exits because no unreleased changesets were found.
+
+So, if you have the GitHub Action set up in the repo, and then create a PR adding a brand new package with a version of `0.0.0` in package.json (common scenario when just starting out, a throwaway version number when still in development), the GitHub Action will run on the PR and `changeset publish` will see that `0.0.0` has not been published on npm for this package, and **it will publish that version.**
+
+You can get around this in two ways:
+
+- By temporarily marking `private: true` in package.json while setting up the new package, and removing it when you're ready to generate a changeset.
+
+- By utilizing the [ignore option](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages) temporarily.
```
Andarist, line 108: This one actually might not work right now (but I think I'd like to make it behave this way)

docs/snapshot-releases.md
```diff
@@ -51,3 +51,58 @@ yarn add your-package-name@bulbasaur
...
+          yarn changeset version --snapshot prerelease-$COMMIT_SHA_SHORT
+          yarn changeset publish --tag prerelease
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+```
+
+This will generate a snapshot on each commit to a PR using the commit `SHA`, and publish it under the `prerelease` tag. This is very handy if you need to incrementally test out changes in your PR and don't want to generate snapshots manually for each commit.
+
+However, there is an **unintended consequence** to this for brand new packages that have never been published to npm before.
```
Andarist, line 96: From this paragraph to the end - this actually describes a behavior that is not related to snapshot versions so I think I would move this to the description of the `changeset publish` itself.

This information is still important for snapshot releases so I would link from here to the description of this caveat in the `changeset publish` doc.

Andarist (commented):
Thank you for the PR! It's wonderful. I've just left some minor comments that would have to be resolved before we can land this.

docs/snapshot-releases.md
melanieseltzer: [`HEAD` refers to the SHA for the merge commit](https://stackoverflow.com/questions/68061051/get-commit-sha-in-github-actions) which doesn't match the SHA shown in the GitHub PR UI. 

IMO, it's not useful to have that SHA in the version name since I want to be able to grab the SHA and do a find in my GitHub PR to find the matching commit.

Screen Shot 2022-03-07 at 8 35 56 AM
Screen Shot 2022-03-07 at 8 38 46 AM

I can change it to `HEAD` here in the docs for brevity though? Up to you 😄 I just included it since it's part of my setup.

docs/snapshot-releases.md
melanieseltzer: That's a good point I hadn't considered. Would adding a repo check (like how [Formik does it](https://github.com/jaredpalmer/formik/blob/master/.github/workflows/release.yml#L10), for example) be acceptable in a non-private repo?

Also just so I'm fully understanding (am a bit new to publishing in general) - wouldn't the action fail in the forked repo since the secrets wouldn't have gotten added there? Or is it an issue if they go and add their own GitHub token/NPM token and attempt to publish under the same package name?

melanieseltzer:
Thanks for the review @Andarist! I have committed the requested changes that don't require more discussion 😄

docs/snapshot-releases.md
melanieseltzer: Looks like you can [turn off workflow runs for forks](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks), and require approval for all outside collaborator PRs. Should we recommend that here in addition to the repo check in the action itself?

Andarist: > Would adding a repo check (like how [Formik does it](https://github.com/jaredpalmer/formik/blob/master/.github/workflows/release.yml#L10), for example) be acceptable in a non-private repo?

I think that the answer is **yes** but I'm not completely sure. I'm not 100% sure about this though. I think I have understood this at one point in time but I completely lost this knowledge since then and it was quite fuzzy to me back then anyway. We have both `pull_request` and `pull_request_target` events and I don't recall exactly how they are supposed to be used. I think this might be a somewhat good explainer: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

> Also just so I'm fully understanding (am a bit new to publishing in general) - wouldn't the action fail in the forked repo since the secrets wouldn't have gotten added there?

I think that ye - it would fail. On the other hand - if we know that a particular action doesn't have a chance of succeeding then we shouldn't start it at all as it will only bring noise to the logs etc.

> Mainly wondering because I'm a bit confused how automatic snapshots would be different from regular publishing in terms of the forks/security concerns.

Regular publishes are usually done from the base branch of the repo and as such they are almost always executed from within that repo's context - so they have access to those secrets and since new commits can arrive on the base branch only because somebody with write permissions has pushed/merged them then it's fine. After all they could steal those secret one way or another anyway by pushing out some malicious action or something.

docs/snapshot-releases.md
Andarist: This makes sense - I totally forgot that HEAD here points to a different commit.

docs/snapshot-releases.md
melanieseltzer: From what I can piece together from the article, workflows triggered via `pull_request` in external forks do not have access to repository secrets and is read-only. Seems that a combination of turning off workflow runs for forks (described above) and adding a repo check (don't even run if the `github.repository` doesn't match) would be sufficient, as a note to have in the docs.

_However_, now I'm thinking it would probably just be simpler to remove the GitHub Action stuff entirely and not include it in the docs. I believe accidental publishing would still occur if a person attempts to do a snapshot manually without an added changeset, is that correct? So it would still be valuable information to have in the docs without all the context around automatic snapshots.

What do you think?

Andarist: You are probably right that a repository check should be enough to avoid unwanted consequences here. I think this part of the PR is good to go and that including this workflow example can be quite handy for people

docs/snapshot-releases.md
Andarist: this comment got resolved but I think that my doubt wasn't addressed, could you take another look here?

melanieseltzer: Oops, good catch! I accidentally removed it from the wrong section. Fixed in my latest commit!

docs/snapshot-releases.md
melanieseltzer: Added to the docs :) Let me know if there's anything else about it you'd want me to include.

docs/snapshot-releases.md
```diff
@@ -51,3 +51,44 @@ yarn add your-package-name@bulbasaur
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+```
+
+This will generate a snapshot on each commit to a PR using the commit `SHA`, and publish it under the `prerelease` tag. This is very handy if you need to incrementally test out changes in your PR and don't want to generate snapshots manually for each commit.
+
+There is an **unintended consequence** to this for brand new packages that have never been published to npm before. You can read more about it in the [changeset publish docs](./command-line-options.md#unintended-first-time-publish).
+
+**Please note:** In addition to the repository check (`if: github.repository == ...`), we recommend that you also [turn off workflow runs for forks](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-required-approval-for-workflows-from-public-forks). This will prevent the workflow from running on public forks, unless you approve it.
```
Andarist, line 94: I wonder though if that wouldn't also turn off testing/listing/whatever jobs for PRs coming from forks. 🤔

docs/command-line-options.md
```diff
@@ -109,6 +109,18 @@ Because this command assumes that last commit is the release commit you should n
 
 `--tag TAGNAME` - for packages that are published, the chosen tag will be used instead of `latest`, allowing you to publish changes intended for testing and validation, not main consumption. This will most likely be used with [snapshot releases](./snapshot-releases.md).
 
+### Unintended first time publish
```
Andarist, line 112: This section is great but it's explained from the perspective of snapshots whereas this is just an implication of the design.

The primary rule that should be explained first is that `changeset publish` just doesn't actually care about changeset files at all. The only thing that it does is:
1. it queries current versions in the repository
2. it compares with published versions
3. it tries to publish what has not been published yet

So this accidental publish can even happen if you have some unpublished package (or just unpublished version of a package), you start using Changesets and use our `changesets/action`. If you don't immediately add a changeset file then the first run of our action just calls `changeset publish` and it publishes what has not been published yet (as described above).

I think that we should start with explaining this principle, then mention this implication for starting using Changesets and `changeset/action` (later we'd link to here from `changeset/action`'s README) and then also mention how it affects snapshot releases.

Snapshot releases are just always prone to this behavior because when creating them on CI we usually run a script like `changeset version && changeset publish`, without checking if `changeset version` even has done anything (checking that would usually complicate stuff beyond what is necessary). So since such action always ends up calling `changeset publish` we are always facing this "risk".

melanieseltzer: Sorry for the delay here, I've re-worded everything as you suggested! I've removed the section I added previously ("Unintended first time publish") and just worked everything into the main section for the command.

Does it read any better?

docs/snapshot-releases.md
melanieseltzer: You're right! I forgot that with GitHub Actions people tend to split up the workflows into different files, rather than having everything in one file. I've removed this recommendation since you can guard against it with the repo check in the one workflow file for release.

beeequeue:
closing as this is stale, feel free to open a PR against the [new docs branch](https://github.com/changesets/changesets/pull/1979)

Issue #754, opened 2022-02-13
Add docs for the fixed mode


Andarist:
We've landed https://github.com/changesets/changesets/pull/690 without the docs but it would be great to have at least basic information about this new feature in the docs.

akphi:
Is the doc basically https://github.com/changesets/changesets/issues/581#issuecomment-830807176 or you want to have something similar to https://github.com/changesets/changesets/blob/main/docs/linked-packages.md with a section on `linked` vs `fixed`?

PR #764, opened 2022-02-24
add docs for fixed packages

akphi:
Closes https://github.com/changesets/changesets/issues/754


README.md
```diff
@@ -50,6 +50,7 @@ To make releasing easier, you can use [this changesets github action](https://gi
 - [Decisions](./docs/decisions.md)
 - [Dictionary](./docs/dictionary.md)
 - [Linked packages](./docs/linked-packages.md)
+- [Fixed packages](./docs/fixed-packages.md)
```
Andarist, line 53: from my experience people were often confused about linked packages - they actually assumed that they would work like fixed ones, so perhaps we should reorder those here to prioritize fixed over linked? the same comment goes for other places were you put fixed below linked within this PR

docs/config-file-options.md
```diff
@@ -65,6 +65,18 @@ If you want to use this option, you should read the documentation on [linked pac
 
 > NOTE: This does not do what some other tools do, which is make sure when any package is published, all other packages are also published with the same version.
 
+## `fixed` (array of arrays of package names)
+
+This option can be used to declare that packages should be version-bumped and published together. As an example, if you have a `@changesets/button` component and a `@changesets/theme` component and you want to make sure that when one gets bumped to `2.0.0`, the other is also bumped to `2.0.0` regardless if it has any change or not. To achieve this you would have the config:
```
Andarist, line 70: It feels that using a major version example could be confusing here cause it's less common situation and somebody could think that the behavior only applies to major versions cause of that.
```suggestion
This option can be used to declare that packages should be version-bumped and published together. As an example, if you have a `@changesets/button` component and a `@changesets/theme` component and you want to make sure that when one gets bumped to `1.1.0`, the other is also bumped to `1.1.0` regardless if it has any change or not. To achieve this you would have the config:
```

docs/dictionary.md
```diff
@@ -17,6 +17,7 @@ Several of these have associated type definitions, which you can find in [our ty
 - **multi-package repo/monorepo** - A repository which contains multiple packages, generally managed by [Bolt](https://github.com/boltpkg/bolt) or [Yarn Workspaces](https://yarnpkg.com/lang/en/docs/workspaces/).
 - **release line generators** - The `getReleaseLine` and `getDependencyReleaseLine` functions which are responsible for creating the lines inserted into changelog. A changelog entry for a particular release can be thought of as `releaseLineGenerators(changesets)`
 - **linked packages** - Linked packages share a semver categorisation, such that all published linked packages have consistent new semver ranges. The logistics of this are best left to our [./linked-packages.md] documentation.
+- **fixed packages** - Fixed packages share a semver categorisation, such that all linked packages have consistent new semver ranges and will be published together. The logistics of this are best left to our [./fixed-packages.md] documentation.
```
Andarist, line 20:
```suggestion
- **fixed packages** - Fixed packages share a semver categorisation, such that all fixed packages have the same semver version and are always published together. The logistics of this are best left to our [./fixed-packages.md] documentation.
```

docs/fixed-packages.md
```diff
@@ -0,0 +1,49 @@
+# Fixed Packages
+
+Fixed packages allow you to specify a group or groups of packages that should be versioned and published together.
+
+> Unlike `linked packages`, all packages in the group of fixed packages will be version-bumped and published together even when there are no changes done to some of the the member packages.
```
Andarist, line 5: a similar note (just "in reverse") could be added to the linked packages doc

Andarist:
❤️ ❤️ ❤️

akphi:
Updated!

docs/fixed-packages.md
akphi: updated!

README.md
akphi: done

Andarist (approved):

PR #768, opened 2022-03-05
Add CommitFunctions and update config with defaults

### LINKED ISSUE #285: Changesets ignoring the skipCI config key
## Affected Packages

`pkg:cli`

## Problem
Setting the config.json this way
```
{
  "$schema": "https://unpkg.com/@changesets/config@0.3.0/schema.json",
  "commit": true,
  "skipCI": false,
  "linked": [],
  "access": "restricted",
  "baseBranch": "master"
}
```
when changesets commits in CI, the comit message still contains `skipCI`

### LINKED ISSUE #688: Ability to set `commit` to different values in add and version commands.

rohit-gohri:
Follow up of #735 

Closes #688 
Closes #285 

Made commit option similar to changelog option but with a default template when "true" is passed (keeping it backward compatible).

~This currently doesn't allow disabling commit for one of add/version. Maybe that can be added too if the respective function returns null or doesn't exist?~
If one of the functions doesn't exist then we will not commit for that only.


packages/types/src/index.ts
```diff
@@ -122,6 +122,21 @@ export type ChangelogFunctions = {
...
+) => Promise<string>;
+
+export type GetVersionLine = (
+  releasePlan: ReleasePlan,
+  commitOptions: null | Record<string, any>
+) => Promise<string>;
+
+export type CommitFunctions = {
+  getAddLine: GetAddLine;
+  getVersionLine: GetVersionLine;
```
Andarist, line 137: IMHO using "message" here makes more sense as that's how this is called by git itself (https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--mltmsggt)
```suggestion
  getAddMessage: GetAddLine;
  getVersionMessage: GetVersionLine;
```

You'd have to adjust other places too (and names of those types here).

packages/config/src/index.ts
```diff
@@ -42,6 +42,21 @@ function getNormalizedChangelogOption(
...
+  if (thing === false) {
+    return false;
+  }
+  if (thing === true) {
+    return ["@changesets/cli/commit", null];
+  }
+  if (typeof thing === "string") {
+    return [thing, null];
+  }
+  return thing;
```
Andarist, line 57: I'm not sure if this is validated for changelogs but we probably shouldn't assume here that this will be valid. At the very least this has to be an array and its first element has to be a string, right?

packages/config/src/index.ts
```diff
@@ -126,13 +141,22 @@ export let parse = (json: WrittenConfig, packages: Packages): Config => {
...
+      typeof json.commit[0] === "string"
+    )
+  ) {
     messages.push(
       `The \`commit\` option is set as ${JSON.stringify(
         json.commit,
         null,
         2
-      )} when the only valid values are undefined or a boolean`
+      )} when the only valid values are undefined or a boolean or a module path(e.g. "@changesets/cli/commit" or "./some-module") or a tuple with a module path and config for the changelog generator(e.g. ["@changesets/cli/commit", { someOption: true }])`
```
Andarist, line 159:
```suggestion
      )} when the only valid values are undefined or a boolean or a module path (e.g. "@changesets/cli/commit" or "./some-module") or a tuple with a module path and config for the changelog generator (e.g. ["@changesets/cli/commit", { someOption: true }])`
```

packages/apply-release-plan/src/index.ts
```diff
@@ -181,7 +179,32 @@ export default async function applyReleasePlan(
...
+      possibleCommitFunc = possibleCommitFunc.default;
+    }
+    if (
+      typeof possibleCommitFunc.getAddLine === "function" &&
+      typeof possibleCommitFunc.getVersionLine === "function"
+    ) {
+      getCommitFuncs = possibleCommitFunc;
+    } else {
+      throw new Error("Could not resolve commit generation functions");
+    }
```
Andarist, line 202: I guess we can start from requiring both to exist - but I wonder if there are people only wanting one of those to kick in 🤔

packages/cli/src/commit/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    .map(release => `  ${release.name}@${release.newVersion}`)
+    .join("\n");
+
+  return outdent`
+    RELEASING: Releasing ${numPackagesReleased} package(s)
+
+    Releases:
+    ${releasesLines}
+
+    [skip ci]
```
Andarist, line 26: What if we would remove this from here and instead add options to this function and only render this based on smth like `options.versionSkipCI === true`. This way we could "translate" current `commit: true` to `["@changesets/cli/commit", { versionSkipCI: true }]`. 

This would be quite nice because one could more easily extend the built-in function without having to strip `[skip ci]` from it if they don't want it. This option would go away in Changesets v3 (if that ever gets published) and would only be supported now to avoid breaking changes.

packages/cli/src/commit/index.ts
```diff
@@ -0,0 +1,67 @@
...
+};
+
+export const defaultCommitFunctions: CommitFunctions = {
+  getAddLine,
+  getVersionLine
+};
+
+export default defaultCommitFunctions;
+
+export function getCommitFuncs(
```
Andarist, line 38: is this actually used anywhere? or just in tests?

packages/apply-release-plan/src/index.ts
rohit-gohri: We can treat not existing as `commit: false` for add/version separately.

packages/cli/src/commit/index.ts
rohit-gohri: This is a great idea, adds back `skipCI` but as a different option

packages/cli/src/commit/index.ts
rohit-gohri: It is used in the `add` function. I wanted to also use it in `apply-release-plan` but couldn't figure out how that import would work

packages/cli/src/commit/index.ts
rohit-gohri: I added for both `versionSkipCI` and `addSkipCI`, with `versionSkipCI` as true for the `true` config

packages/cli/src/commit/index.ts
Andarist: I think that it would make sense to introduce a breaking change to the `@changesets/apply-release-plan` (that would not be a breaking change for the Changesets itself) and just move out the committing logic from it. Based on its return value the "version commit" could be conditionally created within `@changesets/cli`. 

This would make both commit types be performed by a single package - which I think would be a good thing for the codebase.

rohit-gohri: Moved to the cli 🙂

packages/cli/src/commit/index.ts
rohit-gohri: `addSkipCI` seems like a weird option name now that I have added it, WDYT about just having `skipCI` and possible values being `'version' | 'add' | boolean`?

packages/apply-release-plan/src/index.ts
rohit-gohri: Added support for partial functions. If only one of it is supplied then commit will be done for that command only

rohit-gohri:
All done from my side, the only thing remaining is the name of the skipCI options. See [this comment](https://github.com/changesets/changesets/pull/768#discussion_r820182114)

packages/types/src/index.ts
```diff
@@ -122,6 +122,21 @@ export type ChangelogFunctions = {
...
+) => Promise<string>;
+
+export type GetVersionMessage = (
+  releasePlan: ReleasePlan,
+  commitOptions: null | Record<string, any>
+) => Promise<string>;
+
+export type CommitFunctions = {
+  getAddMessage: GetAddMessage;
+  getVersionMessage: GetVersionMessage;
```
Andarist, line 137: Since we allow those two to be optional~ I think this should be reflected in this type
```suggestion
  getAddMessage?: GetAddMessage;
  getVersionMessage?: GetVersionMessage;
```

packages/cli/src/commit/commit.test.ts
```diff
@@ -0,0 +1,217 @@
...
+            type: "minor"
+          }
+        ]
+      },
+      commitOpts
+    );
+    expect(commitStr).toEqual(`docs(changeset): test changeset summary commit`);
+  });
+
+  it("should handle a simple changeset - skipCi", async () => {
```
Andarist, line 82:
```suggestion
  it("should handle a simple changeset - skipCI", async () => {
```

packages/cli/src/commands/version/index.ts
```diff
@@ -71,15 +73,38 @@ export default async function version(
...
+    cwd
+  );
+  if (getVersionMessage) {
+    let newTouchedFilesArr = [...touchedFiles];
+    // Note, git gets angry if you try and have two git actions running at once
+    // So we need to be careful that these iterations are properly sequential
+    while (newTouchedFilesArr.length > 0) {
+      let file = newTouchedFilesArr.shift();
+      await git.add(path.relative(cwd, file!), cwd);
+    }
```
Andarist, line 94: we don't need to touch this here but it's probably better to construct a single call to `git add` since stuff like `git add file1.js file2.js ...` is a totally valid thing to do

packages/apply-release-plan/src/index.test.ts
```diff
@@ -2352,35 +2304,18 @@ describe("apply release plan", () => {
...
 
-      expect(gitCmd.stdout.toString().includes("nothing to commit")).toBe(true);
+      const changesetsDeleted = releasePlan.changesets.reduce(
+        (prev, { id }) => {
+          return (
+            prev &&
+            gitCmd.stdout.toString().includes(`deleted:    .changeset/${id}.md`)
+          );
+        },
+        true as boolean
```
Andarist, line 2314: This is already inferred correctly
```suggestion
        true
```

packages/cli/src/commit/index.ts
Andarist: that works too, I don't mind any of those solutions - I agree that the current names are a little bit weird but since I'd like to remove this in the future anyway... I don't care that much about this :p

Andarist:
The PR is almost ready to be shipped, great work! Once all remaining items (mostly nits really) get addressed I will rethink the rollout strategy as we need to consider how this affects the version of the involved packages and how to ensure that this doesn't break any existing consumers.

packages/cli/src/commands/version/index.ts
rohit-gohri: The add function in `@changesets/git` only accepts one path so I'm going to leave this for now.

packages/cli/src/commit/index.ts
rohit-gohri: Changed to `"skipCI": "version" | "add" | boolean` as the option.

LeviticusMB:
I would love to see this PR merged as soon as possible. What's holding it back?

Really looking forward to being able to turn auto-commit back on without getting that `docs(changeset)` prefix (which makes no sense when the actual change is just one commit) in the git log.

docs/config-file-options.md
```diff
@@ -16,9 +16,30 @@ Changesets has a minimal amount of configuration options. Mostly these are for w
 
 > NOTE: the `linked`, `updateInternalDependencies`, and `ignore` options are only for behaviour in monorepos.
 
-## `commit` (`true` | `false`)
+## `commit` (`true` | `false` or a module path)
 
-This argument sets whether the `changeset add` command and the `changeset publish` command will also add and commit the changed files using git. By default, we do not commit the files, and leave it to the user to commit the files.
+This option is for setting if the `changeset add` command and the `changeset publish` command will also add and commit the changed files using git, and how the commit messages should be generated for them.
```
Andarist, line 21:
```suggestion
This option is for setting if the `changeset add` command and the `changeset version` commands will also add and commit the changed files using git, and how the commit messages should be generated for them.
```

docs/config-file-options.md
```diff
@@ -16,9 +16,30 @@ Changesets has a minimal amount of configuration options. Mostly these are for w
 
 > NOTE: the `linked`, `updateInternalDependencies`, and `ignore` options are only for behaviour in monorepos.
 
-## `commit` (`true` | `false`)
+## `commit` (`true` | `false` or a module path)
```
Andarist, line 19:
```suggestion
## `commit` (`true` | `false` or `[modulePath: string, options: any]`)
```

packages/config/src/index.ts
```diff
@@ -42,6 +42,21 @@ function getNormalizedChangelogOption(
...
+  thing: boolean | readonly [string, any] | string
+): Config["commit"] {
+  if (thing === false) {
+    return false;
+  }
+  if (thing === true) {
+    return ["@changesets/cli/commit", { skipCI: "version" }];
+  }
+  if (typeof thing === "string") {
+    return [thing, null];
```
Andarist, line 55:
```suggestion
    return [thing, {}];
```

packages/types/src/index.ts
```diff
@@ -122,6 +122,21 @@ export type ChangelogFunctions = {
   getDependencyReleaseLine: GetDependencyReleaseLine;
 };
 
+export type GetAddMessage = (
+  changeset: Changeset,
+  commitOptions: null | Record<string, any>
```
Andarist, line 127: it would be more true if this would get annotated as `any`
```suggestion
  commitOptions: any
```

packages/types/src/index.ts
```diff
@@ -122,6 +122,21 @@ export type ChangelogFunctions = {
...
+) => Promise<string>;
+
+export type GetVersionMessage = (
+  releasePlan: ReleasePlan,
+  commitOptions: null | Record<string, any>
```
Andarist, line 132:
```suggestion
  commitOptions: any
```

Andarist:
> What's holding it back?

Lack of the free time of maintainers 😉

Andarist:
@rohit-gohri thank you very much for your contributions and bearing with me. People like you keep OSS projects afloat 🚀

LeviticusMB:
Thanks a lot guys, really awesome! Works as advertised for me.

However, it does look like `config/schema.json` was not updated to reflect the change?

Andarist:
> However, it does look like config/schema.json was not updated to reflect the change?

Yeah, we forgot about it - PR with an update would be appreciated.

PR #769, opened 2022-03-05
Fixed an infinite loop with fixed & ignored package depending on another release

Andarist:
This is fixing a problem reported in this comment: https://github.com/changesets/changesets/issues/571#issuecomment-1057352651

cc @fivethreeo


fivethreeo:
Seems correct to me

mitchellhamilton (approved):

PR #770, opened 2022-03-05
Get correct absolute path of changed files when `.changeset/` is not in the root of the repository

alizeait:
When creating absolute paths from changed files, we assume `cwd` is the repository root but that is not always the case like when .changesets is set in a subfolder. Instead we can ask git for the repository root and create correct absolute path of the files returned from `git diff`

### DIFF
--- a/.changeset/silly-sloths-brush.md
+++ b/.changeset/silly-sloths-brush.md
@@ -2,4 +2,4 @@
 "@changesets/git": patch
 ---
 
-Get correct repository root instead of `cwd`when getting absolute paths of changed files
+`getChangedFilesSince` and `getChangedPackagesSinceRef` will now return the correct absolute paths of the changed files when the passed `cwd` is different from the repository's root.
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -183,14 +183,14 @@ export async function deepenCloneBy({ by, cwd }: { by: number; cwd: string }) {
   await spawn("git", ["fetch", `--deepen=${by}`], { cwd });
 }
 async function getRepoRoot({ cwd }: { cwd: string }) {
-  const { stdout, code } = await spawn(
+  const { stdout, code, stderr } = await spawn(
     "git",
     ["rev-parse", "--show-toplevel"],
     { cwd }
   );
 
   if (code !== 0) {
-    return cwd;
+    throw new Error(stderr.toString());
   }
 
   return stdout
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -217,14 +217,14 @@ export async function getChangedFilesSince({
     );
   }
 
-  const repoRoot = await getRepoRoot({ cwd });
-
   const files = cmd.stdout
     .toString()
     .trim()
     .split("\n")
     .filter(a => a);
   if (!fullPath) return files;
+
+  const repoRoot = await getRepoRoot({ cwd });
   return files.map(file => path.resolve(repoRoot, file));
 }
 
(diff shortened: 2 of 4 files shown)


Andarist:
Ok, I think I understand what you are saying... but we would need a test case for this change, ideally one that would involve some changesets command, or at least a full user story that involves Changesets.

alizeait:
> Ok, I think I understand what you are saying... but we would need a test case for this change, ideally one that would involve some changesets command, or at least a full user story that involves Changesets.

Sure, I added a unit test that asserts the changed behavior. Not sure how adding a Changeset command test case would be beneficial here considering that the git functions are simply mocked in those test files. Would appreciate any pointers if you have something else in mind

Andarist:
I was just wondering how exactly this has manifested in a bug in Changesets. Did `changesets status` report a wrong value~? Or maybe `changesets add` didn't list the packages correctly?

alizeait:
Exactly, `changesets add` would not report the list of changed packages since changed files full paths were not correct. For example in a monorepo with this setup where `.changeset` folder is not in the monorepo root:

    └── monorepo/
        ├── packages/
        │   ├── .changeset/
        │   │   └── config.json
        │   ├── pkg-a/
        │   │   └── package.json
        │   └── pkg-b/
        │       └── package.json
        └── package.json 

When running `changesets add` from `/packages` folder, let's say pkg-a/package.json was modified, the reported full path of that file would be:
`/home/user/monorepo/packages/packages/pkg-a/package.json`, notice the double /packages, this is because we have `cwd` as `/home/user/monorepo/packages` and changed file reported by git is `packages/pkg-a/package.json` (always relative repository root). 
Now when we do `path.resolve(cwd, 'packages/pkg-a/package.json')` we get:
`/home/user/monorepo/packages/packages/pkg-a/package.json` which is not correct. So instead of using `cwd` in `path.resolve`, we can ask git for the repository root and always get the correct full path of the changed file.

packages/git/src/index.ts
```diff
@@ -182,6 +182,22 @@ export async function isRepoShallow({ cwd }: { cwd: string }) {
...
+async function getRepoRoot({ cwd }: { cwd: string }) {
+  const { stdout, code } = await spawn(
+    "git",
+    ["rev-parse", "--show-toplevel"],
+    { cwd }
+  );
+
+  if (code !== 0) {
+    return cwd;
+  }
```
Andarist, line 194: Is this a defensive check or perhaps this can sometimes happen for some legitimate reason?

.changeset/silly-sloths-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/git": patch
+---
+
+Get correct repository root instead of `cwd`when getting absolute paths of changed files
```
Andarist, line 5:
```suggestion
`getChangedFilesSince` will now return the correct absolute paths of the changed files when the passed `cwd` is different from the repository's root.
```

Andarist: We should also add a separate changeset file with such content:
```md
---
"@changesets/cli": patch
---

`changeset add` and `changeset status` should now properly handle the situation where Changesets are managed from a directory different than the root of the repository.
```

packages/git/src/index.ts
alizeait: I doubt It'll fail, but if it does, fallback to the previous behavior

Andarist, line 194: according to the docs:
> Show the (by default, absolute) path of the top-level directory of the working tree. If there is no working tree, report an error.

So I think that we should throw an error here:
```suggestion
  if (code !== 0) {
    throw new Error(gitCmd.stderr.toString());
  }
```

Where for us this actually shouldn't ever throw, but this function is quite generic - so if we ever expose it then somebody could call it with `cwd` outside of a git repo.

I think that it's better to throw here to raise potential issues early, rather than to swallow the error and lead to undetermined behavior.

alizeait: Sounds reasonable

PR #791, opened 2022-04-20
Add commit generator options to schema

rohit-gohri:
Follow up of #768


rohit-gohri:
@Andarist  No worries. As long as the issue is fixed I'm happy! 😄

PR #792, opened 2022-04-25
Add The Guild libraries to list of Changesets adopters

dotansimha:
Hi there! 

We at [The Guild](https://the-guild.dev/) are huge fans of Changesets. 

We are using it in popular libs like [`graphql-yoga`](https://github.com/dotansimha/graphql-yoga), [`GraphQL Code Generator`](https://github.com/dotansimha/graphql-code-generator), [`graphql-tools`](https://github.com/ardatan/graphql-tools) and a lot more (~20 more repos). 

We are using most of Changesets features and also extended it with some custom configurations (like [publishing canary/alpha releases based on changesets for each in-progress PR](https://github.com/dotansimha/graphql-code-generator/blob/master/scripts/canary-release.js)).


Andarist:
Thank you for your PR! I am slightly torn on what I should do with it though. There are two aspects here that I take into consideration:
- The Guild is a company and not a single project, everything else on the list is a project. I understand that this somewhat acts as an umbrella for all of the projects that you maintain but the point stands.
- the list already is somewhat long and it can grow somewhat indefinitely with time

I'm not saying that I don't want to merge this - I'm just considering, in general, what's the best course of action here. I think that ideally I would like to replace the vertical list with a list of logos that could be wrapped horizontally - with such in place I wouldn't object to including The Guild as a whole on the list. But perhaps right now I would prefer this to be replaced with a project or two.

Note that this is mostly me thinking out loud - if you have any other point of view on the matter I would happily discuss this further, don't feel obliged to make any changes right away.

dotansimha:
Thanks @Andarist , I agree with your take on this one. I updated the PR and added a list of projects :)

README.md
```diff
@@ -76,6 +76,13 @@ To make releasing easier, you can use [this changesets github action](https://gi
 - [SvelteKit](https://kit.svelte.dev/)
 - [Hydrogen](https://hydrogen.shopify.dev)
 - [react-pdf](https://github.com/diegomura/react-pdf)
+- [GraphQL Code Generator](https://github.com/dotansimha/graphql-code-generator)
+- [GraphQL Yoga](https://github.com/dotansimha/graphql-yoga)
+- [GraphQL ESLint](https://github.com/B2o5T/graphql-eslint)
+- [Envelop](https://github.com/n1ru4l/envelop)
+- [GraphQL Hive](https://github.com/kamilkisiela/graphql-hive)
```
Andarist, line 83: Those 3 seem to be less popular than the other packages on the list. Of course, those are probably still excellent libraries but this list is kind of a "showcase", and as mentioned - I'd like to try to keep it short(-ish) so I need to use some arbitrary metric to reject a project from time to time. It's just mainly to avoid having too many projects here so the readers are not overwhelmed with the list.
```suggestion
```

dotansimha: Sure, done! Thanks @Andarist

PR #801, opened 2022-05-08
[Feature] Support NX Monorepos

### LINKED ISSUE #800: [Feature Request] Support NX Monorepos
## Affected Packages

[get-workspaces](https://github.com/changesets/changesets/tree/main/packages/get-workspaces)

## Problem A

As of today, in an [NX workspace](https://nx.dev/getting-started/intro), it is not possible for Changesets to discover any related packages that are configured for the workspace. 

There are only [two ways of configuring where packages might be in an NX repo](https://nx.dev/configuration/projectjson#workspace-json):

1. [project.json and nx.json](https://nx.dev/configuration/projectjson)
2. [package.json and nx.json](https://nx.dev/configuration/packagejson) (also requires a [workspace.json](https://nx.dev/configuration/projectjson#workspace-json) file)

With the first option, NX relies on a project.json file existing at the root of any packages in the monorepo, which for publishable packages, will also need to have a package.json alongside them.

With the second option, there is a workspace.json file at the root that references all packages in the monorepo.

> Note: [If there is a `workspaces.json` file, NX will not glob for `project.json` files](https://github.com/nrwl/nx/blob/cce5bfebc008df85a0d2a2c9abde815fe330cb69/packages/nx/src/config/workspaces.ts#L68) in the directories. It appears that `readWorkspaceConfiguration` is available as part of `@nrwl/devkit`.

### Proposed solution

1. If `@manypkg/get-packages` does not return any packages, check for `nx.json` at our monorepo root. If present, use `readWorkspaceConfiguration()` from `@nrwl/devkit` to get the workspace tree instead of trying to implement ourselves.
2. Define `@nrwl/devkit` as an optional dependency, which will already be present in any NX monorepo.

## Problem B

By default, NX does not compile publishable code adjacent to publishable packages, and instead usually has a root `/dist` directory where all build outputs go which are namespaced to the lib/app. This compiled output will also contain a copy of the package.json with all relevant dependencies added to it, and is what we should be publishing.

> The build output is configurable in NX per package and there is likely a function as part of `@nrwl/devkit` that also returns configured output for each package

> Not every package is necessarily publishable, again configurable, and thus will not get built to the `/dist` directory

### Proposed solution

1. Use `@nrwl/devkit` to find which package(s) are publishable and publish their assets directly from the configured output directory

---
It is also worth noting that there are a couple other issues already opened that are asking for support for this, but are asking for ways to configure all of this by hand, versus using the tool set in `@nrwl/devkit`: 
- #654 
- #761 
- #773

brianespinosa:
Resolves #800 

WIP. Creating a draft PR for visibility, and for CI. Includes an NX fixture repo and tests. Two expect statements to get green and then add the NX fixture use case to other packages to verify NX works there as well.

I might get a chance to come back and finish this evening.


packages/get-workspaces/package.json
```diff
@@ -1,16 +1,30 @@
...
+    "fs-extra": "^7.0.1"
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "@nrwl/devkit": "*",
+    "fixturez": "^1.1.0",
+    "nx": "*"
+  },
+  "peerDependencies": {
+    "@nrwl/devkit": "*",
```
Andarist, line 19: is it always **explicitly** installed by Nx users? 🤔

packages/get-workspaces/package.json
```diff
@@ -1,16 +1,30 @@
 {
-  "name": "get-workspaces",
+  "name": "@changesets/get-workspaces",
```
Andarist, line 2: In general `@manypkg/get-packages` is also under our control - so any logic that tries to find packages in a given monorepo should be added there. Given this situation I don't think there is any strong reason to create `@changesets/get-workspaces`

jericopingul:
Hello @brianespinosa 👋 just wondering when would we expect for this change to be available?

packages/get-workspaces/package.json
airtonix: nx monorepos don't always have the required files that `@manypkg/get-packages` needs in order to build the graph of related packages.

In a nx repo, you can in fact have only one `package.json` in the root of the repo and then all the packages defined with `project.json` files throughout.

When it comes to a typescript orientated nx monorepo, the default way to facilitate `import { Foo } from "@ourrepo/foo";` is via the root `tsconfig.base.json` which will contain a `paths` entry describing every package and the page to its main `index.ts`.

This combined with how each package has it's `tsconfig.json` configured with references makes for a blazingly fast monorepo install times and vscode intellisense experience.

sledorze:
@Andarist any news on this?

vemundeldegard:
It would be great to see this come into action. We are currently in the progress of making use of mono-repos and are considering nx. Nx does mention changesets in their documentation for version management, but how they expect this to actually work is not known to me.

brianespinosa:
I am going to go ahead and decline this PR as I think there is a better approach that is more extensible as proposed in #849

PR #820, opened 2022-05-30
Errors resulting from the GitHub API calls should now be properly raised

Andarist:
I got a report from @mikearnaldi that `data` can be returned in a shape like this:
```
{
  data: { a0: null },
  errors: [
    {
      type: 'NOT_FOUND',
      path: [Array],
      locations: [Array],
      message: "Could not resolve to a Repository with the name 'repo-tooling/eslint-plugin-dprint'."
    }
  ]
}
```

This is a simple fix to at least make those errors pop up in the CLI logs.

### DIFF
--- a/packages/get-github-info/src/index.ts
+++ b/packages/get-github-info/src/index.ts
@@ -103,7 +103,9 @@ const GHDataLoader = new DataLoader(async (requests: RequestData[]) => {
   if (data.errors) {
     throw new Error(
       `An error occurred when fetching data from GitHub\n${JSON.stringify(
-        data.errors
+        data.errors,
+        null,
+        2
       )}`
     );
   }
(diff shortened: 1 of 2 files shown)


packages/get-github-info/src/index.ts
```diff
@@ -100,6 +100,14 @@ const GHDataLoader = new DataLoader(async (requests: RequestData[]) => {
     body: JSON.stringify({ query: makeQuery(repos) })
   }).then((x: any) => x.json());
 
+  if (data.errors) {
+    throw new Error(
+      `An error occurred when fetching data from GitHub\n${JSON.stringify(
+        data.errors
```
mitchellhamilton, line 106:
```suggestion
        data.errors,
        null,
        2
```

mitchellhamilton (approved):

PR #830, opened 2022-06-03
Add options to configure snapshot preid format

jakubmazanec:


.changeset/pink-eagles-agree.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/assemble-release-plan": minor
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Adds two options to configure snapshot preid format: which character separates timestamp part from the rest of the preid, and whether the timestamp part comes first in the preid.
```
Andarist, line 7: In general, I'm not a fan of adding so many different options for something that feels very related - it seems like we are aiming to tweak a "format"/pattern of the generated snapshot with both of those.

I might give some alternative suggestions for this later on.

Andarist:
Could you refresh my memory on the motivation behind this being configurable? I know some~ relevant context but would love to hear out your arguments first.

.changeset/pink-eagles-agree.md
jakubmazanec: Sure, I'm open to alternative way how to configure this, because TBH I don't like these two options either...

jakubmazanec:
> Could you refresh my memory on the motivation behind this being configurable? I know some~ relevant context but would love to hear out your arguments first.

AFAIK the reason for having the timestamp first in the preid is to be able to predictably sort prereleases, and using `.` is also related to sorting, because according to [semver](https://semver.org/):

> Precedence for two pre-release versions [...] determined by comparing each dot separated identifier [...].

.changeset/pink-eagles-agree.md
jakubmazanec: I'm thinking about single option `snapshotPreidTemplate`, that would serve as a template for the snapshot preid, e.g. `'{{label}}.{{timestamp}}'`; placeholders would be replaced with actual timestamp and value of `snapshot` CLI option. But the problem is that `snapshot` option returns `string | boolean`, i.e. the label is optional, and that needs little more complex templating logic than simple placeholder replacement.

jakubmazanec: And now I'm thinking that I would actually like to have different templates for the preid for different prereleases. So maybe it would be better to create a new CLI option rather than config option? 🤔

Andarist:
A competing PR has been created, you might find it interesting: https://github.com/changesets/changesets/pull/858

Andarist:
The mentioned completing PR has been merged in and thus we can close this PR. Nevertheless, thank you for your contribution! It's very much appreciated.

PR #843, opened 2022-06-17
fix: changeset filename with status verbose

### LINKED ISSUE #842: changeset filename is not right when use status command with --verbose
## Affected Packages

@changesets/cli

## Problem

image

The changeset file name is not right

## Proposed solution
image

caohuilin:
fixes #842

### DIFF
--- a/.changeset/gold-mangos-search.md
+++ b/.changeset/gold-mangos-search.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+fix: changeset filename with status verbose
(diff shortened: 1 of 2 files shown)


.changeset/gold-mangos-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+fix: changeset filename with status verbose
```
Andarist, line 5:
```suggestion
Fixed the printed filename in the verbose output of `changeset status`.
```

PR #844, opened 2022-06-17
Attempt to log  when publish fails if  is empty and JSON error can't be retrieved from none of those

Andarist:
closes https://github.com/changesets/action/issues/191

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -214,7 +214,7 @@ async function internalPublish(
       );
     }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -214,7 +214,7 @@ async function internalPublish(
       );
     }
 
-    error(stderr.toString());
+    error(stderr.toString() || stdout.toString());
```
Andarist, line 217: I'm actually not sure if this is going to help anything - but IMHO it's worth doing regardless.

Issue #835, opened 2022-06-11
Cannot read properties of null (reading 'prerelease')


unional:
## Affected Packages

`standard-log`

## Problem

```sh
   yarn cs version
🦋  warn ===============================IMPORTANT!===============================
🦋  warn You are in prerelease mode
🦋  warn If you meant to do a normal release you should revert these changes and run `changeset pre exit`
🦋  warn You can then run `changeset version` again to do a normal release
🦋  warn ----------------------------------------------------------------------
🦋  error TypeError: Cannot read properties of null (reading 'prerelease')
🦋  error     at getPreVersion (D:\code\unional\standard-log\node_modules\@changesets\assemble-release-plan\dist\assemble-release-plan.cjs.dev.js:437:27)
🦋  error     at getPreInfo (D:\code\unional\standard-log\node_modules\@changesets\assemble-release-plan\dist\assemble-release-plan.cjs.dev.js:609:43)
🦋  error     at Object.assembleReleasePlan [as default] (D:\code\unional\standard-log\node_modules\@changesets\assemble-release-plan\dist\assemble-release-plan.cjs.dev.js:490:19)
🦋  error     at version (D:\code\unional\standard-log\node_modules\@changesets\cli\dist\cli.cjs.dev.js:599:60)
🦋  error     at async run$2 (D:\code\unional\standard-log\node_modules\@changesets\cli\dist\cli.cjs.dev.js:1337:11)
```

## Proposed solution

Seems like this is a bug with prerelease.

https://github.com/unional/standard-log/runs/6841190529?check_suite_focus=true

Even after I run `changeset pre exit`, the issue persist. It goes away when I delete the `pre.json` file.

Andarist:
Yeah, it looks like a bug in the prerelease mode. The problem is caused by the fact that `@standard-log/example` is included as part of packages and it's a private package without any `version` and thus parsing `undefined` version return `null` and we crash.

It shouldn't be hard to fix but I think there are some other PRs/issues related to version-less packages so I'd like to look through all of them and fix all at once.

wbcs:
> Yeah, it looks like a bug in the prerelease mode. The problem is caused by the fact that `@standard-log/example` is included as part of packages and it's a private package without any `version` and thus parsing `undefined` version return `null` and we crash.
> 
> It shouldn't be hard to fix but I think there are some other PRs/issues related to version-less packages so I'd like to look through all of them and fix all at once.

Is it fixed?

TuiMao233:
[...]

PR #847, opened 2022-06-24
Fixed the prerelease version parse

vzt7:
fix https://github.com/changesets/changesets/issues/835

when i `pnpm changeset version` after enter prerelease stage ( `pnpm changeset pre enter alpha` ), A private package without version field cause error `error TypeError: Cannot read property 'prerelease' of null`. ( It works if no enter pre )

### DIFF
--- a/packages/assemble-release-plan/src/index.ts
+++ b/packages/assemble-release-plan/src/index.ts
@@ -16,10 +16,10 @@ import { Packages, Package } from "@manypkg/get-packages";
 import { getDependentsGraph } from "@changesets/get-dependents-graph";
 import { PreInfo, InternalRelease } from "./types";
 
-function getPreVersion(version: string) {
-  let parsed = semver.parse(version)!;
+function getPreVersion(version?: string) {
+  let parsed = semver.parse(version);
   let preVersion =
-    parsed.prerelease[1] === undefined ? -1 : parsed.prerelease[1];
+    parsed?.prerelease[1] === undefined ? -1 : parsed.prerelease[1];
   if (typeof preVersion !== "number") {
     throw new InternalError("preVersion is not a number");
   }


frankxin:
Hi There，Do we have any update ？

silesky:
Thanks. Can we merge this in?

Andarist:
I think that a better fix for this would be to filter out private packages without a version field. The question is if this should happen in the `assemble-release-plan` or if this should happen before calling it:
https://github.com/changesets/changesets/blob/0fd56e17092c7258de57adf7dd4ba08406c8e440/packages/cli/src/commands/version/index.ts#L67-L69

We probably should also review all of the other places where we call `getPackages` because there is a high chance that we need similar logic in those other places too.

ndaba1 (commented):
Thanks! It works for me

AntoineKM (approved):

xencodes:
any fix ?

Szymon-dziewonski:
Thanks @vzt7 for this fix I just monkey patched on my side with `yarn patch` however it would be nice if this one would be merged :)

kevich:
@Andarist I've [tried to](https://github.com/kevich/changesets/commit/8c3d226837b485da9cf2e382cbd5905487748e47) briefly filter packages in `assembleReleasePlan` function.

But with this filtering approach there are some failed other tests with `Dependency map is incorrect` error.

So I'm not sure that it's worth doing filtering, rather than suggested solution.

Anyway, @Andarist what is the current status of this PR? It seems like this case bothering pretty big amount of users.

equinusocio:
I have the same issue with a private package, is there a reason why this pr isn't merged after 2 years?

melvinmcrn:
I would love to have this PR merged if everything is alright since it could fix a problem for me and some other devs (e.g. https://github.com/changesets/changesets/issues/835)

cc. @AntoineKM @benmccann

bluwy:
Perhaps related: https://github.com/changesets/changesets/pull/705

equinusocio:
I gues is another dead-support package. After 2 years you can close this pr.

fredojose (approved):

melvinmcrn:
Hey @AntoineKM @ndaba1 @fredojose, apologies for the tag! Just wondering when we might expect this PR to be merged?

guesant:
is changesets dead?

PR #852, opened 2022-06-27
Fixed commit type defined in changeset config

caohuilin:


packages/config/schema.json
```diff
@@ -50,7 +50,26 @@
...
+        {
+          "type": "array",
+          "items": [
+            {
+              "type": "string",
+              "title": "Commit Generator Path"
+            },
+            { "title": "Commit Generator Options" }
+          ],
+          "minItems": 2
```
Andarist, line 63: I think that this should be:
```suggestion
          "minItems": 1,
          "maxItems": 2,
```

The same change could be done to the `changelog` option.

.changeset/selfish-buckets-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
+Fixed commit type defined in changeset config.
```
Andarist, line 5: Fixed the declared JSON schema type for the `commit` config option.

PR #853, opened 2022-06-27
Reduce the amount of requests sent to npm

### LINKED ISSUE #427: Weird publish output while publishing a new version?
## Affected Packages

cli?

## Problem

I'm using a custom script for setting our pre-release versions, and I noticed an error *sometimes* from NPM. The error I'm getting is:

```
error an error occurred while publishing @graphql-codegen/flow: E403 403 Forbidden - PUT https://registry.npmjs.org/@graphql-codegen%2fflow - You cannot publish over the previously published versions: 1.17.8-alpha-0f93aa04.0. 
🦋  error a package version that is forbidden by your security policy.
```

I scanned the log, and noticed that `changesets` already checks that, and it seems valid:

```
// ...
🦋  info npm info @graphql-codegen/flow
// ...
🦋  info @graphql-codegen/flow is being published because our local version (1.17.8-alpha-0f93aa04.0) has not been published on npm
// ...
error an error occurred while publishing @graphql-codegen/flow: E403 403 Forbidden - PUT https://registry.npmjs.org/@graphql-codegen%2fflow - You cannot publish over the previously published versions: 1.17.8-alpha-0f93aa04.0. 
🦋  error a package version that is forbidden by your security policy.
// ...
```

Maybe I'm missing something - but it seems like `changesets` looks for that package in that specific version, and it can't find it in NPM. But then when it tries to publish it, it fails because it already exists. 

I checked and the package does exists on NPM, around the same time the pipeline was executed. I can only assume that the publish went fine, and the NPM error reflects something else? 

## Proposed solution

No idea. Back then we had similar issues with E403 from NPM, and the messages was never informative enough, or mistaken. @ardatan suggested once that publishing many packages in short period of time causes issues, and it's better to limit the amount of `publish` command and not run all packages in parallel. We tried to limit the amount of publishing to maximum of 5 in parallel, and that seems to resolve the issue. 

@Andarist

dotansimha:
Closes https://github.com/changesets/changesets/issues/427 

This should lower the number of requests sent to GitHub. With the default config we have today, it seems like large projects like https://github.com/dotansimha/graphql-code-generator are having trouble get this working. 

Also, GitHub returns 403 when you sent too many requests (instead of 429) which makes it super difficult to debug - so the extra few `ms` to the publish commands are worth it.

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -11,8 +11,8 @@ import isCI from "is-ci";
 import { TwoFactorState } from "../../utils/types";
 import { getLastJsonObjectFromString } from "../../utils/getLastJsonObjectFromString";
 
-const npmRequestLimit = pLimit(40);
-const npmPublishLimit = pLimit(10);
+const npmRequestLimit = pLimit(20);
+const npmPublishLimit = pLimit(5);
 
 function jsonParse(input: string) {
   try {
--- a/.changeset/tall-maps-call.md
+++ b/.changeset/tall-maps-call.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Reduce the default amount of concurrent requests sent to `npm`. This should address unexpected 403 issues on large monorepos.


packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -11,8 +11,8 @@ import isCI from "is-ci";
 import { TwoFactorState } from "../../utils/types";
 import { getLastJsonObjectFromString } from "../../utils/getLastJsonObjectFromString";
 
-const npmRequestLimit = pLimit(40);
-const npmPublishLimit = pLimit(10);
+const npmRequestLimit = pLimit(20);
+const npmPublishLimit = pLimit(5);
```
Andarist, line 15: How sure are we that we are hitting rate limits with the current values?

Andarist:
For my own reference - the repo in question has 58 public packages.

packages/cli/src/commands/publish/npm-utils.ts
Andarist: Some old blog post suggests that 429 should be correctly returned in the case of rate limit: https://blog.npmjs.org/post/164799520460/api-rate-limiting-rolling-out.html

dotansimha: Yeah, I was thinking the same thing, but in reality, when a token reaches the rate-limit, it's blocked and you get 403 (see the issue I mentioned, it's for a ~year ago but still relevant...) 
In our libraries that has many packages, we have to use a patch for `changesets` in order to make publishing fully functional and less flakey: 
https://github.com/dotansimha/graphql-code-generator/blob/master/patches/%40changesets%2Bcli%2B2.22.0.patch#L10

dotansimha: Oh, I see what you mean. The alpha release flow was not related to it- I feel like it was a symptom. Also with regular publishing of many packages - at some point, some packages are published, and some got 403 (regardless of what version they have). 
We had that a few times with graphql-code-generator repo (again, not related to the canary/alpha patches we have there - just with regular publishing).
Our solution eventually was to add that patch and reduce the number of NPM requests and it seems to work fine since then (for ~1y now).
Since then, we applied that to more large monorepos like `graphql-tools` and `graphql-mesh`, because otherwise, we get a random 403 when trying to publish a big changeset that changes many packages :(

> Do you recall if the same was true for all of failed packages or if some of them were published successfully (despite returning an error) or if some were published ok but some weren't?

Well, I'm not 100% sure. I think we got some published (no errors), some published (and we had 403 error), and some that was failing completely due to NPM error. All of those happened only when we tried to publish many packages at once...

dotansimha: Btw, should we consider just allowing libraries to configure these limits using env vars? (and keep the defaults as-is?). I feel like there is no concrete number that works well. NPM seems to have their own opinions on when you can publish many packages at once

I-Want-ToBelieve:
```
> yakite@0.1.0 release /home/runner/work/yakite/yakite
> changeset publish

🦋  info npm info krohnkite-core
🦋  info npm info yakite-bridge
🦋  info npm info yakite-config
🦋  info npm info yakite-daemon
🦋  info npm info yakite-message
🦋  info npm info yakite-yabai
🦋  warn Received 404 for npm info "yakite-yabai"
🦋  warn Received 404 for npm info "yakite-bridge"
🦋  warn krohnkite-core is not being published because version 0.1.1 is already published on npm
🦋  warn yakite-config is not being published because version 0.1.1 is already published on npm
🦋  warn yakite-daemon is not being published because version 0.1.1 is already published on npm
🦋  info yakite-bridge is being published because our local version (0.1.1) has not been published on npm
🦋  info yakite-yabai is being published because our local version (0.1.1) has not been published on npm
🦋  warn yakite-message is not being published because version 0.1.1 is already published on npm
🦋  info Publishing "yakite-bridge" at "0.1.1"
🦋  info Publishing "yakite-yabai" at "0.1.1"
🦋  error an error occurred while publishing yakite-yabai: E429 429 Too Many Requests - PUT https://registry.npmjs.org/yakite-yabai - Too Many Requests 
🦋  error npm notice Publishing to https://registry.npmjs.org/ with tag latest and public access
🦋  error npm ERR! code E429
🦋  error npm ERR! 429 Too Many Requests - PUT https://registry.npmjs.org/yakite-yabai - Too Many Requests
🦋  error 
🦋  error npm ERR! A complete log of this run can be found in: /home/runner/.npm/_logs/2023-11-24T05_50_40_929Z-debug-0.log
🦋  error 
🦋  error an error occurred while publishing yakite-bridge: E429 429 Too Many Requests - PUT https://registry.npmjs.org/yakite-bridge - Too Many Requests 
🦋  error npm notice Publishing to https://registry.npmjs.org/ with tag latest and public access
🦋  error npm ERR! code E429
🦋  error npm ERR! 429 Too Many Requests - PUT https://registry.npmjs.org/yakite-bridge - Too Many Requests
🦋  error 
🦋  error npm ERR! A complete log of this run can be found in: /home/runner/.npm/_logs/2023-11-24T05_50_40_968Z-debug-0.log
🦋  error 
🦋  yakite-bridge@0.1.1
🦋  yakite-yabai@0.1.1
 ELIFECYCLE  Command failed with exit code 1.
```

I-Want-ToBelieve:
Click Settings on the corresponding package page of the npm site, invite another account as a maintainer, then log in to another account, log in to its corresponding email address, check the email sent by npm, accept the invitation, 

And then generate a token on github. Replace the original one in the settings, and then you can publish it successfully. If anyone encounters the same problem, you can try this temporary solution.

I-Want-ToBelieve:
Now a new situation has arisen. 
Two github releases have been released in changesets. 

The four previously successful npm packages were not included because the two failed npm packages cut off the subsequent process.

beeequeue:
@dotansimha is this still a problem for you? should we look into making the concurrency configurable?

bertho-zero:
It's still a problem for me, I recently ran into this issue again.
I set the limits to 10 and 1 with a patch.

Issue #855, opened 2022-06-28
Support for canary releases based on changesets? 


dotansimha:
Hi there! 

We at The Guild are working with `changesets` extensively. One of the solution we implement on top of the current implementation of `changesets` is the ability to easily connect canary/alpha releases, based on active `changesets`. 

To do that, we are applying a minimal `patch` file to changesets code, in order to allow flexibility in the release flow. 

I'm sharing an overview of the technical changes we are doing today, and **I can also create a PR if this makes sense**. I was thinking of a workflow of `changesets version canary` in the CLI.

### The `canary` / `alpha` workflows

1. To release a canary based on changesets, we are scanning and reading all relevant changesets that are currently available in the repo.
2. Then, we run `assembleReleasePlan` and builds the complete plan as changesets plan to release it (on the next release planned)
3. Then, we modify the version to be `{plannedVersion}-canary.{COMMIT_ID}` 
4. Run `applyReleasePlan` to apply the change temporarily 
5. Run changesets publishing script with `--tag canary`  

Here's the script we are using: https://github.com/B2o5T/graphql-eslint/blob/master/scripts/canary-release.ts 

This way - changesets are used as the source of truth to the packages that actually was changed and needs to be released - and we apply an empheral change to the versioning before publishing it to npm. 

### Allowing non-ranged publishing

As part of the canary release workflows, we want to make sure to get rid of all possible ranges (`^` / `~`) in the internal packages dependencies, to make sure to get the exact versions of the canary changes. 

To do that, we added a new argument to `versionPackage` function, called `ignoreRange`

Reference: https://github.com/B2o5T/graphql-eslint/blob/master/patches/%40changesets%2Bapply-release-plan%2B6.0.0.patch#L19

mitchellhamilton:
This sounds like [snapshot releases](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md) (or at least the intent behind them).

> {plannedVersion}-canary.{COMMIT_ID}

Using `plannedVersion` instead of `0.0.0` can be done with `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH.useCalculatedVersionForSnapshots` though you can't currently use a commit.

@Andarist I'm kinda thinking we should have a "snapshot version template" config option,

So you could make have some variables that we provide like:
- `tag`
- `commit`
- `timestamp`
- `plannedVersion` (v not sure about this name)

[...]

PR #857, opened 2022-06-29
bugfix: ignore ranges when snapshot versioning is performed (use exact version)

dotansimha:
Related: https://github.com/changesets/changesets/issues/855

### DIFF
--- a/.changeset/stupid-jars-rest.md
+++ b/.changeset/stupid-jars-rest.md
@@ -1,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
+---
+
+bugfix: ignore ranges when snapshot versioning is performed (use exact version)
--- a/.changeset/stupid-jars-rest.md
+++ b/.changeset/stupid-jars-rest.md
@@ -1,5 +1,6 @@
 ---
+"@changesets/cli": patch
 "@changesets/apply-release-plan": patch
 ---
 
-bugfix: ignore ranges when snapshot versioning is performed (use exact version)
+Fixed an issue with dependency ranges still using pre-existing range modifiers instead of fixed package versions when performing a snapshot release. This ensures that installs of snapshot versions are always reproducible.
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -74,7 +74,11 @@ async function testSetup(
   fixtureName: string,
   releasePlan: ReleasePlan,
   config?: Config,
-  setupFunc?: (tempDir: string) => Promise<any>
+  setupFunc?: (tempDir: string) => Promise<any>,
+  snapshotConfig?: {
+    snapshot: string | undefined;
+    useCalculatedVersionForSnapshots: boolean;
+  }
 ) {
   if (!config) {
     config = {
@@ -89,7 +93,8 @@ async function testSetup(
       ___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH: {
         onlyUpdatePeerDependentsWhenOutOfRange: false,
         updateInternalDependents: "out-of-range",
-        useCalculatedVersionForSnapshots: false
+        useCalculatedVersionForSnapshots:
+          snapshotConfig?.useCalculatedVersionForSnapshots ?? false
       }
     };
   }
@@ -108,7 +113,8 @@ async function testSetup(
     changedFiles: await applyReleasePlan(
       releasePlan,
       await getPackages(tempDir),
-      config
+      config,
+      snapshotConfig?.snapshot
     ),
     tempDir
   };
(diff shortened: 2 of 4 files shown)


packages/apply-release-plan/src/version-package.ts
```diff
@@ -29,7 +29,8 @@ export default function versionPackage(
     updateInternalDependencies: "patch" | "minor";
     onlyUpdatePeerDependentsWhenOutOfRange: boolean;
     bumpVersionsWithWorkspaceProtocolOnly?: boolean;
-  }
+  },
+  snapshot?: string | boolean | undefined
```
Andarist, line 33: nit: this probably could be added to the previous argument, any particular reason why a new one has been added?

packages/apply-release-plan/src/version-package.ts
```diff
@@ -91,8 +92,9 @@ export default function versionPackage(
           // leaving those as is would leave the package in a non-installable state (wrong dep versions would get installed)
           semver.prerelease(version) !== null
         ) {
-          let rangeType = getVersionRangeType(depCurrentVersion);
-          let newNewRange = `${rangeType}${version}`;
+          let newNewRange = snapshot
+            ? version
+            : `${getVersionRangeType(depCurrentVersion)}${version}`;
```
Andarist, line 97: This is the core of the change and LGTM

packages/apply-release-plan/src/index.test.ts
```diff
@@ -74,7 +74,11 @@ async function testSetup(
   fixtureName: string,
   releasePlan: ReleasePlan,
   config?: Config,
-  setupFunc?: (tempDir: string) => Promise<any>
+  setupFunc?: (tempDir: string) => Promise<any>,
+  snapshotConfig?: {
+    snapshot: string | undefined;
+    useCalculatedVersionForSnapshots: boolean;
```
Andarist, line 80: isn't this particular argument (`useCalculatedVersionForSnapshots`) sort of already part of the `config`?

packages/apply-release-plan/src/index.test.ts
```diff
@@ -74,7 +74,11 @@ async function testSetup(
   fixtureName: string,
   releasePlan: ReleasePlan,
   config?: Config,
-  setupFunc?: (tempDir: string) => Promise<any>
+  setupFunc?: (tempDir: string) => Promise<any>,
+  snapshotConfig?: {
```
Andarist, line 78: nit: i would move `setupFunc` to the last position to match closer the argument order in other places, like here:
https://github.com/changesets/changesets/blob/cc7807227a357c13c27830eeaccd4522a64b78ac/packages/apply-release-plan/src/index.ts#L62-L63

packages/apply-release-plan/src/version-package.ts
dotansimha: sure, done!

packages/apply-release-plan/src/index.test.ts
dotansimha: yeah, makes sense. done!

packages/apply-release-plan/src/index.test.ts
dotansimha: Yeah, you are right :) done!

dotansimha:
Thanks for the code review @Andarist . I pushed fixes for all notes :)

.changeset/stupid-jars-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
```
Andarist, line 2:
```suggestion
"@changesets/cli": patch
"@changesets/apply-release-plan": patch
```

.changeset/stupid-jars-rest.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+bugfix: ignore ranges when snapshot versioning is performed (use exact version)
```
Andarist, line 5:
```suggestion
Fixed an issue with dependency ranges still using pre-existing range modifiers instead of fixed package versions when performing a snapshot release. This ensures that installs of snapshot versions are always reproducible.
```

packages/apply-release-plan/src/index.test.ts
```diff
@@ -653,6 +655,47 @@ describe("apply release plan", () => {
...
+            id: "some-id",
+            releases: [{ name: "pkg-b", type: "minor" }],
+            summary: "a very useful summary"
+          }
+        ],
+        [
+          {
+            changesets: ["some-id"],
+            name: "pkg-b",
+            newVersion: "1.1.0",
```
Andarist, line 671: While working on https://github.com/changesets/changesets/pull/1008 I've noticed that this test case doesn't reflect how the full thing works in practice. Usually the `newVersion` should already have the snapshot modifier here. This would be computed here when assembling the release plan:
https://github.com/changesets/changesets/blob/4acaff7b5b32cc1d231d3c3eb61958420ba9b112/packages/assemble-release-plan/src/index.ts#L248-L255

This test case still tests that the range modifier gets dropped when `snapshot` is provided:
https://github.com/changesets/changesets/blob/4acaff7b5b32cc1d231d3c3eb61958420ba9b112/packages/apply-release-plan/src/version-package.ts#L96-L98

However... this should, sort of, be an impossible setup. Should we validate this anyhow? Should the test case be tweaked to reflect the real usage better?

PR #858, opened 2022-06-29
New config flag: `snapshot.prereleaseTemplate`, and make `snapshot` feature stable

dotansimha:
Closes https://github.com/changesets/changesets/issues/855 and  https://github.com/changesets/changesets/pull/830

## Stable `snapshot` feature

This PR changes the stability state of the snapshot feature. You can now use and declare the configuration under the root of `config.json`.

Before:

```json
{
    "useCalculatedVersionForSnapshots": true
  }
}
```

After:

```json
{
    "snapshot": {
        "useCalculatedVersion": true
    }
}
```

## New `snapshotPreidTemplate`

Added a new config option: `snapshotPreidTemplate` for customizing the way snapshot release numbers are being composed. 

### Available placeholders

You can use the following placeholders for customizing the snapshot release version:

- `{tag}` - name of the tag, as specified in `--snapshot something`
- `{commit}` - the Git commit ID
- `{timestamp}` - Unix timestamp of the time of the release 
- `{datetime}` - date and time of the release (14 characters, for example: `20211213000730`)

> Note: if you are using `--snapshot` with empty tag name, you cannot use `{tag}` as placeholder - this will result in error.

### Integration with `useCalculatedVersion`

You can still use and pass `snapshot.useCalculatedVersion: boolean` if you wish to have the snapshot releases to based on the planned release of changesets, instead of `0.0.0`.

### Legacy mode

If you are not specifying `snapshot.prereleaseTemplate`, the defualt behaviour will fallback to use the following template: `{tag}-{datetime}`, and in cases where tag is empty (`--snapshot` with no tag name), it will use `{datetime}` only.


Andarist:
re `snapshotVersionTemplate` - I don't think that the whole version field should be configurable like this. It would allow people to generate invalid versions etc. IMHO only the suffix should be configurable - this way we wouldn't have to validate what got generated etc

It should still be possible to switch to using a planned version using a boolean config or something.

> This PR also drops the expetimental useCalculatedVersionForSnapshots and use snapshotVersionTemplate instead.

I don't think we want to drop it just yet - it would be better to raise a warning when it's used. Even though it was an experimental option it is definitely used by some projects.

> {tag} - the name of the tag. Please note that in the case of using --snapshot (without tag name), this will be empty. Otherwise, the value will be replaced with -{tag} to avoid -- in the case of an empty tag name.

I didn't look into the code itself but it seems like those are some funky rules here. Wouldn't it be a safe assumption that people are always using `--snapshot tag` or `--snapshot` but they are never "mixing" it? Could we just throw when `${tag}` is encountered in a pattern but it has not been provided? Could we throw when `${tag}` is not encountered but a custom tag has been provided?

---

I also think that we feel somewhat confident about this feature - it's only a matter of bikeshedding the details and stuff so we should introduce this as a stable option. We don't have to introduce each new thing within the experimental namespace, that is reserved for things that seem useful but for which not everything is clear, has been hashed out, etc

jakubmazanec:
We use `useCalculatedVersionForSnapshots: true` and TBH I think that should be the default, and definitely **not removed**; I very much like using snapshots for prereleases (without committing changed package.jsons and changelogs).

Andarist:
> We use useCalculatedVersionForSnapshots: true and TBH I think that should be the default, and definitely not removed; I very much like using snapshots for prereleases (without committing changed package.jsons and changelogs).

Note that this PR doesn't remove the ability to do this - it just accomplishes it in a different way. I agree though that we shouldn't immediately remove the old way of doing this, to allow people to migrate things at their own pace.

dotansimha:
> re `snapshotVersionTemplate` - I don't think that the whole version field should be configurable like this. It would allow people to generate invalid versions etc. IMHO only the suffix should be configurable - this way we wouldn't have to validate what got generated etc

We do have checks for valid semver at a later step if I remember correctly, but I see your point. If only suffix is configurable, I guess it's only a matter of having `useCalculatedVersionForSnapshots: boolean` and `snapshotsSuffix: 'commitId' | 'timestamp'`? With the `tag` being specified anyway through the CLI. 

> Note that this PR doesn't remove the ability to do this - it just accomplishes it in a different way. I agree though that we shouldn't immediately remove the old way of doing this, to allow people to migrate things at their own pace.

With this note added, I guess the solution I suggested above might be ideal? (avoids the semver issues, supports for customizing the snapshot version, and doesn't deprecate/change existing config flags?) 

If that's the direction you prefer to go, please let me know and I'll update this PR accordingly :)

jakubmazanec:
> > Note that this PR doesn't remove the ability to do this - it just accomplishes it in a different way. I agree though that we shouldn't immediately remove the old way of doing this, to allow people to migrate things at their own pace.
> 
> With this note added, I guess the solution I suggested above might be ideal? (avoids the semver issues, supports for customizing the snapshot version, and doesn't deprecate/change existing config flags?)
> 
> If that's the direction you prefer to go, please let me know and I'll update this PR accordingly :)

I was trying to do something similar in this PR #830 and i realized that - at least for my use case - I would not only like to use simple "templates" with tokens (like in this PR), but also have a new CLI flag allowing to specify snapshot template for that specific run of `changset version`.

Anyway, I'm happy to help.

Andarist:
> If only suffix is configurable, I guess it's only a matter of having useCalculatedVersionForSnapshots: boolean and snapshotsSuffix: 'commitId' | 'timestamp'? With the tag being specified anyway through the CLI.

People have different expectations as to where the tag should be put, some could also want to include both `commitId` and `timestamp` in the template. So maybe a template is still desirable here? I'm not entirely sure about it though - WDYT? cc @mitchellhamilton 

You can also actually attach "random" build metadata (it can be just anything) to your `package.json#version` after a plus sign so by using a template we would allow people to use this. It's not a very popular thing though - I only learned about this possibility a couple of months back. I've seen it being used in the wild though:
https://unpkg.com/browse/@react-aria/toggle@3.0.0-nightly.1651/package.json#L3

>  would not only like to use simple "templates" with tokens (like in this PR), but also have a new CLI flag allowing to specify snapshot template for that specific run of changset version.

I'm curious - what's your use case for configuring a template for a specific `changeset version` run?

mitchellhamilton:
> IMHO only the suffix should be configurable - this way we wouldn't have to validate what got generated etc

I don't feel that strongly about having the template be the entire version but we would still have to validate the template anyway since you could put in characters that aren't valid as a suffix. We could trivially validate it once by replacing the placeholders with some static values and checking if it's a valid version so what's the harm in making the template be for the whole version?

> People have different expectations as to where the tag should be put, some could also want to include both `commitId` and `timestamp` in the template. So maybe a template is still desirable here? I'm not entirely sure about it though - WDYT?

Yeah, this is why I suggested a template. Having the order of things, the separator and what things should appear in a bunch of separate options just seems confusing, having a template makes it easy to see what it'll actually be.

mitchellhamilton:
If the template is only the suffix, you could make it an empty string and get versions like `0.0.0-` or make it some static thing and you'd have a problem if you tried to publish twice, so it's the same, no?

Andarist:
Hm, sure - but with a static suffix, we won't attempt to publish the same version multiple times. It won't do anything useful, it will just skip over the publish step. If we give complete control over the version to users then it will be "easy" to publish `1.0.0` (no prerelease suffix at all!) when your latest version is already `2.3.4` (or whatever). 

Perhaps this is just a territory of "please don't do stupid things" but it seems to me that if we can avoid that then it would be better. The drawback is though that we'd complicate our options API a little bit.

I'll leave the decision about this to you - those are just my concerns, I'm not giving a veto to the single template idea.

mitchellhamilton:
> If we give complete control over the version to users then it will be "easy" to publish 1.0.0 (no prerelease suffix at all!) when your latest version is already 2.3.4 (or whatever).

Yeah, that's fair, happy to have just a template for the suffix

dotansimha:
> Yeah, that's fair, happy to have just a template for the suffix

Cool, so we'll go with: 

```
useCalculatedVersionForSnapshots: boolean (default: false)
snapshotVersionSuffix: 'timestamp' | 'commit' (default: 'timestamp')
```

So this way, the actual version part is controlled internally (can be customized using `useCalculatedVersionForSnapshots`), the `tag` is controller using the CLI option, and the suffix can be either a timestamp of a commit ID. 

Makes sense? @Andarist @mitchellhamilton  Please let me know so I'll change this PR accordingly. thanks!

Andarist:
I think we've agreed on using the template for the suffix itself and I would go with that here, otherwise, we'll end up with people wanting to customize this further.

dotansimha:
> I think we've agreed on using the template for the suffix itself and I would go with that here, otherwise, we'll end up with people wanting to customize this further.

Got it, thanks! I'm working on changing this PR :) Will update soon

jakubmazanec:
> So this way, the actual version part is controlled internally (can be customized using `useCalculatedVersionForSnapshots`), the `tag` is controller using the CLI option, and the suffix can be either a timestamp of a commit ID.

Does that mean I won't be able to use both a timestamp **and** a tag (from the CLI option) at the same time? That would be backwards incompatible with how it currently works.

dotansimha:
> > So this way, the actual version part is controlled internally (can be customized using `useCalculatedVersionForSnapshots`), the `tag` is controller using the CLI option, and the suffix can be either a timestamp of a commit ID.
> 
> Does that mean I won't be able to use both a timestamp **and** a tag (from the CLI option) at the same time? That would be backwards incompatible with how it currently works.

Nope, it means that you can still use `--snapshot` for empty tag, or `--snapshot alpha` for a named tag. 
The suffix will be either timestamp or commit ID. 

Here's an example of the config set and their result, does it makes sense? @jakubmazanec 

```
cli: --snapshot
output: 0.0.0-1656838548676
```

```
cli: --snapshot alpha
output: 0.0.0-alpha.1656838548676
```

```
cli: --snapshot
useCalculatedVersionForSnapshots: true
output: 1.2.3-1656838548676
```

```
cli: --snapshot cabary
useCalculatedVersionForSnapshots: true
output: 1.2.3-canary.1656838548676
```

```
cli: --snapshot canary
useCalculatedVersionForSnapshots: true
snapshotVersionSuffix: commit
output: 1.2.3-canary.abcdef
```

```
cli: --snapshot
snapshotVersionSuffix: commit
output: 0.0.0-abcdef
```

jakubmazanec:
@dotansimha Yes, thank you for the explanation. But I would welcome if I could 1) use both timestamp and commit id at the same time, and 2) sometimes reverse location of timestamp/commit and tag. Also, you use `.` as preid separator, while currently it's `-`, so that needs to be configurable too. I don't think all that can be configured without some simple templating, as mentioned earlier in this discussion (otherwise we could just use #830 right away).

Anyway, I was thinking how to then solve the situations like 1) when the tag from CLI option is empty, resulting in invalid preid (e.g. `0.0.0-.timestamp`), or 2) different snapshot suffix for different "types" of prereleases (e.g. nightly release with only commit id vs feature branch release with only timestamp), without including some sort of if/then/else logic to the template. I'm still coming back to this solution: new CLI option for overriding the default `snapshotPreidTemplate` option from config, e.g.:

```
snapshotPreidTemplate: {{tag}}.{{timestamp}} 
cli: --snapshot foobar
result: 0.0.0-foobar.timestamp

snapshotPreidTemplate: {{tag}}.{{timestamp}} 
cli: --snapshot 
result: ERROR

snapshotPreidTemplate: {{tag}}.{{timestamp}} 
cli: --snapshot --snapshotPreidTemplate {{timestamp}} 
result: 0.0.0-timestamp
```

dotansimha:
Thanks for the feedback @jakubmazanec ! 

> 1) use both timestamp and commit id at the same time

This might be valid use-case (I never tried to mix the both, but I can understand where it's coming from) - @Andarist @mitchellhamilton what do you think? 

>  2) sometimes reverse location of timestamp/commit and tag.

Yeah, there is no clear spec in semver about that - you can put whatever you want after the `-`. I usually do `{tag}.{commit}`.

Regarding next steps - I'm fine with both solutions TBH - regarding empty `tag` - I think we should either choose to prevent it (does it make sense to publish something like `1.2.3-test` as `tag=alpha` on NPM? It's allowed but kinda weird), or keep it as-is (where the code treats it as part of the suffix). Today the CLI can handle empty tag because there is a default timestamp added, so you get a valid `suffix` always (even when it's empty).

@Andarist @mitchellhamilton @jakubmazanec I guess now it's a matter of a decision? Are we going with a new config param for the suffix (timestamp/commit), or with a template for the entire suffix (tag + timestamp/commit/anything)?

Andarist:
> or with a template for the entire suffix (tag + timestamp/commit/anything)?

This

jakubmazanec:
I vote, of course, for template for the entire suffix (with CLI option to override it). I think it will provide the greatest number of possible prerelease version formats and thus satisfying most users. We can then check if the preid is nonempty (so no `0.0.0-` version) and then use [Semver regexp](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) to make sure there are no illegal characters. The rest is users' responsibility - if they want to publish version like `1.2.3-test`, it's their choice.

> > 1. use both timestamp and commit id at the same time
> 
> This might be valid use-case (I never tried to mix the both, but I can understand where it's coming from)

Timestamp is useful because if placed at the beginning of the preid, it solves the [prerelease sorting issue](https://semver.org/#spec-item-11). And commit id identifies the commit, that's obviously useful for testing, debugging and stuff.

Andarist:
I agree with @jakubmazanec. It seems the most straightforward to do this. We can't have a single default without some special logic around it anyway, so best to treat the lack of a template in a special way. 

I will also quote what I've written a couple of days ago here as I feel it's still relevant:
> Wouldn't it be a safe assumption that people are always using --snapshot tag or --snapshot but they are never "mixing" it? Could we just throw when {tag} is encountered in a pattern but it has not been provided? Could we throw when {tag} is not encountered but a custom tag has been provided?

dotansimha:
> Wouldn't it be a safe assumption that people are always using --snapshot tag or --snapshot but they are never "mixing" it? Could we just throw when {tag} is encountered in a pattern but it has not been provided? Could we throw when {tag} is not encountered but a custom tag has been provided?

Yeah, that's a fair assumption. I guess it's never mixed. We can throw, but for the default, as you said, we'll have to go with `undefined` and fallback to the current behavior. 

I'm working on an improved version of this PR, will update soon! Thanks :)

dotansimha:
@Andarist @jakubmazanec @mitchellhamilton I implemented the `snapshotPreidTemplate` along with testing. This seems to work pretty well :) Review/feedback is welcome! :)

.changeset/strong-geckos-divide.md
```diff
@@ -0,0 +1,27 @@
...
+
+> Note: if you are using `--snapshot` with empty tag name, you cannot use `{tag}` as placeholder - this will result in error.
+
+### Integration with `useCalculatedVersionForSnapshots`
+
+You can still use and pass `useCalculatedVersionForSnapshots: boolean` if you wish to have the snapshot releases to based on the planned release of changesets, instead of `0.0.0`.
+
+### Legacy mode
+
+If you are not specifying `snapshotPreidTemplate`, the defualt behaviour will fallback to use the following template: `{tag}-{datetime}`, and in cases where tag is empty (`--snapshot` with no tag name), it will use `{datetime}` only.
```
william-will-angi, line 27: These are good docs - I think it would be valuable to include them somewhere outside of the changelog. I think it would be valuable to start a doc that describes experimental options.

william-will-angi, line 27:
```suggestion
If you are not specifying `snapshotPreidTemplate`, the default behaviour will fallback to use the following template: `{tag}-{datetime}`, and in cases where tag is empty (`--snapshot` with no tag name), it will use `{datetime}` only.
```

.changeset/strong-geckos-divide.md
```diff
@@ -0,0 +1,27 @@
+---
+"@changesets/assemble-release-plan": minor
+"@changesets/cli": minor
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Added a new config option: `snapshotPreidTemplate` for customizing the way snapshot release numbers are being composed.
```
william-will-angi, line 8: Being nitpicky - I think this should be named just `snapshotTemplate` - I think snapshot templates could have use cases outside of just pre-releases.

.changeset/strong-geckos-divide.md
```diff
@@ -0,0 +1,27 @@
...
+- `{timestamp}` - Unix timestamp of the time of the release
+- `{datetime}` - date and time of the release (14 characters, for example: `20211213000730`)
+
+> Note: if you are using `--snapshot` with empty tag name, you cannot use `{tag}` as placeholder - this will result in error.
+
+### Integration with `useCalculatedVersionForSnapshots`
+
+You can still use and pass `useCalculatedVersionForSnapshots: boolean` if you wish to have the snapshot releases to based on the planned release of changesets, instead of `0.0.0`.
+
+### Legacy mode
```
william-will-angi, line 25: It feels like calling this `Legacy mode` is a misnomer (especially given this is an experimental feature) - this is just the default behavior/template from what I can tell

william-will-angi (commented):
Take my comments with a grain of salt because I'm not a maintainer of this repo. I saw this PR and was just curious 😄

packages/apply-release-plan/src/index.test.ts
```diff
@@ -52,7 +52,8 @@ class FakeReleasePlan {
       ___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH: {
         onlyUpdatePeerDependentsWhenOutOfRange: false,
         updateInternalDependents: "out-of-range",
-        useCalculatedVersionForSnapshots: false
+        useCalculatedVersionForSnapshots: false,
+        snapshotPreidTemplate: null
```
Andarist, line 56: In general, I think that those are options that have proven to be needed and that there is a very low risk of us wanting to take another stab at the problem in the near future. We could promote this stuff to stable options.

One open question - should those two be grouped in the config under a single key or should both be separate? I would probably lean towards the former.

packages/assemble-release-plan/src/index.ts
```diff
@@ -27,51 +27,63 @@ function getPreVersion(version: string) {
...
+  // and the resulting version needs to be composed without a tag.
+  if (!snapshotParameters.preidTemplate) {
+    const legacySuffix = [snapshotParameters.tag, snapshotParameters.datetime]
+      .filter(Boolean)
+      .join("-");
 
-  let dateAndTime = new Date()
-    .toISOString()
-    .replace(/\.\d{3}Z$/, "")
-    .replace(/[^\d]/g, "");
```
Andarist, line 38: does this mean that previously we could end with different snapshot versions when releasing a set of packages? 😱

packages/assemble-release-plan/src/index.ts
```diff
@@ -80,7 +92,15 @@ function assembleReleasePlan(
   config: Config,
   // intentionally not using an optional parameter here so the result of `readPreState` has to be passed in here
   preState: PreState | undefined,
-  snapshot?: string | boolean
+  // snapshot: undefined            ->  not using snaphot
+  // snapshot: { tag: undefined }   ->  --snapshot (empty tag)
+  // snapsgot: { tag: "canary" }    ->  --snapshot canary
```
Andarist, line 97:
```suggestion
  // snapshot: { tag: "canary" }    ->  --snapshot canary
```

packages/assemble-release-plan/src/index.ts
```diff
@@ -27,51 +27,63 @@ function getPreVersion(version: string) {
...
 
-  let dateAndTime = new Date()
-    .toISOString()
-    .replace(/\.\d{3}Z$/, "")
-    .replace(/[^\d]/g, "");
-  let tag = "";
+    return [baseVersion, legacySuffix].join("-");
+  } else {
+    const composedSuffix = snapshotParameters.preidTemplate
+      .replace("{timestamp}", snapshotParameters.timestamp)
```
Andarist, line 61: while it doesn't make a lot of sense to use the same placeholder multiple times - we probably should replace all of its occurrences. Since `replaceAll` might not be available in the node version running Changesets - we have to resort to using a regex with `g` modifier here.

packages/cli/src/commands/version/version.test.ts
```diff
@@ -13,6 +13,16 @@ import pre from "../pre";
 import version from "./index";
 import humanId from "human-id";
 
+global.Date = class MockDate extends Date {
```
Andarist, line 16: since this is patching the global env I would prefer to keep it per test case, and let the unrelated tests run as normal, avoiding potential weird problems

.changeset/strong-geckos-divide.md
Andarist: Agreed - this should be moved to the docs, and we should only mention this feature briefly in the changeset file itself. We could mention that you can learn more about this in the docs though.

.changeset/strong-geckos-divide.md
Andarist: what those use cases could be though? while somebody might use it for something else - this part/component of the version string is described by SemVer as "pre-release": https://semver.org/#spec-item-9

.changeset/strong-geckos-divide.md
Andarist: Agree, this is probably here to stay as the default - the template is only meant to allow more control over the format. I would refrain from calling this "legacy", both in the changesets/docs and in the code.

.changeset/strong-geckos-divide.md
william-will-angi: The only use case I was thinking was when you want to publish changes to test, but not necessarily to be communicated to consumers as a pre-release. 

Rethinking this, I think the current naming is good. The use case I described is probably best handled by an empty template.

packages/assemble-release-plan/src/index.ts
```diff
@@ -80,7 +92,15 @@ function assembleReleasePlan(
...
   preState: PreState | undefined,
-  snapshot?: string | boolean
+  // snapshot: undefined            ->  not using snaphot
+  // snapshot: { tag: undefined }   ->  --snapshot (empty tag)
+  // snapsgot: { tag: "canary" }    ->  --snapshot canary
+  snapshot?: {
+    tag: string | undefined;
+    commit: string;
+    timestamp: string;
+    datetime: string;
```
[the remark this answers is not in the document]
dotansimha: I wanted to avoid depending on `git` or `fs` in that library (because at the moment it doesn't do any fs work). 
I needed that mainly in order to get the git commit ID. 
I can move the rest of the params to be calculated inside `assemble-release-plan`, but it's still going to change the signature because the Git commit is calculated and passed from the caller. Should I take another approach here?

.changeset/strong-geckos-divide.md
dotansimha: Yeah, that makes sense. I'll move this to a new Markdown file with initial docs for all experimental options. Thanks!

.changeset/strong-geckos-divide.md
dotansimha: Sure. I changed it to `Default behavior` and moved to a new markdown file documenting all experimental features. Thanks!

packages/assemble-release-plan/src/index.ts
dotansimha: Yeah, makes perfect sense. Fixed :)

packages/apply-release-plan/src/index.test.ts
dotansimha: Well, that's a good question. I think both effects the same feature (snapshot releases) so it might make sense to group it together when we are moving it to a stable (and rethink names maybe?) 

```
"snapshot": {
  "useCalculatedVersion": boolean,
  "preidTemplate": string
}
```

I guess this can be done in a separate PR once this one is merged? (Because I guess it will require a "bigger" release of Changesets due to the new features formally added? I can help with that)

packages/assemble-release-plan/src/index.ts
```diff
@@ -27,51 +27,63 @@ function getPreVersion(version: string) {
...
+  } else {
+    const composedSuffix = snapshotParameters.preidTemplate
+      .replace("{timestamp}", snapshotParameters.timestamp)
+      .replace("{datetime}", snapshotParameters.datetime)
+      .replace("{commit}", snapshotParameters.commit)
+      .replace("{tag}", () => {
+        if (snapshotParameters.tag === undefined) {
+          throw new Error(
+            'Failed to compose snapshot version: "{tag}" placeholder is used without specifying a tag name'
+          );
```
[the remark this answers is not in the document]
dotansimha: Sure, great catch. I added it :)

dotansimha:
> Please don't forget to add CLI option to specify `snapshotPreidTemplate` (and override value from config); it is needed to change the template when you use `--snapshot` without tag.

Sure, done!

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
mitchellhamilton: This approach is right, assemble-release-plan shouldn't do git or fs stuff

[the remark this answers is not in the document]
Andarist: Sure, that I agree with - having to provide the current timestamp+datetime seems to be a little bit cumbersome from the caller's perspective though. OTOH, this is mainly an internal package so it doesn't matter that much. However, those two are highly related and one can be derived from another so I'd only accept a single date-related input (`timestamp`) and derive the rest. And to avoid a breaking change I'd at least default this value to `Date.now()` if we want to start accepting it as an actual input (which I don't care about that much, seems to be unnecessary to me though, I'm fine either way).

packages/apply-release-plan/src/index.test.ts
Andarist: The proposed config shape sounds good to me (cc @mitchellhamilton - last chance to stop us here :P).

> I guess this can be done in a separate PR once this one is merged? (Because I guess it will require a "bigger" release of Changesets due to the new features formally added? I can help with that)

I don't think this is necessarily true - we just need to be mindful about being backward-compatible but that isn't related to the actual release pace. While it might require some changes here and there I think it's best to just handle this all in a single sweep here. 

Once you make changes I can take care of making sure that this won't cause any issues after the release.

packages/assemble-release-plan/src/index.ts
```diff
@@ -27,51 +27,72 @@ function getPreVersion(version: string) {
...
 
-  let dateAndTime = new Date()
-    .toISOString()
-    .replace(/\.\d{3}Z$/, "")
-    .replace(/[^\d]/g, "");
-  let tag = "";
+    const composedSuffix = snapshotParameters.preidTemplate
+      .replace(/\{timestamp\}/g, snapshotParameters.timestamp)
+      .replace(/\{datetime\}/g, snapshotParameters.datetime)
+      .replace(/\{commit\}/g, snapshotParameters.commit)
```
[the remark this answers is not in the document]
dotansimha: Yeah, I can see your point here, but I think `--tag` has some special use-case here. 

I added validation, but it's a bit less strict - meaning that you can have all placeholders' values provided, but you don't have to use them all. 

So `tag` -> extra validation because we want to make sure the user uses it in the template when it's specified. 
`tag` + all others -> validation that if you use it, it must really have a value. 

So this makes sure that all values used in the template are really there, and validate the special case of `tag` where it must be used when specified (we can't apply this to others, because otherwise it introduces errors when `commit` is not used, and it does make sense not to use it in all times).

I also made the flags optional as you requested, so now it should be even easier to use -> you can either specify all the snapshot parameters or not, and validation makes sure that you use only what's really available.

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
dotansimha: Yeah, I agree, pass a `Date` seems ideal here? also it moves the logic of calculating the date format/timestamp format to the right place, and not just passed externally. 
I also cleared the code around this packages to make sure the commit id is actually being loaded only when snapshot release is done.

packages/apply-release-plan/src/index.test.ts
dotansimha: Sure, no problem. Waiting for @mitchellhamilton feedback here, I can do this as part of this PR if this is approved.

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
Andarist: > I added validation, but it's a bit less strict - meaning that you can have all placeholders' values provided, but you don't have to use them all.

Ok, I can roll with that.

packages/assemble-release-plan/src/index.ts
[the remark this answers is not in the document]
Andarist: > Yeah, I agree, pass a Date seems ideal here? also it moves the logic of calculating the date format/timestamp format to the right place, and not just passed externally.

`timestamp` is serializable out of the box and thus it could potentially be passed from the CLI arguments or smth without converting it to `Date`. I don't feel super strong about this though. For simplicity, I would just not accept it at all though - until a use case is presented to make it configurable I'm perfectly fine with just computing the current date-related stuff within `assemble-release-plan`, based on `Date.now()`. YAGNI.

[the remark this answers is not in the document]
dotansimha: Oh, that makes perfect sense. I agree with your take - I changed it to be just a `new Date` created within `assembleReleasePlan` and we can make it configurable if there will be an actual need. 
Thanks!

dotansimha:
@mitchellhamilton feedback on https://github.com/changesets/changesets/pull/858#discussion_r915745466 ? 

@Andarist maybe we can continue with this one and then see if we should change the config structure before the next release? (just to unblock this PR..)

packages/apply-release-plan/src/index.test.ts
mitchellhamilton: > ```
> "snapshot": {
>   "useCalculatedVersion": boolean,
>   "preidTemplate": string
> }
> ```

nesting it definitely sounds good to me. just on name bikeshedding, `preid` is probably wrong because I believe it refers to just the `beta` in `0.0.0-beta.0` whereas the option we have is about `beta.0` (https://github.com/npm/node-semver/#prerelease-identifiers)

dotansimha: Sure. `versionTemplate` is better in this case?

mitchellhamilton: Yeah, maybe or maybe `prereleaseTemplate`, `preTemplate`? @Andarist thoughts?

dotansimha: I did `prereleaseTemplate` :)

dotansimha:
@Andarist @jakubmazanec @mitchellhamilton now snapshot configuration are grouped under a single flag.  Anything else is missing to complete this PR? :)

.changeset/olive-ducks-camp.md
```diff
@@ -0,0 +1,32 @@
+---
+"@changesets/cli": patch
+"@changesets/config": patch
+---
+
+**NOTE: BREAKING CHANGE**
```
Andarist, line 6: Let's not do this as a breaking change. I think that it should be fairly easy to introduce the new `snapshot` option as a **stable** one while still supporting the old experimental flag (just translate it to a new format when loading the config). We could print a deprecation warning with a migration suggestion when we find the old experimental flag.

Andarist:
@dotansimha sorry, I was on a company retreat and didn't have much time to look into GitHub. I've only left a single comment about the breaking change stuff. Otherwise this LGTM.

.changeset/olive-ducks-camp.md
mitchellhamilton: > I think that it should be fairly easy to introduce the new snapshot option as a stable

Just want to explicitly say that I agree that making this stable seems good to me

dotansimha: Sure. I pushed changes to support the previous state (`useCalculatedVersionForSnapshots` under experimental) and also the new state that also makes the snapshot feature stable now (`snapshot.useCalculatedVersion` under the root).

dotansimha:
> @dotansimha sorry, I was on a company retreat and didn't have much time to look into GitHub. I've only left a single comment about the breaking change stuff. Otherwise this LGTM.

No problem! I pushed a fix to avoid the breaking changes and make the `snapshot` config stable :) I think we should good to go now 🚀

Andarist:
Thank you @dotansimha for bearing with me :) The PR LGTM - atm I'm in the middle of things and I would like to merge this with a clear head, gonna take another look at this soon, just double-checking everything etc. If I notice anything to tweak then I will do it myself to avoid the extra cycle of 🏓 I expect this to get merged in within the next 3 days or so.

.changeset/strong-geckos-divide.md
```diff
@@ -0,0 +1,8 @@
+---
+"@changesets/assemble-release-plan": minor
+"@changesets/cli": minor
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Added a new config option: `snapshotPrereleaseTemplate` for customizing the way snapshot release numbers are being composed.
```
Andarist, line 8:
```suggestion
Added a new config option: `snapshot.prereleaseTemplate` for customizing the way snapshot release numbers are being composed.
```

.changeset/olive-ducks-camp.md
```diff
@@ -0,0 +1,28 @@
+---
+"@changesets/cli": patch
+"@changesets/config": patch
+---
+
+Snapshot feature is now stable 🥳 All config parameters are grouped under a single property called `snapshot`.
```
Andarist, line 6:
```suggestion
A possibility to use the calculated version for snapshot releases is now stable 🥳 All snapshot-related config parameters are now grouped under a single property called `snapshot`.
```

docs/experimental-options.md
```diff
@@ -0,0 +1,54 @@
...
+
+The config flag can be used to add dependent packages to the release (if they are not already a part of it) with patch bumps.
+
+## `onlyUpdatePeerDependentsWhenOutOfRange` (type: `boolean`)
+
+Default value: `false`
+
+When set to `true`, Changesets will only bump peer dependents when `peerDependencies` are leaving the range.
+
+## `snapshot` (type: `object`)
```
Andarist, line 19: This has to be moved elsewhere as it's no longer an experimental option.

packages/config/src/index.ts
```diff
@@ -315,12 +315,42 @@ export let parse = (json: WrittenConfig, packages: Packages): Config => {
...
+    if (
+      snapshot.prereleaseTemplate !== undefined &&
+      typeof snapshot.prereleaseTemplate !== "string"
+    ) {
+      messages.push(
+        `The \`snapshot.prereleaseTemplate\` option is set as ${JSON.stringify(
+          snapshot.prereleaseTemplate,
+          null,
+          2
+        )} when the only valid values are undefined, or a template string.`
```
Andarist, line 342: types specify that `null` is also valid - we should adjust the types or this validation here. `undefined` is only implicitly allowed because of the optional (`?`) modifier on the property

packages/cli/src/index.ts
```diff
@@ -11,7 +11,7 @@ const { input, flags } = meow(
   Commands
     init
     add [--empty] [--open]
-    version [--ignore] [--snapshot <?name>]
+    version [--ignore] [--snapshot <?name>] [--snapshot-preid-template <template>]
```
Andarist, line 14:
```suggestion
    version [--ignore] [--snapshot <?name>] [--snapshot-prerelease-template <template>]
```

packages/assemble-release-plan/src/index.ts
```diff
@@ -16,6 +16,15 @@ import { Packages, Package } from "@manypkg/get-packages";
 import { getDependentsGraph } from "@changesets/get-dependents-graph";
 import { PreInfo, InternalRelease } from "./types";
 
+type SnapshotReleaseParameters = {
+  preidTemplate: string | null | undefined;
```
Andarist, line 20:
```suggestion
  prereleaseTemplate: string | null | undefined;
```

Andarist (approved):

dotansimha:
Amazing, thank you for taking this one to the finish line @Andarist !

Andarist:
Thank you for the amazing contribution and for bearing with me 😉

PR #867, opened 2022-07-11
fix: use correct prettier configuration

Max-Dbv:
### Problem

Using `overrides` in `prettierrc.js` doesn't work on generated files. It is an issue for us, as we want a specific configuration for changesets generated files (large `printWidth`).

### Why

We use `cwd` to resolve the current configuration file. The configuration object returned doesn't resolve overrides as it doesn't match anything.

`.prettierrc.js`
```js
{
  printWidth: 120,
  overrides: [
    {
      files: '*.md',
      options: {
        printWidth: 300,
      },
    },
  ],
}
```

Resolved configuration object using `cwd`:
```js
{
  printWidth: 120,
}
```

### Fix

Resolve configuration's file specific to the current file path.

Resolved configuration object using file path:
```js
{
  printWidth: 300,
}
```

### DIFF
--- a/.changeset/flat-jars-talk.md
+++ b/.changeset/flat-jars-talk.md
@@ -1,0 +1,6 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/write": patch
+---
+
+Fix prettier configuration file resolution using correct path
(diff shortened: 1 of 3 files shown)


bluwy:
I think this has been fixed at https://github.com/changesets/changesets/pull/1351. Going to close this for now. Thanks for the contribution!

PR #871, opened 2022-07-13
Fix: Changes not detected by changeset cli 2.23.1

NicoKam:
Not include ignore.

### DIFF
--- a/packages/cli/src/commands/add/index.ts
+++ b/packages/cli/src/commands/add/index.ts
@@ -42,7 +42,7 @@ export default async function add(
     });
     const changePackagesName = changedPackages
       .map(pkg => pkg.packageJson.name)
-      .filter(pkgName => config.ignore.includes(pkgName));
+      .filter(pkgName => !config.ignore.includes(pkgName));
 
     newChangeset = await createChangeset(changePackagesName, packages);
     printConfirmationMessage(newChangeset, packages.length > 1);
--- a/.changeset/slimy-dolls-hang.md
+++ b/.changeset/slimy-dolls-hang.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Fixed filtering of the ignored packages when adding a changeset.


packages/cli/src/commands/add/index.ts
```diff
@@ -42,7 +42,7 @@ export default async function add(
     });
     const changePackagesName = changedPackages
       .map(pkg => pkg.packageJson.name)
-      .filter(pkgName => config.ignore.includes(pkgName));
+      .filter(pkgName => !config.ignore.includes(pkgName));
```
walt1992, line 45: That's why

walt1992 (approved):

packages/cli/src/commands/add/index.ts
```diff
@@ -42,7 +42,7 @@ export default async function add(
...
 
     newChangeset = await createChangeset(changePackagesName, packages);
```
walt1992, line 47: Is this line correct? Could you check this as well?

Because I got an error below 
```
$ /Users/ms.park/Desktop/workspace/payment.ohou.se/node_modules/.bin/changeset
🦋  error TypeError: Cannot read properties of undefined (reading 'packageJson')
🦋  error     at createChangeset (/Users/ms.park/Desktop/workspace/payment.ohou.se/node_modules/@changesets/cli/dist/cli.cjs.dev.js:415:75)
🦋  error     at add (/Users/ms.park/Desktop/workspace/payment.ohou.se/node_modules/@changesets/cli/dist/cli.cjs.dev.js:507:26)
🦋  error     at async run$2 (/Users/ms.park/Desktop/workspace/payment.ohou.se/node_modules/@changesets/cli/dist/cli.cjs.dev.js:1254:5)
error Command failed with exit code 1.

```

And this error is from this line.

This line was originally same with before

```
    newChangeset = await createChangeset(changePackagesName, packages.packages);
```

But this had been changed since 2.23.1

packages/cli/src/commands/add/index.ts
Andarist: Ouch 😬 sorry for that!

packages/cli/src/commands/add/index.ts
Andarist: With what version do you experience this problem? With the 2.23.1 or with the version of the code from this PR?

PR #874, opened 2022-07-14
Fix: Log errors if git tag fails during publish

silesky:
This PR adds better error logging, and will make `changeset publish` only show "tag created" on success. Under the previous logic, the following error from `git tag` was being suppressed, making  something like this hard to debug:
```
--
await tag("v1", cwd)
await tag("v1", cwd)

// => "fatal: tag 'v1' already exists"
```
 
... or errors like this (which can happen because the `git.tag` method uses _annotated_ tags):
```

*** Please tell me who you are.
 
Run
 
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
 
to set your account's default identity.
Omit --global to set the identity only in this repository.
 
fatal: empty ident name (for <buildkite-agent@localhost>) not allowed

```

### DIFF
--- a/.changeset/plenty-colts-join.md
+++ b/.changeset/plenty-colts-join.md
@@ -1,0 +1,6 @@
+---
+"@changesets/git": patch
+"@changesets/cli": patch
+---
+
+Fix tag step in `changeset publish` so it can display errors as well as successful operations.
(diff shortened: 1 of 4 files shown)


silesky:
@Andarist (or any other mantainers) let me know if you have any thoughts -- I ran into this footgun yesterday where I was completely vexed about why on CI it said "tag created" when no tags were created, so was hoping to make the debugging story a bit nicer.

PR #878, opened 2022-07-18
Add Polaris to projects using Changesets section

samrose3:
[Polaris](https://github.com/Shopify/polaris) loves Changesets 🦋 ❤️


gwyneplaine:
Very cool :D

PR #881, opened 2022-07-19
Fix grammer

gamertike:


Andarist:
I think that the current version is better/more correct. Note that this part of the sentence is followed by a bullet point and with your suggested version we'd end up with:
> We have a [github action](https://github.com/changesets/action) for that creates a `version` PR, then keeps it up to date, recreating it when merged. This PR always has an up-to-date run of `changeset version`

And that isn't correct.

gamertike:
> I think that the current version is better/more correct. Note that this part of the sentence is followed by a bullet point and with your suggested version we'd end up with:
> > We have a [github action](https://github.com/changesets/action) for that creates a `version` PR, then keeps it up to date, recreating it when merged. This PR always has an up-to-date run of `changeset version`
> 
> And that isn't correct.

My bad, you are correct.

PR #889, opened 2022-07-25
Fix return value of `getCurrentCommitId`

### LINKED ISSUE #888: Commit hash with quotation marks results in invalid snapshot version
## Affected Packages

@changesets/git, @changesets/cli, @changesets/assemble-release-plan

## Problem

Function `getCurrentCommitId` from package `@changesets/git` returns commit hash with quotation marks. This value is then [used](https://github.com/changesets/changesets/blob/e4204820cdb93d22a3c7a27f452a0ab30db6369a/packages/cli/src/commands/version/index.ts#L78) when generating snapshot version: with config option `prereleaseTemplate` set to `'${tag}.{commit}'` it results to e.g.: `'2.1.0-foobar."41b793f"'` (instead of `'2.1.0-foobar.41b793f'`) and because such semver is invalid, error is thrown.

## Proposed solution

`getCurrentCommitId` should return commit hash with quotation marks removed, but I'm not sure if any other package depends upon quotation marks actually being present. Anyway, the quick solution is in #889.

jakubmazanec:
fixes #888

### DIFF
--- a/.changeset/kind-beds-divide.md
+++ b/.changeset/kind-beds-divide.md
@@ -2,4 +2,4 @@
 "@changesets/git": patch
 ---
 
-Fixes `getCurrentCommitId` so that the returned value doesn't contain quotation marks.
+Fixed `getCurrentCommitId` so that the returned value doesn't contain quotation marks on some Windows machines.
--- a/.changeset/kind-beds-divide.md
+++ b/.changeset/kind-beds-divide.md
@@ -1,0 +1,5 @@
+---
+"@changesets/git": patch
+---
+
+Fixes `getCurrentCommitId` so that the returned value doesn't contain quotation marks.
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -296,5 +296,6 @@ export async function getCurrentCommitId({
     await spawn("git", ["show", `--format="%h"`, "--no-patch"], { cwd })
   ).stdout
     .toString()
-    .trim();
+    .trim()
+    .replace(/"/g, "");
 }
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -292,10 +292,7 @@ export async function getCurrentCommitId({
 }: {
   cwd: string;
 }): Promise<string> {
-  return (
-    await spawn("git", ["show", `--format="%h"`, "--no-patch"], { cwd })
-  ).stdout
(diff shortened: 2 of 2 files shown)


packages/git/src/index.ts
```diff
@@ -296,5 +296,6 @@ export async function getCurrentCommitId({
     await spawn("git", ["show", `--format="%h"`, "--no-patch"], { cwd })
   ).stdout
     .toString()
-    .trim();
+    .trim()
+    .replace(/"/g, "");
```
Andarist, line 300: Interesting, `git show --format="%h" --no-patch` doesnt add quotation marks in the stdout for me. What system and git version are you using?

packages/git/src/index.ts
```diff
@@ -296,5 +296,6 @@ export async function getCurrentCommitId({
     await spawn("git", ["show", `--format="%h"`, "--no-patch"], { cwd })
```
Andarist, line 296: I wonder if the reason we end up with those quotation marks isn't that we include them here in the `format`. I've checked on my machine and this works fine too:
```suggestion
    await spawn("git", ["show", "--format=%h", "--no-patch"], { cwd })
```

Alternatively we could also just switch to using this:
```suggestion
    await spawn("git", ["rev-parse", "--short", "HEAD"], { cwd })
```
It seems to be better suited for this job anyway.

packages/git/src/index.ts
jakubmazanec: Windows 10 Enterprise 21H1, git 2.30.1.windows.1.
When i run `git show --format="%h" --no-patch` in PowerShell, I also see only the hash 🤔

packages/git/src/index.ts
jakubmazanec: Both suggestion seem to solve the problem. I agree that the second one seems to be better for this. I updated the PR.

.changeset/kind-beds-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/git": patch
+---
+
+Fixes `getCurrentCommitId` so that the returned value doesn't contain quotation marks.
```
Andarist, line 5:
```suggestion
Fixes `getCurrentCommitId` so that the returned value doesn't contain quotation marks on some Windows machines.
```

Andarist (approved):

Issue #896, opened 2022-08-01
Config.changelog set to false, still writes CHANGELOG.MD files after exiting prerelease


sdirosa:
## Affected Packages

As far as I can tell the logic of this is in `@changesets/apply-release-plan`. 

## Problem

We have recently switched to changesets in our monorepo (very happy with it!) but we don't want to use the changelog feature (yet). 

I've set `"changelog": false` in our `.changeset/config.json`.  This is the full config:

```json
{
  "$schema": "https://unpkg.com/@changesets/config@2.1.0/schema.json",
  "changelog": false,
  "commit": false,
  "fixed": [["@objectiv/*"]],
  "linked": [],
  "access": "public",
  "baseBranch": "main",
  "updateInternalDependencies": "patch",
  "ignore": []
}
```

When I exit a prerelease and run `yarn changeset version` to apply my changesets, I also end up with changelog files being created in all the packages of the monorepo. I am not sure if this happens also without `pre`, I haven't tested that, as we only do prerelease flows.

### Some superficial investigation
From a quick check on the source code it seems the `getNewChangelogEntry` function correctly checks whether changelogs have been disabled to skip some initial logic, but then proceeds to return a list of packages with `changelog`s anyway.

Later in the code, the bit of code that processes the list of `finalisedRelease` entries doesn't check the config either. Instead it validates the presence of the changelog attribute. E.g. `changelog && changelog.length > 0`. 

Unless I am completely off (I only spent a few minutes reading the code), since `finalisedRelease` is partially based on the output of `getNewChangelogEntry`, I'd guess that most probably the changelog attribute is always present and filled with something.

It's not a big deal for us, we can run a `find` and delete the changelogs before committing. That said, I think quite a bit of logic runs to process changelogs and it could be entirely skipped when that setting is set to false.

## Proposed solution

Would be really cool if setting `changelog` to `false` in the config.json would result in no changelogs being generated at all, just package.json updates.

A quick fix could be to either skip `getNewChangelogEntry` when config.changelog is false, or double check whether changelogs have been disabled when processing the list of `finalisedRelease` and skip the changelog logic there.

I'd be very happy to try and fix this up, or test it if necessary.

Andarist:
I've checked the PR that introduced `schema.json` and it seems that `false` was always a valid value there:
https://github.com/changesets/changesets/pull/132/files#diff-d3013abc54530a5bedc30c2f340c829e5dc49923a70bdf4cc469180bf047cb38R18-R21
[...]

PR #900, opened 2022-08-04
Fix: skip generating changelogs when config.changelog is set to false

sdirosa:
Closes https://github.com/changesets/changesets/issues/896

### DIFF
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -180,9 +180,13 @@ async function getNewChangelogEntry(
   config: Config,
   cwd: string
 ) {
-  // Skip generating changelog entries if config.changelog is false
   if (!config.changelog) {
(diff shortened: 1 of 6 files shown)


packages/apply-release-plan/src/index.ts
```diff
@@ -180,28 +180,31 @@ async function getNewChangelogEntry(
   config: Config,
   cwd: string
 ) {
+  // Skip generating changelog entries if config.changelog is false
+  if (!config.changelog) {
+    return releasesWithPackage;
+  }
```
Andarist, line 186:
```suggestion
  // Skip generating changelog entries if config.changelog is false
  if (!config.changelog) {
    return Promise.resolve(
      releasesWithPackage.map(release => ({
        ...release,
        changelog: null
      }))
    );
  }
```

PR #905, opened 2022-08-10
Prefer local Prettier when formatting files

### LINKED ISSUE #616: prettier old version
## Affected Packages

apply-release-plan

## Problem

It's relying on an old version of prettier: https://github.com/atlassian/changesets/blob/main/packages/apply-release-plan/package.json#L20

## Proposed solution

Is it necessary to ship prettier as a dependency of this package? Could it be moved to devDependencies? If not, upgrade prettier to v2.

Andarist:
closes #616

it's the best to just take a look at the first commit (the one without formatting changes): https://github.com/changesets/changesets/pull/905/commits/85fa88920e13a188e478824eebaf4c81e5b0d90a


packages/apply-release-plan/src/index.ts
```diff
@@ -19,14 +19,25 @@ import prettier from "prettier";
 import versionPackage from "./version-package";
 import getChangelogEntry from "./get-changelog-entry";
 
+function getPrettierInstance(): typeof prettier {
+  try {
+    return require("prettier");
```
mitchellhamilton, line 24: Could we `createRequire` and resolve it from the repo root? Without that, this doesn't change anything, right?

mitchellhamilton: Looks like `require.resolve("prettier", { paths: [projectDir] })` works on more versions so let's probably use that

Andarist: Ah, ye - this is totally what I had in mind when I started working on this and just forgot it in the rush 🤣

packages/write/src/index.ts
```diff
@@ -4,6 +4,17 @@ import prettier from "prettier";
 import humanId from "human-id";
 import { Changeset } from "@changesets/types";
 
+function getPrettierInstance(): typeof prettier {
+  try {
+    return require("prettier");
```
mitchellhamilton, line 9: This also needs to use `require.resolve`

Andarist: fixed

mitchellhamilton (approved):

Samira900i8uu (approved):

PR #907, opened 2022-08-10
Allow publishing of packages with ignored devDependencies

### LINKED ISSUE #906: Allow versioning/publishing of packages with ignored devDependencies
## Affected Packages

@changesets/changesets

## Problem

Currently you can't version packages with `devDependencies` that point at ignored packages.

`devDependencies` are typically only for local development of the package (from the monorepo) so we should allow you to point at ignored packages here.

For my workflow I have an `internal` folder in my monorepo where each internal package is prefixed with the `@internal` namespace. I store things like typescript configurations etc in here that most of my packages is use to build but not at install time.

I want to ignore this whole namespace however I can no longer version or publish any of my packages.

## Proposed solution

Add an option to allow the versioning of publishing of packages that have ignored `devDependencies`.

georeith:
Resolves: https://github.com/changesets/changesets/issues/906


KasJones (approved):

georeith:
didn't mean to close and reopen/mark this for automerge I thought I was in a different PR

kachkaev:
Might be worth looking at `"private": true` instead. Context: https://github.com/changesets/changesets/issues/906#issuecomment-1239468046 (**UPD2**)

packages/get-dependents-graph/src/get-dependency-graph.test.ts
```diff
@@ -115,4 +115,39 @@ describe("getting the dependency graph", function() {
     `);
     })
   );
+
+  it("should skip devDependencies if specified", function() {
```
Andarist, line 119: It would also be great to have a test for the version command (I assume that it's where the original problem originated). While this package here might be directly responsible for the issue - it is an implementation detail and I would like to have a test for what people usually interact with (so `@changesets/cli`)

packages/types/src/index.ts
```diff
@@ -71,6 +71,8 @@ export type Config = {
   /** The minimum bump type to trigger automatic update of internal dependencies that are part of the same release */
   updateInternalDependencies: "patch" | "minor";
   ignore: ReadonlyArray<string>;
+  /** Allows packages to be published if they have ignored packages in their devDependencies */
+  allowIgnoredDevDependencies?: boolean;
```
Andarist, line 75: q: what would be the downsides of removing this flag entirely and instead just ignoring such dev dependencies at all times?

wycats: This would be fine, and certainly would be fine if the dev dependency is marked as private.

wycats:
This change would really improve my workflow 😄

Andarist:
@wycats I'm OK with this "feature" request - but I think that the additional config option that has been introduced here is redundant. Would you be able to prepare a PR targeting this PR that would introduce the requested changes? I'm a little bit busy right now and it might take me some days to get to this myself.

georeith:
@Andarist I reworked this removing the extra config option and made it so it just ignores devDependencies when looking at ignored files. I added a test for the version command.

I opted to not make it ignore devDependencies everywhere as devDependencies can have an effect on your files output (they may affect how it is built) but it is fine for them to be ignored.

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -193,6 +193,7 @@ async function internalPublish(
           ["publish", opts.publishDir, "--json", ...publishFlags],
           {
             env: Object.assign({}, process.env, envOverride),
+            cwd: opts.cwd,
```
georeith, line 196: Without this change my snapshot publishes fail with:

```
🦋  info Publishing "@marvelapp/crdt" at "0.0.0-20230614163206"
🦋  error an error occurred while publishing @marvelapp/crdt: E403 403 Forbidden - PUT https://registry.npmjs.org/@marvelapp%2fcrdt - You cannot publish over the previously published versions: 7.1.0.
🦋  error a package version that is forbidden by your security policy, or
🦋  error on a server you do not have access to.
🦋  error npm notice Publishing to https://registry.npmjs.org with tag dev and restricted access
🦋  error npm ERR! code E403
🦋  error npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/@marvelapp%2fcrdt - You cannot publish over the previously published versions: 7.1.0.
🦋  error npm ERR! 403 In most cases, you or one of your dependencies are requesting
🦋  error npm ERR! 403 a package version that is forbidden by your security policy, or
🦋  error npm ERR! 403 on a server you do not have access to.
🦋  error
🦋  error npm ERR! A complete log of this run can be found in: /Users/george/.npm/_logs/2023-06-15T09_47_41_993Z-debug-0.log
🦋  error
```

package.json
```diff
@@ -73,5 +73,6 @@
       "scripts/*"
     ]
   },
-  "prettier": {}
+  "prettier": {},
+  "packageManager": "yarn@1.22.19"
```
georeith, line 77: For those of us who use another version of yarn than classic and have corepack enabled this will use the right yarn version when working on this repo.

Netail: Bump to 1.22.22

Netail:
Any update on this?

0xMurage:
What is preventing this from being accepted or merged?
---

bluwy:
Looks like this is already fixed by #1370

Issue #862, opened 2022-07-05
Streamline changeset generation with conventional commit support


william-will-angi:
## Affected Packages

`@changesets/cli`, perhaps a new package would be more appropriate for this functionality

## Problem

Our organization has been versioning packages with lerna + conventional commits for a while now, but we are looking at moving towards changesets, mainly because the `.changeset/config.json` is just awesome. I love how you can make explicit associations between different packages - this is exactly what the conventional commits workflow has been missing imo.

Our developers have become accustomed to the conventional commit workflow and I think that providing an integration could go a long way in terms of facilitating adoption for other organizations as well.

## Proposed solution

I am hoping something similar to the following would be possible.

Given some number of conventional commits since master `n`, and some arbitrary number of packages in a mono-repo, running `changeset version` should:
1. Visit each of the `n` commits and determine the proper version bump based on the commit message and affected package. For example, if you have a commit that affects package A: `feat: Add some feature to package A`, the associated version bump should be translated to a minor bump of package A. Similarly if you have a commit to package B: `fix: Fix something`, the associated version bump should be translated to a patch bump of package B.

2. Once the version bumps from conventional commits are determined, pre-populate the `changeset version` prompts with the respective bumps. In the example above with package A, when the prompt asks `Which packages should have a minor bump?`, package A should already be pre-selected. I would also see this pre-populating the prompt. `Please enter a summary for this change (this will be in the changelogs)`.

I would also hope for a flag that would allow us to auto-generate the changesets from conventional commits (without any additional prompting). That being said, I can see how this specifically may be at odds with the generally recommended changeset workflow.

I am new to changesets, so I am currently unfamiliar with how it is set up, or how much of a shift this would be. Hoping I can get an idea of whether this something that others would find valuable, and/or what changes might be necessary to facilitate this. I am happy to help out however possible with this effort. Alternatively, if there is any sort of plugin-based API for third-party integrations, that may be more appropriate for this use case, but I am unaware of such an API.

Cheers!

unional:
dup of https://github.com/changesets/changesets/issues/577

oncet:
> dup of #577

If I understand correctly, #577 is about the format of the PR message and what OP is looking for is something more similar to sementic release. I.e. having Changsets to automatically determine if changes should be major, minor or patch.

sangdth:
I have the same understanding with @oncet , and looking forward to this feature.

zigang93:
as happy user from semantic release, this is what I am looking for..

[...]

PR #928, opened 2022-08-25
Add support for conventional-recommended-bump

heygrady:
# Add conventional recommended bump

Picking versions manually is boring when you're already using conventional commits in your repository.

This PR adds some flags to enable using [conventional-recommended-bump](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-recommended-bump).

The purpose is to cover the use-case outlined in https://github.com/changesets/changesets/issues/862 (but in a way that matches how Changesets currently works). 

This feature was heavily inspired by [@lerna/conventional-commits](https://github.com/lerna/lerna/tree/main/core/conventional-commits). People asking for this feature have already adopted conventional commits in their workflows.

**Goals**:
1. Enable conventional commits for version bumps
2. Enable creation of changesets without user interaction
3. Don't break the existing workflow

```sh
# Create a changeset without answering any questions
yarn changeset add --all-changed --recommend --message "Amazing!" --yes
```

### New flags:

- `--all` -- selects all changed and unchanged packages
- `--all-changed` -- selects only changed packages
- `--all-unchanged` -- selects only the unchanged packages
- `--message` -- used for the summary of the changeset
- `--recommend` -- automatically recommends the version bump using conventional commits
- `--yes` -- skips all confirmations to enable adding a changeset with zero interactions

### New Config

Users will need to opt-in to using conventional commits by adding a preset to the `.changesets/config.json`. This preset is only used for recommending version bumps and will not otherwise affect the changelog that gets generated.

- `conventionalCommits` -- Defaults to `false`. Otherwise you should add a conventional-changelog preset ([official presets](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages)). Setting it to `true` will presume [conventional-changelog-angular](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular), which is the default that Lerna uses.

### Example: Default preset

This example shows the `.changeset/config.json` from the changesets repo with conventional commits enabled.

```
{
  "changelog": [
    { "repo": "changesets/changesets" }
  ],
  "conventionalCommits": true,
  "baseBranch": "main",
  "commit": false,
  "access": "public",
  }
}
```


packages/cli/package.json
```diff
@@ -44,17 +44,21 @@
     "@changesets/read": "^0.5.7",
     "@changesets/types": "^5.1.0",
     "@changesets/write": "^0.1.9",
+    "@lerna/conventional-commits": "^5.4.3",
```
heygrady, line 47: This package provides a function for resolving conventional-changelog presets using short names.

packages/cli/package.json
```diff
@@ -44,17 +44,21 @@
...
     "chalk": "^2.1.0",
+    "conventional-changelog-angular": "^5.0.13",
+    "conventional-recommended-bump": "^6.1.0",
     "enquirer": "^2.3.0",
     "external-editor": "^3.1.0",
     "fs-extra": "^7.0.1",
     "human-id": "^1.0.2",
     "is-ci": "^3.0.1",
     "meow": "^6.0.0",
+    "npmlog": "^6.0.2",
```
heygrady, line 61: This package is required to disable lerna logging output. It's probably more desirable to drop the dependency on `@lerna/conventional-commits`, since it comes with extra baggage.

packages/cli/package.json
heygrady, line 47: This package provides a method for resolving conventional changelog presets using common conventions. This is probably a bonus feature.

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -101,76 +142,71 @@ function getPkgJsonByName(packages: Package[]) {
...
+          path: pkgDirsByName.get(pkgName)
+        };
+        const data = await conventionalRecommendedBumpAsync(
+          conventionalOptions
+        );
+        if (options.recommend === true) {
+          releases.push({ name: pkgName, type: data.releaseType ?? "patch" });
+        } else {
+          bumpSets[data.releaseType ?? "patch"].add(pkgName);
         }
```
heygrady, line 190: This is the main magic in this PR. It resolves the conventional changelog preset and queries for the recommended version bump for each package being included in this release.

chathu-novade:
This is nice. I was missing those features

william-will-angi:
Psyched to see this implemented!

william-will-angi (commented):
Are the added flags related to the conventional commits workflow or unrelated?

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -101,76 +138,91 @@ function getPkgJsonByName(packages: Package[]) {
...
+      );
+      for (const pkgName of pkgsLeftToGetBumpTypeFor) {
+        const conventionalOptions = {
+          config: changelogConfig,
+          lernaPackage: pkgName,
+          path: pkgDirsByName.get(pkgName)
+        };
+        const data = await conventionalRecommendedBumpAsync(
+          conventionalOptions
+        );
```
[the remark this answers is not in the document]
heygrady: ## Are the added flags related to the conventional commits workflow?

The most beneficial new flag here is `--recommend`, as it skips the prompts for choosing the major/minor/patch versions. The other flags are bonus and simply allow you to skip all of the confirmation prompts. We could get rid of all new flags if we presumed that the existence of `conventionalCommits` in the config meant that the user _always_ wanted to skip the major/minor/patch prompts.

The only workflow I've used that utilized conventional commits was Lerna. There is also Semantic Release and a few others. Automation of version numbers is a big part of conventional commits. I added the `--yes` flag to make it possible to generate a changeset without answering any prompts.

## I don't think that changesets should necessarily be coupled to lerna concepts.

These aren't really "lerna" concepts, Lerna is just a popular example of conventional commits being used. Many similar tools rely on conventional-changelog to implement Conventional Commits.

Conventional Commits is an open standard and the conventional-changelog package ecosystem builds upon that standard. Other release management tools use the low-level conventional-changelog packages.

- Conventional Commits Standard: https://www.conventionalcommits.org/en/v1.0.0/
- Conventional Changelog Tools: https://github.com/conventional-changelog/conventional-changelog
- Popular implementation: https://github.com/semantic-release/semantic-release

## More beneficial to abstract this to a generic function that can be referenced in the changeset config?

I'm not sure that passing in a generic function from the config would meet the user's expectations. Also, it is important to note that the changesets config is a json file by default. That makes passing a function from the config impossible.

The customization you are asking for is accomplished here:

```js
const changelogConfig = await getChangelogConfig(options.conventionalCommits)
```

The `options.conventionalCommits` value should be a string that refers to a preset package (resolvable using  [conventional-changelog-preset-loader](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-preset-loader)). The lerna example supported a wild west of cases using [npm-package-arg](https://www.npmjs.com/package/npm-package-arg).

In practice, most users are going to be using a standard preset. I decided to go with the standard preset loader to keep the dependencies lighter.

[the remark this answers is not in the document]
heygrady: Another companion tool is [Commitlint](https://commitlint.js.org/#/reference-configuration).

Adopting Conventional Commits means adopting several tools to work in concert. The standardized presets enable this interoperability.

1. Husky -- run git precommit hooks
2. Commitlint -- enforce valid conventional commits for all commit messages using a husky precommit hook
3. Changesets / Lerna / Semantic Release -- utilizes the commit messages to bump version numbers and generate changelog files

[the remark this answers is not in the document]
heygrady: I'm going to be rethinking my approach here. I've been playing more with changesets on a personal project and I can see some issues.

You make an excellent point about the difficulty of testing conventional commits behavior. It may be considered out of scope to do full integration testing of conventional commits as it can be presumed that conventional commits is testing itself.

For clarity... "would meet the user's expectations" was referring to the exact rig you are recommending. Most competing release management tools that support conventional commits also support named presets. I'm simply not aware of any competing standards that would warrant making the version bump interface within changesets wholly generic. People are either using conventional commits or they aren't.

I did fine one case where [release-please](https://github.com/googleapis/release-please) supports conventional commits but does not support changing the preset. However, they _are_ using a conventional-changelog preset under the hood.

There are lots of examples of tools similar to changesets that support conventional-changelog presets.

For example, semantic release allows you to provide a `preset` key [like this](https://github.com/semantic-release/commit-analyzer#usage):

```json
{
  "plugins": [
    ["@semantic-release/commit-analyzer", {
      "preset": "angular",
    }]
  ]
}
```

Other tools, like commitlint ([see here](https://commitlint.js.org/#/reference-configuration?id=javascript)) offer similar settings.

```js
module.exports = {
  parserPreset: 'conventional-changelog-atom',
}
```

Similarly, Lerna ([see here](https://github.com/lerna/lerna/blob/6cb8ab2d4af7ce25c812e8fb05cd04650105705f/commands/version/README.md#--changelog-preset)) supports a `changelogPreset` like this:

```json
{
  "changelogPreset": "angular"
}
```

Usually tools default to using the angular preset as the Angular community originally popularized the modern notion of conventional commits.

Andarist:
Thank you for your PR - I will try to get to reviewing this in a couple of days.

JulianCataldo:
Hey there!
I wonder if I can clone this experimental branch for testing out conventional commit based bumps?
Not sure if it is useable, that's why I'm asking.
I admit I'm waiting for this specific feature before pulling the trigger on changeset, as I'm an avid lerna+tools or semantic-release user.
Thanks for all.

maximveksler:
Happy to help test this, what's the status please?

AlexandrosMelissas:
Hello, any status on this? Also, has anyone made a script for conventional commits with changesets until this is merged?

willwill96:
Hey @AlexandrosMelissas It isn't published as a package, but I put together a script [here](https://github.com/willwill96/mono-repo-tools/blob/9cc80932c804e0e8e5887db9fd532865e8f5adf9/packages/changeset-conventional-commit/src/index.ts) which translates your conventional commits directly to changesets. It's just a standalone script though (It doesn't integrate with the changeset CLI).

AlexandrosMelissas:
> Hey @AlexandrosMelissas It isn't published as a package, but I put together a script [here](https://github.com/willwill96/mono-repo-tools/blob/9cc80932c804e0e8e5887db9fd532865e8f5adf9/packages/changeset-conventional-commit/src/index.ts) which translates your conventional commits directly to changesets. It's just a standalone script though (It doesn't integrate with the changeset CLI).

This is exactly what i was thinking. Could be useful for ci cd.

PR #949, opened 2022-09-17
Fixed an issue with incorrect dependent updates on none changesets with `updateInternalDependents: "always"`

### LINKED ISSUE #947: updateInternalDependents incorrectly updates version of transitive devDependencies
## Problem

updateInternalDependents currently incorrectly triggers version updates on packages with transitive dependencies on devDependencies. I appreciate that is too terse to make sense, a demo is clearer:

The following demonstration with reproduction steps is available at https://github.com/BPScott/changesets-transitive-deps-bug

Consider a repository with 3 packages:

- `package-a`
- `package-b` - this has a devDependency on `package-a`
- `package-c` - this has a runtime dependency on `package-b`

`package-a` is the only package with a changeset associated with it.

### Expected behaviour

When running `npm run changset version`, I would expect changsets to:

- Update `package-a`: Increase the version in the package.json and generate a CHANGELOG entry, as there is a changeset associated with this package.
- Update `package-b`: Update the devDependency version of package-a to the new version i the package.json. The version of `package-b` should be untouched as only a devDependency changed. No CHANGELOG entry is created as the version does not change.
- Not touch `package-c`'s package.json. Because package-b's version was not updated there should be no change to `package-c`'s package.json or CHANGELOG.

### Actual Behaviour

The expected changes to `package-a` and `package-b`'s both occur as expected.

However modifications are made to `package-c`, when there should be none, because there package-b did not receive a version bump.

`package-c`'s version number is increased in the package.json. A changelog heading for the new version is generated, but no content is present.

## Proposed solution

Changsets is currently smart enough to skip bumping the version of a package is when one of its devDependencies has changed. Because that package does not recieve a version bump, neither should any of its dependencies. I.e. in the above example - because `package-b` did not receive a version bump, then `package-c` should be have a version bump either.

BPScott:
This PR adds a pair of failing tests, showcasing the issue described in #947.

When `updateInternalDependents:'always'` is set then transitive dependencies of devDependencies should not be part of the release plan, howevever they currently are.

I believe the root cause of this issue is somewhere around [this logic](https://github.com/changesets/changesets/blob/main/packages/assemble-release-plan/src/determine-dependents.ts#L85-L93). There needs to be some way of stripping out dependencies of devDependencies from the release plan. I've tried and failed to work out how to do that with small adjustments, but I hope these tests can serve as a useful starting point for an actual fix.

---

UPDATE: fixes #947


packages/cli/src/commands/version/version.test.ts
```diff
@@ -1066,6 +1066,59 @@ describe("updateInternalDependents: always", () => {
...
+    );
+    expect(getPkgJSON("pkg-c", spy.mock.calls)).toEqual(
+      expect.objectContaining({
+        name: "pkg-c",
+        version: "1.0.0",
+        dependencies: {
+          "pkg-b": "1.0.0",
+        },
+      })
+    );
```
Andarist, line 1105: Correct me if I'm wrong but this one should not be part of the valid expectation, right? In this situation the `pkg-c` should be left untouched

Andarist:
I've pushed out a fix for this here - still need to recheck some other things though before I can land this.

Andarist (approved):

packages/cli/src/commands/version/version.test.ts
BPScott: Yes, you're correct - pkg-c's package.json should be untouched. I didn't spot that getPkgJSON behaves in the same way as `getChangelog` and throws an error if it wasn't touched.

BPScott:
Fixes and extra tests all look good to me. Thank you for taking this over @Andarist!

PR #951, opened 2022-09-19
Fix types to avoid `@ts-ignore`

Andarist:


packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -123,36 +122,38 @@ export default function determineDependents({
...
-        // @ts-ignore - I don't know how to make typescript understand that the filter above guarantees this and I got sick of trying
-        ({ name, type, pkgJSON }: Release & { pkgJSON: PackageJSON }) => {
-          // At this point, we know if we are making a change
-          updated = true;
+      .filter(
+        (
+          dependentItem
+        ): dependentItem is typeof dependentItem & { type: VersionType } =>
+          !!dependentItem.type
+      )
```
Andarist, line 130: the fix is here - I've added a type predicate here and removed `@ts-ignore` comment targeting the binding pattern in the parameter below

mitchellhamilton (approved):

PR #952, opened 2022-09-19
Upgrade TypeScript and Babel deps

Andarist:


babel.config.js
```diff
@@ -7,5 +7,14 @@ module.exports = {
       },
     ],
   ],
+  plugins: [
+    [
+      "@babel/plugin-transform-runtime",
```
Andarist, line 12: An extra fix for using the externalized Babel helpers

Andarist:
The CI failure is weird:
- it doesn't happen to me on my machine
- optional chaining should be enabled by default (see [here](https://github.com/babel/babel/blob/aad7eb743d1354a5475f9d43226bfedc5f149e4b/packages/babel-preset-env/src/available-plugins.ts#L99))

bluwy:
Looks like the dependencies have been upgraded to higher versions now on main. Should we close this?

PR #967, opened 2022-10-01
fix: No longer bumps dependent unversioned packages

willwill96:


packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1230,6 +1230,28 @@ describe("version update thoroughness", () => {
     expect(releases[3].name).toEqual("pkg-d");
     expect(releases[3].newVersion).toEqual("1.0.1");
   });
+
+  it("should not bump unversioned dependent packages", () => {
+    // @ts-ignore
```
willwill96, line 1235: I started to create new types to address this, but it started to spiral out of scope of the original issue, because multiple packages in this repo are dependent on the `@manypkg/get-packages` type definitions. In case anyone is curious this is what my types looked like:
```tsx
import {
  Package as _Package,
  Packages as _Packages,
} from "@manypkg/get-packages";

export interface Packages extends Omit<_Packages, "root" | "packages"> {
  root: Package;
  packages: Package[];
}

export interface Package extends Omit<_Package, "packageJson"> {
  packageJson: Omit<_Package["packageJson"], "version"> & {
    version: string | null | undefined;
  };
}

export interface Packages extends Omit<_Packages, "root" | "packages"> {
  root: Package;
  packages: Package[];
}

```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1230,6 +1230,28 @@ describe("version update thoroughness", () => {
...
+    );
+
+    expect(releases.length).toBe(3);
+    expect(releases[0].name).toEqual("pkg-a");
+    expect(releases[0].newVersion).toEqual("1.0.1");
+    expect(releases[1].name).toEqual("pkg-b");
+    expect(releases[1].newVersion).toEqual("1.0.1");
+    expect(releases[2].name).toEqual("pkg-e");
+    expect(releases[2].type).toEqual("none");
+    expect(releases[2].newVersion).toEqual(undefined);
```
Andarist, line 1253: What would be the result here before your change?

willwill96: I actually had a misunderstanding of the problem initially and have updated the test to reflect the use case better. Before the change I introduced, the release type would be `none` and `newVersion` would be `null`

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -91,22 +90,26 @@ export default function determineDependents({
...
-                  // We don't need a version bump if the package is only in the devDependencies of the dependent package
-                  if (
-                    type !== "major" &&
-                    type !== "minor" &&
-                    type !== "patch"
-                  ) {
-                    type = "none";
+              if (!dependentPackage.packageJson.version) {
+                type = "none";
+              } else {
```
Andarist, line 95: Could you tell me more about your latest commit (10c832ce0f361ecf79822751fdcf52625e4dbab3)? I wonder why you have pushed the check further down the road.

packages/assemble-release-plan/src/index.test.ts
Andarist: Perhaps you could lift this test to the CLI level to avoid this problem

packages/assemble-release-plan/src/index.test.ts
willwill96: Good idea, I think you are right. I'll try to add the test for that this evening

packages/assemble-release-plan/src/determine-dependents.ts
willwill96: This means that the unversioned package only gets a release of type none if the semversion is no longer valid for the repo or there is some other config (always update dependents for example)

Andarist: roger that - it would be cool to have a test case showcasing that difference

packages/assemble-release-plan/src/index.test.ts
willwill96: I have removed this test in favor of one at the CLI level which still encompasses the result of this test as well as validating the dependency ranges were updated as expected.

PR #985, opened 2022-10-19
Fix(cli): the problem of isListablePackage condition

mino01x:


packages/cli/src/commands/add/isListablePackage.ts
```diff
@@ -8,7 +8,7 @@ export function isListablePackage(config: Config, packageJson: PackageJSON) {
     return false;
   }
 
-  if (!config.privatePackages && packageJson.private) {
+  if (!config.privatePackages?.version && packageJson.private) {
```
Andarist, line 11: Could you add a test case that would cover the fix?

Isn't `config.privatePackages` always an object? If yes, shouldn't it be this?
```suggestion
  if (!config.privatePackages.version && packageJson.private) {
```

mino01x: i have added test case, is that ok?

JakeGinnivan (approved):
lgtm

JakeGinnivan:
Just needs a changeset

packages/cli/src/commands/add/__tests__/add.ts
```diff
@@ -230,4 +230,25 @@ describe("Changesets", () => {
     const { choices } = askCheckboxPlus.mock.calls[0][1][0];
     expect(choices).toEqual(["pkg-a", "pkg-c"]);
   });
+
+  it("should not include private packages with a version in the prompt", async () => {
```
Andarist, line 234:
```suggestion
  it("should not include private packages with a version in the prompt if private packages are configured to be not versionable", async () => {
```

PR #993, opened 2022-10-27
Add support for CLI parameters to specify the version and message in the add() command

### LINKED ISSUE #979: [New Feature] support command way to add changelog
## Affected Packages

@changesets/cli@2.25.0

## Problem

Wanna support command way to add changelog

Now If I add a changelog, I need `Select the packages you want to include in the changeset using ↑ and ↓ to navigate to packages, and space to select a package. Hit enter when all desired packages are selected.`

It doesn't suitable on ci pipeline

I wanna support command way to add a changelog and autoly execute the command by CI pipeline, like this:

```
changeset add --major -m "this release info and description"
```

```
# then I can get branch commit description to add changelog when the branch merged
git log --pretty=format:"%b" HEAD -2 | xargs changeset add --major -m
```

## Proposed solution

I havent

TomOConnor95:
First time contributor

Saw this Issue and thought I'd have a go at solving it:  https://github.com/changesets/changesets/issues/979

I'm not sure of the steps to actually run this locally to test, but I think it should work ok :)

The --version parameter will only work for single package repos. 
The --message parameter should work for monorepos with several packages


packages/cli/README.md
```diff
@@ -137,6 +137,11 @@ If you set the `commit` option in the config, the command will add the updated c
 
 - `--open` - opens the created changeset in an external editor
 
+- `--version` - Creates a changeset with the specified level,
```
JakeGinnivan, line 140: I think this should be called `change` which lines up with what the CLI asks

image

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -218,12 +222,21 @@ export default async function createChangeset(
...
-    let type = await cli.askList(
-      `What kind of change is this for ${green(
-        pkg.packageJson.name
-      )}? (current version is ${pkg.packageJson.version})`,
-      ["patch", "minor", "major"]
-    );
+
+    let type: string | undefined;
+    if (options?.version) {
+      type = options?.version;
```
Andarist, line 228: nit:
```suggestion
      type = options.version;
```

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -233,12 +246,21 @@ export default async function createChangeset(
...
   }
 
-  log(
-    "Please enter a summary for this change (this will be in the changelogs)."
-  );
-  log(chalk.gray("  (submit empty line to open external editor)"));
+  let summary: string;
+  if (typeof options?.message === 'string') {
+    log(
+      `--message paramter passed: ${options.message}`
```
Andarist, line 252:
```suggestion
      `--message argument passed: ${options.message}`
```

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -233,12 +246,21 @@ export default async function createChangeset(
...
+    );
+    summary = options.message;
```
Andarist, line 254: we call it summary internally, but also in the prompts - should we then use `--message` as a CLI parameter? I see how it might be easier for users so I'm not entirely opposed to this but this makes the used terminology ambiguous

packages/cli/src/commands/add/index.ts
```diff
@@ -15,9 +15,17 @@ import printConfirmationMessage from "./messages";
 import { ExternalEditor } from "external-editor";
 import { isListablePackage } from "./isListablePackage";
 
+
+function isValidVersion(input?: string): input is 'major' | 'minor' | 'patch' {
```
Andarist, line 19: `'none'` is also a valid "bump type"

packages/cli/README.md
Andarist: Even in this very description here it's also called a "level". I'm not sure if that's the best name for the CLI parameter though - I'd like to have cohesive terminology though. So we should call this with a single name all over the place. I agree that `version` could be misleading here so I wouldn't chose that

Andarist: We call this a "bump type" here:
https://github.com/changesets/changesets/blob/1d3f9f29f4cfe3c32d18a98a1dc16b1bfce9e7e6/packages/apply-release-plan/src/utils.ts#L7
but a "version type" here:
https://github.com/changesets/changesets/blob/1d3f9f29f4cfe3c32d18a98a1dc16b1bfce9e7e6/packages/types/src/index.ts#L10

packages/cli/src/commands/add/index.ts
```diff
@@ -41,7 +49,18 @@ export default async function add(
...
 
-    newChangeset = await createChangeset(changedPackagesName, packages);
+    const changesetOptions: {
+      message?: string,
+      version?: 'major' | 'minor' | 'patch'
+    } = {
+      message,
+    }
+
+    if (isValidVersion(version)) {
```
Andarist, line 59: is an invalid argument is given by the user then I'd prefer throwing an error

packages/cli/src/commands/add/createChangeset.ts
TomOConnor95: Yeah I chose --message to be consistent with Git, but I'm happy to change it to --summary

The CLI asks this question "Please enter a summary for this change (this will be in the changelogs)." so I'll go with summary

TomOConnor95:
Thanks for the feedback on this PR. Made all the requested changes. 
The bits saying `const bumpType: VersionType` make me think the parameter should maybe be called --version-type, but I'm interested to hear your feedback, and am happy to make another change if necessary

Also I'm not sure how to actually test this change locally. Can you give me some pointers?
(note: I couldn't find a contribution guide in the repo)

bilibiliou:
I review these code, nice solution, thx u contribution @TomOConnor95

TomOConnor95:
@Andarist / @JakeGinnivan Would be great if you could give this PR another review when you have a chance. I think all issues have been addressed, and it'd be great to get this released some time soon

thomasmery:
Hi there,

was looking for a way to add changesets during a CI pipeline

this PR looked liked it would allow to do so

is this abandoned?

thnks

TomOConnor95:
> Hi there,
> 
> was looking for a way to add changesets during a CI pipeline
> 
> this PR looked liked it would allow to do so
> 
> is this abandoned?
> 
> thnks

Hey, I'm still very keen to get this deployed, but have been waiting for next steps on this PR having fixed the previous requests.
I'm very happy to make any more changes necessary, just waiting on some direction.

@Andarist / @JakeGinnivan please could you give some clarity on the next steps for this?

Tom

thomasmery:
hi,

would really like to see this happen,

unless I'm mistaken this would be the only way to generate changesets automatically and have a full automated release cycle

I'm guessing this might not be desirable in many cases but in the case of some repetitive workflow with the same kind of changes it would be very helpful

is there any other way to not have to go through the manual steps of producing a changeset?

thanks

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -218,12 +222,21 @@ export default async function createChangeset(
     }
   } else {
```
joekur, line 223: It looks like this option would only be available for single package repos. I would personally like an option like this compatible with monorepos, if it would also accept a cli argument such as `--package my-package` (or even better, some form that allows you to provide a list of packages `--package first-package --package second-package`). At the very least I think it should be otherwise documented, and possibly an error thrown if you try to use `--bump-type` within a monorepo.

For my use-case it would be an acceptable limitation that only a single `--bump-type` be allowed for creating a changeset against multiple packages.

aalv1234:
> hi,
> 
> would really like to see this happen,
> 
> unless I'm mistaken this would be the only way to generate changesets automatically and have a full automated release cycle
> 
> I'm guessing this might not be desirable in many cases but in the case of some repetitive workflow with the same kind of changes it would be very helpful
> 
> is there any other way to not have to go through the manual steps of producing a changeset?
> 
> thanks

+1 would be great to be able to version the change via the command line so we could automate our releases

TomOConnor95:
> +1 would be great to be able to version the change via the command line so we could automate our releases

It'd be great to see this merged. I've been waiting on approval from @Andarist/ @JakeGinnivan  , having made the requested changes.

I have seen the comment from @joekur about multi-package repos:
I'd be happy to improve this PR to handle those cases better. But it'd be great to have some guidance form @Andarist / @JakeGinnivan on which options they'd like for this, to avoid implementing the wrong thing.

packages/cli/src/commands/add/createChangeset.ts
aalv1234: For our monorepo the majority of the time you're only allowed to update single packages at a time.

When you update multiple packages they're required to be the same bump version because it will be just be updating configs and the version type ('major', 'minor', 'patch') will be the same for all packages. 

makes it possible to specify a single version for all changes

aalv1234:
@TomOConnor95

is your PR here compatible with monorepos, meaning via the cli we can specify the version for all the packages that require version bumps?

ekans:
It's exactly what I need to automate the creation of a changeset a bit more!

IHMO this PR should be split in 2 parts: one with `--bum-type` and one with `--summary` because `--summary` works with monorepos but `--bum-type` doesn't.

In this case, the PR with `--summary` would be ready to merge. WDYT?

Andarist:
Overlapping changes were implemented by https://github.com/changesets/changesets/pull/1121 and we just landed this PR.

I know this PR here was first - but the other one caught my attention and it I liked its `--patch` approach better as that supports better a mix of package names and bump types (each `--patch`, `--minor` and `--major` can separately provide a list of package names).

Nevertheless, thank you for your PR and the contribution. I'm sorry we didn't get to yours earlier.

TomOConnor95:
Glad that this has ended up being sorted out in the end :) thanks!

PR #1005, opened 2022-11-05
fix(changelog-git): typo on variable name

tbor00:
In this pr, fixed a typo in a variable name. It is a really very small change that does not affect any functionality or process.

### DIFF
--- a/packages/changelog-git/src/index.ts
+++ b/packages/changelog-git/src/index.ts
@@ -37,11 +37,11 @@ const getDependencyReleaseLine = async (
       }`
   );
 
-  const updatedDependencies = dependenciesUpdated.map(
+  const updatedDependenciesList = dependenciesUpdated.map(
     (dependency) => `  - ${dependency.name}@${dependency.newVersion}`
   );
 
-  return [...changesetLinks, ...updatedDependencies].join("\n");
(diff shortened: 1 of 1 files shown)


packages/changelog-git/src/index.ts
```diff
@@ -37,11 +37,11 @@ const getDependencyReleaseLine = async (
       }`
   );
 
-  const updatedDepenenciesList = dependenciesUpdated.map(
+  const updatedDependencies = dependenciesUpdated.map(
```
Andarist, line 40:
```suggestion
  const updatedDependenciesList = dependenciesUpdated.map(
```

packages/changelog-git/src/index.ts
```diff
@@ -37,11 +37,11 @@ const getDependencyReleaseLine = async (
...
     (dependency) => `  - ${dependency.name}@${dependency.newVersion}`
   );
 
-  return [...changesetLinks, ...updatedDepenenciesList].join("\n");
+  return [...changesetLinks, ...updatedDependencies].join("\n");
```
Andarist, line 44:
```suggestion
  return [...changesetLinks, ...updatedDependenciesList].join("\n");
```

PR #1012, opened 2022-11-19
chore: Update TypeScript, ESLint, Babel and Jest dependencies

chaance:
Just figured I'd do some housekeeping! Updating Jest results in some new snapshot formatting that required updates to those tests, though I checked the underlying assertions and they should be unchanged.

I updated TypeScript to 4.8 because ESLint's parser hasn't quite caught up to support 4.9 yet (it technically works fine, but I figured it's safer to just chill for a bit on that one).

I didn't include any Changesets here since I only touched dev dependencies. Let me know if you'd prefer to include one anyway.


chaance:
It does look like CircleCI is running a Node 12 image which is incompatible with this change. Any harm in updating to 16 since we're well beyond EOL for 12?

Andarist:
> It does look like CircleCI is running a Node 12 image which is incompatible with this change. Any harm in updating to 16 since we're well beyond EOL for 12?

I think it's fine. The supported node version for Changesets is not well-defined (our mistake but at the moment it's hard to know the node requirements of our whole dependency graph).

PR #1029, opened 2022-12-12
Cleanup `@changesets/git` tests

Andarist:
This PR:
- removes `beforeEach` hooks in favor of explicit setup per test
- unifies all methods to work on full commit SHAs by default (this wasn't consistent and it created annoying issues in tests when comparing full and short SHAs of the same commit). You can still request a short SHA explicitly from `getCurrentCommitId`

TODO:
- [ ] add appropriate changesets for the `@changesets/git` package since the second change could be seen as a breaking change

### DIFF
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -57,22 +57,15 @@ export async function getDivergedCommit(cwd: string, ref: string) {
   return cmd.stdout.toString().trim();
 }
 
-export const getCommitThatAddsFile = deprecate(
-  async (gitPath: string, cwd: string) => {
-    return (await getCommitsThatAddFiles([gitPath], cwd))[0];
-  },
-  "Use the bulk getCommitsThatAddFiles function instead"
-);
-
 /**
  * Get the SHAs for the commits that added files, including automatically
  * extending a shallow clone if necessary to determine any commits.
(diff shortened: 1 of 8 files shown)


emmatown (commented):
Seems fine besides missing missing changesets

packages/git/src/index.ts
```diff
@@ -57,24 +56,17 @@ export async function getDivergedCommit(cwd: string, ref: string) {
...
+ * Get the SHAs for the commits that added files, including automatically
  * extending a shallow clone if necessary to determine any commits.
  * @param gitPaths - Paths to fetch
- * @param cwd - Location of the repository
+ * @param options - `cwd` and `short`
  */
 export async function getCommitsThatAddFiles(
   gitPaths: string[],
-  cwd: string
+  { cwd, short = false }: { cwd: string; short?: boolean }
```
chaance, line 67: Since this is exposed API, should we include a changeset to document it?

chaance (changes requested):
Looks good overall. Just wondering if we need a changeset to top it off.

packages/git/src/index.ts
Andarist: Yeap, gonna add changesets before merging this - a few minor breaking changes got included here as part of the refactor.

Issue #1020, opened 2022-12-01
`changeset status`: allow JSON output to stdout


Dzieni:
## Affected Packages

@changesets/cli

## Problem

For scripting, I want to parse the JSON output of `changeset status`. The most convenient way for me would be to simply print that JSON to stdout, but now the tool does not allow me to do that (or it's not well documented).

## Proposed solution

IMO the best would be the `--format` flag that prints to stdout in any given format, starting from `json`. It would allow to easily support any other kind of output if needed.

Andarist:
We definitely want to support something like that. Would you be open to providing a PR for this?

Dzieni:
Sure! Will try to find a moment to make that feature :)

PR #1030, opened 2022-12-13
add new --format flag to status command to enable json output

simonxabris:
This PR adds the functionality described in #1020.

I added a new `--format` flag that accepts a file format (for now `json`, but `yaml` or whatever else could be added anytime) and prints the result of the `status` command to stdout.

The test for it is not the most resilient as I couldn't figure out how to do it properly. I wanted to spy on `console.log` and verifiy it was called with the right JSON, but couldn't do it as the `Expected` and `Received` JSONs never matched, if anyone has any idea how to write a better test for this please let me know.


MageJohn:
Hey, it would be great to have this feature. What needs to happen for it to get merged?

nedredmond:
How can I help get this over the line? Parsing the out all the 🦋 from the output is easy but painful when every other CLI tool provides a way to get usable JSON.

nedredmond:
How can I help get this over the line? Parsing the out all the 🦋 from the output is easy but painful when every other CLI tool provides a way to get usable JSON. @Andarist ?

Kworz:
Hey! is there something to do to get this merged ? lmk if i can do something 😉

onlywei:
It looks like there are conflicts that need to be resolved. That's not surprising since this PR is 2 years old. I still want this feature just like everyone else, though!

PR #1033, opened 2022-12-14
 Add support for the `changedFilePatterns` config

Andarist:
This PR introduces a new config option: `changedFilesPatterns`.

At first, I thought about introducing this just as a CLI argument for `changeset status`. However, I realized that `changeset add` should also use this to split packages between changed and not changed in the prompt. That's why I went with the config option approach.

The main purpose of this new feature is to have a way to optionally validate PRs that changesets are added for changed packages but without requiring changesets to be added for test file changes, docs changes etc.


packages/config/schema.json
```diff
@@ -100,6 +100,14 @@
...
       "description": "Determines the branch that Changesets uses when finding what packages have changed."
     },
+    "changedFilesPatterns": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "default": ["**"],
+      "description": "Glob patterns for changed files that should determining the package as changed."
+    },
```
Andarist, line 110: I went with an array because it's much easier to write a couple of small globs than one big one - especially if some negated pattern is involved.

packages/git/src/index.ts
```diff
@@ -250,26 +249,36 @@ export async function getChangedChangesetFilesSinceRef({
 export async function getChangedPackagesSinceRef({
   cwd,
   ref,
+  changedFilesPatterns = ["**"],
 }: {
   cwd: string;
   ref: string;
-}) {
+  changedFilesPatterns?: readonly string[];
+}): Promise<Package[]> {
```
Andarist, line 257: The core of the change is really in this method. I also rewrote this looping to be more efficient.

scripts/test-utils/src/index.ts
```diff
@@ -102,3 +103,33 @@ export async function testdir(dir: Fixture) {
 }
 
 export const tempdir = f.temp;
+
+export async function gitdir(dir: Fixture) {
```
Andarist, line 107: Extracted this here to reuse in the refactored tests for `changeset status`

packages/config/schema.json
Andarist: perhaps this should be named `changedFilePatterns`? WDYT?

Bunos07 (approved):

PR #1039, opened 2022-12-16
issue#815: exit with exit code 1 when cancelled

mbbillz:


packages/cli/src/utils/cli-utilities.ts
```diff
@@ -42,7 +42,7 @@ const limit = Math.max(termSize().rows - 5, 10);
 
 let cancelFlow = () => {
   success("Cancelled... 👋 ");
-  process.exit();
+  process.exit(1);
```
Andarist, line 45: Could you link to some prior art that does this on cancellation? I've just checked with `npm init` and it doesn't do that - it still returns with `0`.

Issue #1025, opened 2022-12-08
Integration with Rush: how to use with no root package.json file?


elliot-nelson:
## Problem

In a [Rush](https://rushjs.io) monorepo, there is no root `package.json` (it's generally forbidden and/or discouraged), and instead all of the paths for your individual projects are stored in a root `rush.json` file.

I'm not sure exactly what package or tool is looking for the root package:

```console
❯ changeset
🦋  error NoPkgJsonFound [Error]: No package.json could be found upwards from the directory /Users/enelson/dev/rushstack2
🦋  error     at _callee4$ (/Users/enelson/.nvm/versions/node/v16.15.0/lib/node_modules/@changesets/cli/node_modules/@manypkg/find-root/dist/find-root.cjs.dev.js:217:19)
🦋  error     at tryCatch (/Users/enelson/.nvm/versions/node/v16.15.0/lib/node_modules/@changesets/cli/node_modules/@babel/runtime/helpers/regeneratorRuntime.js:44:17)
🦋  error     at Generator.<anonymous> (/Users/enelson/.nvm/versions/node/v16.15.0/lib/node_modules/@changesets/cli/node_modules/@babel/runtime/helpers/regeneratorRuntime.js:125:22)
🦋  error     at Generator.next (/Users/enelson/.nvm/versions/node/v16.15.0/lib/node_modules/@changesets/cli/node_modules/@babel/runtime/helpers/regeneratorRuntime.js:69:21)
🦋  error     at asyncGeneratorStep (/Users/enelson/.nvm/versions/node/v16.15.0/lib/node_modules/@changesets/cli/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
🦋  error     at _next (/Users/enelson/.nvm/versions/node/v16.15.0/lib/node_modules/@changesets/cli/node_modules/@babel/runtime/helpers/asyncToGenerator.js:22:9) {
🦋  error   directory: '/Users/enelson/dev/rushstack2'
🦋  error }
```

## Proposed solution

I think first off, if I could get it to the point where at least it is not erroring, maybe I would be able to add a changeset if I explicitly told it what package it was? But ideally, I could add code to get the package list from `rush.json` `projects: []` field, and pass that into changeset to inform it of the projects in the repo.

elliot-nelson:
Ah ok, I see it: https://github.com/Thinkmill/manypkg/tree/master/packages/get-packages

[...]

PR #1044, opened 2022-12-21
fix: Support monorepos with no root package

elliot-nelson:
### SUMMARY

 - Do not require a "root package.json" when calculating the dependency graph

(Related issues: #1025)

### DETAILS

This change prepares the way for support for monorepo tools like Rush, which do not have a root package.json file.

> **NOTE:** Intended to merge after the type changes coming in https://github.com/Thinkmill/manypkg/pull/151. As-is, the unit tests pass but this change would fail tsc compilation.

### DIFF
--- a/.changeset/slow-berries-tie.md
+++ b/.changeset/slow-berries-tie.md
@@ -1,0 +1,5 @@
+---
+"@changesets/get-dependents-graph": patch
+---
+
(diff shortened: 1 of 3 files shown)


elliot-nelson:
Closing this PR in favor of #1069

PR #975, opened 2022-10-06
Rework dependency ranges updates

Andarist:

PR #1047, opened 2022-12-22
fix: Invalid comparator semver error when using tag versions

patzick:
Hey, great job with the project!

Small fix to cover the case when you're using tag versions.

Let's say you publish snapshots on npm with our `bulbosaur` example from docs :)
You have examples in monorepo which can be fired from web container like stackblitz or codesandbox, but we don't want (and should not) commit snapshot releases with versions back to repo. Tags are solutions to that problem, yet they're not a valid semver ranges, so we have for example (my case is `canary` tag):
image

which breaks whole release. I'll be very grateful for quickly releasing this fix if all looks good for you here. Cheers!


patzick:
hey @Andarist!
Would you mind taking a look at this PR? It'd be super helpful to have this fix released :)

Andarist:
Hmm, the goal of the change looks OK to me. if you are using a tag - that's your choice and Changesets should leave this alone. However, I wonder if your use case is correct - snapshot releases were not designed to be committed. Whatever changes they make to `package.json`/`CHANGELOG.md`/etc files should be treated as disposable.

packages/apply-release-plan/src/version-package.ts
```diff
@@ -84,14 +84,15 @@ export default function versionPackage(
           depCurrentVersion = workspaceDepVersion;
         }
         if (
+          semver.validRange(depCurrentVersion) !== null &&
```
Andarist, line 87: tags should already be allowed (and ignored), please take a look at the logic here:
https://github.com/changesets/changesets/blob/87e9bb819d7b08260de4bd8eec5cc71dd656b9a7/packages/get-dependents-graph/src/get-dependency-graph.ts#L114-L118

So if this isn't working for you... I wonder what went wrong in between. Why do we end up with a release for a package here if it's not part of the dep graph? Or is the package in question (example package) versioned on its own as well - so it wasn't pulled into the whole release by its dependencies but it's rather released on its own? If that's the case - why is it released/versioned? Could it be avoided?

patzick:
@Andarist thank you for your response. Maybe I wasn't clear but definitely we're not committing snapshots, but on the other hand we have stackblitz templates which should work with current master branch (so snapshot) published on npm, that's why templates are using snapshot tag, but on local envs it resolves to packages from monorepo. 

Quick real life case:
https://github.com/shopware/frontends/blob/main/templates/vue-demo-store/package.json#L15 template package file, which on stackblitz takes tag version from npm and locally from package.

But having this setup it crasches during `changelog version` and that's why I creates this PR :)

packages/apply-release-plan/src/version-package.ts
patzick: probably because the package itself is correct and we're upgrading this package version; but then it detects other dependency packages which are in monorepo and tries to upgrade this dependencies as well, which are marked as tags - and this is where the problem is. If you'd revert change that test will not pass with exactly the error I'm having while `changeset version` is invoked

Andarist:
Could you take a look at [my other comment](https://github.com/changesets/changesets/pull/1047#discussion_r1059352494)? I feel like I'm missing something about the nature of this bug (or your setup) - and I would like to commit this change only after understanding the full picture.

packages/apply-release-plan/src/version-package.ts
Andarist: So if we take [`vue-demo-store`](https://github.com/shopware/frontends/blob/65d6647c5b3ca94733c4660dfa54bfec6292afa4/templates/vue-demo-store/package.json#L2) as an example of this issue - do you explicitly create a changeset file for this `vue-demo-store` package? If yes - what for?

patzick: so you have

```
/packages
 /package-a --> 0.1.0
/templates
  /some-template --> 0.2.3
    dependency to package-a -> not workspace:* but TAG_NAME
```

so when you add changeset to `some-template` we should have it as for example `0.2.4` and nothing else. andwhen you update `package-a` then dependency at `some-template` should not change even if `some-template` has a correct range version

patzick: > So if we take [`vue-demo-store`](https://github.com/shopware/frontends/blob/65d6647c5b3ca94733c4660dfa54bfec6292afa4/templates/vue-demo-store/package.json#L2) as an example of this issue - do you explicitly create a changeset file for this `vue-demo-store` package? If yes - what for?

yes, we do create changeset for that, reasoning:

vue-demo-store is a template, it's not published on npm as it's not upgradable, but also templates have their issues, so thatnks to that we want to keep changelog for templates which features/fixes were added and when. It helps to keep cleaner history of changes and inform our users what has changed since last time they used this template

Andarist: Could you add a version-level test for this that would depict the whole flow more closely? You can add it [here](https://github.com/changesets/changesets/blob/87e9bb819d7b08260de4bd8eec5cc71dd656b9a7/packages/cli/src/commands/version/version.test.ts). Th test case that you already added shouldn't be removed.

patzick:
Great, I appreciate you taking the time to dive deeper into that to understand my issue. I'm aware this setup is not a common one, but I have good reasons for that, which I hope are also making sense to you :)

packages/apply-release-plan/src/version-package.ts
patzick: sure! I'm on it

packages/cli/src/commands/version/version.test.ts
```diff
@@ -1839,6 +1839,85 @@ describe("updateInternalDependents: always", () => {
...
+    expect(getChangelog("pkg-b", spy.mock.calls)).toMatchInlineSnapshot(`
+      "# pkg-b
+
+      ## 1.0.1
+
+      ### Patch Changes
+
+      - g1th4sh: This is some fix
+      - Updated dependencies [g1th4sh]
+        - pkg-a@1.1.0
```
patzick, line 1917: @Andarist thanks for the suggestion on adding an additional test here. This case covers our case when versioning crashed before, although this part is the part which made me think if that's the result you want to have.

From my perspective having this info in the changelog, even if the actual version is still a tag, is desired and helpful, as the tag was chosen not by mistake; but I can understand that could also be confusing for someone.
Could you give me your thoughts on that?

patzick:
hey @Andarist, can I ask you to re-review this one? :)

patzick:
hello @Andarist, you're probably busy, it there anyone else I could ping to make this go forward?

patzick:
Hey @Andarist, as a patch it's working for 5 months now. Is there anything blocking you from merging that? Just to refresh memory - it's just making sure the whole process is not failing when we have an npm tag version instead of semver.

Andarist:
@patzick could you resolve conflicts here? Sorry for not getting back to you earlier here

patzick:
@Andarist, no problem, thanks for getting back at this!

The conflict was caused by this: https://github.com/changesets/changesets/pull/1176/files#diff-73bc94618d4ce1bc263e6ecb9e90dee49072a5d789137c6b530dccd1f99fbcabL91-R95
all resolved now and my import also changed in the same tree-shake convention. Test passing locally

patzick:
Hey @Andarist, can you merge it? After the new release, I needed to patch it once again in our code and I see new merge movements so probably new release is coming soon :)

packages/cli/src/commands/version/version.test.ts
Andarist: Yeaaah, I think that's not what we want here. The fact that `pkg-a` was bumped in the same release is somewhat incidental. Since `pkg-b` depends on `pkg-a@bulbasaur` it might depend on a completely different version than `pkg-a@^1.x`, it could be just anything. So we shouldn't print misleading information like this here.

packages/apply-release-plan/src/version-package.ts
```diff
@@ -85,14 +86,15 @@ export default function versionPackage(
           depCurrentVersion = workspaceDepVersion;
         }
         if (
+          semverValidRange(depCurrentVersion) !== null &&
```
Andarist, line 89: I think that this check should be moved to this condition:
https://github.com/changesets/changesets/blob/207a532d321687e508484bf1457ce014c9fe0787/packages/apply-release-plan/src/version-package.ts#L48C1-L50C1

The other checks there are already "discarding" the ranges that we should skip over and it makes sense to include this one as part of that group

Andarist: I also wonder why we just don't use the computed graph-based information here but well, it is what it is and let's not try to refactor things as part of this PR

packages/cli/src/commands/version/version.test.ts
Andarist: To fix this we need to discard those entries in a similar manner [here](https://github.com/changesets/changesets/blob/207a532d321687e508484bf1457ce014c9fe0787/packages/apply-release-plan/src/get-changelog-entry.ts#L65). It's kinda messy how all of those places are using different conditions. IIRC I started [this PR](https://github.com/changesets/changesets/pull/975) in an attempt to clean it up... perhaps it's time for me to revive that work 😅

patzick: perfect place - worked like a charm! Thanks for the hint here ❤️

packages/apply-release-plan/src/version-package.ts
patzick: That would be too early, as we also want to allow processing`workspace:` ranges, which technically are not correct semver ranges, so we need to check it first. Moving it sooner will be here: https://github.com/changesets/changesets/blob/207a532d321687e508484bf1457ce014c9fe0787/packages/apply-release-plan/src/version-package.ts#L67-L70
 I'll commit this change

patzick:
@Andarist, all changes applied to suggestions and tests are passing

packages/apply-release-plan/src/get-changelog-entry.ts
```diff
@@ -64,6 +65,7 @@ export default async function getChangelogEntry(
     const versionRange = dependencyVersionRange || peerDependencyVersionRange;
     return (
```
Andarist, line 66: this check here should allow `workspace:` dependencies - could you check if this is working alright after the change?

patzick: Right, I added a test scenario to cover that case as well!

Btw, that would be cool if `- Updated dependencies` entry also has other updated dependencies (external ones) in te same format.

For example, I update dependencies and add changeset `bump dependencies` but only internal dependencies are included in `- Updated dependencies` entry. Would detecting other dependency changes between changeset releases be something in the scope of this project? I'd create a feature request issue with a proper description for  that if you think that'd be useful

Andarist: The problem is that we don't have a good place to store this kind of information. On top of that, `- Updated dependencies` are automated and you propose to feed some extra manual information. I wonder if you couldn't build this using a custom changelog generator? It seems plausible that you could just detect all dependency bumps since the last release and generate what you need based on that.

packages/cli/src/commands/version/version.test.ts
```diff
@@ -1916,6 +1916,77 @@ describe("updateInternalDependents: always", () => {
       "
     `);
   });
+
+  it("should bump dependant if it's a 'workspace' range", async () => {
```
Andarist, line 1920: A test case like this already exists [here](https://github.com/changesets/changesets/blob/9c09b3d4a64e84237a0ac6c03a04be70d5318c92/packages/cli/src/commands/version/version.test.ts#L997). This one is specific to the changelog change so I'd focus on that in the test title. Please also only assert on CHANGELOG updates to make the test smaller and more readable. You could also move it to a different `describe` block since this test isn't really that specific to the `updateInternalDependents` option

Andarist:
LGTM, it only needs some minor code shuffling related to the added test case.

packages/apply-release-plan/src/get-changelog-entry.ts
patzick: I'll experiment with that at some point probably and get back here if I discover some non-invasive way of doing it :) Thanks!

packages/apply-release-plan/src/index.test.ts
```diff
@@ -1356,6 +1356,111 @@ describe("apply release plan", () => {
             },
           });
         });
+        it("should not update internal version for tags", async () => {
```
Andarist, line 1359: I wonder, is this case right now? Shouldn't it be just using a `pkg-a: patch` changeset or smth like that to test that pkg-b **doesn't** get updated transitively? Perhaps I misunderstand the intention of this test though

patzick: The purpose of this test is that `dependencies` are not changed (not bumped) as they're using the `tag` version, not a numeric version. So even if both packages are updated, they should not change anything inside dependencies as tags should be static, not semver specific :)

patzick:
Woah, I somehow missed that there was another question here. As I thought `LGTM` covered all changes and it's merged already :)
 @Andarist do you think we can merge it so I won't have to patch it again when new version will be released? :)

patzick:
Hey @Andarist - is there anything else keeping you from merging this PR? 
Unfortunately, with every new release, I need to patch it with that fix to avoid crashes in the process :(

Let me know if I can assist somehow more to go through with that!

patzick:
hey @Andarist!
I'm coming back with the question if we could merge this fix, it has been tested (as patches) in our project for over a year and the only thing is doing is preventing crashes by additional checks; nothing fancy, very useful though :)

patzick:
Hello, @Andarist, I see some movements in the repo.
Could you please go back to this fix and merge it?

packages/apply-release-plan/src/get-changelog-entry.ts
```diff
@@ -3,6 +3,7 @@ import { ChangelogFunctions, NewChangesetWithCommit } from "@changesets/types";
 import { ModCompWithPackage } from "@changesets/types";
 import startCase from "lodash.startcase";
 import { shouldUpdateDependencyBasedOnConfig } from "./utils";
+import validRange from "semver/ranges/valid";
```
Andarist, line 6: We could also try doing smth about invalid tags (as in - continue to throw). To check what a valid tag is we could do:
```ts
function isValidTag(tag: string) {
  return encodeURIComponent(tag) === tag;
}
```
This can be verified [here](https://github.com/npm/npm-package-arg/blob/74d06ae66df6b8a91f5b5e4b8ff001a887cf4851/lib/npa.js#L391-L411). 

But honestly, I don't think it's worth it.

.changeset/famous-poems-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
+---
+
+Fixed `TypeError: Invalid comparator:` error when using tag versions.
```
Andarist, line 5:
```suggestion
Fixed a crash that could occur when depending on a tagged version of another workspace package.
```

.changeset/famous-poems-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
```
Andarist, line 2:
```suggestion
"@changesets/cli": patch
"@changesets/apply-release-plan": patch
```

Andarist:
better late than never

rmakara:
Respect! 💙

PR #1053, opened 2022-12-31
Replace depdendency `is-ci` by `ci-info`

paescuj:
`is-ci` is just a simple wrapper around `ci-info` and mainly intended to be used as a command line tool.
This makes it easier to follow updates of `ci-info` and as a nice side effect no additional package is needed for the types anymore.


.changeset/many-dodos-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Replace the dependency `is-ci` by `ci-info`.
```
Andarist, line 5:
```suggestion
Replace the dependency `is-ci` with `ci-info`.
```

PR #1069, opened 2023-01-27
Upgrade manypkg/get-packages

elliot-nelson:
### SUMMARY

 - Upgrade to latest major release of `@manypkg/get-packages`
 - This includes the reorganized `@manypkg/tools` collection

### DETAILS

 - Adjust various call points and checks to reflect the updated interface of `get-packages`.

For the most part, there should be no functional changes in this PR _except_ those introduced in the latest version of `get-packages`:

 - Rush monorepos are now parsed correctly and work for changeset creation + changeset version commands.
 - Bolt monorepos are no longer supported.

Note that for Rush specifically, changeset publish needs more work, which will come in a separate dedicated PR.

### TESTING

 - Updated existing unit tests
 - Verified behavior in local Rush monorepo


packages/release-utils/src/run.ts
```diff
@@ -44,7 +44,7 @@ export async function runPublish({
   let { packages, tool } = await getPackages(cwd);
   let releasedPackages: Package[] = [];
 
-  if (tool !== "root") {
+  if (tool.type !== "root") {
```
Andarist, line 47: We will have to also update this in `changesets/action`:
https://github.com/changesets/action/blob/595655c3eae7136ff5ba18200406898904362926/src/run.ts#L95

Andarist: `changesets/action` should be built on top of `@changesets/release-utils` - but this wasn't yet unified.

packages/apply-release-plan/src/index.ts
```diff
@@ -73,7 +73,7 @@ export default async function applyReleasePlan(
   config: Config = defaultConfig,
   snapshot?: string | boolean
 ) {
-  let cwd = packages.root.dir;
+  let cwd = packages.rootDir;
```
Andarist, line 76: If I'm not mistaken - this is a breaking change for `@changesets/apply-release-plan`. This is fine - we just have to create a major changeset for it

Andarist:
Could we also add a test that would check `changeset add` and `changeset version` in Rush monorepos?

packages/apply-release-plan/src/index.ts
elliot-nelson: Is it major? I might misunderstand the ramifications but I think apply-release-plan's external behavior for all existing monorepo implementations doesn't change (it's just a dependency bump + corresponding implementation tweak).

packages/release-utils/src/run.ts
elliot-nelson: Got it, will check it out.

As an aside, could be interesting to explore moving the source code for `actions` into the monorepo as a package. We've had some success with this with action collections in monorepos.

The "publish" command, instead of being NPM-based, becomes git-based: you build the action, bundle `dist` up into a single file (with `@vercel/ncc` or your favorite equivalent), and then your calculated version (e.g. `4.1.0`) becomes the branch name... you push the files directly into branch `v4.1.0` in the target repo, `changesets/action`.

(To follow GH conventions, you usually also have branches for major versions, so you follow it up by chopping off and fast-forwarding, e.g. `git checkout v4 && git reset --hard v4.1.0 && git push -f` in the target repo.)

This also brings up the whole question of "well... what if I wanted changesets to _manage_ publishing all my actions, using this alternate publishing scheme?".  That's a question we never really dug into Rush but I've always thought it would be cool if it was as simple as:

```
[
  {
    "packageName": "@example/core",
    "shouldPublish": true,
    "publishMethod": "pnpm"
  },
  {
    "packageName": "@example/custom-action",
    "shouldPublish": true,
    "publishMethod": "github-action",
    "targetRepo": "example-company/custom-action"
  }
]
```

And then you'd just "publish all" and everything is done for you.

Andarist: Ye, it would be kinda neat if the `changesets/action` repo could just become a publish target.

packages/git/src/index.test.ts
```diff
@@ -345,11 +345,11 @@ describe("git", () => {
         )[0];
         expect(commit).toEqual(originalCommit);
 
-        // It should not have completely unshallowed the clone; just enough.
+        // It should not have to completely unshallowed the clone; just enough.
         const originalRepoDepth = await getCommitCount(cwd);
         expect(await getCommitCount(clone)).toBeGreaterThan(5);
         expect(await getCommitCount(clone)).toBeLessThan(originalRepoDepth);
-      });
+      }, 10_000 /* 10 seconds */);
```
Andarist, line 352: Hm, quite interesting that you are hitting timeouts here - somewhat similarly to what happened in manypkg repo. There is likely some overhead in `spawn` here but I never had problems with this - even on the older machine.

packages/apply-release-plan/src/index.ts
Andarist: The problem is in the API surface - not the behavior. It now accepts an argument of a different shape

packages/git/src/index.test.ts
elliot-nelson: 🤔 I wonder why... on the other repo, upgrading that ts wrapper tool made the timeout issues vanish.

But in this case, it's only this specific test that consistently goes over 5 seconds for me, and it's the test that calls git 50+ times so it made sense to me that it was a bit slower.

Andarist: Out of curiosity - could you check how long those `spawn('git', ...)` calls take on your machine? It would be interesting to learn more about this particular bottleneck. This is, of course, just a sideways exploration - not quite related to this PR.

packages/apply-release-plan/src/index.ts
Andarist: 👋

akphi:
@elliot-nelson just curious, is this effort still being pursued?

Andarist:
@akphi ye, definitely

elliot-nelson:
Thanks for ping though looks like this is stalled on me 😅 I'll get those 2 additional tests added in next couple days.

akphi:
@elliot-nelson did you have chance to look at this again?

danez:
I think this will also fix this bug: https://github.com/changesets/changesets/issues/1209 
I was investigating this and created more tests for manypkgs, but there it worked fine. Only then I realized changesets was using an old version.

What is left here? I cannot see the circleCI pipeline, so not sure what's broken. Can I somehow help get this over the finish line?

brad-mccormick:
Anything we can do to help get this released? I'm at a standstill with a pnpm monorepo. @elliot-nelson

elliot-nelson:
@Andarist 👋 Hi Andarist, long time no chat :). I've added tests that shows basic `rush add` and `rush version` commands working in a Rush-style monorepo with no root package file, and all existing tests for non-Rush repos should work as well.

lachlancollins:
Hi @Andarist , I'm wondering if you could have a look at this old PR? The latest [manypkg release](https://github.com/Thinkmill/manypkg/pull/217) cuts a bunch of dependencies.

- @manypkg/get-packages@1.1.3 - [46 dependencies](https://npmgraph.js.org/?q=%40manypkg%2Fget-packages%401.1.3)
- @manypkg/get-packages@2.2.2 - [23 dependencies](https://npmgraph.js.org/?q=%40manypkg%2Fget-packages%402.2.2)

bluwy:
I know it's been a while, but if there's still interest in the PR, we've created a [`next` branch](https://github.com/changesets/changesets/tree/next) to combine all the breaking changes, and perhaps this PR can be part of it too. Happy to leave a review too if so.

PR #1085, opened 2023-02-09
Migrate to GitHub Actions

Andarist:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    name: Prepare Cache
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-node
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+  test:
```
Andarist, line 19: I skipped the install step in those jobs on purpose - because that's what Babel does:
https://github.com/babel/babel/blob/0c97749e0fe8ad845b902e0b23a24b308b0bf05d/.github/workflows/ci.yml#L55-L73

But it turns out that it's still required. I thought that maybe the builtin cache in `actions/setup-node` would somehow make it obsolete. I'm not sure why they don't need it - maybe it's some Yarn 3 magic?

package.json
```diff
@@ -51,7 +51,6 @@
     "@types/semver": "^6.0.0",
     "@typescript-eslint/eslint-plugin": "^5.43.0",
     "@typescript-eslint/parser": "^5.43.0",
-    "codecov": "^3.6.5",
```
Andarist, line 54: I'm using `codecov/codecov-action` now - so we don't need this dependency

.github/workflows/ci.yml
Andarist: I just included the install script in every job and the problem is fixed but I still wonder how does it work for Babel :P

Andarist: Oh, I've missed that they actually run the install script in other steps - just not in all of them:
https://github.com/babel/babel/blob/0c97749e0fe8ad845b902e0b23a24b308b0bf05d/.github/workflows/ci.yml#L176-L177

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
+name: CI
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  prepare-cache:
```
Andarist, line 9: The install step without the cache took... 23s. And now the install steps with the cache take like 13-15s. So there is no real gain here. I've checked some other repos using pnpm and the cache-based installs still take around 10+ seconds, so overall... I think that perhaps this "prepare" step could be removed altogether. 

I guess that the only thing that this prepare step saves us is the multiple runs of the cache upload step when in case of the cache bust. If we remove it then all 3 jobs might be "racing" to save the (identical) new artifact.

elliot-nelson: I personally wouldn't bother with this, it seems like in best case the only thing it can do is _add_ 10+ seconds to the tallest pole in your CI build.

elliot-nelson (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
+name: CI
+
+on: [push, pull_request]
```
elliot-nelson, line 3: Note: I think you want:

```yaml
on:
  pull_request:
  push: [main]
```

This way it will run on PR updates _or_ pushes to main (but won't double-run on branch push).

elliot-nelson:
Suggestion:

When you add, update, or refactor your jobs in your pull request workflow, there's an unfortunate issue where you then need to go and revamp the Required Status Checks in your Branch Protection settings as well (which requires admin approval).

There's a pattern to fix that, which I recommend: use a token job called `pr-ready` to indicate that a PR is ready to merge, and make _that_ your Required Status Check. The jobs this `pr-ready` token job depends on becomes your quality gate for merging, which can be adjusted in the future and only requires approval from the codeowners of your workflows, and not any Admin Settings adjustments.

Here's a typical example of this token job:

```yaml
  pr-ready:
    name: PR Ready
    runs-on: ubuntu-latest
    if: always()
    needs: [lint, test-coverage, build]
    env:
      FAILURE: ${{ contains(join(needs.*.result, ','), 'failure') }}
    steps:
      - name: Check for failure
        run: |
          echo $FAILURE
          if [ "$FAILURE" = "false" ]; then
            exit 0
          else
            exit 1
          fi
```

(You can add this to the bottom, replace the job ids in your "needs" section with whatever your job ids are in the yaml, and now you have a single "source of truth" status check for merging.)

PR #1095, opened 2023-02-20
add ability to publish multiple versions per package

yaacovCR:
useful when build script produces cjs/dual/esm versions

sample config from main "package.json":

```json
{
  ...
  "scripts": {
    "changeset:publish": "npm run build && changeset publish"
  },
  ...
}
```

and from individual packages:
```json
{
  ...
  "publishConfig": {
    "tag": "alpha",
    "versions": {
      "npm": {
        "directory": "npmDist"
      },
      "esm-only": {
        "suffix": "-esm",
        "directory": "npmEsmDist"
      }
    }
  }
  ...
}

### DIFF
--- a/packages/cli/src/commands/publish/publishPackages.ts
+++ b/packages/cli/src/commands/publish/publishPackages.ts
@@ -24,6 +24,13 @@ export type PublishedResult = {
   published: boolean;
 };
 
+interface PublishVersions {
+  [version: string]: {
+    suffix: string;
+    directory: string;
+  };
+}
+
 function getReleaseTag(pkgInfo: PkgInfo, preState?: PreState, tag?: string) {
   if (tag) return tag;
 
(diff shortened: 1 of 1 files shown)


yaacovCR:
maybe there is a better approach?

yaacovCR:
Goal: Allow `changesets` to kick off multiple new releases with different versions/tags whenever a release must be published.

Options:
1. write custom publish script instead of `changeset publish`

pro: complete control
con: we have to re-implement all functionality within `changeset publish`

2. call `changeset publish` more than once with different inputs

pro: nothing need change within `changeset publish`
cons:
(a) have to carefully update `changeset publish` in terms of detecting "unpublished packages" -- which version?
(b) seems confusing in a monorepo where the publishing behavior may differ per package

3. allow individual packages to declare that they should be published as separate versions

the approach of this PR

Andarist:
Hmm, I think this is rather a niche use case - usually, people are publishing different flavors of a package within the package itself. Any particular reason why do you want to publish multiple package names?

yaacovCR:
Yes we want an esm only version for purists and a dual cjs/esm version for broader compatibility.

Esm only reduces package size and avoids the dual package hazard when all dependents use it.

See https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c?permalink_comment_id=3850849#gistcomment-3850849

yaacovCR:
I started this issue for a project of mine, but it looks to me that this workflow and https://github.com/changesets/action/issues/262 are the technical impediments to introducing changesets at https://github.com/graphql/graphql-js so I wouldn’t call it niche per se, although there are perhaps better solutions.

another perhaps better option is for the esm only package to track releases of the dual version and extract the esm only version…

Andarist:
I realize that it would be easier for you if we'd support this within Changesets. However, this is the first time that somebody has asked for this and I still consider this to be a rather niche use case. I'd like to avoid needless complexity, configuration options etc.

That being said - this doesn't have to stop you from using Changesets in your project. All of this depends on how your current setup looks like... but I would do smth like this:
1. write a "prepare" script that builds all packages into multiple publish dirs etc
2. in that script adjust the root's `package.json#workspaces` to point to your publish dirs
3. call `changeset publish`
4. revert step 2

On top of that, you might want to actually auto-generate some changeset files for your extra ESM variants. I'm not sure if you want to simply duplicate changeset content or if you want to point to the dual package content or what - that's up to you.

Note that the content of this PR doesn't cover everything that would have to be done to land this as a feature here. All dependent packages should also get published, specified dependencies should likely be adjusted to their ESM-only variants etc. That's why I think this is, at the moment, a complexity that I'd like to avoid in the core. All of that can be scripted around Changesets.

yaacovCR:
That seems like a reasonable approach. I am actually using changesets in "non monorepo mode" so I haven't encountered all the extra complexity that dependent packages might entail. I'll give it a go -- thanks!

Any thoughts on https://github.com/changesets/action/issues/262 ?

yaacovCR:
It worked! Thanks again @Andarist !!!

PR #1096, opened 2023-02-22
README: Order "Cool projects" alphabetically, minor typo fixes

0liver:


bluwy:
Thanks for the PR, but I think the projects here had historically been first-come-first-serve so it's probably better to leave it as is for now.

Issue #568, opened 2021-04-13
Allow internal dependents to not be automatically bumped


mondash:
## Affected Packages

First time issue submitter here, so I'm not 100% sure, but I think only these packages would be affected:
- `apply-release-plan`
- `assemble-release-plan`
- `cli`
- `config`
- `types`

## Problem

I'll construct a contrived example here to illustrate my real-world problem.

I've got a monorepo set up with several shared libraries, `foo@1.0.0` and `bar@1.0.0`, to be used in other applications. These libraries may have internal dependencies. So let's say that `foo@1.0.0` has a dependency on `bar@^1.0.0`. I make a breaking change to `bar@1.0.0` and create a major version changeset bumping to `bar@2.0.0`.

The next time I run `yarn changeset version` (or the changeset action runs in CI), in addition to bumping `bar@2.0.0`, `foo@1.0.0` is automatically bumped to `foo@1.0.1`, and its internal dependency is bumped to `bar@^2.0.0`. As far as I understand, this is intended behavior.

But what if I needed to make some changes to the `foo` library before it can use `bar@2.0.0`? Sure, I could make those changes before releasing `bar@2.0.0` and publish both releases together, but should the release of `bar@2.0.0` be blocked by the dependent `foo` library? This problem compounds if there are many packages which all depend on `bar@1.0.0`, such that coordinating the update of each dependent to be compliant with the new version of `foo@2.0.0` _before even releasing_ `foo@2.0.0` becomes near impossible.

Here's my `.changesets/config.json` for reference:

```json
{
  "$schema": "https://unpkg.com/@changesets/config@1.6.0/schema.json",
  "changelog": "@changesets/cli/changelog",
  "commit": false,
  "linked": [],
  "access": "restricted",
  "baseBranch": "main",
  "updateInternalDependencies": "patch",
  "ignore": []
}
```

## Proposed solution

What I am suggesting is a new config option (perhaps `"updateInternalDependencies": "never"` or something) which would no longer update internal dependencies to a package when a new version of it is published. This means that each package would need to manually update it's internal dependencies, but it would allow the releases of packages with dependents to be decoupled from the releases of their dependents.

I've sorted this out in theory in my TS monorepo. Since yarn workspaces will install the outdated versions of the internal dependencies in the packages' `node_modules`, I used path aliases to point to either the internal dependencies location in that package's `node_modules` or to that packages `src` folder or to . This handles the case where we're using the current version of the internal dependency _and_ the case where we're using an outdated version.

[...]

PR #1108, opened 2023-03-13
Fix: use option for only workspace:^ pkgs (#568)

eihli:
Fixes https://github.com/changesets/changesets/issues/568

The bumpVersionsWithWorkspaceProtocolOnly was added in commit 12f9a433a6c3ac38f9405fcd77c9108c423d7101 in January 2021 but it never made its way to the `config` package.

The state of the world before this fix was that `changeset add` would respect the option and not bump packages that didn't have the `workspace` range when the option was true, but `changeset version` would log a warning:

	Package "foo" must depend on the current version of "bar:2.0.0"
vs "bar:1.0.0"

This fix just stops printing that message.


eihli:
What should the semver bump be here? I would expect it to be backwards compatible. But it does change something that dependents see. I suppose this could break someones scripts if they were depending on certain output. I hate for something so minor to be a major bump. But I think that's the safest bet.

eihli:
@Andarist I saw your comments and request for PR on the issue at https://github.com/changesets/changesets/issues/568. Can I ask you for your thoughts on this change?

PR #1122, opened 2023-03-28
Sort packages alphabetically in changeset add prompt

### LINKED ISSUE #1077: CLI changed packages should be sorted alphabetically
## Affected Packages

CLI

## Problem

Ordering of "changed packages" is by **length**

eg:

```
changed packages
  @foo/dddd
  @foo/aaa
  @foo/bb
```

## Proposed solution

Sort packages by alphabetical order

```
changed packages
  @foo/aaa
  @foo/bb
  @foo/dddd
```

Filipoliko:
Resolves #1077


packages/cli/src/commands/add/__tests__/add.ts
```diff
@@ -259,6 +259,44 @@ describe("Changesets", () => {
     );
   });
 
+  it("should sort packages alphabetically in the prompt", async () => {
```
Andarist, line 262: That likely won't test this behavior. I would probably create such layout to test it:
```
packages-1/pkg-b/package.json
packages-2/pkg-a/package.json
```

The point here is that `pkg-a` is contained in a directory that comes *after* the directory in which `pkg-b` is contained.

Filipoliko: You are right, thanks to a different mistake (the packages location did not match packages names), it seemed like it was doing the right thing, but not really.

I created 2 testing scenarios, one for `changedPackages` and one for `unchangedPackages` and named the directories `packages-1`/`packages-2` as suggested. It seems like the `unchangedPackages` are already sorting alphabetically, so the new `.sort` call does not really change anything, but I feel like it is still a good reassurance, that it will still work even if something in the background changes.

I tried removing the `.sort` call and at least for `uchangedPackages` use-case the tests failed, so it seems to be working.

Filipoliko:
@Andarist Thanks for the quick review, I updated the test, so hopefully it will work as expected. Feel free to check it when you have some more time :)

bluwy (approved):
I wonder if we should cut this as a minor since it could affect existing users' muscle memory, and is sort of a feature-ish.

bluwy:
I think it might be better to merge this in next altogether. In #1879 it'll already sort though

Andarist:
Yeah, @Filipoliko - I'm sorry this got stale and didn't get in. As mentioned, https://github.com/changesets/changesets/pull/1879 revamps this area quite a bit and kinda includes this change anyway. So it's better for us to close this PR to avoid future conflicts and stuff.

PR #1127, opened 2023-03-30
feat: add getVersionLine to changelog functions, update docs

cellog:
Our team needs to be able to have more control over the changelog format than what is provided. By adding `getVersionLine` as an optional new changelog generator, this gives us the flexibility we need without introducing any risk for existing users.

I also took it upon myself to document changelog functions because it wasn't that hard.

In addition, I fixed a crash that I encountered while trying to test the github changelog generator locally.

### DIFF
--- a/.changeset/cyan-geese-play.md
+++ b/.changeset/cyan-geese-play.md
@@ -1,0 +1,8 @@
+---
+"@changesets/apply-release-plan": minor
+---
+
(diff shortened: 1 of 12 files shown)


adhorodyski:
I like it, can be helpful for producing more user friendly backlogs. Is there any way we can push this forward?

mdorda:
Any plans to merge that?

Issue #960, opened 2022-09-24
Unexpected peerdependent bump to 1.0.0 in pre mode


hmans:
## Problem

I have a branch in `pre` mode, and I don't understand why Changesets is bumping this peerdependent to 1.0.0, even if the dependency's new version perfectly satisfies the specified range (0.11 is >= 0.9, right?)

![image](https://user-images.githubusercontent.com/1061/192098589-77765274-cdea-4526-bf24-a1d33c48db28.png)

I would have expected the specified peer dependency range to remain intact, and the dependent's version not to be bumped to 1.0.0.

My config.json:

```json
{
  "$schema": "https://unpkg.com/@changesets/config@2.0.0/schema.json",
  "changelog": "@changesets/cli/changelog",
  "commit": false,
  "fixed": [],
  "linked": [],
  "access": "public",
  "baseBranch": "main",
  "updateInternalDependencies": "minor",
  "ignore": [],
  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
    "onlyUpdatePeerDependentsWhenOutOfRange": true
  }
}
```

Version of changesets in use is 2.24.4.

Link to PR and repository: https://github.com/hmans/miniplex/pull/126/files

Thanks!

hmans:
Addendum: setting `"updateInternalDependencies": "patch"` didn't make a difference.

yishayweb:
Hi

I have a similar issue, just in regular mode
I am using [this repo](https://github.com/yishayweb/monorepo_pnpm_changesets) as a playground and I have managed to reproduce this issue there

## Affected Packages

I have two packages p13 and p14 where p13 is a peer dependency of p14

## Problem

When I make changes to p13 and make a patch bump to it, only the p13 is updated as expected.
But when I make changes to p13 and make a minor bump to it (like in [this PR](https://github.com/yishayweb/monorepo_pnpm_changesets/pull/59/files)) it also bumps p14 with major bump (see [this PR](https://github.com/yishayweb/monorepo_pnpm_changesets/pull/60/files) which is the versioning PR of the previous one)

Is this a bug or something expected?
If it's not a bug, is there a way I can prevent p14 from being bumped when changes are made only to p13?

I would really appreciate some help here
Thanks !!

Andarist:
Thanks for the repro case - I will try to analyze this case soon

hmans:
Thanks for taking a closer look!

[...]

Issue #1126, opened 2023-03-30
Do not but to major when peerDependency bumped to minor


B2o5T:
## Affected Packages

`@changesets/assemble-release-plan`

## Problem

Currently, when `"fixed"` option is specified with `"onlyUpdatePeerDependentsWhenOutOfRange": true`
```json
  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
    "onlyUpdatePeerDependentsWhenOutOfRange": true
  }
```

Packages are still bumped to major, to fix it I patch `assemble-release-plan` source code with the following patch https://github.com/shuding/nextra/blob/main/patches/%40changesets__assemble-release-plan%405.2.3.patch

## Proposed solution

New option or rethink behaviour

unional:
https://github.com/justland/just-web/pull/295

Got the same issue. @B2o5T thanks for the patch info!

PR #1132, opened 2023-04-07
Add a new bump strategy for peers bump

### LINKED ISSUE #1011: Unexpected major version bumps on one package
## Problem

Changesets has repeatedly bumped one of my packages' major version when it only receives patch or minor changes. Sometimes I'm paying attention in the PR and catch it, but at least 3 times it's gotten past me, so it's now 3 major versions ahead of other packages in the project.

An example PR (`@lo-fi/react` is bumped from 3.0.0 to 4.0.0 despite only having patch changes): https://github.com/a-type/lo-fi/pull/59

I'm not sure if this is intended behavior, but it's not desired. I'd prefer to have all my packages be in the same version range without having to manually edit changeset PRs.

I'm assuming the source of this behavior is in the main project and not the Github Action, but I could be wrong.

thynson:
If merged, resolve #1011, #960, #822

### DIFF
--- a/packages/config/schema.json
+++ b/packages/config/schema.json
@@ -122,6 +122,11 @@
       "description": "The minimum bump type to trigger automatic update of internal dependencies that are part of the same release.",
       "default": "patch"
     },
+    "updatePeerDependencies": {
+      "enum": ["major", "follow"],
+      "type": "string",
+      "description": "The strategy used to trigger automatic update of peer dependencies that are part of the same release, \"follow\" means if peer dependencies has a major/minor/patch bumps, this package will have a major/minor/patch bump too, respectively"
+    },
     "bumpVersionsWithWorkspaceProtocolOnly": {
       "type": "boolean",
       "default": false,
--- a/packages/config/schema.json
+++ b/packages/config/schema.json
@@ -132,6 +132,17 @@
       "default": false,
       "description": "Determines whether Changesets should only bump dependency ranges that use workspace protocol of packages that are part of the workspace."
     },
+    "autoBumpPeerDependentInSameChangeset": {
+      "type": "boolean",
+      "default": true,
+      "description": "Determines whether Changesets should automatically bump peer dependencies that are part of the same changeset."
+    },
+    "autoBumpPeerDependentCondition": {
+      "enum": ["always", "out-of-range"],
+      "type": "string",
+      "default": "always",
+      "description": "Determines the condition to automatically bump peer dependencies"
+    },
     "snapshot": {
       "type": "object",
       "properties": {
--- a/packages/config/schema.json
+++ b/packages/config/schema.json
@@ -122,26 +122,26 @@
       "description": "The minimum bump type to trigger automatic update of internal dependencies that are part of the same release.",
       "default": "patch"
     },
-    "updatePeerDependencies": {
-      "enum": ["major", "follow"],
-      "type": "string",
-      "description": "The strategy used to trigger automatic update of peer dependencies that are part of the same release, \"follow\" means if peer dependencies has a major/minor/patch bumps, this package will have a major/minor/patch bump too, respectively"
-    },
     "bumpVersionsWithWorkspaceProtocolOnly": {
       "type": "boolean",
       "default": false,
       "description": "Determines whether Changesets should only bump dependency ranges that use workspace protocol of packages that are part of the workspace."
     },
-    "autoBumpPeerDependentInSameChangeset": {
+    "autoBumpPeerDependentsInSameChangeset": {
       "type": "boolean",
       "default": true,
       "description": "Determines whether Changesets should automatically bump peer dependencies that are part of the same changeset."
     },
-    "autoBumpPeerDependentCondition": {
+    "autoBumpPeerDependentsCondition": {
       "enum": ["always", "out-of-range"],
       "type": "string",
       "default": "always",
       "description": "Determines the condition to automatically bump peer dependencies"
+    },
+    "autoBumpPeerDependentsStrategy": {
+      "enum": ["major", "follow"],
+      "type": "string",
+      "description": "The strategy used to trigger automatic update of peer dependencies that are part of the same release, \"follow\" means if peer dependencies has a major/minor/patch bumps, this package will have a major/minor/patch bump too, respectively"
     },
     "snapshot": {
       "type": "object",
--- a/.changeset/short-geese-taste.md
+++ b/.changeset/short-geese-taste.md
@@ -4,11 +4,15 @@
 "@changesets/types": minor
 ---
 
-Add a new bump strategy for peers bump.
+Add a serials of config options to control the behaviour of auto bumping of peers dependents.
 
-This adds a new option "updatePeersDependencies" to the config, and can be set
-to "major" (default) or "follow". When set to "major", it preserves the 
-origin behaviour that when peers dependencies has a minor or major bump,
-the dependent need to perform a major bump; when set to "follow", when peers 
-have a major/minor/patch bump, the dependent will need to perform **at least**
-a major/minor/patch bump, respectively.
+- `autoBumpPeerDependentsInSameChangeset`: boolean, default to `true` for compatibility reason,
+  if set to `false`, auto bumping of peer dependents will be disabled.
+
+- `autoBumpPeerDependentsCondition`: string enum of `"always"` or `"out-of-range"`, default to `"always"`
+  for compatibility reason, auto bumping of peer dependents will only be triggered when the bumped version
+  is out of semver range specified by the peer dependent.
+
+- `autoBumpPeerDependentsStrategy`: string enum of `"major"` or `"follow"`, default to `"major"` for
+  compatibility reason, if set to `"follow"`, the bump type of peer dependents will follow the bump type
+  of the bumped package.
--- a/.changeset/short-geese-taste.md
+++ b/.changeset/short-geese-taste.md
@@ -1,0 +1,14 @@
+---
+"@changesets/assemble-release-plan": minor
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Add a new bump strategy for peers bump.
+
+This adds a new option "updatePeersDependencies" to the config, and can be set
+to "major" (default) or "follow". When set to "major", it preserves the 
+origin behaviour that when peers dependencies has a minor or major bump,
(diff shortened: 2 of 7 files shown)


Andarist:
Thank you very much for your PR! I will review this thoroughly after the weekend.

packages/config/schema.json
```diff
@@ -122,6 +122,11 @@
       "description": "The minimum bump type to trigger automatic update of internal dependencies that are part of the same release.",
       "default": "patch"
     },
+    "updatePeerDependencies": {
```
Andarist, line 125: I think that we need a little bit more config for this - the last time I was thinking about it I got like 3 different options for the "what to do with the peer-dependent **version**" and a different setting for what to do about the peer-dependent's **dependency range**. 

I hope to gather more thoughts about this over the weekend and gonna get back to you with the desired config shape for this feature.

thynson: Thanks for review. I do also concern about what the correct config name/shape should be. And after writing the test, I came up with a question: is it necessary to follow the peers patch bumps too? 
Also, I believe this strategy is more reasonable than the current behavior, since a major bump is only required if a peer had a major bump, so would you consider make it the default strategy?

thynson:
>  "what to do with the peer-dependent version" and a different setting for what to do about the peer-dependent's dependency range.

I'm not clear if you were asking what to do when peers' bump exceeds the dependent's range.
If it is what you concern about, my thought to this question is still to follow the bump type of peer, despite if the dependent pinned the peer, or restricted the version of peers by tilde or caret, as long as we preserve the range type of peer dependencies after the bump. 
For example, if `pinned-peer` has a major bump, the dependent shall have a major bump too, and update the specifier from `pinned-peer@1.0.1` to `pinned-peer@2.0.0`.

thynson:
My current thought about shape of config:
- `updatePeerDependencies: "patch" | "minor"`:  The minimum bump type to trigger automatic update of peer dependencies that are part of the same release. This is consisted with `updateInternalDependencies`.
- `bumpStrategyForPeerUpdate: "major" | "follow"`:  The strategy used to trigger automatic update of peer dependencies that are part of the same release, `"follow"` means if a peer dependency has a major/minor/patch bumps, this package will have a major/minor/patch bump at minimum, respectively.

When `bumpStrategyForPeerUpdate` set to `follow`, with different `updatePeerDependencies` and different peers bump type, the result of bump type for the dependent should be:
| updatePeerDependencies |  Bump type of peer | Bump type for dependent
|-|-|-|
| `patch` | `major` | `major` |
| `patch` | `minor` | `minor` |
| `patch` | `patch` | `patch` |
| `minor` | `major` | `major` |
| `minor` | `minor` | `minor` |
| `minor` | `patch` | `minor` |

@Andarist What do you think about it?

hasparus:
@Andarist Is there anything an outsider can do to help this move past the finish line?

I have a pnpm patch, so I don't wanna rush, but I'm wondering if there's anything to do to support this. (e.g. write a PR updating docs? some tests?)

BTW https://github.com/changesets/changesets/issues/1126 seems related too.

Andarist:
> updatePeerDependencies: "patch" | "minor": The minimum bump type to trigger automatic update of peer dependencies that are part of the same release. This is consisted with updateInternalDependencies.

Hm, quite frankly - I'm personally almost always confused by `updateInternalDependencies` and I need to constantly remind myself what does it actually do. So I'm worried if using it as a prior art for this setting here is the best thing that we can do.

In XState we currently **always** update ranges in peer dependents (with a custom script). It feels quite safe to do that, we didn't really encounter any problems with this approach. The main difference is that we do it despite a peer dependent not being released in any particular release. Potential downsides of this approach:
- the current state of `package.json` might not exactly reflect what's published on npm but that can already happen in so many situations that I don't care about this
- if we end up releasing the peer dependent in the future - we won't exactly have a good way of knowing that a range in its `package.json` was touched so there won't be an easy way to add this information to its CHANGELOG etc. This can probably be alleviated by always updating the range when the peer dependent is published. And perhaps we should do the same for `updateInternalDependencies`?

> bumpStrategyForPeerUpdate: "major" | "follow": The strategy used to trigger automatic update of peer dependencies that are part of the same release, "follow" means if a peer dependency has a major/minor/patch bumps, this package will have a major/minor/patch bump at minimum, respectively.

I feel even stronger about this thing not relying on packages being part of the same release. This setting might even sometimes make a package to be included as part of the release. 

I feel like the proposed options need to be extended with smth like `out-of-range`. This should allow people to use `changeset version` in projects like Babel or React. In those projects there is a strong desire to leave the peer dep range untouched, often "pinned" to a specific major version (like `^7.0.0`) or to a version that contains some specific API (like `^16.8.0` if we think about React hooks). 

I think that perhaps the names of those options could use some bike-shedding but that's also the least interesting part at this stage of this PR. We can always rename things just before landing this. What's needed right now is the implementation and the docs.

thynson:
> I feel like the proposed options need to be extended with smth like out-of-range. 

Indeed. The version of peer dependent is still out of control.

The point people felt pain at most is that it requires major bump unexpectedly. People probably can tolerate it result in some patch bump, but ultimately we should give people more precisely control.

And during the time with the experience of my modified changeset, I felt `bumpStrategyForPeerUpdate` is not enough for this issue. 

I felt like we people should have a way to specified the the result bump type for a peer dependent, and probably the most intuitive way to achieve this is specify both bump type of peer dependency and peer dependent in a same changeset. 

That means, we should not bump the peer dependent automatically, when it was listed together with the peer dependency in the same changeset, as the developer already specified a desired bump type. 

Suppose `foo` is a peer dependencies of `bar`, when making a new release:

- If there is a changeset affects both `foo` and `bar`, e.g. a major bump for `foo` and a patch bump for `bar`, the developer expected `bar` should have a patch bump as he/she has the knowledge that the change does not result in breaking change of `bar`.  If there is some breaking change in `bar`, it's his/her responsibility to specify `bar` should have a major bump in the changset. 

- If there is a changeset affects only `foo`, we should automatically bump `bar`, depite whether there is another changeset that affects `bar` or not. The result bump for `bar` could be none under some configuration if `foo` still fit into the version range specified by `bar`.

We need an option to enable the behavior described in the first situation, e.g. `"autoBumpPeerDependentInSameChangeset": true | false`(default to `true` for compatibility), and when set to `"false"`, it just apply the kind of bump specified in the changeset.

An additional option `"autoBumpPeerDependent": "always" | "out-of-range"`(default to `"always"` for compatibility) may introduced for the second situation, and if it bumps automatically, consult `autoBumpStrategy="major" | "follow"`(default to `"major"` for compatibility) for what kind of bump it will be. 

Probably we could introduce `"never"` to `"autoBumpPeerDependent"` too, to disable auto bump totally, but it seems to be dangerous unless we change the interactive cli to force the user to specify bump types for all affected peer dependents.

@Andarist If finally we reach an agreement, I could probably work on it this weekend, but it could takes some times longer than expected.

Guergeiro:
Any update on this PR? Currently I have pipelines disabled for auto version bump because of this. Would really be nice if a simple patch version of `Package A` didn't cause a major bump on `Package B` if `Package A` is **_peerDependency_** of `Package B`.

thynson:
> We need an option to enable the behavior described in the first situation, e.g. `"autoBumpPeerDependentInSameChangeset": true | false`(default to `true` for compatibility), and when set to `"false"`, it just apply the kind of bump specified in the changeset.
> 
> An additional option `"autoBumpPeerDependent": "always" | "out-of-range"`(default to `"always"` for compatibility) may introduced for the second situation, and if it bumps automatically, consult `autoBumpStrategy="major" | "follow"`(default to `"major"` for compatibility) for what kind of bump it will be.
> 

The proposed changes were implemented in the latest commit. 
Three config were added:

- `autoBumpPeerDependentsInSameChangeset: true|false`, default to `true` for compatibility. When false, the auto bump behavior is completely disabled. And it's responsibility of end user to ensure that all affected peer dependents must be appear in the same changeset that bumps a peer dependency.
- `autoBumpPeerDependentsCondition: "always" | "out-of-range"`, default value depends on `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH.onlyUpdatePeerDependentsWhenOutOfRange` which defaults to `false`, so it defaults to `"always"` if neither option was configured; but the old experimental configuration can be removed after this PR merged.
- `autoBumpPeerDependentsStrategy: "major" | "follow"`: which was `updatePeerDependencies` that described before.

@Andarist Do we have a chance to get it reviewed and merged?

stevejcox:
Anything we can do to help this one along? This would be extremely helpful for my org

frodehansen2:
Any updates on this?

cef62:
Is there anything we can do to help merge this PR?

siuvdlec:
Any updates on this?

v0rs4:
Any updates on this as well?

fredericoo:
Anything we can do? I’m ready to sacrifice a goat to get this PR merged
@Andarist

EXDCUR1:
@Andarist +1 on this change being needed, any update on it getting merged?

JuniYadi:
any update?

our peer update with minor, and another package related to this peer has bump to major version

thynson:
We can stay on a fork before the author merges this PR.

igoldny:
@Andarist please review, this feature is really needed!

yonayarin:
Issue release a non wanted major in our company. please review it soon @Andarist

basilmdrecords:
any update on this?

erprilianoAbbas:
Hello? Any updates on this?

thynson:
Rebased upon the lastest HEAD. Wish for review. @Andarist

DRiFTy17:
It doesn't seem like there has been any movement on this in a while... Worth asking again, does anyone have any updates or workarounds?

I've been working around it by removing packages from `peerDependencies` for now, but that's obviously not great. This one is very frustrating after spending time converting many of our repos to a monorepo and then running into this problem while publishing at the end.

tylerbutler:
> I'd like to throw out an alternative option - allowing a .js configuration file that allows us to provide a function to configure the desired behavior.

See the discussion at https://github.com/changesets/changesets/pull/1367.

rikez:
Any updates on this?

trystan2k:
This is also affecting us. Any idea when this can be merged ?

xdebbie:
Any news? It's been two and a half years... it's too bad because changesets has been working really great for our design system project but the automatic major bumps are **_really_** undesirable and I can't stress this enough...

PR #1140, opened 2023-04-21
Fix `status --since` when .changeset isn't in the repo root

MageJohn:
The `filterChangesetsSinceRef` function assumed that (a) the path delimiter is a `/` character, and (b) that the `.changesets` folder is in the root of the repository. The first assumption doesn't hold on Windows, and the second assumption is doesn't hold in our repository, where we're using changesets to manage the changelogs of a subset of our packages.


sergunich:
Any chances to make it merged?

sergunich:
@Andarist, looks like this is a tidy and nicely done PR with a one-liner fix for the bug and even with additional test cases. And @MageJohn was kind enough to rebase it just recently!

I hope that I don't sound pesky, but it would be superb if you would have a chance to take a look and maybe merge this PR 🙂

MageJohn:
Hey @sergunich, just wanted to let you know that for our internal repo where we found this bug we worked around it by patching the package. yarn and pnpm have a `patch` subcommand for this, and for npm there is a package called `patch-package` available. 

I would love for this to be merged ASAP as well, but if this bug is causing you issues right now then I thought you'd appreciate a workaround. 😃

eimerreis:
How can this not be merged after 1 year?  

Is there any reason for it?

PR #1155, opened 2023-05-07
fix typo of automating-changesets.md

odanado:


benmccann (approved):
this will need to match line 32, so you're correct

PR #1177, opened 2023-06-26
chore: Correcting typos

Hansanghyeon:
There was a typo in the document 😄

스크린샷 2023-06-26 오후 4 23 01


Andarist:
`--since-master` is actually supported (you can't use this flavor with just any branch name though, there is an explicit flag supporting the `master` branch there). It is deprecated though in favor of `--since=BRANCH_NAME` so this is a fine change. Thanks!

PR #1183, opened 2023-07-02
feat: Support Publish since flag

igoldny:
Following my conversation with @Andarist - https://github.com/changesets/changesets/discussions/1083#discussioncomment-4892376
@Andarist suggested implementing a since flag for publish cli as well.

Using this flag, you will be able to release tag versions in your pull requests.
By adding the following flow in your ci

```
changeset version --snapshot=[TAG] 
changeset publish --since=[PULL_REQUEST_BRANCH]  --tag=[TAG] --snapshot
```


.github/workflows/changeset-version.yml
```diff
@@ -1,38 +0,0 @@
-name: Release
```
Andarist, line 1: this file shouldn't be removed

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,11 +0,0 @@
-## Affected Packages
```
Andarist, line 1: this file shouldn't be removed

packages/cli/src/commands/publish/index.ts
```diff
@@ -56,19 +61,24 @@ export default async function run(
     showNonLatestTagWarning(tag, preState);
   }
 
-  const { packages, tool } = await getPackages(cwd);
+  const { tool } = await getPackages(cwd);
+  const changedPackages = await git.getChangedPackagesSinceRef({
+    cwd,
+    ref: since || config.baseBranch,
+    changedFilePatterns: config.changedFilePatterns,
```
Andarist, line 68: `changedFilePatterns` shouldn't be used here - those patterns are only meant to filter lists in the CLI prompt and to be used in `changeset status`. They are intended to help answer the question of what changes in a package might have to result in a changeset. For example, you could ignore all `.md` files with it or scope "changes" to `src/**`

packages/cli/src/commands/publish/index.ts
```diff
@@ -56,19 +61,24 @@ export default async function run(
     showNonLatestTagWarning(tag, preState);
   }
 
-  const { packages, tool } = await getPackages(cwd);
+  const { tool } = await getPackages(cwd);
+  const changedPackages = await git.getChangedPackagesSinceRef({
```
Andarist, line 65: Similarly, this particular filtering shouldn't be done here at all. I expect that this would break quite a lot.

I think that what you want is to add `--since` to the `changeset version` and in there you might want to filter the existing changes by that `since` parameter. This way you'd allow the rest of the algorithm to work as expected - transitive dependencies etc would get pulled into the release which is a good thing (as per the comment [here](https://github.com/changesets/changesets/discussions/1083#discussioncomment-4892376))

igoldny: Thanks for the suggestion,
Agree it dose sounds safer,
Will make the adjustments...

igoldny: @Andarist Here is new pr with suggested change - https://github.com/changesets/changesets/pull/1206

Issue #1046, opened 2022-12-22
ESM packages doesn't work


boyeborg:
## Affected Packages

pkg:read

(probably more, if not all, but I've only looked into `read`)

## Problem

Reproduction: https://github.com/boyeborg/changesets-esm-issue-repro

When importing `@changesets/read` in an ESM module, and using the default exported function:

```javascript
import readChangesets from "@changesets/read";

const cwd = process.cwd();

readChangesets(cwd).then(console.log);
```

You'll get the following error:

```
readChangesets(cwd).then(console.log);
^

TypeError: readChangesets is not a function
```

It seems like it only uses the CJS version. You can work around it by accessing the function through `readChangesets.default`.

## Proposed solution

I'm no expert at this, but to me it looks like a misconfiguration. However, there are several issues.

First off, I don't think it's sufficient just to set `"module": "<path>"` in `package.json`. I think you also need to add some exports, like:

```
"exports": {
  ".": {
    "require": "./dist/read.cjs.js",
    "import": "./dist/read.esm.js"
  }
}
```

Secondly, I don't think you can use the `.esm.js` extension. I believe you have to use `.mjs` since you don't have `"type": "module"`in your package.json-file.

Lastly, there are some issues with the ESM file (`dist/read.esm.js`) itself where it's doing some named CJS imports, which are not allowed in ESM. Hence, it's not a valid ESM file.

I guess all of these issues might originate from Preconstruct, which I'm not familiar with, so I cannot tell if it's some configuration change that can be made, or if it's a downstream issue.

boyeborg:
Ah, just discovered this is a duplicate of https://github.com/changesets/changesets/issues/622.

I'll leave it open anyway, since there are aspects in this issue that's not covered in #622

Issue #622, opened 2021-08-12
Settings of `@changesets/read` are incorrect and cause problems when using native esm


JounQin:
## Affected Packages

pkg:@changesets/read

## Problem

`@changesets/read` is not set as `type: "module"` currently, so `dist/read.cjs.js` will be used in native esm, but [`dist/read.cjs.prod.js`](https://unpkg.com/browse/@changesets/read@0.4.7/dist/read.cjs.prod.js) is still wrapped into `exports.default` statement.

## Proposed solution

Change `dist/read.cjs.js` like following:

```js
'use strict';

if (process.env.NODE_ENV === "production") {
  module.exports = require("./read.cjs.prod.js").default;
} else {
  module.exports = require("./read.cjs.dev.js").default;
}
```

That would be nicer to migrate to `type: "module"` at the same time.

Andarist:
That's matching the declared interface. It's sort-of a node's problem that they have chosen to break compatibility with the existing tooling.

We'll be "fixing" this - but we can't say when exactly this will happen because it's a breaking change.

JounQin:
If you use `exports` field in `package.json` without any other changes, that would be still compatible for commonjs and native esm at same time.

```jsonc
{
  "exports": {
    "import": "xx.mjs", // notice `.mjs` is required
    "require": "xxx.js" // current commonjs entry
  }
}
```

No `type: "module"` required.

Andarist:
Unfortunately, to my great despair, this is not as easy. Right now you can use `@changesets/read` with ESM in node like this:
```js
import read from '@changesets/read'

read.default()
```
After the proposed change the semantics would change and only this would be possible:
```js
import read from '@changesets/read'

read()
```

So for existing users, this would classify as a breaking change.

JounQin:
I suppose no one will be using this pattern as it's just more rediculous than the original issue.

```js
import read from '@changesets/read'

read.default()
```

HenryNguyen5:
Just ran into this, I was able to work around it via:
```ts
import _readChangesets from '@changesets/read'
// See https://github.com/changesets/changesets/issues/622
const readChangesets =
  _readChangesets.default as typeof import('@changesets/read/dist/declarations/src/index.js')['default']
```

PR #1185, opened 2023-07-03
Use `package.json#exports`

Andarist:
closes https://github.com/changesets/changesets/issues/1046
closes https://github.com/changesets/changesets/issues/622

### DIFF
--- a/.changeset/healthy-cobras-worry.md
+++ b/.changeset/healthy-cobras-worry.md
@@ -21,4 +21,4 @@
 "@changesets/pre": minor
 ---
 
-`package.json#exports` have been added to limit what (and how) code might be imported from the package
+`package.json#exports` have been added to limit what (and how) code might be imported from the package.
--- a/.changeset/healthy-cobras-worry.md
+++ b/.changeset/healthy-cobras-worry.md
@@ -1,0 +1,24 @@
+---
+"@changesets/get-version-range-type": minor
+"@changesets/assemble-release-plan": minor
+"@changesets/get-dependents-graph": minor
+"@changesets/apply-release-plan": minor
+"@changesets/changelog-github": minor
+"@changesets/get-release-plan": minor
+"@changesets/get-github-info": minor
+"@changesets/changelog-git": minor
+"@changesets/release-utils": minor
+"@changesets/test-utils": minor
+"@changesets/config": minor
+"@changesets/errors": minor
+"@changesets/logger": minor
+"@changesets/parse": minor
+"@changesets/types": minor
+"@changesets/write": minor
+"@changesets/read": minor
+"@changesets/cli": minor
+"@changesets/git": minor
+"@changesets/pre": minor
+---
+
+`package.json#exports` have been added to limit what (and how) code might be imported from the package
--- a/packages/apply-release-plan/package.json
+++ b/packages/apply-release-plan/package.json
@@ -4,6 +4,18 @@
   "description": "Takes a release plan and applies it to packages",
   "main": "dist/changesets-apply-release-plan.cjs.js",
   "module": "dist/changesets-apply-release-plan.esm.js",
+  "exports": {
+    ".": {
+      "types": {
+        "import": "./dist/changesets-apply-release-plan.cjs.mjs",
+        "default": "./dist/changesets-apply-release-plan.cjs.js"
+      },
+      "module": "./dist/changesets-apply-release-plan.esm.js",
+      "import": "./dist/changesets-apply-release-plan.cjs.mjs",
+      "default": "./dist/changesets-apply-release-plan.cjs.js"
+    },
+    "./package.json": "./package.json"
+  },
   "license": "MIT",
   "repository": "https://github.com/changesets/changesets/tree/main/packages/apply-release-plan",
   "dependencies": {
--- a/packages/apply-release-plan/package.json
+++ b/packages/apply-release-plan/package.json
@@ -2,8 +2,8 @@
   "name": "@changesets/apply-release-plan",
   "version": "6.1.4",
   "description": "Takes a release plan and applies it to packages",
-  "main": "dist/apply-release-plan.cjs.js",
-  "module": "dist/apply-release-plan.esm.js",
+  "main": "dist/changesets-apply-release-plan.cjs.js",
+  "module": "dist/changesets-apply-release-plan.esm.js",
   "license": "MIT",
   "repository": "https://github.com/changesets/changesets/tree/main/packages/apply-release-plan",
   "dependencies": {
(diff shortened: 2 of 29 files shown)


.changeset/healthy-cobras-worry.md
```diff
@@ -0,0 +1,24 @@
+---
+"@changesets/get-version-range-type": minor
```
Andarist, line 2: Some of those are in the `0.x` range so I could consider using `patch` for them - it doesn't quite matter for them though anyway so I'm just using `minor` for every pkg here

emmatown (approved):
Look good besides the usual "technically, this could break people if they relied on .default" but it seems like that went fine for Emotion so probably ✅ ?

Andarist:
> Look good besides the usual "technically, this could break people if they relied on .default" but it seems like that went fine for Emotion so probably ✅ ?

Ye, based on that experience I'm totally on board with just shipping this in a minor. Can it break some people? Technically yes. Was the previous way of importing those packages in strict ESM env intentional? Not quite. Do we hear complaints from Emotion users? Not really - IMHO that's a sign that whoever we broke with this change actually welcomed the fix with arms wide open 😅

packages/apply-release-plan/package.json
```diff
@@ -2,8 +2,20 @@
   "name": "@changesets/apply-release-plan",
   "version": "6.1.4",
   "description": "Takes a release plan and applies it to packages",
-  "main": "dist/apply-release-plan.cjs.js",
-  "module": "dist/apply-release-plan.esm.js",
+  "main": "dist/changesets-apply-release-plan.cjs.js",
+  "module": "dist/changesets-apply-release-plan.esm.js",
```
benmccann, line 6: this field will have no effect. it will be ignored in favor of `exports` and can be removed

packages/apply-release-plan/package.json
```diff
@@ -2,8 +2,20 @@
...
+  "exports": {
+    ".": {
+      "types": {
+        "import": "./dist/changesets-apply-release-plan.cjs.mjs",
+        "default": "./dist/changesets-apply-release-plan.cjs.js"
+      },
+      "module": "./dist/changesets-apply-release-plan.esm.js",
```
benmccann, line 13: this is totally non-standard. is it really necessary? it would be much nicer not to have it

packages/apply-release-plan/package.json
```diff
@@ -2,8 +2,20 @@
...
+      "import": "./dist/changesets-apply-release-plan.cjs.mjs",
```
benmccann, line 14: I mean this in the nicest possible way, but it is absolutely crazy to use `.cjs.mjs` as a file extension since a file cannot be both CJS and ESM. Assuming this file is actually ESM, can the extension be just `.mjs`?

packages/apply-release-plan/package.json
```diff
@@ -2,8 +2,20 @@
...
+      "default": "./dist/changesets-apply-release-plan.cjs.js"
+    },
+    "./package.json": "./package.json"
```
benmccann, line 17: I'm afraid we started this convention because of some ill-behaved versions of `rollup-plugin-svelte` that would tell users to inform package authors to add it. It's really not necessary though and `rollup-plugin-svelte` no longer has that behavior

packages/apply-release-plan/package.json
Andarist: It's here to support older bundlers that don't support `exports`

packages/apply-release-plan/package.json
Andarist: It is totally standard :p it's just not described by node - the `exports` "standard" is meant to be extensible. This condition was born out of the collaboration between webpack and Rollup maintainers when it became clear that the dual package hazard is a thing. To mitigate the problem in bundlers that would like to apply node semantics closely they invented this condition that kinda works like the "old" `package.json#module`. Both require and import loaders in bundlers are able to load this, even though this file is authored in ESM. This is also adopted by other bundlers like Vite, bun, Parcel, esbuild and more - so I'd call it pretty standard :)

packages/apply-release-plan/package.json
Andarist: This is an MJS wrapper - it's an ESM file that reexports CJS file. I know it looks funny but it kinda makes perfect sense to us. It's also kinda irrelevant rly - the extension is `.mjs` here. The other dotted part is **not** considered to be an extension, we just use it to denote some additional meaning that is separate from the package name (hence the need to use a different delimiter than a dash)

packages/apply-release-plan/package.json
Andarist: It's to support `require('pkg/package.json')` in node. We think that there is no big cost associated with supporting this and some `node_modules` crawlers (and similar tools) are using this to inspect package metadata.

packages/apply-release-plan/package.json
benmccann: Ah, thank you for educating me! I had no idea Rollup supported this, but it looks like it's been in `@rollup/plugin-node` since 11.0.0. Very much appreciate the informative responses!

packages/apply-release-plan/package.json
benmccann: That's helpful to know. While I agree it's irrelevant to consumers of the package, it isn't necessarily intuitive to people who might want to contribute to it. Something like `-wrapper.mjs` might be a little clearer

packages/apply-release-plan/package.json
Andarist: I agree but this is auto-generated by a somewhat universal library builder ( https://github.com/preconstruct/preconstruct/ ) and for a lot of npm packages it might still make sense to have proper compat with webpack 4 etc.

packages/apply-release-plan/package.json
Andarist: Well, usually you just don't touch those files at all. When contributing you just interact directly with sources and all of this here is auto-generated for us. I appreciate the feedback but at the end of the day, I consider this particular thing a very minor detail that is kinda irrelevant. I can see how it might be surprising at first glance though.

mblackrittr:
Super thanks, @Andarist, @emmatown and @benmccann for this "life-savør"! 😻 🥳🎉

I hit the ESM issue while being at some improvements we're working on for [Changeset-Conventional-Commits](https://github.com/iamchathu/changeset-conventional-commits) and was worried at first (it was still set to `"@changesets/read": "^0.5.9"`), but then found `6.0.0` and this PR to my final relief, so now it also runs the latest `Meow` (pure ESM) and all - **great!** 😹💦

mblackrittr:
Super thanks, @Andarist, @emmatown and @benmccann for this "life-savør"! 😻 🥳🎉

I hit the ESM issue while being at some improvements we're working on for [Changeset-Conventional-Commits](https://github.com/iamchathu/changeset-conventional-commits/pulls) and was worried at first (it was still set to `"@changesets/read": "^0.5.9"`), but then found `6.0.0` and this PR to my final relief, so now it also runs the latest [Meow](https://github.com/sindresorhus/meow) (pure ESM) and all - **great!** 😹💦

PR #1186, opened 2023-07-05
Adjust bump types related to the `exports` change

Andarist:
follow up to https://github.com/changesets/changesets/pull/1185


emmatown (approved):
(I'm assuming these are the correct bumps, I haven't checked the current versions of all these things)

Andarist:
I tried my best to go through all of those and adjust the ones that had to be changed 😉

PR #1198, opened 2023-08-01
Add commit step before publish

SevenOutman:
I followed this guide to publish packages but the tag was assigned to the commit before I run `changeset version`.

So I want to add a "commit changes" step before running `changeset publish`.

### DIFF
--- a/docs/intro-to-using-changesets.md
+++ b/docs/intro-to-using-changesets.md
@@ -56,15 +56,17 @@ yarn changeset version
 
 This consumes all changesets, and updates to the most appropriate semver version based on those changesets. It also writes changelog entries for each consumed changeset.
 
-We recommend at this step reviewing both the changelog entries and the version changes for packages. Once you are confident that these are correct, and have made any necessary tweaks to changelogs, you can publish your packages:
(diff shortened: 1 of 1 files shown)


altano:
This contradicts [the documentation for publish](https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish):

> Because this command assumes that the last commit is the release commit, you should not commit any changes between calling version and publish. These commands are separate to enable you to check if the release changes are accurate.

SevenOutman:
> Because this command assumes that the last commit is the release commit, you should not commit any changes between calling version and publish. These commands are separate to enable you to check if the release changes are accurate.

@altano To be honest I don't quite understand why this statement makes sense given `changesets version` doesn't commit changes. Here's what happens in sequence if you follow this statement.

1. You have your changes to your package ready. The last commit is the last change you've made, and the `version` in `package.json` is not bumped yet.
2. You run `changesets version` and `version` in `package.json` is bumped, changelogs are generated - but these changes are not committed because `version` command doesn't do that.
3. You don't commit these changes either, following the guide. The last commit is still the last change you've made, with the `version` in `package.json` unbumped.
4. You run `changesets publish` and the last commit with old `version` in `package.json` is tagged as new version.

Note that in step 3, if you don't commit changes your self, then the last commit does not satisfy `publish` command's assumption that it's the release commit - it's not.

PR #1240, opened 2023-09-30
fix branch name

chronark:
The code snippet is using `main` already but the next line wasn't updated I think.


ArnavK-09 (approved):

bluwy:
Looks like this was already updated in https://github.com/changesets/changesets/pull/1155

PR #1248, opened 2023-10-19
Allow opting-out of username links in release notes

lilnasy:
This PR adds a config to `@changesets/changelog-github` to remove links from GitHub usernames, allowing contributors to appear prominently in releases in the "Contributors" section.

To make use of this feature, set `linkUsernames` to `false` in your .changeset/config.json file:

```diff
{
    "$schema": "https://unpkg.com/@changesets/config@1.7.0/schema.json",
    "changelog": [
        "@changesets/changelog-github",
        {
            "repo": "your-org/your-repo",
+           "linkUsernames": false
        }
    ]
}
```

See [GitHub documentation](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository) for more details.


Andarist:
I just wonder if we shouldn't transform the release notes and unlinkify those there. The downside of the proposed approach here is that the generated markdown won't have those links in it. GitHub doesn't automatically linkify those mentions in markdown files

lilnasy:
I was under the impression that the `getReleaseLine` function would only be ised for releases not changelog. I agree CHANGELOG.md does not need to change. I will look into your suggestion.

Andarist:
Those are called to generate the CHANGELOG entries and the latest update is extracted by [this function](https://github.com/changesets/action/blob/2bb9bcbd6bf4996a55ce459a630a0aa699457f59/src/utils.ts#L37). So to do what I suggested you'd have to transform the extracted content there

PR #1255, opened 2023-11-03
Added `disableThanks` option

Kauhsa:
Hi,

While thanking users in changelogs is commendable and makes sense for public projects or projects with a large amount of contributors, it served no purpose in our private packages and we felt that it cluttered our changelogs (for example, we had dozen entries in a single release, all prefixed with `Thanks @Kauhsa!`).

So, I added option `disableThanks` to `@changesets/changelog-github` to disable this behaviour.


TheMikeyRoss:
@Andarist @emmatown any update on this one?

bluwy (approved):

beeequeue:
related to my pr (#460) from very long ago :^)

beeequeue (commented):
could you add support for a list of names as well as a boolean?

```json
{
  "disableThanks": ["beeequeue"]
}
```

then one could disable the thanks for people who commit often and still keep thanking rare contributors. :)

_this is what my original PR was trying to do_

bluwy (changes requested):
Just noticed that I messed up the cherry-pick before and the option isn't called `disableThanks` anymore. Temporarily blocking this for now.

@beeequeue I think that idea makes sense too. Seems not too complicated to add so we can do it together in this PR.

bluwy:
Actually it's a bit tricky to refactor to handle the username filtering due to how it fetches the sources now, so I'm leaving it out from the changes.

Andarist (approved):

bluwy (approved):

PR #1260, opened 2023-11-09
Improve error handling when an invalid release type is specified in a changeset

askoufis:
When specifying a release type, if an invalid type is specified, such as `feat`, then a relatively unhelpful error is shown to the user:

```
TypeError: Cannot read properties of undefined (reading 'push')
```

This occurs because the key `feat` is not present in [this object](https://github.com/changesets/changesets/blob/0c8829bf8729ed5080bd7da843a717f60ea4b84b/packages/apply-release-plan/src/get-changelog-entry.ts#L41-L45), so it tries to `push` to `undefined` rather than an empty array.

I've added a small check to validate the release type and show an immediately actionable error if the type is invalid:

```
Invalid release type of "feat" for package "my-package" in changeset "foo-bar-baz"
```

Happy to work with a maintainer on an alternate solution if this is too naive/doesn't handle some edge cases I'm unaware of.


askoufis:
@bluwy @Andarist I don't like tagging people directly, but I've just been bitten by this same bug ~2 years later. Would appreciate some eyes on this PR. I'll fix up the code coverage stuff if it's agreed that this is a useful fix.

bluwy:
It looks like this has been handled in #1831. Sorry for not getting to this earlier.

askoufis:
No worries. Thanks for closing the loop.

PR #1263, opened 2023-11-19
docs: rename master to main

thdk:
Small rename of `master` to `main` as the relevant code snippet this is referring to also uses `main`.


simonhaenisch (approved):

bluwy:
It looks like this has been updated in https://github.com/changesets/changesets/pull/1155

PR #1291, opened 2024-01-12
Adhere to semver rules when using workspace:^ dependencies, fixes #1290

### LINKED ISSUE #1290: Dependency range `workspace:^` isn't treated as semver range, bumping dependent packages
## Affected Packages

* assemble-release-plan

## Problem

Dependents that are specifying their dependency as `workspace:^` get their version bumped regardless of the versions.

This is due to `workspace:^` being mapped to the dependency range `^` in

https://github.com/changesets/changesets/blob/93ee679537fb43111c29359fe6c82164b359e4d1/packages/assemble-release-plan/src/determine-dependents.ts#L197

and that's not a valid `Range` in `semver` land. And then the check in 

https://github.com/changesets/changesets/blob/93ee679537fb43111c29359fe6c82164b359e4d1/packages/assemble-release-plan/src/determine-dependents.ts#L89-L92

fails to detect that the semver satisfies.

## Proposed solution

Probably better to replace `workspace:^` with `"^" + dependencyRelease.oldVersion`

Pike:
This fixes #1290.

I've adjusted the tests, and also ran my local code against the [backstage repository](https://github.com/backstage/backstage).

On my current version, the `main` version updates some 200 packages, and with my patch it goes down to some 90.

CC @Rugvip , @freben from that side.

PS: I went for a major release of all affected packages, as this is breaking backwards compat. `yarn changeset status` gives me:

```
🦋  info NO packages to be bumped at patch
🦋  ---
🦋  info NO packages to be bumped at minor
🦋  ---
🦋  info Packages to be bumped at major:
🦋  info 
🦋  - @changesets/assemble-release-plan
🦋  - @changesets/cli
🦋  - @changesets/get-release-plan
```

### DIFF
--- a/packages/assemble-release-plan/src/determine-dependents.ts
+++ b/packages/assemble-release-plan/src/determine-dependents.ts
@@ -188,18 +188,18 @@ function getDependencyVersionRanges(
     if (versionRange.startsWith("workspace:")) {
       // intentionally keep other workspace ranges untouched
       // this has to be fixed but this should only be done when adding appropriate tests
-      let workspaceRange = versionRange;
-      if (versionRange === "workspace:*") {
-        // workspace:* actually means the current exact version, and not a wildcard similar to a reguler * range
-        workspaceRange = dependencyRelease.oldVersion;
-      } else if (versionRange === "workspace:^") {
-        // Use ^oldVersion for workspace:^.
-        // The version range might have changed in dependent package, but that should have its own changeset bumping that package.
-        workspaceRange = `^${dependencyRelease.oldVersion}`;
-      } else {
-        workspaceRange = versionRange.replace(/^workspace:/, "");
+      let workspaceRange = versionRange.replace(/^workspace:/, "");
+      switch (workspaceRange) {
+        case "*":
+          // workspace:* actually means the current exact version, and not a wildcard similar to a reguler * range
+          workspaceRange = dependencyRelease.oldVersion;
+          break;
+        case "~":
+        case "^":
+          // Use ^oldVersion for workspace:^ or ~oldVersion for workspace:~.
+          // The version range might have changed in dependent package, but that should have its own changeset bumping that package.
+          workspaceRange += dependencyRelease.oldVersion;
       }
-
       dependencyVersionRanges.push({
         depType: type,
         versionRange: workspaceRange,
--- a/.changeset/clever-candles-obey.md
+++ b/.changeset/clever-candles-obey.md
@@ -4,12 +4,11 @@
 "@changesets/get-release-plan": major
 ---
 
-Use semver compat for workspace:^ dependencies.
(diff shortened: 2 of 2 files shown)


packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -186,15 +186,23 @@ function getDependencyVersionRanges(
     if (!versionRange) continue;
 
     if (versionRange.startsWith("workspace:")) {
+      // intentionally keep other workspace ranges untouched
+      // this has to be fixed but this should only be done when adding appropriate tests
+      let workspaceRange = versionRange;
+      if (versionRange === "workspace:*") {
+        // workspace:* actually means the current exact version, and not a wildcard similar to a reguler * range
+        workspaceRange = dependencyRelease.oldVersion;
+      } else if (versionRange === "workspace:^") {
```
Rugvip, line 195: Afaik we'll want to handle `~` in a similar way as well

Pike:
I started to dabble into the `~` part, and then got stuck in a rabbit hole.

I think the culprit here are pre-releases, which is going to limit the impact this can have for projects like backstage:

```
> semverSatisfies('1.0.1','~1.0.0')
true
> semverSatisfies('1.0.1-alpha.0','~1.0.0')
false
> semverSatisfies('1.0.1-alpha.0','^1.0.0')
false
```

This goes back to https://github.com/npm/node-semver?tab=readme-ov-file#prerelease-tags, which reads in backstage lingo: If we bump `@backstage/core-plugin-api` in a pre-release, we need to also ship a pre-release version of `@backstage/plugin-todo`, which references that core package with the specific pre-release version.

More details on why that also then affects the non-pre releases is in Emma's comment in https://github.com/changesets/changesets/pull/382#discussion_r434434182.

adbayb:
Another possible workaround, until Changesets code owners review the pull request, can be to patch (via [PNPM](https://pnpm.io/cli/patch) if it's your default package manager or the [patch-package](https://github.com/ds300/patch-package) module) `@changesets/assemble-release-plan@6.0.0` dependency with @Pike proposed fix (thanks to him). 
Patching is working well confirming the issue fix 🎉 .

I hope this workaround can unblock projects while not rushing @Andarist and the Changesets team 👍 .

Rugvip:
@adbayb indeed, we'll try it out and see how it works: https://github.com/backstage/backstage/pull/24614. In particular, I expect this to improve our patch release process where we currently see too many package bumps.

timofei-iatsenko:
I think this fix is already good enough to unblock the vast majority of cases with "workspace:" protocol. Even without a proper support of `~`, support workspace caret range is already a big improvement. 

@Andarist could you take a look at this PR and prioritize it?

petebacondarwin:
We have also hit this problem with peer dependencies using `workspace:^` unnecessarily causing major bumps to the dependent package.

Issue #1292, opened 2024-01-17
When releasing a pre-release version, the pre-release version number from the remote repository was not obtained.


ahstufl:
Failure to retrieve the pre-release version number from the remote repository during the release of a pre-release version results in an error when attempting to publish a pre-release version that already exists in the remote repository, causing the CI to interrupt during execution.

atsikov:
I believe it is related to https://github.com/changesets/changesets/issues/1285

PR #1293, opened 2024-01-17
When releasing a pre-release version, the pre-release version number from the remote repository was not obtained.

ahstufl:
Failure to retrieve the pre-release version number from the remote repository during the release of a pre-release version results in an error when attempting to publish a pre-release version that already exists in the remote repository, causing the CI to interrupt during execution.


ahstufl:
This fixes https://github.com/changesets/changesets/issues/1292.
image

ahstufl:
image

PR #1296, opened 2024-01-20
docs(README.md): add suspensive as user of changesets

manudeli:
Thank you for providing such a great library. I'm really enjoying using changesets


bluwy:
Thanks for the PR! At the moment, we try to keep the list not too large, so we have a soft minimum of 1k stars to add to the list. Happy to revisit again in the future. I'll close this for now.

PR #1320, opened 2024-03-09
Output errors in console when git commit return code is not 0 (changesets#1277)

### LINKED ISSUE #1277: Changesets ran into trouble committing your files (CI only)
## Affected Packages

`@changesets/cli`
`@changesets/git`

## Problem

`npx changeset version` won't commit the files and doesn't give any details about the actual error.

```
🦋  error Changesets ran into trouble committing your files
...
🦋  info npm info x
🦋  info npm info y
🦋  warn x is not being published because version 0.3.0 is already published on npm
🦋  warn y is not being published because version 0.3.0 is already published on npm
🦋  warn No unpublished projects to publish
$ git status
HEAD detached at 1273ea3
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	deleted:    .changeset/silver-plums-study.md
	modified:   packages/assets/CHANGELOG.md
	modified:   packages/assets/package.json
	modified:   packages/ui/CHANGELOG.md
	modified:   packages/ui/package.json
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   .npmrc
```

## Proposed solution

- process should exit if commit fails so we don't publish anything that's not commit / pushed to the main branch
- the error message with output of stderr should be logged

I saw some git commands throw an error while others (tag, commit) simply return `false` when something went wrong.

minenwerfer:
fix https://github.com/changesets/changesets/issues/1277

This PR adds verbosity to when git commits fails (see [this comment](https://github.com/changesets/changesets/issues/1277#issuecomment-1986758500]). It shouldn't have no problems since I just copied the code from the function above.
Would it be better to change both to output to `console.error`?


minenwerfer:
Unrelated: I also noticed @changesets/* packages include a lot of seemingly unecessary source files because `files` property is lacking in `package.json`. Would remove them do any harm?

.changeset/large-ladybugs-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/git": patch
+---
+
+Output errors in console when git commit return code is not 0 (changesets#1277)
```
bluwy, line 5:
```suggestion
Output errors in console when git commit return code is not 0
```

I don't think we need to specify the issue here. Only what has changed.

bluwy (commented):
Looks like we already do the same for the `add()` function above, so this makes sense I think

Zigonja:
Would love to see this merged

danguilherme:
Is something missing here?

PR #1324, opened 2024-03-12
fix: avoid version changing for release type "none"

AkatQuas:
Before, it would assemble unnecessary version change for packages, output example:

```
{
  changesets: [
    {
      releases: [Array],
      summary: 'composite',
      id: 'feat+image-new_2024-03-05'
    }
  ],
  releases: [
    {
      name: '@scopeA/image-filter',
      type: 'patch',
      oldVersion: '13.0.0',
      changesets: [Array],
      newVersion: '13.0.1'
    },
    {
      name: '@scopeB/app1',
      type: 'none',
      oldVersion: '1.0.0',
      changesets: [],
      newVersion: '1.0.0'
    },
    {
      name: '@scopeB/app2',
      type: 'none',
      oldVersion: '1.0.0',
      changesets: [],
      newVersion: '1.0.0'
    },
  ],
  preState: undefined
}
```

It would touch `package.json` for `@scopeB/app1` and `@scopeB/app2` with unnecessary file change.

This Pull Request just skip this event totally.

### DIFF
--- a/packages/assemble-release-plan/src/index.ts
+++ b/packages/assemble-release-plan/src/index.ts
@@ -242,7 +242,7 @@ function assembleReleasePlan(
 
   return {
     changesets: relevantChangesets,
-    releases: [...releases.values()].map((incompleteRelease) => {
+    releases: Array.from(releases.values()).filter(r => r.type !== 'none').map((incompleteRelease) => {
       return {
         ...incompleteRelease,
         newVersion: snapshotSuffix


AkatQuas:
Would you please review this at your convenience ? @Andarist  . Thanks.

bluwy:
Can you point out which code is doing the IO event? Changing the output of `assemble-release-plan` seems like a breaking change to me. Also, the tests are failing.

AkatQuas:
@bluwy sorry for the delay.

In my case, the changeset cli would re-write the package.json files even if there's no actual version bumping.

So I use the `filter` to skip these unnecessary I/O change.  Especially I got 500 projects in my pnpm workspace. This might save a lot.

As for testing, I don't inspect to much so it failed.

Andarist:
It can't be treated as a trivial improvement if tests are failing with this change. It introduces a semantic difference that makes tests break. I don't recall the exact intricacies around `"none"` changesets but I do remember they are, at times, crucial for some of the logic. It might be that `apply-release-plan` (that receives output of `assemble-release-plan`) adjusts the bump types of packages and it might "promote" a none release to another type.

This design is not set in stone, it could be revisited. But to do that we need to understand why tests are failing here and how we could make them pass with a proposed change.

PR #1331, opened 2024-03-22
fix: throw error on unrecognized type

### LINKED ISSUE #1301: Cannot read properties of undefined (reading 'push')
### When i try to run the following command:

```shell
yarn changeset version
```
I'm getting the following giant error

```shell
The following error was encountered while generating changelog entries
We have escaped applying the changesets, and no files should have been affected
🦋  error TypeError: Cannot read properties of undefined (reading 'push')
🦋  error     
```
Here is the version of `changesets/cli` i'm using:

```json
{
"@changesets/cli": "^2.27.1",
}
```
I'm running node version: `v18.17.1`

What is the issue here?

unional:
fixes #1301

### DIFF
--- a/packages/apply-release-plan/src/get-changelog-entry.ts
+++ b/packages/apply-release-plan/src/get-changelog-entry.ts
@@ -51,6 +51,9 @@ export default async function getChangelogEntry(
   changesets.forEach((cs) => {
     const rls = cs.releases.find((r) => r.name === release.name);
     if (rls && rls.type !== "none") {
+      if (!changelogLines[rls.type]) {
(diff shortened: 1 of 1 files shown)


bluwy:
I wonder if we should do the fix in this function instead: https://github.com/changesets/changesets/blob/7ccb22e302dfc46d18e01f81f7979455960c0db9/packages/parse/src/index.ts#L6

The loaded yaml isn't being validated, which it probably should?

bluwy:
This is already handled in https://github.com/changesets/changesets/pull/1831

PR #1344, opened 2024-04-18
fix(cli): Typo in `pre` error and add test

kunalnagar:
When a tag is not passed in to `changeset pre enter`, the error message thrown as a typo:

```
$ yarn changeset pre enter
🦋  error A tag must be passed when using prerelese enter
```

This PR fixes the typo in `prerelease` and adds a test for it as well.


kunalnagar:
cc: @Andarist for review

bluwy (approved):

Issue #1036, opened 2022-12-15
"throws error if no new changesets" - tried it and does not throw and exits


orenmizr:
i needed a CI check (sadly we use azure) - created a branch from master and ran the check without adding a change sets. 
It did not errored out.  

https://github.com/changesets/changesets/blob/c606dc6b1b80f5a6ad4e4a9f0a557709702ccbea/docs/checking-for-changesets.md?plain=1#L10

Andarist:
I'm afraid that you'd have to provide a repro case for this. I can't investigate something that I can't run.

orenmizr:
can't you just verify on any repo the CI command "changeset status --since-master" does throw error if not changeset is commited ? it can ben done in any repo... to me it just show "no changes to x package" without a exit 1.

Andarist:
We have [tests](https://github.com/changesets/changesets/blob/c606dc6b1b80f5a6ad4e4a9f0a557709702ccbea/packages/cli/src/commands/status/__tests__/status.ts) for this command and while this test file uses some mocks I just recently refactored those tests to use git in one of my PRs:
https://github.com/changesets/changesets/pull/1033/files#diff-3c30dfc6aadfa4054a9df7f10ce9f405e9fe6ddd02a15d7418b93ccf5329a50b

So I'm somewhat confident that it works OK. Your use case has to be different from the ones that we cover with tests - so no, I can't "just verify" it. I don't know what you are **exactly** doing and I don't intent to do random things in hope to reproduce this. It might turn out that the repro will be trivial - I can't know that though. Providing a repro case is a courtesy for maintainers and issue reporters should try their best to provide them regardless of the complexity of the issue at hand. It ensures that we are going to look at the very same thing, no double guessing.

orenmizr:
appreciate the quick response. perhaps i expected the same behavior as the beachball package.  i edited the initial query but feel free to delete this thread.

Andarist:
According to your repro steps etc - this doesn't look alright. It should error - however, this is kinda a basic scenario for which we have tests and many users relying on this command. I would gladly investigate this but I can't do much without a proper repro case.

stefan1010:
@Andarist 

Observing the same behavior. No errorcode on `yarn changeset status` with the current version. With this setup version v2.12.0 seems the last state where it behaves as documented.

To reproduce please check out https://github.com/stefan1010/changesets-demo

Andarist:
[...]

PR #1345, opened 2024-04-26
feat: add failOnNoChanges flag to status cli command

runningrandall:
Addresses issue found in https://github.com/changesets/changesets/issues/1036.

This adds a flag of `failOnNoChanges` to the status cli command to force an exit code 1 if no uncommitted changes exist.


packages/cli/src/commands/status/__tests__/status.ts
```diff
@@ -133,6 +133,39 @@ describe("status", () => {
     expect(process.exit).toHaveBeenCalledWith(1);
   });
 
+  it("should exit early with a non-zero error code when there are no changed packages but no changesets", async () => {
```
Andarist, line 136: why is this a desired behavior? I mean, what's the use case? It's not quite clear to me.

runningrandall: During our CI process I would like the `changeset status --since=main` command to fail when no changeset is detected, even if no uncommitted changes are detected. There will never be any uncommited changes during our CI process (and probably most others).

I went with a cli option to maintain existing behavior.

The [docs](https://github.com/changesets/changesets/blob/main/docs/checking-for-changesets.md?plain=1#L10) indicate the above command will fail in CI when there are no detected changesets. This is not the behavior as demonstrated in https://github.com/stefan1010/changesets-demo.

See open issue https://github.com/changesets/changesets/issues/1036.

armaaar: The current behaviour is confusing others as well like in #1362 and it confused me at the beginning.

I would love this to be merged. At the very least the docs needs to be improved regarding the current behaviour

docs/checking-for-changesets.md
```diff
@@ -9,3 +9,5 @@ alert users of missing changesets.
 
 If you want to cause a failure in CI on missing changesets (not recommended), you can run `changeset status --since=main`,
 which will exit with a status code of 1 if there are no new changesets.
+
+> [!CAUTION] > **This will still pass with an exit code 0 in CI unless an uncommitted change exists in one of your packages. If you need this to fail in CI without an uncommitted change use the `--failOnNoChanges` flag **
```
armaaar, line 13: It's not only `uncommitted changes`, but also committed changes that doesn't affect any packages would also exit successfully. Your test description is describing it better.

PR #1351, opened 2024-05-08
fix: prettier config resolving

### LINKED ISSUE #1350: changeset version does not respect prettier config since prettier 3.1.1
## Affected Packages

`apply-release-plan`

## Problem

When having `prettier` version `3.1.1` or higher installed. It will not respect the prettier configuration when running `changeset version`. 

This results in `CHANGELOG.md` files being created with invalid formatting, resulting in broken CI checks on formatting.

The problem happens at this line: https://github.com/changesets/changesets/blob/0bf89b3709e3e3df6ed5dbb8ece0fb000a55d5f4/packages/apply-release-plan/src/index.ts#L136

`prettierConfig` is null even though a `.prettierrc` file is present.

This is caused by changes introduced in `3.1.1` to how `.prettierrc` config is found. See changelog notes: https://github.com/prettier/prettier/blob/main/CHANGELOG.md#311

## Proposed solution

Changing it to this fixes the prettier config resolving:
```
let prettierConfig = await prettierInstance.resolveConfig(cwd + '/package.json');
```

TheHolyWaffle:
fixes #1350

### DIFF
--- a/.changeset/breezy-moles-behave.md
+++ b/.changeset/breezy-moles-behave.md
@@ -1,0 +1,6 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/write": patch
+---
+
+Fix prettier config resolving
--- a/.changeset/breezy-moles-behave.md
+++ b/.changeset/breezy-moles-behave.md
@@ -1,4 +1,5 @@
 ---
+"@changesets/cli": patch
 "@changesets/apply-release-plan": patch
 "@changesets/write": patch
 ---
--- a/.changeset/breezy-moles-behave.md
+++ b/.changeset/breezy-moles-behave.md
@@ -3,4 +3,4 @@
 "@changesets/write": patch
 ---
 
-Fix `changeset version` not applying a custom .prettierrc configuration with `prettier` >= 3.1.1
+Fix an issue with not applying a custom `.prettierrc` configuration with `prettier@>= 3.1.1` 
--- a/.changeset/breezy-moles-behave.md
+++ b/.changeset/breezy-moles-behave.md
@@ -3,4 +3,4 @@
 "@changesets/write": patch
 ---
 
-Fix prettier config resolving
+Fix `changeset version` not applying a custom .prettierrc configuration with `prettier` >= 3.1.1
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -133,9 +133,6 @@ export default async function applyReleasePlan(
   });
 
   let prettierInstance = getPrettierInstance(cwd);
-  let prettierConfig = await prettierInstance.resolveConfig(
-    path.join(cwd, "package.json")
-  );
 
   for (let release of finalisedRelease) {
     let { changelog, packageJson, dir, name } = release;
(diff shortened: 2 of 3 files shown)


TheHolyWaffle:
cc @Andarist request for review 🙏

packages/apply-release-plan/src/index.ts
```diff
@@ -133,7 +133,9 @@ export default async function applyReleasePlan(
   });
 
   let prettierInstance = getPrettierInstance(cwd);
-  let prettierConfig = await prettierInstance.resolveConfig(cwd);
+  let prettierConfig = await prettierInstance.resolveConfig(
+    path.join(cwd, "package.json")
```
Andarist, line 137: I don't think this is completely correct. We are going to use the resolved config to format markdown files here so we should resolve a config for a markdown file. 

We probably should resolve the config file separately for each file that we want to update but that's a separate issue.

TheHolyWaffle: Aha you're right. Pushed a change!

packages/write/src/index.ts
```diff
@@ -31,7 +31,9 @@ async function writeChangeset(
   });
 
   const prettierInstance = getPrettierInstance(cwd);
-  const prettierConfig = await prettierInstance.resolveConfig(cwd);
+  const prettierConfig = await prettierInstance.resolveConfig(
+    path.join(cwd, "package.json")
```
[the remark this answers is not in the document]
TheHolyWaffle: Aha you're right. Pushed a change!

packages/apply-release-plan/src/index.ts
```diff
@@ -326,32 +310,25 @@ async function prependFile(
...
-  prettierConfig: prettier.Options | null
+  prettierInstance: typeof prettier
 ) {
   await fs.writeFile(
     filePath,
     // Prettier v3 returns a promise
     await prettierInstance.format(content, {
-      ...prettierConfig,
+      ...(await prettierInstance.resolveConfig(filePath)),
       filepath: filePath,
```
Andarist, line 332: it's interesting that we pass a `filepath` here but we don't do that in the other place. It's either redundant or we should add it in that other place. I don't know which one is more correct so 🤷‍♂️

Andarist (approved):

.changeset/breezy-moles-behave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/write": patch
+---
+
+Fix `changeset version` not applying a custom .prettierrc configuration with `prettier` >= 3.1.1
```
Andarist, line 6:
```suggestion
Fix an issue with not applying a custom `.prettierrc` configuration with `prettier@>= 3.1.1` 
```

.changeset/breezy-moles-behave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/write": patch
```
Andarist, line 3:
```suggestion
"@changesets/cli": patch
"@changesets/apply-release-plan": patch
"@changesets/write": patch
```

PR #1354, opened 2024-05-16
Fixed `changeset status` combined with ignored and non-versionable private packages

Andarist:


jakebailey (approved):
With no knowledge of the codebase, this seems correct, though it does seem like anywhere `getChangedPackagesSinceRef` is used would need this exact filtering, right? More or less I'm looking for a way to just tell changesets to never worry about private packages period.

emmatown (approved):

Andarist:
> though it does seem like anywhere getChangedPackagesSinceRef is used would need this exact filtering, right?

yeah, I just made sure that it's the case. It turns out that we had almost the same filtering in another place. I didn't make the change in `getChangedPackagesSinceRef` itself though since that's a lower-level util. It has no notion of Changesets today.

PR #1360, opened 2024-05-21
chore: add flake.nix to ease development for nix users

tim-smart:


Andarist:
Why does this need to be included in Changesets?

benmccann:
This looks like a script to do general node setup of installing node, corepack, etc. to setup a development environment. It doesn't seem appropriate to me to add this type of PR to every open source project and would suggest it should be closed

PR #1361, opened 2024-05-21
Fix unversioned private package handling, tagging of ignored

jakebailey:
This works for `version` and `tag`, at least in my testing for https://github.com/microsoft/TypeScript-Website/pull/3116.

~However, there seem to be other places that should be checking this, but have less convenient access to actual Package objects. That and it's not clear to me if the other cases actually matter.~

~Also, needs tests...~ And to figure out where to move this shared code as there is no common package for these besides `types` which is types only.

### DIFF
--- a/.changeset/sharp-humans-battle.md
+++ b/.changeset/sharp-humans-battle.md
@@ -1,0 +1,7 @@
+---
+"@changesets/assemble-release-plan": patch
+"@changesets/apply-release-plan": patch
+"@changesets/cli": patch
+---
+
+Ensure that version/tag do not touch private packages with privatePackages config
--- a/.changeset/sharp-humans-battle.md
+++ b/.changeset/sharp-humans-battle.md
@@ -4,4 +4,4 @@
 "@changesets/cli": patch
 ---
 
-Ensure that version/tag do not touch private packages with privatePackages config
+Ensure that `version`/`tag` do not touch private packages with when versioning/tagging is turned off using `versionPackages` config
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -191,7 +191,6 @@ export default async function applyReleasePlan(
 }
 
 // Note: if updating this, also update the other copies of createIsVersionablePackage.
-// TODO(jakebailey): don't copy paste
 function createIsVersionablePackage(
   ignoredPackages: readonly string[],
   allowPrivatePackages: boolean
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -163,6 +163,7 @@ export default async function applyReleasePlan(
           // At this point, we know there is no such changeset, because otherwise the program would've already failed,
           // so we just check if any ignored package exists in this changeset, and only remove it if none exists
           // Ignored list is added in v2, so we don't need to do it for v1 changesets
+          // TODO(jakebailey): should this check isVersionablePackage?
           if (
             !changeset.releases.find((release) =>
               config.ignore.includes(release.name)
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -9,7 +9,7 @@ import {
 import { defaultConfig } from "@changesets/config";
 import * as git from "@changesets/git";
 import resolveFrom from "resolve-from";
-import { Packages } from "@manypkg/get-packages";
+import { Package, Packages } from "@manypkg/get-packages";
 import detectIndent from "detect-indent";
 
 import fs from "fs-extra";
@@ -78,6 +78,11 @@ export default async function applyReleasePlan(
 
   const packagesByName = new Map(
     packages.packages.map((x) => [x.packageJson.name, x])
+  );
+
+  const isVersionablePackage = createIsVersionablePackage(
+    config.ignore,
+    config.privatePackages.version
   );
 
   let { releases, changesets } = releasePlan;
@@ -163,10 +168,10 @@ export default async function applyReleasePlan(
           // At this point, we know there is no such changeset, because otherwise the program would've already failed,
           // so we just check if any ignored package exists in this changeset, and only remove it if none exists
           // Ignored list is added in v2, so we don't need to do it for v1 changesets
-          // TODO(jakebailey): should this check isVersionablePackage?
           if (
-            !changeset.releases.find((release) =>
-              config.ignore.includes(release.name)
+            !changeset.releases.find(
+              (release) =>
+                !isVersionablePackage(packagesByName.get(release.name)!)
             )
           ) {
             touchedFiles.push(changesetPath);
@@ -183,6 +188,26 @@ export default async function applyReleasePlan(
 
   // We return the touched files to be committed in the cli
   return touchedFiles;
+}
+
+// TODO(jakebailey): don't copy paste
+export function createIsVersionablePackage(
+  ignoredPackages: readonly string[],
+  allowPrivatePackages: boolean
+): (pkg: Package) => boolean {
+  const ignoredPackagesSet = new Set(ignoredPackages);
+  return ({ packageJson }: Package) => {
+    if (ignoredPackagesSet.has(packageJson.name)) {
+      return false;
+    }
+
+    if (packageJson.private && !allowPrivatePackages) {
+      return false;
+    }
+
+    const hasVersionField = !!packageJson.version;
+    return hasVersionField;
+  };
 }
 
 async function getNewChangelogEntry(
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -190,8 +190,9 @@ export default async function applyReleasePlan(
   return touchedFiles;
 }
 
+// Note: if updating this, also update the other copies of createIsVersionablePackage.
 // TODO(jakebailey): don't copy paste
-export function createIsVersionablePackage(
+function createIsVersionablePackage(
   ignoredPackages: readonly string[],
   allowPrivatePackages: boolean
 ): (pkg: Package) => boolean {
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -1,23 +1,21 @@
-import {
-  ReleasePlan,
-  Config,
-  ChangelogFunctions,
-  NewChangeset,
-  ModCompWithPackage,
-} from "@changesets/types";
-
 import { defaultConfig } from "@changesets/config";
 import * as git from "@changesets/git";
-import resolveFrom from "resolve-from";
-import { Package, Packages } from "@manypkg/get-packages";
+import { shouldSkipPackage } from "@changesets/should-skip-package";
+import {
+  ChangelogFunctions,
+  Config,
+  ModCompWithPackage,
+  NewChangeset,
+  ReleasePlan,
+} from "@changesets/types";
+import { Packages } from "@manypkg/get-packages";
 import detectIndent from "detect-indent";
-
 import fs from "fs-extra";
 import path from "path";
 import prettier from "prettier";
-
-import versionPackage from "./version-package";
+import resolveFrom from "resolve-from";
 import getChangelogEntry from "./get-changelog-entry";
+import versionPackage from "./version-package";
 
 function getPrettierInstance(cwd: string): typeof prettier {
   try {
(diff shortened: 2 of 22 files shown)


jakebailey:
Tests fail for me, but also fail on main. Not sure what I'm doing wrong.

packages/cli/src/utils/versionablePackages.ts
```diff
@@ -32,6 +32,14 @@ export function filterVersionablePackages(config: Config, packages: Package[]) {
   return packages.filter((pkg) => isVersionablePackage(pkg, options));
 }
 
+export function filterTaggablePackages(config: Config, packages: Package[]) {
+  const options = {
+    ignoredPackages: new Set(config.ignore),
+    versionPrivatePackages: config.privatePackages.tag,
```
Andarist, line 38: `versionPrivatePackages` should probably now become smth like `ignorePrivatePackages`

romankulkovsf:
App

packages/apply-release-plan/src/index.ts
```diff
@@ -184,6 +190,26 @@ export default async function applyReleasePlan(
   return touchedFiles;
 }
 
+// TODO(jakebailey): don't copy paste
```
jakebailey, line 193: Noting this TODO; there is no common package for this besides `types`. So maybe a copy/paste plus a note to keep them the same is sufficient?

packages/assemble-release-plan/src/utils.ts
```diff
@@ -58,3 +58,23 @@ export function getCurrentHighestVersion(
 
   return highestVersion!;
 }
+
+// TODO(jakebailey): don't copy paste
```
jakebailey, line 62: Also here.

packages/cli/src/run.ts
```diff
@@ -146,9 +146,11 @@ export async function run(
           bumpVersionsWithWorkspaceProtocolOnly:
             config.bumpVersionsWithWorkspaceProtocolOnly,
         });
+        // TODO(jakebailey): should this check isVersionablePackage?
         for (const ignoredPackage of config.ignore) {
```
jakebailey, line 150: This is some sort of early warning in the CLI; this may be challenging to fix as we can't directly iterate "all private packages", so here I'd need to instead pull all packages and then check each, which may be okay.

packages/cli/src/run.ts
```diff
@@ -146,9 +146,11 @@ export async function run(
...
           const dependents = dependentsGraph.get(ignoredPackage) || [];
           for (const dependent of dependents) {
+            // TODO(jakebailey): should this check isVersionablePackage?
             if (!config.ignore.includes(dependent)) {
```
jakebailey, line 154: Unlike the one above, this one can just be turned into a plain call so would be fine.

packages/apply-release-plan/src/index.ts
```diff
@@ -164,8 +169,9 @@ export default async function applyReleasePlan(
           // so we just check if any ignored package exists in this changeset, and only remove it if none exists
```
Andarist, line 169: this comment has to be updated

packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -37,6 +38,10 @@ export default function determineDependents({
   let updated = false;
   // NOTE this is intended to be called recursively
   let pkgsToSearch = [...releases.values()];
+  const isVersionablePackage = createIsVersionablePackage(
```
Andarist, line 41: if we are already passing `isVersionablePackage` to other functions here then we could also pass it down here so we don't have to recreate it on each run (IIRC this function can get re-called in a loop)

packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -8,15 +8,15 @@ import { InternalRelease } from "./types";
...
   changesets: NewChangeset[],
   packagesByName: Map<string, Package>,
-  ignoredPackages: Readonly<string[]>
+  isVersionablePackage: (pkg: Package) => boolean
 ): Map<string, InternalRelease> {
   let releases: Map<string, InternalRelease> = new Map();
 
   changesets.forEach((changeset) => {
     changeset.releases
       // Filter out ignored packages because they should not trigger a release
```
Andarist, line 17: this comment needs to be updated

packages/assemble-release-plan/src/index.ts
```diff
@@ -270,11 +278,7 @@ function getRelevantChangesets(
     const ignoredPackages = [];
     const notIgnoredPackages = [];
```
Andarist, line 279: nit: those names probably still work in this context but maybe we could rename them to avoid confusion

packages/assemble-release-plan/src/match-fixed-constraint.ts
```diff
@@ -10,6 +14,11 @@ export default function matchFixedConstraint(
 ): boolean {
   let updated = false;
 
+  const isVersionablePackage = createIsVersionablePackage(
+    config.ignore,
+    config.privatePackages.version
+  );
```
Andarist, line 20: similarly here, we could pass down this from the caller

packages/assemble-release-plan/src/utils.ts
```diff
@@ -58,3 +58,23 @@ export function getCurrentHighestVersion(
 
   return highestVersion!;
 }
+
+// Note: if updating this, also update the other copies of createIsVersionablePackage.
```
Andarist, line 62: i think i'm leaning towards creating a new package for this util. Don't worry about it though, it's fine right now and I can extract it any time

packages/assemble-release-plan/src/utils.ts
```diff
@@ -58,3 +58,23 @@ export function getCurrentHighestVersion(
...
+export function createIsVersionablePackage(
+  ignoredPackages: readonly string[],
+  allowPrivatePackages: boolean
+): (pkg: Package) => boolean {
+  const ignoredPackagesSet = new Set(ignoredPackages);
```
Andarist, line 67: i really wish we'd be normalizing this when reading the config 😬 it's totally our fault and I had that thought already in the past ;p

packages/cli/src/commands/tag/__tests__/index.test.ts
```diff
@@ -30,7 +31,7 @@ describe("tag command", () => {
       (git.getAllTags as jest.Mock).mockReturnValue(new Set());
 
       expect(git.tag).not.toHaveBeenCalled();
-      await tag(cwd);
+      await tag(cwd, defaultConfig);
```
Andarist, line 34: i'd probably go with `readConfig` like in the status' tests but this is fine too. Our test suite isn't particularly consistent when it comes to this

packages/cli/src/commands/tag/__tests__/index.test.ts
```diff
@@ -81,9 +82,27 @@ describe("tag command", () => {
       (git.getAllTags as jest.Mock).mockReturnValue(new Set());
 
       expect(git.tag).not.toHaveBeenCalled();
-      await tag(cwd);
+      await tag(cwd, {
+        ...defaultConfig,
+        privatePackages: { version: true, tag: true },
```
Andarist, line 87: hm, doesn't this change mean it's a breaking change? 🤔 note to myself: i need to recheck the current defaults

packages/cli/src/commands/tag/index.ts
```diff
@@ -1,13 +1,15 @@
 import * as git from "@changesets/git";
 import { getPackages } from "@manypkg/get-packages";
 import { log } from "@changesets/logger";
+import { filterTaggablePackages } from "../../utils/versionablePackages";
+import { Config } from "@changesets/types";
 
-export default async function run(cwd: string) {
+export default async function run(cwd: string, config: Config) {
```
Andarist, line 7:
```suggestion
export default async function tag(cwd: string, config: Config) {
```

and the same change could be made to `publish` (+ status command could get use `status` and not `getStatus`)

packages/cli/src/run.test.ts
```diff
@@ -65,6 +65,41 @@ describe("cli", () => {
       );
     });
 
+    it("should not throw if dependents of unversioned private packages are not explicitly listed in the ignore array", async () => {
```
Andarist, line 68: note to myself: recheck what this test is about and how it behaves today

packages/cli/src/utils/versionablePackages.ts
```diff
@@ -2,34 +2,40 @@ import { Config } from "@changesets/types";
...
-  }: {
-    ignoredPackages: Set<string>;
-    versionPrivatePackages: boolean;
-  }
-) {
-  if (ignoredPackages.has(packageJson.name)) {
-    return false;
-  }
+// Note: if updating this, also update the other copies of createIsVersionablePackage.
+export function createIsVersionablePackage(
```
Andarist, line 6: this name mentions "versionable" but it's also used with `.privatePackages.tag`, thinking out loud: is there a better name for this?

packages/cli/src/run.test.ts
Andarist: i guess what i'm surprised here is that this test has no changesets, it just tests a "dry" version run

packages/cli/src/utils/versionablePackages.ts
jakebailey: isIgnoredPackage then flip the condition?

packages/cli/src/commands/tag/__tests__/index.test.ts
Andarist: the current default is `privatePackages: { version: true, tag: false }`. The current behavior was broken and this adjustment just reflects that the bug was fixed

packages/cli/src/run.test.ts
Andarist: So it turns out that `version` has some validation logic in it already so this throws an error on `main`:
> "The package "pkg-a" depends on the ignored package "pkg-b", but "pkg-a" is not being ignored. Please pass "pkg-a" to the \`--ignore\` flag."

The test, indeed, validates a fixed behavior 👍

.changeset/sharp-humans-battle.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/assemble-release-plan": patch
+"@changesets/apply-release-plan": patch
+"@changesets/cli": patch
+---
+
+Ensure that version/tag do not touch private packages with privatePackages config
```
Andarist, line 7:
```suggestion
Ensure that `version`/`tag` do not touch private packages with when versioning/tagging is turned off using `versionPackages` config
```

packages/cli/src/commands/tag/__tests__/index.test.ts
WoodyWoodsta: Hey @Andarist - Would this be the cause of my private packages no longer receiving tags with the default value of `privatePackages` (i.e. unset in my config)?

Some documentation for `privatePackages` would be a massive help - I wasn't able to find any.

jakebailey: Yeah, that's effectively what this change did... I didn't realize the default option here was `tag: false`, but if it's tag: false, surely that means it shouldn't be tagging, right? So you should just set it to true?

WoodyWoodsta: > surely that means it shouldn't be tagging, right? So you should just set it to true?

Yes, I'm just asking to make sure because a default behaviour has changed on a patch version increment, on an undocumented property :)

I have tested it with `tag: true`, and we're back to the previous behaviour.

Andarist: > Some documentation for privatePackages would be a massive help - I wasn't able to find any.

I agree... PRs are more than welcome 😉 

Sorry for the breaking your workflow here. I had to recheck this myself when reviewing this PR but I concluded it was a bug. IIRC, older versions of the CLI didn't tag private packages - at some point that was broken.

WoodyWoodsta: Not a problem! Was just checking I wasn't going mad. It must have been broken for a fair few months (i.e. since before we integrated changesets!)

Andarist: Yeah, it might be. The last year wasn't a particularly active year when it comes to the development of this package 😅

kernwig: Hi. This was a breaking change because:

1. My repo which I don't publish (thus private) stopped tagging versions.
2. The referenced `privatePackages` configuration is [not documented](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md), so it doesn't matter what the default was if nobody using the library knew the configuration existed.

Do I add `privatePackages: { version: true, tag: true }` to my `.changeset/config.json` to make it tag again? I see it is in the $schema.

drianoaz: Hi guys, 

An issue #1383 and a pull request #1385 were created to talk about the documentation of this item

PR #1370, opened 2024-05-28
Fixed a regression that caused `changeset version` to fail on packages having a dev dependency on a skipped package

Andarist:


packages/cli/src/run.test.ts
```diff
@@ -100,6 +100,41 @@ describe("cli", () => {
       expect(loggerErrorCalls.length).toEqual(0);
     });
 
+    it("should not throw on a dev dependent on an unversioned private package", async () => {
```
Andarist, line 103: a similar scenario was already buggy before this recent regression, see:
https://github.com/changesets/changesets/pull/1361#discussion_r1617118563

PR #1381, opened 2024-06-12
Stop requiring `tool` as part of `Packages` in certain packages

Andarist:
It turns out that it's difficult for `changesets/bot` to create the proper `Packages` object. It's only forced to do it because types expect the `tool` property but it's not actually needed at runtime

This should be cleaned up further but I don't have the energy to figure out the more proper code organization right now.


Andarist:
Actually, this might not be needed right now as Changesets are still using manypkg v1

PR #1390, opened 2024-06-23
Remove tty-table dependency

bluwy:
The `tty-table` dependency is only used for `changeset status --verbose`. The dependency has a [very big dependency graph](https://npmgraph.js.org/?q=tty-table#zoom=w) of 115 transitive dependencies with many polyfill packages.

This PR re-implements a simpler table formatting manually to remove `tty-table`, as I can't find a similar table printing table that would suffice the current requirements. I don't think the large dependency footprint is worth keeping it around either.

The main difference is that this PR's implementation does not shrink the table on small console widths, but I think it's a fine tradeoff for now. The shrinking behaviour also sometimes cuts off package names and changeset paths.

Before:

image

image

After:

<details>
<summary>Previous refactor of table form</summary>

image

image

</details>

*Now updated as a list display from https://github.com/changesets/changesets/pull/1390#issuecomment-2185949156

image


bluwy:
Looks like `status/index.ts` wasn't covered by tests before, so adding more lines here seem to caused more missing coverage. Should I add some sort of test to cover it?

bluwy:
I've discussed with @jakebailey and @benmccann of some alternative table libraries:
- [`cli-table3`](https://www.npmjs.com/package/cli-table3) - tradeoffs:
  - No word wrapping for small console widths
  - No direct option for table indenting (can be workaround with string manipulation)
- [`table-layout`](https://github.com/75lb/table-layout) - tradeoffs:
  - Larger dependency graph, but [there's a PR](https://github.com/75lb/table-layout/pull/18) to slim it down
  - Has a limited set of styling options

Other libraries like `console-table-printer` seem to also share the same or more tradeoffs of the above two libraries. Or libraries like `cliui` could work but they're CLI focused.

Andarist:
In a sense, the only thing that `--verbose` does is print the information about related changesets. The table layout is kinda nice at presenting this but maybe it's unnecessary? Can we think about a different of presenting this data that would result in a simpler code and that wouldn't suffer from wrapping limitations/complexity?

bluwy:
Perhaps a list form like this?
image

A bit similar to non-verbose mode but with the added version and changeset files.

Andarist:
It works for me, I think. Since a lot of changesets can be related to any particular release, I wonder if this should be a bullet~ list or just a comma-delimited list. None of them is perfect as it all depends on how many packages, changesets, etc u have... Let's just roll with what u have above.

.changeset/thirty-glasses-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Remove `tty-table` dependency and re-implement simpler table printing
```
Andarist, line 5: this has to be tweaked now

bluwy:
```suggestion
Display `changeset status --verbose` in list form and remove `tty-table` dependency
```

Issue #1383, opened 2024-06-19
Add privatePackages info in docs/config-file-options.md


luiz-araujo:
## Affected Packages

docs/config-file-options.md

## Problem

Recently there was an update to the project adding `privatePackages` to the configuration file, but the documentation was not updated.

And I work on a project set as private in `package.json` and which uses tagging in its deployment process. After the update, our deploy stopped working and the lack of this information in the documentation made resolving this scenario a little more complex than it should have been.

## Proposed solution

Update the example in `docs/config-file-options.md`

```json
{
  "commit": false,
  "updateInternalDependencies": "patch",
  "linked": [],
  "access": "restricted",
  "baseBranch": "master",
  "ignore": [],
  "changelog": "@changesets/cli/changelog",
  "privatePackages": {
    "version": true,
    "tag": true
  }
}
```

And then add the details to the documentation as below:

## `privatePackages` (object)

This option is used to configure the versioning and tagging of private packages. It can be set as follows:

```json
{
  "privatePackages": {
    "version": true,
    "tag": true
  }
}
```

- `version` (boolean): Determines whether private packages can be versioned. Set it to true to enable versioning.
- `tag` (boolean): Determines whether private packages can be tagged. Set it to true to enable tagging.

By default changesets will only update the changelog and version (ie `{ version: true, tag: false }`).

 You can also opt private packages out of versioning entirely by setting `privatePackages: false`.

drianoaz:
related: https://github.com/changesets/changesets/pull/1361#discussion_r1643409054

jakebailey:
Should the defaults for the privatePackages options just be set to true to undo the break? Docs are good but clearly people were assuming the actual behavior (which never actually looked at the option!) were intended.

luiz-araujo:
> Should the defaults for the privatePackages options just be set to true to undo the break? Docs are good but clearly people were assuming the actual behavior (which never actually looked at the option!) were intended.

Hi @Andarist,

Could you please share your opinion on the following comment?

Andarist:
Hi @luiz-araujo , I'm sorry that this has happened. When merging the PR, I assessed that this was a bug that was being fixed. The broken behavior was out there for an extended period of time though and I understand that people could depend on it (Hyrum's law and all). 

[...]

PR #1394, opened 2024-06-24
Default privatePackages.tag to true to match old buggy behavior

jakebailey:
Per https://github.com/changesets/changesets/issues/1383#issuecomment-2181280964

Due to a bug fixed in https://github.com/changesets/changesets/pull/1361, `tag: false` did not actually have any effect, so private packages were still tagged. This surprised many people, so I think it may be worthwile to just default this to "true" to match the buggy behavior.

### DIFF
--- a/.changeset/spicy-bottles-try.md
+++ b/.changeset/spicy-bottles-try.md
@@ -1,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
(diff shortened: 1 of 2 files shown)


snitovets:
That's why my packages stopped to have tags and therefor breaks the other steps. I'm glad that I found the reason.

PR #1396, opened 2024-06-26
upgrade packages from the @manypkg org

benmccann:


packages/apply-release-plan/package.json
```diff
@@ -25,7 +25,7 @@
     "@changesets/git": "^3.0.0",
     "@changesets/should-skip-package": "^0.1.0",
     "@changesets/types": "^6.0.0",
-    "@manypkg/get-packages": "^1.1.3",
+    "@manypkg/get-packages": "^2.2.1",
```
Andarist, line 28: This might actually be problematic because the returned packages are no longer serializable there. IIRC this will create problems in the Changesets bot which might not be able to construct the required arguments for the `assembleReleasePlan` call here:
https://github.com/changesets/bot/blob/126c61d7997442d240ab26d7d43aaf5e393dbc8d/get-changed-packages.ts#L182-L187

.changeset/honest-trees-tell.md
```diff
@@ -0,0 +1,14 @@
...
+"@changesets/should-skip-package": patch
+"@changesets/apply-release-plan": patch
+"@changesets/get-release-plan": patch
+"@changesets/release-utils": patch
+"@changesets/config": patch
+"@changesets/git": patch
+"@changesets/pre": patch
+---
+
+upgrade @manypkg/get-packages
```
Andarist, line 14: this change changes the required API of some of those packages - for those it's a breaking change and they should be annotated as major bumps here.

benmccann:
thanks for the feedback! looks like this will need to be reworked a bit. I'll close it for now

packages/apply-release-plan/package.json
benmccann: Ah, it looks like `Tool` that isn't serializable: https://github.com/Thinkmill/manypkg/blob/045f7d7dabfa8a25b9d160661184a35eb4b5eaa7/packages/tools/src/Tool.ts#L116

Issue #1397, opened 2024-06-29
Issue while `pnpm changeset`


wojtekKrol:
## Problem

I'm creating a package and wanted to add changeset to it, I did and once I want to run `pnpm changeset` I got error:

```
Package "zod-to-fields" must depend on the current version of "react": "0.0.0" vs "^18.2.0"
Package "react" must depend on the current version of "react": "0.0.0" vs "^18.2.0"
🦋  error TypeError: Cannot read properties of undefined (reading 'dependents')
🦋  error     at /Users/wojciechkrol/WebstormProjects/my-projects/zod-to-fields/node_modules/.pnpm/@changesets+get-dependents-graph@2.1.0/node_modules/@changesets/get-dependents-graph/dist/changesets-get-dependents-graph.cjs.js:139:38
🦋  error     at Array.forEach (<anonymous>)
🦋  error     at /Users/wojciechkrol/WebstormProjects/my-projects/zod-to-fields/node_modules/.pnpm/@changesets+get-dependents-graph@2.1.0/node_modules/@changesets/get-dependents-graph/dist/changesets-get-dependents-graph.cjs.js:138:20
🦋  error     at Array.forEach (<anonymous>)
🦋  error     at Object.getDependentsGraph (/Users/wojciechkrol/WebstormProjects/my-projects/zod-to-fields/node_modules/.pnpm/@changesets+get-dependents-graph@2.1.0/node_modules/@changesets/get-dependents-graph/dist/changesets-get-dependents-graph.cjs.js:132:21)
🦋  error     at parse (/Users/wojciechkrol/WebstormProjects/my-projects/zod-to-fields/node_modules/.pnpm/@changesets+config@3.0.1/node_modules/@changesets/config/dist/changesets-config.cjs.js:255:50)
🦋  error     at Object.read (/Users/wojciechkrol/WebstormProjects/my-projects/zod-to-fields/node_modules/.pnpm/@changesets+config@3.0.1/node_modules/@changesets/config/dist/changesets-config.cjs.js:138:10)
🦋  error     at async run (/Users/wojciechkrol/WebstormProjects/my-projects/zod-to-fields/node_modules/.pnpm/@changesets+cli@2.27.6/node_modules/@changesets/cli/dist/changesets-cli.cjs.js:1279:16)
```

my package is `zod-to-fields`

this is my `package.json`

```
{
  "name": "zod-to-fields",
  "version": "0.1.2",
[...]

PR #1400, opened 2024-06-30
Fixed a crash when root workspace has dependents

Andarist:
fixes https://github.com/changesets/changesets/issues/1397

### DIFF
--- a/packages/get-dependents-graph/src/index.ts
+++ b/packages/get-dependents-graph/src/index.ts
@@ -14,7 +14,12 @@ export function getDependentsGraph(
 
   const dependentsLookup: {
     [key: string]: { pkg: Package; dependents: Array<string> };
-  } = {};
+  } = {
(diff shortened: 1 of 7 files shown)


packages/get-dependents-graph/src/index.ts
```diff
@@ -14,7 +14,12 @@ export function getDependentsGraph(
 
   const dependentsLookup: {
     [key: string]: { pkg: Package; dependents: Array<string> };
-  } = {};
+  } = {
+    [packages.root.packageJson.name]: {
+      pkg: packages.root,
+      dependents: [],
+    },
+  };
```
Andarist, line 22: this mirrors what we already do in `getDependencyGraph`:
https://github.com/changesets/changesets/blob/main/packages/get-dependents-graph/src/get-dependency-graph.ts#L75-L77

emmatown (approved):

PR #1410, opened 2024-07-07
Detect pnpm manually. Drops 34 dependencies

benmccann:
This drops 34 dependencies (28.8% of the `@changesets/cli` total). The functionality is pretty simple so we can just do it ourselves as opposed to pulling in a library that does a bunch of stuff we don't need. E.g. we don't need a yaml parser to detect what version of yarn is being used, etc.

https://npmgraph.js.org/?q=preferred-pm@3.0.0

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -2,8 +2,9 @@ import { ExitError } from "@changesets/errors";
 import { error, info, warn } from "@changesets/logger";
 import { AccessType, PackageJSON } from "@changesets/types";
 import pLimit from "p-limit";
-import preferredPM from "preferred-pm";
 import chalk from "chalk";
+import * as fs from "fs";
+import * as path from "path";
 import spawn from "spawndamnit";
 import semverParse from "semver/functions/parse";
 import { askQuestion } from "../../utils/cli-utilities";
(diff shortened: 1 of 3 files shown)


yarn.lock
Andarist, line 1: to be fair, i don't quite see 34 dependencies dropped from here

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -44,8 +45,7 @@ function getCorrectRegistry(packageJson?: PackageJSON): string {
 async function getPublishTool(
   cwd: string
 ): Promise<{ name: "npm" } | { name: "pnpm"; shouldAddNoGitChecks: boolean }> {
-  const pm = await preferredPM(cwd);
-  if (!pm || pm.name !== "pnpm") return { name: "npm" };
+  if (!isPnpm(cwd)) return { name: "npm" };
```
Andarist, line 48: At the moment we only detect pnpm but we have some needs to detect yarn berry too. I wouldn't like to lose that ability here. I see that you also have sent accepted PRs to the `preferred-pm` package... what are we gaining here by inlining it here if Zoltan is willing to improve the package size on his end?

yarn.lock
TheDevMinerTV: This will be dropping these dependencies: https://npmgraph.js.org/?q=preferred-pm@^3.0.0#select=exact%3Apreferred-pm%403.1.4

benmccann: It could also be that they're used somewhere in changesets `devDependencies`. But I can say that for me as a user it really does drop a ton of dependencies because we don't use them elsewhere

packages/cli/src/commands/publish/npm-utils.ts
benmccann: Yes, trying to reduce the size of `preferred-pm` was my initial approach. However, when I tried to upgrade changesets to use the latest version (https://github.com/changesets/changesets/pull/1409), I realized that won't work because the latest versions of `preferred-pm` require Node 18 and changesets supports older versions of Node.

yarn.lock
benmccann: Yeah, the devDependencies are pulling in the removed dependencies. E.g. jest ends up pulling in `pkg-dir`. However, I use vitest in my projects rather than jest, so when I install `@changesets/cli` into my project with these changes I do end up seeing the savings even if they're not visible in yarn lockfile here.

benmccann:
closed by https://github.com/changesets/changesets/pull/1446

PR #1413, opened 2024-07-12
Support lowercase changelog file

### LINKED ISSUE #277: Change changelog filepath/name
## Affected Packages
@changesets/cli

## Problem
I'm using a lowercase file `changelog.md` in my projects and would like to use the changesets CLI with this file instead of having an uppercase named changelog file. I mainly see this uppercased changelog as a relic and would like to use lowercase naming.

## Proposed solution
A field in the configuration to set the name of the file would be nice.

Siilwyn:
Closes #277

Finally got around to this. :) The code stood the test of time.

Unsure about testing this as the current tests fail for me locally:
```
    Cannot find module '/home/siilwyn/codeground/a-forks/changesets/packages/apply-release-plan/src/test-utils/simple-get-changelog-entry'
    Require stack:

      214 |   const changelogOpts = config.changelog[1];
      215 |   let changesetPath = path.join(cwd, ".changeset");
    > 216 |   let changelogPath = resolveFrom(changesetPath, config.changelog[0]);
          |                                  ^
      217 |
      218 |   let possibleChangelogFunc = require(changelogPath);
      219 |   if (possibleChangelogFunc.default) {

      at resolveFileName (node_modules/resolve-from/index.js:29:39)
      at resolveFrom (node_modules/resolve-from/index.js:43:9)

```

However I figured this is quite a simple change so it might not be needed.

### DIFF
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -140,7 +140,13 @@ export default async function applyReleasePlan(
     touchedFiles.push(pkgJSONPath);
 
     if (changelog && changelog.length > 0) {
-      const changelogPath = path.resolve(dir, "CHANGELOG.md");
+      const changelogPathUppercase = path.resolve(dir, "CHANGELOG.md");
+      const changelogPathLowercase = path.resolve(dir, "changelog.md");
+      let changelogPath = changelogPathUppercase;
+      if (fs.existsSync(changelogPathLowercase)) {
+        changelogPath = changelogPathLowercase;
+      }
+
       await updateChangelog(changelogPath, changelog, name, prettierInstance);
       touchedFiles.push(changelogPath);
     }


bluwy:
Seems good to me, but it looks like they were conflicting opinions in the past about the changelog file name: https://github.com/changesets/changesets/pull/457

sebastian-nowak:
What if someone calls it `Changelog.md`? What about `ChangeLog.md`? Hardcoding only two options in the source (`CHANGELOG.md` and `changelog.md`) is just asking for case-sensitivity issues on platforms like Windows.

If this change is supposed to work reliably, it has to do a case-insensitive search for an existing `changelog.md` file in the package root.

Siilwyn:
First of just because I want it to work with lowercase doesn't mean it should work with all 'casings', I think having two reasonable options is enough.
Secondly working without case sensitivity is already the case on filesystems with no case-sensitivity support, as per Node.js guidelines I think it is good to respect the 'casing' when possible: https://nodejs.org/en/learn/manipulating-files/working-with-different-filesystems#adopt-a-superset-approach

Issue #1414, opened 2024-07-15
Replace `chalk` dependency with a lighter alternative


trivikr:
## Affected Packages

* `pkg:cli`
* `pkg:get-dependents-graph`
* `pkg:logger`
* `pkg:read`

## Problem

Recently, the [Ecosystem Performance](https://e18e.dev/) was formed to cleanup, speedup, levelup the JavaScript Ecosystem.
Socket Security team wrote a blog post about it https://socket.dev/blog/javascript-community-launches-e18e-initiative-to-improve-ecosystem-performance

On the e18e cleanup channel on Discord, multiple solutions are being discussed. One of them is to replace `chalk` with `picocolors`.

I see that changesets uses chalk:
* https://github.com/changesets/changesets/blob/b59375614b1b3dabdf67806cd202defb314686a8/packages/read/package.json#L27
* https://github.com/changesets/changesets/blob/b59375614b1b3dabdf67806cd202defb314686a8/packages/logger/package.json#L22
* https://github.com/changesets/changesets/blob/b59375614b1b3dabdf67806cd202defb314686a8/packages/cli/package.json#L87
* https://github.com/changesets/changesets/blob/b59375614b1b3dabdf67806cd202defb314686a8/packages/get-dependents-graph/package.json#L24

## Proposed solution

Replace `chalk` with a lighter alternative, like `picocolors`.

Andarist:
We can consider a PR implementing this change. Ideally, a PR wouldn't change the minimal supported node version of Changesets

trivikr:
> We can consider a PR implementing this change

Posted at https://github.com/changesets/changesets/pull/1417

> Ideally, a PR wouldn't change the minimal supported node version of Changesets.

It doesn't.
* `changeset` supports Node.js 14+
* `picocolors@1.x` supports Node.js 6+

PR #1417, opened 2024-07-15
Replace `chalk` with `picocolors`

trivikr:
Fixes: https://github.com/changesets/changesets/issues/1414

### DIFF
--- a/packages/cli/src/commands/add/messages.ts
+++ b/packages/cli/src/commands/add/messages.ts
@@ -1,5 +1,5 @@
-import chalk from "chalk";
 import outdent from "outdent";
+import pc from "picocolors";
 import { log } from "@changesets/logger";
 import { Release, VersionType } from "@changesets/types";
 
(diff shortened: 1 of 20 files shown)


packages/cli/src/commands/add/messages.ts
```diff
@@ -21,18 +21,19 @@ export default function printConfirmationMessage(
...
-    log(`${chalk.bold.green("minor")}:  ${minorReleases.join(", ")}`);
+    log(`${pc.bold(pc.green("minor"))}:  ${minorReleases.join(", ")}`);
   if (patchReleases.length > 0)
-    log(`${chalk.bold.green("patch")}:  ${patchReleases.join(", ")}`);
+    log(`${pc.bold(pc.green("patch"))}:  ${patchReleases.join(", ")}`);
 
   log("");
 
   if (repoHasMultiplePackages) {
+    // ToDo: Change to pc.redBright when https://github.com/alexeyraspopov/picocolors/pull/55 is published.
```
Andarist, line 33: I wouldn't block this PR on this and I would be OK with just switching to `pc.red` here. But given that it seems that the referenced PR might get published soon-ish... it feels like it's best to wait for it.

It that won't happen in like a week or two then we can drop this TODO and land this PR as-is

trivikr: Yeah. Let's wait for new version of piccocolors to be published for a week or two.

benmccann: @trivikr the new release of picocolors is now out!

trivikr: PR is updated to use `pc.redBright` in [`dcc7754` (#1417)](https://github.com/changesets/changesets/pull/1417/commits/dcc77542a324d290cb5a950553fe8f895b75a16c)

trivikr:
@Andarist The PR is ready for review post picocolors@1.1.0 update.

PR #1419, opened 2024-07-18
Fix logging --version flag

### LINKED ISSUE #1418: Is `--version` flag no longer supported?
I was using code to check the exact version of an installed changeset CLI with `changeset --version`.

I'm not sure if there was an announcement, but I just noticed that its behavior changed to be same as `changeset version` and broke my workflow.

Is that an intended change or a bug?

bluwy:
Fixes #1418

I didn't realize that `meow` would [read the nearest package.json](https://github.com/sindresorhus/meow/blob/87d1bc394d9a1d73b76a81d4874db4f1190259ef/source/options.js#L67-L70) and auto support the `--version` flag, so this PR adds that support manually.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -64,6 +64,13 @@ if (parsed.help && args.length === 1) {
   process.exit(0);
 }
 
+// Version should only be shown if it's the only argument passed
+if (parsed.version && args.length === 1) {
+  // eslint-disable-next-line import/no-extraneous-dependencies
+  console.log(require("@changesets/cli/package.json").version);
+  process.exit(0);
+}
+
 const cwd = process.cwd();
 
 run(parsed._, parsed, cwd).catch((err) => {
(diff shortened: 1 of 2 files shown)


packages/cli/src/index.ts
```diff
@@ -64,6 +64,13 @@ if (parsed.help && args.length === 1) {
   process.exit(0);
 }
 
+// Version should only be shown if it's the only argument passed
+if (parsed.version && args.length === 1) {
+  // eslint-disable-next-line import/no-extraneous-dependencies
+  console.log(require("@changesets/cli/package.json").version);
```
Andarist, line 70: would it work if we'd import it? IIRC Preconstruct (the tools that we use for building packages) uses the `json` plugin. I'm hesitant when it comes to mixing `import`s and `require`s in the same files

bluwy: I copied this from below https://github.com/bluwy/changesets/blob/919bcff06059182c2488c1e5b2c3d1d52c37a888/packages/cli/src/index.ts#L93-L96

I can give import a try though. The best case would also seem to be inlining the version directly in build-time

bluwy:
Hey, checking in if this is ready to be merged now? I think the `require` discussion above shouldn't be a problem as there's also the same code currently.

PR #1423, opened 2024-07-22
docs: Update README.md

SaadBazaz:
Add Material Web Components for React


bluwy:
Thanks for the PR! At the moment we try to not have the list grow too big, so we have a soft minimum of 1k stars. Happy to revisit again in the future. I'll close this for now.

PR #1430, opened 2024-07-30
Fixed version check used when updating internal dependencies

galargh:
This PR fixes the behaviour of the `shouldUpdateDependencyBasedOnConfig` function such that when a `depVersionRange` passed to the function is prefixed with `workspace:`, the function strips the prefix before deciding whether the release version satisfies it or not.

This affects the `changeset version` outcomes as the function is called before the version range is sanitised in: https://github.com/changesets/changesets/blob/b59375614b1b3dabdf67806cd202defb314686a8/packages/apply-release-plan/src/version-package.ts#L51 

The particular user story that led me to the discovery of this issue was that I set `updateInternalDependencies` to `minor` in a monorepo with package A and package B where package B depends on package A (with `"A": `workspace:^X.Y.Z`), and I was surprised to see the version range of A being bumped when I was releasing patch versions of A and B.

I verified the CI passes on this branch in https://github.com/galargh/changesets/pull/2 and added a test case that verifies the fix.

I added a major changeset since this might be a breaking change for some users, but I'd appreciate guidance on whether that's correct for the proposed change.


galargh:
Hey! Do you think there's a chance to get this contribution incorporated?

PR #1431, opened 2024-07-30
Added support for major to updateInternalDependencies config option

galargh:
I'd like to propose adding support for `major` value for the `updateInternalDependencies` config file option (https://github.com/galargh/changesets/blob/8c38d904991e3ff2dc2f41d3cbe3dedbca98d5d3/docs/config-file-options.md#updateinternaldependencies).

The use-case I'm trying to cover is a monorepo with packages A and B, where A depends on B with `workspace:^vX.Y.Z` version range. I'd like the version range to only be automatically updated on major version bumps. Otherwise, the decision should be at the discretion of contributors and should be made explicitly.

I'm aware that the possibility of adding the `major` value to the list of options was mentioned in https://github.com/changesets/changesets/pull/358. However, I'd like to argue that it would be useful to have it to cover the workflows as described above. Especially since it doesn't add much complexity to the implementation.

I added a set of tests to cover internal dependencies update process when `updateInternalDependencies` is set to `major`. The new tests group I added was inspired by similar ones for `updateInternalDependencies=minor` and `updateInternalDependencies=patch`.

I verified the CI is 🟢 in https://github.com/galargh/changesets/pull/1

I added minor changesets to modified packages since this is a new functionality but should not pose a breaking change to the users, but I'd appreciate guidance on whether that's correct for the proposed change.

### DIFF
--- a/.changeset/sour-bottles-judge.md
+++ b/.changeset/sour-bottles-judge.md
@@ -1,0 +1,7 @@
+---
+"@changesets/apply-release-plan": minor
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Added `major` to the list of acceptable values for the `updateInternalDependencies` config file option
--- a/docs/config-file-options.md
+++ b/docs/config-file-options.md
@@ -128,7 +128,9 @@ pkg-b @ version 1.0.1
   depends on pkg-a at range `^1.0.0
 ```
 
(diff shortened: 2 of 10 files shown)


galargh:
Hey! Do you think there's a chance to get this contribution incorporated?

PR #1436, opened 2024-08-11
clean up unlisted binaries and unused dependencies, exports, and files

with-heart:
This PR adds [`knip`](https://knip.dev/) and uses it to clean up unlisted binaries and unused dependencies, exports, and files in the workspace.


package.json
```diff
@@ -13,7 +13,6 @@
     "lint:prettier": "prettier --list-different \"**/*.{js,ts,tsx,md}\"",
     "types:check": "tsc",
     "prettier": "prettier --write \"**/*.{js,ts,tsx,md}\"",
-    "changeset": "packages/cli/bin.js",
```
Andarist, line 16: I believe this is what allows us to call `yarn changeset` from other scripts

packages/apply-release-plan/package.json
```diff
@@ -19,7 +19,6 @@
   "license": "MIT",
   "repository": "https://github.com/changesets/changesets/tree/main/packages/apply-release-plan",
   "dependencies": {
-    "@babel/runtime": "^7.20.1",
```
Andarist, line 22: this is an implicit dependency, Babel might inject references to this module in its generated output

package.json
with-heart: I double checked to make sure that adding `@changesets/cli` to the root would add the binary to `node_modules/.bin` but I'll triple-check it 😀

packages/apply-release-plan/package.json
with-heart: Gotcha, I'll readd those and add them to the ignore

package.json
Andarist: ye, it might do it - i only noticed that extra addition after adding this comment

package.json
```diff
@@ -54,14 +52,13 @@
...
     "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-jest": "^27.1.5",
     "eslint-plugin-n": "^15.5.1",
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-promise": "^6.1.1",
-    "eslint-plugin-standard": "^5.0.0",
     "jest": "^29.3.1",
     "jest-junit": "^15.0.0",
     "jest-watch-typeahead": "^2.2.1",
+    "knip": "^5.27.2",
```
benmccann, line 61: do we need to install `knip`? it adds a lot of dependencies and most people will never run it. I would add a script above like `"knip": "yarn dlx knip"` instead

Andarist: I prefer to keep it in the lockfile and all. It's a dev-only tool. I don't want us to go crazy over keeping the install size small for *maintainers*.

bluwy:
It looks like we hadn't followed up on this well and this has a lot of conflicts now. We're open to adding knip by default though, so if you're still interested in this, feel free to send a new PR. Appreciate the contribution!

PR #1437, opened 2024-08-12
upgrade node to v18 for csb

with-heart:
This PR upgrades the version of `node` for CodeSandbox to v18 LTS.

Needed for #1436

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,4 +1,4 @@
 {
   "packages": ["packages/*"],
-  "node": "14"
+  "node": "18"
 }


Andarist:
tbf, i'd probably just remove this integration altogether instead of updating it ;p

with-heart:
It looks like it's a GitHub App so I'm not sure what will happen if the config is completely removed

benmccann:
Rather than doing this, I'd avoid adding `knip` as a dependency: https://github.com/changesets/changesets/pull/1436/files#r1740307139

Andarist:
> It looks like it's a GitHub App so I'm not sure what will happen if the config is completely removed

Good point. I just uninstalled it from this org. Could you remove the config file and retitle this PR?

bluwy (approved):

PR #1445, opened 2024-08-27
Remove unused `@babel/runtime` dependency

bluwy:
I might be doing something silly, but after checking the files, it doesn't look like `@babel/runtime` is used anywhere and in `dist`, so I think we can remove it.

I also:
- moved `human-id` as dev dep in `cli`
- removed `fs-extra` in `get-dependents-graph`
- removed `@changesets/types` in `git`

They don't seem to be used in the published code. Even though these deps are used elsewhere in transitive deps in practice, I think it's nice to keep them concise if unused.

### DIFF
--- a/.changeset/selfish-flowers-work.md
+++ b/.changeset/selfish-flowers-work.md
@@ -1,0 +1,14 @@
+---
+"@changesets/assemble-release-plan": patch
+"@changesets/get-dependents-graph": patch
+"@changesets/should-skip-package": patch
+"@changesets/apply-release-plan": patch
+"@changesets/get-release-plan": patch
+"@changesets/write": patch
+"@changesets/read": patch
+"@changesets/cli": patch
+"@changesets/git": patch
+"@changesets/pre": patch
+---
+
+Remove unused `@babel/runtime` dependency
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "@babel/core": "^7.20.2",
     "@babel/preset-env": "^7.20.2",
     "@babel/preset-typescript": "^7.18.6",
-    "@babel/runtime": "^7.20.1",
     "@manypkg/cli": "^0.19.1",
     "@preconstruct/cli": "^2.8.1",
     "@types/fs-extra": "^5.1.0",
--- a/packages/apply-release-plan/package.json
+++ b/packages/apply-release-plan/package.json
@@ -19,7 +19,6 @@
   "license": "MIT",
   "repository": "https://github.com/changesets/changesets/tree/main/packages/apply-release-plan",
   "dependencies": {
-    "@babel/runtime": "^7.20.1",
     "@changesets/config": "^3.0.2",
     "@changesets/get-version-range-type": "^0.4.0",
     "@changesets/git": "^3.0.0",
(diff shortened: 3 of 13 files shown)


Andarist:
> I might be doing something silly, but after checking the files, it doesn't look like @babel/runtime is used anywhere and in dist, so I think we can remove it.

This is a little bit surprising to me cause our compilation target is pretty low:
https://github.com/changesets/changesets/blob/4efc0382fcf7ef0ebd0d8e0515b655b6870728f1/babel.config.js#L6

From what I can see this node version already supports things like object rest spread and classes so perhaps this is actually suspected. We don't use any fancy/esoteric new syntaxes that would usually require those helpers.

bluwy:
After fiddling with https://github.com/Thinkmill/manypkg today too, I think I understand better now. The babel config is probably missing `plugins: ["@babel/plugin-transform-runtime"],` which will use from `@babel/runtime`. So right now all the runtime helpers are inlined into the package instead.

So I think we have 2 directions here:
1. (this pr) remove `@babel/runtime` and inline the runtime helpers
2. Add `@babel/plugin-transform-runtime` to use the runtime package

Andarist:
Tbf, im not that sure if Changesets would work today on such an old version of node. If we couod confirm that it doesnt… then we could bump the compilation target and drop this dependency

benmccann:
> Tbf, im not that sure if Changesets would work today on such an old version of node. If we couod confirm that it doesnt… then we could bump the compilation target and drop this dependency

Since the same question came up in https://github.com/changesets/changesets/pull/1446, I've already investigated this a bit. There are 9 dependencies of changesets that specify they only support Node 10 in their `engines` field, so I expect changesets would not run on Node 8

Andarist:
At times, those `engines` declarations are not accurate/up to date. You could try to run `changeset version && changeset publish` on node 10 and node 12. There is non-zero chance that it would fail during init time or smth in one of them.

bluwy:
I haven't got the chance to test on node 8 and above since it's a hard to run them on macos arm (and nvm needs to compile them locally due to lack of prebuilt binaries). Maybe we can setup running CI on them to check it?

In any case, I think this PR wouldn't be too bad if merged too. The dependencies are already unused today.

benmccann:
The CI passes for me on Node 8, but running `changeset version` on an actual project doesn't:
```
$ npm run changeset:version

> kit-monorepo@0.0.1 changeset:version ~/src/kit
> changeset version

node_modules/globby/index.js:28
	} catch {
	        ^

SyntaxError: Unexpected token {
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (node_modules/@manypkg/get-packages/dist/get-packages.cjs.dev.js:16:14)
```

It fails trying to run `globby@11.1.0` which has:
```
  "engines": {
    "node": ">=10"
  }
```

Andarist:
Does it run on node 10? ;p

benmccann:
As far as I can tell, it does run on Node 10. I don't have a GitHub token setup locally in order to be able to test it the whole way through

I wonder if we should just remove babel? It doesn't really seem like it's necessary since object spreading has been fully supported since Node 8. This PR seems like a good first step

Andarist:
Node 10 doesn't support optional chaining syntax. I'd like to be able to use it :P

PR #1455, opened 2024-09-08
feat: update package-lock.json version

### LINKED ISSUE #1139: Version not updated in package-lock.json
## Affected Packages

Don't know

## Problem

When using npm, running `changeset version` does not update the version in the lock file. 

This does not appear to have any ill effect and may be purely cosmetic

## Proposed solution

Running `npm install` ensures that the version is updated

almahdi404:
Fixes #1139

### DIFF
--- a/.changeset/shy-islands-type.md
+++ b/.changeset/shy-islands-type.md
@@ -1,0 +1,5 @@
+---
+"@changesets/apply-release-plan": minor
+---
+
+feat: update package-lock.json version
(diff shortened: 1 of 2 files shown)


almahdi404:
@Andarist . please review

bluwy:
I don't think it should be reading/parsing/writing the lockfile this way 🤔 The package manager (npm) owns it and the format could always change. Maybe this needs to be documented, or `changeset version` could offer a prompt / CLI flag to invoke `npm i` again.

PR #1458, opened 2024-09-15
chore: move `@types/semver` to `devDependencies`

benmccann:
This is not used in the distributed types of `changesets`, so doesn't need to be listed in `dependencies`

### DIFF
--- a/.changeset/red-walls-poke.md
+++ b/.changeset/red-walls-poke.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/red-walls-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+chore: move @types/semver to devDependencies
```
Andarist, line 5:
```suggestion
Moved `@types/semver` to `devDependencies`
```

PR #1460, opened 2024-09-15
fix: include missing peer depedency updates in release notes

### LINKED ISSUE #1459: Empty changelog entries when using experimental options
## Affected Packages

- `@changesets/apply-release-plan`

## Problem

When using the following experimental config options:

```json
{
  // ... general config
    "onlyUpdatePeerDependentsWhenOutOfRange": true,
  }
}
```

The changelogs (and Release Candidate PRs) include empty sections for packages that are bumped due to one of their `peerDependencies` having bumped.

For example, see the following PR screenshot, this is from a real PR, where the names of the packages have been renamed for privacy reasons, and to make things easier to explain:

Screenshot 2024-09-15 at 21 03 25

As you can see, we have included a changeset for `package-a` due to a bug fix. We have `package-a` under `peerDependencies` in `package-b`, `package-c` and `package-d`; and it's under `dependencies` for `app-a` and `app-b`.

The changelog correctly includes that `package-b`, `package-c` and `package-d` will be bumped, however it doesn't specify why. Whereas for `app-a` and `app-b` it _does_ specify why it's being bumped.

## Proposed solution

The `apply-release-plan` package needs updating to take into account the `updateInternalDependents: always` option when detecting whether a dependency should be updated.

breadadams:
- Fixes #1459
- Minor rename of `test` to `it` in test suite, to match others.
- Minor typo fix in issue template.

### DIFF
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -2736,7 +2736,7 @@ describe("apply release plan", () => {
       - Hey, let's have fun with testing!`);
     });
 
-    it("should add an updated peerDependencies line when peerDependencies have been updated and updateInternalDependents is `always`", async () => {
+    it("should add an updated dependencies line when peerDependencies have been updated and updateInternalDependents is `always`", async () => {
       let { changedFiles } = await testSetup(
         {
           "package.json": JSON.stringify({
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -2833,20 +2833,12 @@ describe("apply release plan", () => {
         a.endsWith(`pkg-c${path.sep}CHANGELOG.md`)
       );
 
-      console.log("readmePath", readmePath);
-      console.log("readmePathB", readmePathB);
-      console.log("readmePathC", readmePathC);
-
       if (!readmePath || !readmePathB || !readmePathC)
         throw new Error(`could not find an updated changelog`);
 
       let readme = await fs.readFile(readmePath, "utf-8");
       let readmeB = await fs.readFile(readmePathB, "utf-8");
       let readmeC = await fs.readFile(readmePathC, "utf-8");
-
-      console.log("readme", readme.trim());
-      console.log("readmeB", readmeB.trim());
-      console.log("readmeC", readmeC.trim());
 
       expect(readme.trim()).toEqual(outdent`# pkg-a
 
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -2196,7 +2196,7 @@ describe("apply release plan", () => {
       expect(pkgAChangelogPath).toBeUndefined();
     });
 
-    test("should list multi-line same-type summaries correctly", async () => {
+    it("should list multi-line same-type summaries correctly", async () => {
       const releasePlan = new FakeReleasePlan([
         {
           id: "some-id-1",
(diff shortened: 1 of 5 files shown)


packages/apply-release-plan/src/utils.ts
```diff
@@ -45,7 +47,9 @@ export function shouldUpdateDependencyBasedOnConfig(
   let shouldUpdate = getBumpLevel(release.type) >= minLevel;
 
   if (depType === "peerDependencies") {
-    shouldUpdate = !onlyUpdatePeerDependentsWhenOutOfRange;
+    shouldUpdate =
+      !onlyUpdatePeerDependentsWhenOutOfRange ||
+      updateInternalDependents === "always";
   }
```
Andarist, line 53: hm... the alternative would be to say that peer dependencies are special (cause they kinda are...) and skip updating them altogether.

The reason why we have introduced `updateInternalDependents === "always"` was to make it easier for dependency updates to "flow" into their dependants. But with peers... that's not even a thing. A peer dependency and its peer dependant always form a kind of an implicit relationship. It's explicitly specified by the contract in the `package.json` but it's enforced in a way different way than with regular dependencies.

I also somewhat question if `'*'` range should work with `updateInternalDependents === "always"`. But then I guess if it says always... then it means always 🤣 But again, this really was meant to work with ranges that specific some *minimum* compatible version and `*` is so loose.

packages/apply-release-plan/src/index.test.ts
```diff
@@ -2735,6 +2735,137 @@ describe("apply release plan", () => {
...
+      - Hey, let's have fun with testing!`);
+
+      expect(readmeC.trim()).toEqual(outdent`# pkg-c
+
+      ## 2.0.1
+
+      ### Patch Changes
+
+      - Updated dependencies
+        - pkg-b@1.2.1`);
```
Andarist, line 2867: With `*` this is a big fat lie. The dependency itself was *not* updated. How we are even meant to reason about this?

When installing `pkg-c@2.0.1` the package manager is free to leave `pkg-b@1.2.0` in `node_modules`. It doesn't have to bother updating it because the `*` range says it's OK to leave it alone.

packages/apply-release-plan/src/utils.ts
breadadams: I see where you're coming from. The fact that our packages are being bumped when a peer is bumped made me think this was intentional behaviour.

We actually have some custom logic running in a `prepack` script for all our packages (to _fix_ that issue) which manually modifies the peer dep. versions from being `"*"` to the latest version of said dependency.

We did this because we're using the packages within a monorepo **and** on some projects that are (currently) outside of the monorepo. If we were to leave the `"*"` in the published version nothing forces us to update the peers when updating a package in one of our "outside-monorepo" projects.

We're not using ranges anywhere for our internal peer deps, because we ensure that whenever we update a package we also update its (internal) peer deps.

Having said that, this requirement might be specific to our use case - however like I say, since packages are bumped when a peer is, we took it as intentional behaviour and really dislike the empty changelogs 😕.

packages/apply-release-plan/src/index.test.ts
breadadams: Oof, right, so this behaviour we're seeing could actually be considered a bug? 😅

As mentioned in my other comment, we sort of depend on it right now - and it works quite nicely, minus the empty changesets.

There's no way for us to hook into this behaviour I believe? i.e. to make this functionality intentional on our side.

What about a new config option (maybe under experimental?) that when `true` would add a changelog entry for bumps due to peer dep. bumps... what do you think?

I get this is an important decision, but would **love** to have something thay solves our problem asap to stop getting these darn empty changelog entries 🥺.

breadadams:
Hey @Andarist, is there anything we can do to get this moving?

Whilst I understand your concerns expressed in the previous comments, these changes are only updating the "changelog logic" to match the "release logic" that is already implemented, i.e. using `"*"` versions with the `updateInternalDependents` option **does** create new releases.

PR #1465, opened 2024-09-18
Run CI on multiple node versions

bluwy:
From https://github.com/changesets/changesets/pull/1445, maybe it's useful to run the CI on multiple node versions to see how well older node versions are still supported and prevent breaking them.


Andarist:
It would be good to add this but I think it's enough that we added this now to the `next` branch in: https://github.com/changesets/changesets/pull/1480

Issue #841, opened 2022-06-16
TypeError: Cannot read property 'packageJson' of undefined


fupengl:
## Affected Packages

```json
{
  "devDependencies": {
    "@changesets/cli": "^2.23.0",
    "typescript": "4.7.3"
  }
}
```
node `v14.19.3`
pnpm `v7.2.1`
pc `Apple M1 Pro`

## Problem

<img width="1151" alt="image" src="https://user-images.githubusercontent.com/20211964/174001558-b165de9e-db3d-4c30-b00d-d513b4af5cbe.png">

```sh
🦋  error TypeError: Cannot read property 'packageJson' of undefined
🦋  error     at createChangeset (/Users/fupeng/Workspace/github/silly-girl/node_modules/.pnpm/@changesets+cli@2.23.0/node_modules/@changesets/cli/dist/cli.cjs.dev.js:415:75)
🦋  error     at add (/Users/fupeng/Workspace/github/silly-girl/node_modules/.pnpm/@changesets+cli@2.23.0/node_modules/@changesets/cli/dist/cli.cjs.dev.js:507:26)
🦋  error     at processTicksAndRejections (internal/process/task_queues.js:95:5)
🦋  error     at async run$2 (/Users/fupeng/Workspace/github/silly-girl/node_modules/.pnpm/@changesets+cli@2.23.0/node_modules/@changesets/cli/dist/cli.cjs.dev.js:1254:5)
```

mahboobmonnamd:
The reason may be for this
1. Package.json file is not at the root level
2. If it is mono repo, package.json should be available within the each service folder

fupengl:
mono repo project
changesets are installed in the root directory
The rest of the subpackages also have package.json

mahboobmonnamd:
Check do you have workspace file. I use pnpm. So i have pnpm-workspace.yaml.

fupengl:
I am also pnpm, workspace related declarations are normal
I was wondering if it was a problem with my nodejs version, I migrated from lerna

mahboobmonnamd:
I had this issue, i resolved after doing the above steps. writing it again to ensure you too follow the same
1. have pnpm-workspace.yaml, should have references to those repos
2. install @changeset/cli
3. Added package.json inside the service repos

fupengl:
<img width="905" alt="image" src="https://user-images.githubusercontent.com/20211964/178671429-a440da61-5bb2-4082-8e1c-60f0d86dcdd3.png">

[...]

PR #1466, opened 2024-09-21
Add error message and exit when there are no versionable packages

### LINKED ISSUE #1170: Cannot read properties of undefined (reading 'packageJson')
## Affected Packages
pkg:@changesets/cli v2.26.1

## Problem
Running `changeset add` throws this error.
Ive been trying to use it in my repo [here](https://github.com/Yakiyo/dsm). Its a rust based project but i think changeset should work with it too. I'm using pnpm as a package manager but ive tried with npm and yarn too. Both approach failed. 
My `.changeset/config.json`
```json
{
  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
  "commit": false,
  "fixed": [],
  "linked": [],
  "access": "restricted",
  "baseBranch": "main",
  "ignore": []
}
```
Its a single package repo, not a mono repo. Im on windows but ive also tried on a linux container in github codespaces. Not sure what exactly is going wrong.

image

marcalexiei:
Close #1170
fix https://github.com/changesets/changesets/issues/841

I have added an informative message in the scenario where no versionable packages are found.
This should cover also the error reported in the issue linked above (public packages without `version` field in package).


packages/cli/src/commands/add/__tests__/add-no-versionable-packages.ts
marcalexiei, line 1: I created a separate test file in the add command to avoid clash with `silenceLogsInBlock` mocks

packages/cli/src/commands/add/index.ts
```diff
@@ -19,20 +19,29 @@ export default async function add(
...
         allowPrivatePackages: config.privatePackages.version,
       })
   );
+
+  if (versionablePackages.length === 0) {
+    info("No versionable package detected");
+    info('- Check "ignore" setting info configuration');
+    info('- Ensure that relevant package.json file(s) have "version" field');
+    return;
+  }
```
bluwy, line 43: I wonder if we should be throwing an error here like above, since it seems unexpected that there's no versionable packages to add changesets for.

Maybe something like this? (update the wording too, let me know what you think)
```suggestion
  if (versionablePackages.length === 0) {
    error("No versionable packages found");
    error('- Ensure the packages to version are not in the "ignore" option');
    error('- Ensure that relevant package.json files have the "version" field');
    throw new ExitError(1);
  }
```

marcalexiei: I thought about throwing an error, but IMHO it was too "scary" (Don't know a better term 😢).
But if you handle similar scenarios using errors where you throw the error I will update the code and the test accordingly.

marcalexiei: Additional thought: if changesets are created via automated tool, have no versionable packages should trigger an error to interrupt any following command / tasks.

In the end, I agree with you, an error is better to handle this scenario, I'll the code to issue an error.

bluwy (approved):
I think the flow and fix here looks good. Not sure if others have thoughts on the error logged info.

packages/cli/src/commands/add/__tests__/add-no-versionable-packages.ts
```diff
@@ -0,0 +1,71 @@
+import { defaultConfig } from "@changesets/config";
+import * as logger from "@changesets/logger";
+import { testdir } from "@changesets/test-utils";
+
+import addChangeset from "..";
+
+describe("Add command - no versionable packages", () => {
```
Andarist, line 7: i'd prefer to move this to regular tests for `add`, even with `silenceLogsInBlock` u can still assert on the logs as all log methods should be jest spies

packages/cli/src/commands/add/__tests__/add-no-versionable-packages.ts
```diff
@@ -0,0 +1,71 @@
...
+  const loggerErrorSpy = jest
+    .spyOn(logger, "error")
+    .mockImplementation(jest.fn(() => {}));
+
+  it("should print an informative message when in a polyrepo", async () => {
```
Andarist, line 12:
```suggestion
  it("should print an informative message when in a single-package repo", async () => {
```

packages/cli/src/commands/add/index.ts
```diff
@@ -19,20 +20,29 @@ export default async function add(
...
         ignore: config.ignore,
         allowPrivatePackages: config.privatePackages.version,
       })
   );
+
+  if (versionablePackages.length === 0) {
+    error("No versionable packages found");
+    error('- Ensure the packages to version are not in the "ignore" config');
+    error('- Ensure that relevant package.json files have the "version" field');
+    throw new ExitError(1);
```
Andarist, line 43: note for myself: this is nicely handled here:
https://github.com/changesets/changesets/blob/7ccb22e302dfc46d18e01f81f7979455960c0db9/packages/cli/src/index.ts#L105-L107

so 👍

packages/cli/src/commands/add/__tests__/add.ts
```diff
@@ -367,4 +368,70 @@ describe("Add command", () => {
     const { choices } = askCheckboxPlus.mock.calls[0][1][0];
     expect(choices).toEqual(["pkg-a", "pkg-c"]);
   });
+
+  describe("no versionable packages", () => {
```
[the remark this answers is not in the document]
marcalexiei: I try to check that the describe block are positioned in the relevant  block but I understand your point of view.
No problem rewriting the titles adding more info, right now they were outdated 😅.

PR #1467, opened 2024-09-21
Update and align `@types/jest` with `jest` runtime and remove ts-ignores

marcalexiei:
- `jest` and `@types/jest` have been updated to the latest version (`29.7.0`)
  Previously the types version wasn't matching the runtime version
- code now no longer contains `@ts-ignore`
  - In unit test they have been replaced with "type checkable" code using [just.mocked](https://jestjs.io/docs/mock-function-api#jestmockedsource-options) helper
  - source code has some `@ts-ignore` that were no longer valid
  - `@ts-ignore` that cannot be removed have been replaced with `@ts-expect-error` directive followed by an explanation 

> [!WARNING]
> I had to enable `skipLibCheck` in `tsconfig.json` due to some types error coming from `enquirer` package.
> I updated `enquirer` to `2.4.1` and the error is no longer present.
> However updating this package in this PR seems out of scope so I opted to temporary disable typecheck.
> (If you wish I can open a new PR after this one is merged)


packages/cli/src/commands/publish/__tests__/releaseCommand.test.ts
```diff
@@ -1,15 +1,15 @@
-import publishPackages from "../publishPackages";
 import * as git from "@changesets/git";
 import { defaultConfig } from "@changesets/config";
 import { silenceLogsInBlock, testdir } from "@changesets/test-utils";
+
 import runRelease from "..";
+import publishPackages from "../publishPackages";
```
bluwy, line 6: order

packages/write/src/index.test.ts
```diff
@@ -1,12 +1,13 @@
...
 import path from "path";
+import fs from "fs-extra";
 import parse from "@changesets/parse";
-import writeChangeset from "./";
-
-import humanId from "human-id";
 import { testdir } from "@changesets/test-utils";
+import humanId from "human-id";
+
+import writeChangeset from "./";
```
bluwy, line 7: order

bluwy (commented):
The refactor for mock types looks really good. I've noticed some imports shifted around, do you mind putting the order back to reduce the diff? (I've noted them in the comments below)

Mainly this will help merging future changes that we're working on in the `next` branch. Thanks!

bluwy (approved):

marcalexiei (commented):
Requested file imports have been reverted to their original order

bluwy (approved):

bluwy:
I'm not the main maintainer around here, but I'd say go ahead! Maybe it's better to send the PR first? So we don't have a short timeframe where we have to use skipLibCheck. The other maintainers may have opinions about `skipLibCheck` too that could delay merging this PR.

marcalexiei:
I rechecked the type errors and adding `@types/node` to the root with the version already installed on `main` solved the issue. 

https://github.com/changesets/changesets/blob/7ccb22e302dfc46d18e01f81f7979455960c0db9/yarn.lock#L2016-L2019

So now there is no need to use `skipLibCheck`!

marcalexiei:
@bluwy 
Just re-asking the review since I changed the code after your previous approval. 
When you have time, let me know if everything checks out for you 🙂

marcalexiei:
I noticed that `enquirer` has been updated, alongside `@types/node`, in the `next` branch  via #1476.

If you prefer we can merge this PR into `next` instead of `main`, 
if that is the case let me know and I'll rebase this branch on `next`.

bluwy (approved):
You don't need to re-request review if there isn't any significant code change. It'll also be easier to review what changed if you push a new commit instead of force pushing.

PR #1474, opened 2024-10-03
fix: Throw descriptive error if a package is not found

sidharthv96:
The present error thrown doesn't have anything which could help the maintainer with identifying what's wrong. 

This PR will add the unidentified package name, along with the changeset ID to the error. 

## Current error

```
❯ npx changeset version
🦋  error TypeError: Cannot destructure property 'packageJson' of 'undefined' as it is undefined.
🦋  error     at Object.shouldSkipPackage
...
```

Example of a [failed CI run at mermaid](https://github.com/mermaid-js/mermaid/actions/runs/11163955917)

## Updated error

```
"big-cats-delight" changeset mentions a release for a package "impossible-package" but such a package could not be found."
```

This should help the user identify what exactly the problem is.

### DIFF
--- a/packages/assemble-release-plan/src/index.test.ts
+++ b/packages/assemble-release-plan/src/index.test.ts
@@ -578,28 +578,6 @@ describe("assemble-release-plan", () => {
     expect(releases[1].newVersion).toEqual("1.0.0");
   });
 
-  it("should throw an error when a changeset contains a package that is not in the workspace", () => {
-    setup.addChangeset({
-      id: "big-cats-delight",
-      releases: [{ name: "pkg-a", type: "major" }],
-    });
-    setup.addChangeset({
-      id: "small-dogs-sad",
-      releases: [{ name: "pkg-z", type: "minor" }],
-    });
-
-    expect(() =>
-      assembleReleasePlan(
-        setup.changesets,
-        setup.packages,
-        defaultConfig,
-        undefined
-      )
-    ).toThrow(
-      "Found changeset small-dogs-sad for package pkg-z which is not in the workspace"
-    );
-  });
-
   describe("fixed packages", () => {
     it("should assemble release plan for fixed packages", () => {
       setup.addChangeset({


packages/assemble-release-plan/src/index.ts
```diff
@@ -276,8 +276,14 @@ function getRelevantChangesets(
     const skippedPackages = [];
     const notSkippedPackages = [];
     for (const release of changeset.releases) {
+      const pkg = packagesByName.get(release.name);
+      if (!pkg) {
+        throw new Error(
+          `Could not find matching package for release of: "${release.name}" in changeset ${changeset.id}`
+        );
```
bluwy, line 283: Looks like there's already a similar error being reported at https://github.com/changesets/changesets/blob/7ccb22e302dfc46d18e01f81f7979455960c0db9/packages/assemble-release-plan/src/flatten-releases.ts#L31-L33

Maybe we can reuse that message instead. But also, seems like maybe it should be doing the validation earlier. Perhaps it can pre-check for this issue after the `packagesByName` initialization here?

https://github.com/changesets/changesets/blob/7ccb22e302dfc46d18e01f81f7979455960c0db9/packages/assemble-release-plan/src/index.ts#L154-L163

And we could remove the handling in the two places.

sidharthv96: Thanks for the pointers! Will investigate.

sidharthv96: As `getRelevantChangesets` changes is called immediately after `packagesByName` is initialized, the check inside `getRelevantChangesets` should be enough. Otherwise, we'd have to do an extra loop just for validation. 

I think moving the error from `flattenReleases` to `getRelevantChangesets` should be enough, as the data will be validated before going to `flattenReleases`. We can use a `!` inside `flattenReleases`, as done in numerous places. I initially wrote a function to get the package or throw the error, but it felt unnecessary in our use case.

bluwy: I would kinda prefer "a function to get the package or throw the error" since that makes the function call ordering less fragile 🤔 I don't think it's too bad though and it could be places in the utils file.

sidharthv96: Added in https://github.com/changesets/changesets/pull/1474/commits/9af3631b6f06ef36884ce3d6b2448a4b4eca99a5

bluwy (approved):

bluwy:
This handles a few more missing places with a unified error message compared to https://github.com/changesets/changesets/pull/1508, hence I've removed the specific test from that PR as it's already covered in here.

PR #1476, opened 2024-10-03
Remove `fs-extra`

ziebam:
Part of the v3 cleanup: https://github.com/changesets/changesets/discussions/1473. `fs-extra` is effectively unnecessary in modern Node versions, and can be replaced with the native functionality.

### DIFF
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -5,7 +5,7 @@ import {
   ComprehensiveRelease,
 } from "@changesets/types";
 import * as git from "@changesets/git";
-import { readFile } from "fs/promises";
+import fsp from "fs/promises";
 import path from "path";
 import outdent from "outdent";
 import spawn from "spawndamnit";
@@ -131,7 +131,7 @@ async function testSetup(
 }
 
 async function readJson(path: string) {
-  return JSON.parse(await readFile(path, "utf8"));
+  return JSON.parse(await fsp.readFile(path, "utf8"));
 }
 
 describe("apply release plan", () => {
@@ -155,7 +155,7 @@ describe("apply release plan", () => {
         let pkgPath = changedFiles.find((a) => a.endsWith(`package.json`));
 
         if (!pkgPath) throw new Error(`could not find an updated package json`);
-        let pkgJSON = await readFile(pkgPath, { encoding: "utf-8" });
+        let pkgJSON = await fsp.readFile(pkgPath, { encoding: "utf-8" });
 
         expect(pkgJSON).toStrictEqual(`{
   "name": "pkg-a",
@@ -184,7 +184,7 @@ describe("apply release plan", () => {
         let pkgPath = changedFiles.find((a) => a.endsWith(`package.json`));
 
         if (!pkgPath) throw new Error(`could not find an updated package json`);
-        let pkgJSON = await readFile(pkgPath, { encoding: "utf-8" });
+        let pkgJSON = await fsp.readFile(pkgPath, { encoding: "utf-8" });
 
         expect(pkgJSON).toStrictEqual(`{
 \t"name": "pkg-a",
@@ -206,7 +206,7 @@ describe("apply release plan", () => {
         let pkgPath = changedFiles.find((a) => a.endsWith(`package.json`));
 
         if (!pkgPath) throw new Error(`could not find an updated package json`);
-        let pkgJSON = await readFile(pkgPath, { encoding: "utf-8" });
+        let pkgJSON = await fsp.readFile(pkgPath, { encoding: "utf-8" });
 
         expect(pkgJSON).toStrictEqual(`{
   "name": "pkg-a",
@@ -229,7 +229,7 @@ describe("apply release plan", () => {
         let pkgPath = changedFiles.find((a) => a.endsWith(`package.json`));
 
         if (!pkgPath) throw new Error(`could not find an updated package json`);
-        let pkgJSON = await readFile(pkgPath, { encoding: "utf-8" });
+        let pkgJSON = await fsp.readFile(pkgPath, { encoding: "utf-8" });
 
         expect(pkgJSON).toStrictEqual(`{
   "name": "pkg-a",
@@ -2043,7 +2043,7 @@ describe("apply release plan", () => {
       );
 
       if (!readmePath) throw new Error(`could not find an updated changelog`);
-      let readme = await readFile(readmePath, "utf-8");
+      let readme = await fsp.readFile(readmePath, "utf-8");
 
       expect(readme.trim()).toEqual(outdent`# pkg-a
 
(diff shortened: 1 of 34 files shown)


packages/apply-release-plan/src/index.test.ts
```diff
@@ -128,6 +129,19 @@ async function testSetup(
...
+}
+
+async function outputFile(
+  filePath: string,
+  content: string,
+  encoding = "utf8" as ObjectEncodingOptions
+) {
+  await mkdir(path.dirname(filePath), { recursive: true });
+  await writeFile(filePath, content, encoding);
+}
```
ziebam, line 143: I've implemented this function in a few files. Might be a good idea to have in some shared directory, but I'm not sure what would be the best place.

packages/cli/src/commands/version/version.test.ts
```diff
@@ -113,6 +113,7 @@ beforeEach(() => {
 
 afterEach(() => {
   console.error = consoleError;
+  jest.restoreAllMocks();
```
ziebam, line 116: Some tests were failing without this when run as a suite, but passed individually. Not sure why.

packages/apply-release-plan/src/index.test.ts
```diff
@@ -2950,7 +2964,9 @@ describe("apply release plan", () => {
       );
 
       // @ts-ignore this is possibly bad
-      let pathExists = await fs.pathExists(changesetPath);
+      let pathExists = await access(changesetPath)
+        .then(() => true)
+        .catch(() => false);
```
ziebam, line 2969: Same here, could be a common utility function somewhere. This implementation is also what `fs-extra.pathExists` is doing under the hood, so we might need to credit the author here? Not sure since even Node documentation suggests using `access` for asynchronous existence checks.

packages/apply-release-plan/src/index.test.ts
Andarist: it likely could be put in https://github.com/changesets/changesets/tree/7ccb22e302dfc46d18e01f81f7979455960c0db9/scripts/test-utils

bluwy:
This looks great to me too. I don't know what others think, but I usually prefer `import fs from 'node:fs/promises` rather than named imports because the API surface is large and names can likely collide or misinterpreted as custom functions. But I don't think it's a big deal if others are fine by the named imports too.

For the access/then/catch pattern, I wonder if it's also fine to simply use `fsSync.existsSync`. It's not exactly the same as `access` also checks for file permissions rather than just existing, but I don't know if it'd matter in most case. But again also not a big deal for me 😄

ziebam:
> This looks great to me too. I don't know what others think, but I usually prefer `import fs from 'node:fs/promises` rather than named imports because the API surface is large and names can likely collide or misinterpreted as custom functions. But I don't think it's a big deal if others are fine by the named imports too.
> 
> For the access/then/catch pattern, I wonder if it's also fine to simply use `fsSync.existsSync`. It's not exactly the same as `access` also checks for file permissions rather than just existing, but I don't know if it'd matter in most case. But again also not a big deal for me 😄

I'm fine with changing to default imports as my preference for named imports is very shallow, as in I can't even quite put my finger on why I prefer them. I'm gonna hold off on making the change for now though, in case someone with a stronger opinion pops in! 😄 

As for `existsSync`, that's what I've usually ended up doing in other projects where I've worked on removing `fs-extra`, but I usually try to keep the behavior as close to the original `fs-extra` usage as possible on the first pass, and then make the change if the maintainers are fine with it.

packages/cli/src/commands/version/version.test.ts
Andarist: hm, interesting - i'll take a look into this before landing htis

Andarist:
i softly prefer dotted access to those but it's harder to set up autoimports for that than for named exports so 🤷

Andarist:
@ziebam could you add a changeset to this PR?

ziebam:
> @ziebam could you add a changeset to this PR?

I initially branched off off `main` (the `next` branch didn't exist yet), and later caught up with `next` manually, so now I get this error:
```bash
$ packages/cli/bin.js
🦋  error Error: Failed to find where HEAD diverged from next. Does next exist?
🦋  error     at getDivergedCommit (/home/pralky/dev/changesets/packages/git/src/index.ts:51:11)
🦋  error     at getChangedFilesSince (/home/pralky/dev/changesets/packages/git/src/index.ts:196:22)
🦋  error     at getChangedPackagesSinceRef (/home/pralky/dev/changesets/packages/git/src/index.ts:258:24)
🦋  error     at getVersionableChangedPackages (/home/pralky/dev/changesets/packages/cli/src/utils/versionablePackages.ts:15:27)
🦋  error     at add (/home/pralky/dev/changesets/packages/cli/src/commands/add/index.ts:47:7)
🦋  error     at run (/home/pralky/dev/changesets/packages/cli/src/run.ts:72:5)
```
Should I temporarily change the `"baseBranch"` in the config to `main` to solve this? Also, do I just do the regular patch bump in this case? Not sure what's the correct approach here since this is technically working towards a new major release.

Andarist:
Hm, it might be worth investigating why this happens because it's not a good experience for the users. I'll try to repro it later using this branch. To recheck a thing - do you have a `next` branch locally? What is its tip, is it [ac917ca](https://github.com/changesets/changesets/commit/ac917caa5537773a1ab2932d6244fd7d5ccf4535)?

ziebam:
> Hm, it might be worth investigating why this happens because it's not a good experience for the users. I'll try to repro it later using this branch. To recheck a thing - do you have a `next` branch locally? What is its tip, is it [ac917ca](https://github.com/changesets/changesets/commit/ac917caa5537773a1ab2932d6244fd7d5ccf4535)?
> 
> > Also, do I just do the regular patch bump in this case? Not sure what's the correct approach here since this is technically working towards a new major release.
> 
> I'd use patch since it's not a significant change and we want it to be listed in the patch section of changes in the final v3 release

I thought I had (I pulled the refs earlier), but I had to actually checkout `next` and then back to my branch to get it to work. Seems like a mistake on my part then.

Andarist:
I'm glad that you have resolved the issue. I think this could be improved in the tool though - likely we don't consult the remote at all in this kind of a situation, the local branch could also be outdated. It might be worth rechecking how Changesets behave in cases like this.

ziebam:
> I'm glad that you have resolved the issue. I think this could be improved in the tool though - likely we don't consult the remote at all in this kind of a situation, the local branch could also be outdated. It might be worth rechecking how Changesets behave in cases like this.

Indeed. Should I file an issue for that or would you prefer to have it in the v3 discussion?

Related: pushed the changeset and changed the named imports to default ones per your and @bluwy's suggestions.

Andarist:
> Should I file an issue for that or would you prefer to have it in the v3 discussion?

Let's keep it in that v3 discussion for now.

bluwy (approved):
Looks great! I think the `fs.access` thing is probably worth revisiting, but we could do that later.

PR #1478, opened 2024-10-05
chore(deps): replace resolve-from package with native Node API

Namchee:
## Overview

This pull request replaces `resolve-from` with native Node implementation of `require.resolve`.

The main difference is only the error message as `resolve-from` provides friendlier message compared to native.

> [!NOTE]
> `resolve-from` itself doesn't require any dependencies, maybe we shouldn't remove it?

### DIFF
--- a/packages/apply-release-plan/package.json
+++ b/packages/apply-release-plan/package.json
@@ -30,7 +30,6 @@
     "lodash.startcase": "^4.4.0",
     "outdent": "^0.5.0",
     "prettier": "^2.7.1",
-    "resolve-from": "^5.0.0",
     "semver": "^7.5.3"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


Apshakil (approved):

bluwy (approved):

bluwy:
Not sure if we need a changeset for this. Maybe good to have though so the changelogs have a list of dependencies that's being refactored out.

Andarist:
I think it might be worth waiting for https://github.com/changesets/changesets/pull/1482 before landing this

bluwy:
The next branch now uses `import-meta-resolve` instead of `resolve-from`. I think it might be worth sticking with `import-meta-resolve` though since we want to also match ESM where we can. Node is taking a long time to stabilize `import.meta.resolve` with parent.

bluwy:
I'll close this for now. Thanks for contributing!

PR #1479, opened 2024-10-05
Set engines field to node 18 and above

bluwy:
Add

```json
  "engines": {
    "node": ">=18.0.0"
  }
```

to all `package.json`


.changeset/deep-coins-attend.md
```diff
@@ -0,0 +1,25 @@
...
+"@changesets/assemble-release-plan": major
+"@changesets/get-dependents-graph": major
+"@changesets/should-skip-package": minor
+"@changesets/apply-release-plan": major
+"@changesets/changelog-github": minor
+"@changesets/get-release-plan": major
+"@changesets/get-github-info": minor
+"@changesets/changelog-git": minor
+"@changesets/release-utils": minor
+"@changesets/test-utils": patch
```
Andarist, line 12: this one is internal so we can skip it
```suggestion
```

.codesandbox/ci.json
Andarist, line 1: we don't quite use this integration anymore so I feel like this whole file can be removed

babel.config.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
     [
       "@babel/preset-env",
       {
-        targets: { node: 8 },
+        targets: { node: 18 },
```
Andarist, line 6: likely we can remove `@babel/preset-env` now but we can't remove this file as a whole as we need to continue transpiling TS files

Andarist:
I think even supporting odd versions shouldn't be a problem here and we kinda declare support for some of them anyway with that loose `>=22.0.0`. So I'd probably just keep it simple and use `>=18.0.0` - I don't feel super strong about it though

bluwy:
I set it as `>=18.0.0` for now and we could probably iterate it again as we go.

bluwy:
It looks like removing `@babel/preset-env` affected jest too, causing this [CI fail](https://github.com/changesets/changesets/actions/runs/11191282017/job/31113991950?pr=1479#step:6:1986), so I reverted it for now

PR #1480, opened 2024-10-05
Run CI on node 18, 20, and 22

bluwy:
This is the same thing as https://github.com/changesets/changesets/pull/1465 but for the `next` branch. Maybe we should do this only for the `next` branch for now.

### DIFF
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -1,14 +1,18 @@
 name: Setup CI
 
+inputs:
+  node-version:
+    description: "Node.js version"
+    required: false
+    default: 20
+
 runs:
   using: composite
   steps:
-    - name: Setup Node.js 20.x
+    - name: Setup Node.js ${{ inputs.node-version }}
       uses: actions/setup-node@v4
       with:
-        node-version: 20.x
+        node-version: ${{ inputs.node-version }}
         cache: yarn
 
     - name: Install dependencies
-      shell: bash
-      run: yarn install --frozen-lockfile
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -16,3 +16,5 @@ runs:
         cache: yarn
 
     - name: Install dependencies
+      shell: bash
+      run: yarn install --frozen-lockfile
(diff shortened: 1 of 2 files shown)


Andarist:
> Maybe we should do this only for the next branch for now.

I'm fine with that 👍

.github/actions/ci-setup/action.yml
```diff
@@ -1,12 +1,18 @@
 name: Setup CI
 
+inputs:
+  node-version:
+    description: "Node.js version"
+    required: false
+    default: 20
```
Andarist, line 7: out of curiosity, why 20 and not 22? ;p

bluwy: The initial action sets 20 so I reused that. And in practice I think 22 is still prone to accidental breaking changes from node itself, so sticking with 20 feels safer.

.github/workflows/ci.yml
```diff
@@ -3,19 +3,27 @@ name: CI
...
+      - main
+      - next
 
 permissions:
   contents: read
 
 jobs:
   test:
-    name: Test
+    name: "Test: node-${{ matrix.node_version }}"
```
Andarist, line 15: q: what would be the auto-generated titles without this change?

bluwy: I don't quite remember, but searching around and landing at https://stackoverflow.com/questions/69105449/github-actions-job-name-based-on-matrix-index-instead-of-matrix-value. I assume it could look like `Test (18)`

Andarist, line 15:
```suggestion
    name: "Test using Node ${{ matrix.node_version }}"
```

.github/workflows/ci.yml
```diff
@@ -24,10 +32,17 @@ jobs:
         run: git config --global user.email "you@example.com" && git config --global user.name "Your Name"
 
       - name: Jest tests
+        if: matrix.node_version != 20
+        run: yarn jest --ci --color --runInBand --reporters=default --reporters=jest-junit
+
+      # Run coverage only once on node 20 to avoid duplicate codecov uploads
+      - name: Jest tests with coverage
+        if: matrix.node_version == 20
         run: yarn jest --ci --color --runInBand --coverage --reporters=default --reporters=jest-junit
```
Andarist, line 41:
```suggestion
        run: yarn jest --ci --color --runInBand --reporters=default --reporters=jest-junit --coverage
```

PR #1482, opened 2024-10-05
Switch all packages to being ESM-only

Andarist:


bluwy:
Looks like we also need to be a good citizen now and not use plain `require`, but `createRequire` or `import` instead 😄 Which is probably a good thing anyways. Interestingly this seems to imply that the pure ESM output we had in v2 never quite worked in node, only in bundlers that could handle bundling `require()`.

Andarist:
> Interestingly this seems to imply that the pure ESM output we had in v2 never quite worked in node

That's fine - node wasn't receiving any ESM files from us in v2 beyond short reexporting `.mjs` wrappers. They were just proxying to CJS files.

Andarist:
@emmatown @bluwy this is ready for a proper review :)

Andarist:
@remorses is there any specific reason why you have downvoted this?

scripts/test-utils/package.json
```diff
@@ -2,27 +2,21 @@
   "name": "@changesets/test-utils",
   "version": "0.0.8",
   "private": true,
+  "type": "module",
   "description": "Utils for testing changesets packages",
   "main": "dist/changesets-test-utils.cjs.js",
   "module": "dist/changesets-test-utils.esm.js",
```
bluwy, line 8: I think we can remove these fields too (and in other packages) as they shouldn't be used now.

packages/release-utils/package.json
```diff
@@ -3,23 +3,17 @@
...
-    },
+    ".": "./dist/changesets-release-utils.js",
     "./package.json": "./package.json"
   },
   "dependencies": {
     "@changesets/pre": "^2.0.2",
     "@changesets/read": "^0.6.3",
     "@changesets/types": "^6.1.0",
     "@manypkg/get-packages": "^1.1.3",
+    "execa": "^9.5.2",
```
bluwy, line 16: Could we reuse spawndamnit here instead of adding execa?

bluwy (commented):
Looks great to me! I didn't expect Jest to be swapped out with Vitest here, but I welcome it 😄

I tested this locally a bit and all seems to be working well. Should we add the major changesets too?

Andarist:
> Looks great to me! I didn't expect Jest to be swapped out with Vitest here, but I welcome it 😄

I almost had it working with Jest but it tried to execute ESM `node_modules` through `script.runInContext` (or similar) and that just didn't work. I probably could make it transform `node_modules` but I gave up and just switched to Vitest

> Should we add the major changesets too?

Yeah. Thanks for the reminder

packages/release-utils/package.json
Andarist: This specifically allows us to use shell-executable scripts like:
1. `a && b`
2. `node -e "const a = 'look, I have whitespace'"`

I'm open to other ideas though but I think that `spawndamnit` won't cut it here.

scripts/test-utils/package.json
Andarist: cc @emmatown, this should likely be automated in Preconstruct as part of `preconstruct fix`. I'll remove those manually for now here

packages/release-utils/package.json
bluwy: I think no2 should also work if you pass `"node"` and `["-e", "const a = 'look, I have whitespace'"]`. Maybe the previous code wasn't handling the last part right, which may had passed as `"\"const a = 'look, I have whitespace'\""`.

Node has an API to support no1 with `child_process.exec` I believe. Maybe we can try that if we want to support this case? Though IIUC we only need no2 currently for the tests, so maybe it's enough to handle the quotes part for now.

Andarist: > I think no2 should also work if you pass "node" and ["-e", "const a = 'look, I have whitespace'"]. Maybe the previous code wasn't handling the last part right, which may had passed as "\"const a = 'look, I have whitespace'\"".

Ye, the problem is that the previous version was splitting a string by whitespace to separate the command and the args and that naturally didn't work with this example.

> Node has an API to support no1 with child_process.exec I believe. Maybe we can try that if we want to support this case?

#brainfart, we definitely should use that

Andarist: I got rid of `execa`. Could you take a look?

bluwy: Looks good!

packages/release-utils/src/utils.ts
```diff
@@ -1,11 +1,16 @@
-import { getPackages, Package } from "@manypkg/get-packages";
-// @ts-ignore
-import mdastToString from "mdast-util-to-string";
+import { getPackages, type Package } from "@manypkg/get-packages";
+import cp, { ChildProcess } from "node:child_process";
```
bluwy, line 2:
```suggestion
import cp, { type ChildProcess } from "node:child_process";
```

This might need a `type` here (IIUC following the rest of the code patterns)

bluwy (approved):
Changes look good to me 👍

packages/release-utils/src/utils.ts
Andarist: Good catch. It's not strictly required by tests right now but we want to maintain this style. With TS 5.8 (soon to be released) we'll be able to configure `erasableSyntaxOnly` and then the CI will validate that the code has correct imports, exports and more

emmatown (approved):

PR #1483, opened 2024-10-05
Avoid weird parsing issue

Andarist:


bluwy (approved):
Doesn't seem to fix it for me, but doesn't hurt to merge this if needed though

bluwy:
I'll close this since we probably don't need it now.

PR #1484, opened 2024-10-05
Setup initial tests for package manager publishing

bluwy:
Basing on https://github.com/changesets/changesets/pull/674 to setup some test infra for package manager publishing. I extracted a part of `internalPublish` where it builds the command and args so we can test it directly.

Still working on this halfway. I'm not sure if this is the testing setup you have in mind, but I notice many of the tests are unit tests with mocks rather than e2e-style tests, so trying to replicate that.

Need to probably mock some commands where it checks for certain pnpm version (or any version like in #674) so it doesn't require users to install the package manager for testing.

### DIFF
--- a/packages/cli/src/commands/publish/__tests__/npm-utils.test.ts
+++ b/packages/cli/src/commands/publish/__tests__/npm-utils.test.ts
@@ -1,0 +1,89 @@
+import { initPackageManager, testdir } from "@changesets/test-utils";
+import { getPublishCommand, PublishOptions } from "../npm-utils";
+import { TwoFactorState } from "../../../utils/types";
+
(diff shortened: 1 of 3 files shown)


Andibaso01:
> Basing on #674 to setup some test infra for package manager publishing. I extracted a part of `internalPublish` where it builds the command and args so we can test it directly.
> 
> Still working on this halfway. I'm not sure if this is the testing setup you have in mind, but I notice many of the tests are unit tests with mocks rather than e2e-style tests, so trying to replicate that.
> 
> Need to probably mock some commands where it checks for certain pnpm version (or any version like in #674) so it doesn't require users to install the package manager for testing.

PR #1485, opened 2024-10-05
Ignore find changed packages error in `changeset add`

bluwy:
Resolves the issue noted in https://github.com/changesets/changesets/discussions/1473#discussioncomment-10851419. If `baseBranch` can't be found locally, you'd no longer get a hard error, but a warning.

Before: 
image

After:
image


Theghost256 (approved):

packages/git/src/index.ts
```diff
@@ -49,7 +49,7 @@ export async function getDivergedCommit(cwd: string, ref: string) {
   const cmd = await spawn("git", ["merge-base", ref, "HEAD"], { cwd });
   if (cmd.code !== 0) {
     throw new Error(
-      `Failed to find where HEAD diverged from ${ref}. Does ${ref} exist?`
+      `Failed to find where HEAD diverged from "${ref}". Does "${ref}" exist?`
```
Andarist, line 52:
```suggestion
      `Failed to find where HEAD diverged from "${ref}". Does "${ref}" exist and it's synced with remote?`
```

packages/cli/src/commands/add/index.ts
```diff
@@ -43,11 +43,19 @@ export default async function add(
...
-      })
-    ).map((pkg) => pkg.packageJson.name);
+    let changedPackagesNames: string[] = [];
+    try {
+      changedPackagesNames = (
+        await getVersionableChangedPackages(config, {
+          cwd,
+        })
+      ).map((pkg) => pkg.packageJson.name);
+    } catch (e: any) {
```
Andarist, line 53: I think we should strive to handle specific errors here. We can throw custom errors in `@changesets/git` and handle them here

bluwy: Do you mean we should only handle the specific error from `getDivergedCommit`, and otherwise we re-throw the other errors?

Andarist: Hm, I see your point. It feels like it would be good to at least avoid displaying the error stack for those errors that we know of.

I'm fine with this change as-is though - it's up to you if you think it's worth changing anything here. I'd add a code comment about the catch-all rationale here though

Andarist (approved):

PR #1493, opened 2024-10-10
Update changelog badges

bluwy:
close https://github.com/changesets/changesets/pull/1309

changelogs.xyz doesn't seem to properly show the changelog anymore, so I updated the badge and link it to the `CHANGELOG.md` file directly.

I also added an npm badge to quickly go to the npm page. Let me know if you don't want that, I'm fine either ways 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 </p>
 <br/>
 
-[![npm package](https://img.shields.io/npm/v/@changesets/cli.svg)](https://npmjs.com/package/@changesets/cli)
+[![npm package](https://img.shields.io/npm/v/@changesets/cli?label=%40changesets%2Fcli)](https://npmjs.com/package/@changesets/cli)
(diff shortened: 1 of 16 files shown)


Andarist:
@emmatown is that service still maintained?

emmatown:
I don't see it going down but also probably won't be spending time making it better. I believe it's not working here because the changelog isn't published to npm

I think the change here makes sense anyway tbh. though the links in the package readmes should probably point to GitHub so they work when viewing the package on the npm website

Andarist:
Huh, I was under impression the CHANGELOGs would always be published regardless of `files` and all. I even found mentions of that in an old npm blog post but the current docs don't mention this: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files

So we definitely should fix all `package.json#files` to include our CHANGELOGs. It would be cool to validate this in manypkg but oh well ;p

bluwy:
It happened since a few years ago I believe. Personally I don't like publishing changelog since you can read it online anyways, and if offline mattered we wouldn't have noticed it only now 😅

PR #1494, opened 2024-10-10
Remove changeset v1 support

bluwy:
This removes support for changeset in v1 format in the `.changeset` directory.

The previous warning mentions that it'll error instead of completely not supporting it (like in this PR), I don't know if we _need_ to keep that behaviour though. It already feels like a long time ago.

At https://github.com/changesets/changesets/discussions/1473, there's a mention about this change that could affect https://github.com/changesets/bot, however I can't tell how this affects that repo when reading the source.


Andarist:
> At https://github.com/changesets/changesets/discussions/1473, there's a mention about this change that could affect https://github.com/changesets/bot, however I can't tell how this affects that repo when reading the source.

I thought it could affect it in a sense that it would no longer be able to read those old changeset files. This is an integration and the user can't chose to use a bot compatible with old changesets as that bot is updated out of their control.

Now, when looking through its code... I'm not sure if it handles that at all today:
https://github.com/changesets/bot/blob/a443d065be109aeb890cb97c14e5b34e3c4f4a88/get-changed-packages.ts#L96-L112

It might be able to read them (nested `.md` files should still be read by this logic) but the retrieved ID might be incorrect - I'm not sure what would happen then but there is a risk that it wouldn't quite work correctly.

But either way, that bot doesn't use `@changesets/read` - so this change here should be completely OK from its PoV.

Andarist (approved):

PR #1495, opened 2024-10-10
add `format` script

### LINKED ISSUE #1492: Maintenance: prettier is running twice
## Affected Packages

N/A, the problem affects only repository maintenance 

## Problem

Right now, during `lint` task, `prettier` is running twice:

https://github.com/changesets/changesets/blob/9956976b7cfee85cc79488476dc5455e76370caa/package.json#L11

1. by `lint:eslint` task, which run `eslint` configured with `prettier` plugin and the associated rule:
  https://github.com/changesets/changesets/blob/9956976b7cfee85cc79488476dc5455e76370caa/.eslintrc#L2-L6
  https://github.com/changesets/changesets/blob/9956976b7cfee85cc79488476dc5455e76370caa/.eslintrc#L10
3. as part of `lint:prettier` script, which runes directly `prettier`
  https://github.com/changesets/changesets/blob/9956976b7cfee85cc79488476dc5455e76370caa/package.json#L13

> [!NOTE]
> Removing `prettier` from eslint brough down `lint` task from 9s to 6s on my machine.
> The result has been obtained by turning off `prettier/prettier` rule from `.eslintrc`

### 1. Separate format task from lint task

1. Remove `eslint-plugin-prettier` plugin (keep `eslint-config-prettier` to disable formatting rules)
4. rename `lint:prettier` script to `format`
5. rename `prettier` script to `format:fix`
6. run `format` script alongside `lint` script or in a separate job in CI

### 2. Remove redundant task

As an alternative I can simply remove `lint:prettier` script and let `format` to be part of `lint` script through `eslint`

---

> [!NOTE]
> If you are interested I can quickly setup a PR with the solution you prefer.

## Follow up

As part of the [v3 tasks ](https://github.com/changesets/changesets/discussions/1473) I see that you plan to update project dependencies. 
There is nothing specific about dev dependencies, but, if you wish I can try to update  `eslint` to `v9`, 
in addition I would also consider to enable some type checking rules (after upgrading to v9). 

Let me know if you are interested in discussing this further (in a separated issue or whatever place you prefer 😆).

marcalexiei:
Close #1492

- Remove `eslint-plugin-prettier`,  `eslint-config-prettier` is retained in the config to disable formatting rules
- Rename `lint:prettier` script to `format`
- Rename `prettier` script to `format:fix`
- CI `lint` task renamed to `lint_and_format`, I think that there is no need to create a new job for just run `format` script
- In `eslintcrc` I removed the `packages/cli/default-files/config.js` override since the folder no longer contains any js file

---

I also noticed that `eslint-plugin-jest` is not used, so I removed it from the dependencies.

Since the project is using `jest` I would consider to enable this plugin.
For instance, it includes a [max-nested-describe rule](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/max-nested-describe.md) that could resolve the describe "issue" reported in this [review comment](https://github.com/changesets/changesets/pull/1466#discussion_r1793679057).

If you agree,  I'm willing to work on a PR to achieve this,
Ideally I'll start working on it after  #1467 is merged (it already change some test related code).


.github/workflows/ci.yml
```diff
@@ -53,11 +53,14 @@ jobs:
...
 
+      - name: Format
+        run: yarn format
+
   ci-ok:
     name: CI OK
     runs-on: ubuntu-latest
     if: always()
-    needs: [test, typecheck, lint]
+    needs: [test, typecheck, lint_and_format]
```
Andarist, line 63:
```suggestion
    needs: [test, typecheck, lint]
```

.github/workflows/ci.yml
```diff
@@ -42,8 +42,8 @@ jobs:
       - name: Typecheck
         run: yarn types:check
 
-  lint:
-    name: Lint
+  lint_and_format:
```
Andarist, line 45:
```suggestion
  lint:
```

.github/workflows/ci.yml
```diff
@@ -42,8 +42,8 @@ jobs:
...
+    name: Lint & Format
```
Andarist, line 46: I consider formatting here just as a kind of linting so i think that "lint" is still OK here and if we ever add a new lint type I'd rather like to avoid renaming this to ``lint_and_format_and_xxx` ;p
```suggestion
    name: Lint
```

package.json
```diff
@@ -8,14 +8,13 @@
...
-    "lint:prettier": "prettier --list-different \"**/*.{js,ts,tsx,md}\"",
+    "lint": "yarn eslint . --ext .ts,.tsx,.js",
     "types:check": "tsc",
-    "prettier": "prettier --write \"**/*.{js,ts,tsx,md}\"",
+    "format": "prettier --list-different \"**/*.{js,ts,tsx,md}\"",
+    "format:fix": "prettier --write \"**/*.{js,ts,tsx,md}\"",
     "changeset": "packages/cli/bin.js",
-    "check-all": "yarn test && yarn lint && yarn types:check",
-    "version-packages": "changeset version && yarn prettier",
+    "check-all": "yarn test && yarn format && yarn lint && yarn types:check",
```
Andarist, line 16:
```suggestion
    "check-all": "yarn test && yarn types:check && yarn lint && yarn format",
```

Andarist:
Does format on save still work in VS Code and uses Prettier config to do its job?

marcalexiei:
> Does format on save still work in VS Code and uses Prettier config to do its job?

Yes, of course! You just need to have [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) extension installed and select it as formatter

-----


Also, let me know what do you think about [`eslint-plugin-jest` usage proposal](https://github.com/changesets/changesets/pull/1495#issue-2579361021) 🙂

Andarist:
> Also, let me know what do you think about https://github.com/changesets/changesets/pull/1495#issue-2579361021 🙂

Yeah, this could be nice

PR #1496, opened 2024-10-11
Lint: add `eslint-plugin-jest`

marcalexiei:
Followup to https://github.com/changesets/changesets/pull/1495#issuecomment-2405804284

I added the last version of `eslint-plugin-jest` which requires at least `eslint` to `8.57.0` so I updated it.

I extended `jest/recommended` adding two rules to it:
- `jest/no-standalone-expect`  which requires to be set in order to work with `jestWithCase`.
- `jest/prefer-expect-resolves` which IMHO improves async expect readability

---

A complete list of rules can be found here:
https://github.com/jest-community/eslint-plugin-jest?tab=readme-ov-file#rules

Let me know if you want to enable additional rules.

> [!NOTE]
> I haven't enabled `max-nested-describe` rule because it involves a lot of changes,
> especially in the following files:
> - `packages/apply-release-plan/src/index.test.ts`
> - `packages/cli/src/commands/version/version.test.ts`
> - `packages/assemble-release-plan/src/index.test.ts`
>
> I'll create a PR separate to include changes related to this rule to simplify the review process


.eslintrc
```diff
@@ -39,8 +39,14 @@
...
-      "env": {
-        "jest": true
+      "extends": ["plugin:jest/recommended"],
+      "plugins": ["jest"],
+      "rules": {
+        "jest/no-standalone-expect": [
+          "error",
+          { "additionalTestBlockFunctions": ["jestInCase"] }
+        ],
+        "jest/prefer-expect-resolves": "error"
```
marcalexiei, line 49: Enabled also `jest/prefer-expect-resolves` (PR description updates)

PR #1501, opened 2024-10-19
Lint: enable `jest/max-nested-describe` rule

marcalexiei:
This PR is a based on top of #1496 changes,
it aims to enable `jest/max-nested-describe` rule as per review of https://github.com/changesets/changesets/pull/1466#discussion_r1793679057.

> [!NOTE]
> I'm creating a separate PR to simplify review process since there a lot of code changes (mostly indentation 😅).

> [!CAUTION]
> This is marked as draft because should be reviewed  after #1496  is merged.

> [!TIP]
> If there are too many changes I can disable the rule in  biggest file (like `packages/cli/src/commands/version/version.test.ts`) and restore their original content.


packages/apply-release-plan/src/index.test.ts
marcalexiei, line 1: This file contains a large amount of tests so I created multiple `describe` blocks in the file root.

packages/cli/src/commands/version/version.test.ts
marcalexiei, line 1: I addition to edit tests title I used `#region` and `#endregion` vscode markers to keep related test together

Issue #1137, opened 2023-04-17
Changesets for the monorepo root workspace


sarneeh:
## Problem

Currently, I can only handle changes and maintain changelogs for packages/apps that are under a monorepo. The problem is, that there are plenty of things that are mostly related to the monorepo itself - like CI workflows, tooling etc. and it would be great to be able to maintain a changelog for it as well.

## Proposed solution

Perhaps something like `changesets add --root`, or a configuration entry that would make `changesets` also check for changes that are not related to the apps/packages in the monorepo.

jmatsushita:
I am also missing this feature. Another approach would be to list the mono repo when running `changeset add` if the root repo's `package.json` has a `name` field.

wmertens:
I added `- '.'` to my pnpm-workspaces.yaml and made the root have a name + private:true and it versions it as expected. (I added the monorepo name as a fixed package with the main package)

ghost91-:
@wmertens I just tried that solution, and it works fine for changesets, but it seems to break other tools. For example, turborepo doesn't work this way anymore. Imo, a proper solution inside changesets to include the workspace root is needed.

GauBen:
Somewhat related: without Wout's fix, changeset will crash with this error

```
🦋  error TypeError: Cannot read properties of undefined (reading 'packageJson')
🦋  error     at createChangeset (.../dist/changesets-cli.cjs.js:417:75)
🦋  error     at add (.../dist/changesets-cli.cjs.js:514:26)
🦋  error     at async run (.../dist/changesets-cli.cjs.js:1347:5)
```

It happens when all packages but the workspace root are private

PR #1506, opened 2024-10-25
fix: prevent crash when trying to version the root of the monorepo

GauBen:
Hey! Trying to address the bug I encountered in a monorepo where only the root package is published to npm

Before:

```
🦋  error TypeError: Cannot read properties of undefined (reading 'packageJson')
🦋  error     at createChangeset (.../dist/changesets-cli.cjs.js:417:75)
🦋  error     at add (.../dist/changesets-cli.cjs.js:514:26)
🦋  error     at async run (.../dist/changesets-cli.cjs.js:1347:5)
```

After:

```
🦋  warn This workspace does not contain any packages that can be versioned
🦋  warn If you intend to version the root project, add "." to the list of workspaces
🦋  warn See https://github.com/changesets/changesets/issues/1137 for more details
```

Related: #1137

### DIFF
--- a/packages/cli/src/commands/add/index.ts
+++ b/packages/cli/src/commands/add/index.ts
@@ -53,6 +53,11 @@ export default async function add(
       releases: [],
       summary: ``,
     };
+  } else if (versionablePackages.length === 0) {
+    warn("This workspace does not contain any packages that can be versioned");
+    warn(`If you intend to version the root package, add ${pc.green('"."')} to the list of workspaces`);
+    warn(`See ${pc.blue('https://github.com/changesets/changesets/issues/1137')} for more details`);
+    newChangeset = { confirmed: false };
   } else {
     const changedPackagesNames = (
       await getVersionableChangedPackages(config, {


packages/cli/src/commands/add/index.ts
```diff
@@ -53,6 +53,11 @@ export default async function add(
       releases: [],
       summary: ``,
     };
+  } else if (versionablePackages.length === 0) {
+    warn("This workspace does not contain any packages that can be versioned");
+    warn(`If you intend to version the root package, add ${pc.green('"."')} to the list of workspaces`);
+    warn(`See ${pc.blue('https://github.com/changesets/changesets/issues/1137')} for more details`);
+    newChangeset = { confirmed: false };
```
GauBen, line 60:
```suggestion
    newChangeset = {
      confirmed: false,
      releases: [],
      summary: ``,
    };
```

KaminskiDavid:
Hey @GauBen,

I think I am encountering the same error. I have no monorepo but a single package repo. Now I want to introduce pre-release versions, so I `run npm run changeset pre enter PRERELEASE-XYZ`. That works fine. But when I `run npm run changeset version`, I always get the error: `TypeError: Cannot destructure property 'packageJson' of 'undefined' as it is undefined.`

Basically, is this use case not supported right now and this PR would "fix" it?

GauBen:
Hey @KaminskiDavid! I'm not very familiar with changesets and unfortunately this PR does not fix anything, it just replaces a crash with a warning about a missing feature.

From my short experience, however, I would recommend you to turn your project into a monorepo by adding `"workspaces": ["."]` to your package.json (npm, yarn) or `packages: ['.']` to your pnpm-workspace.yaml file (pnpm)

If that still does not work, you should open an issue and hope for the best 😬

GauBen:
I created https://github.com/GauBen/chachalog, it solves this issue 👀

PR #1517, opened 2024-11-20
relax parameters for read()

akphi:
Hi,

@bluwy We have found that the low-level utilities like `read()` from `@changesets/config` are super useful building blocks for customizations we need to build [on top of the default changeset behavior](https://github.com/finos/legend-studio/blob/master/packages/legend-dev-utils/ChangesetUtils.js).

One small problem we have is that these methods do require `Packages` information that is extracted using `@manypkg/get-packages` explicitly in their signature, so to maintain our custom logic, we have to add a dependency on `@manypkg/get-packages`. We would like to remove this dependency, thinking that it comes with `changesets` anyway. Are you ok with us making the backward-compatible change like in this PR?

> A very other small reason is that we would prefer to be able to rely on the latest possible version of `@manypkg/get-packages` in case we need to keep this dependency around, but we don't want to wait for https://github.com/changesets/changesets/pull/1069 . Ideally, we like to bump all of our dependencies to the latest and we're stuck with depending on a fairly old version of `@manypkg/get-packages` so we would like to nolonger have to care about this dependency. 

Please let us know what you think! Thank you for maintaining this helpful library 🙏


Andarist:
This isn't backwards compatible. Those functions have to be synchronous. You could consider adding alternative ones that would just accept `cwd`

akphi:
@Andarist my bad, I should only really make the change for `read()`

packages/config/src/index.ts
```diff
@@ -91,7 +91,8 @@ function isArray<T>(
   return Array.isArray(arg);
 }
 
-export let read = async (cwd: string, packages: Packages) => {
+export let read = async (cwd: string, _packages?: Packages | undefined) => {
+  const packages = _packages ?? (await getPackages(cwd));
```
Andarist, line 95:
```suggestion
export let read = async (cwd: string, packages?: Packages) => {
  packages ??= await getPackages(cwd);
```

akphi: Accepted, but I did `?: Packages | undefined` because some project might set `exactOptionalPropertyTypes=true` for Typescript.

akphi:
@Andarist @bluwy is there any other changes you would like me to make?

romankulkovsf:
And my apple new ID ***@***.***

Andarist:
@akphi it would be great to have tests for this functionality

akphi:
> @akphi it would be great to have tests for this functionality

@Andarist Done!

packages/config/src/index.test.ts
```diff
@@ -65,6 +65,45 @@ test("read reads the config", async () => {
   });
 });
 
+test("read reads the config with package infos collected from current context", async () => {
```
Andarist, line 68:
```suggestion
test("read can read config based on the passed in `cwd`", async () => {
```

akphi:
@Andarist do we have plan to cut a release any time soon?

Andarist:
done

PR #1535, opened 2024-12-05
Respect `bumpVersionsWithWorkspaceProtocolOnly` value from config

joerobot:
### The issue

When `getDependentsGraph` is called during `config.read`, the config value for `bumpVersionsWithWorkspaceProtocolOnly` isn't passed along, creating unnecessary console errors.

### More detail

`getDependentsGraph` has a default argument that sets `bumpVersionsWithWorkspaceProtocolOnly` to `false`. During `changesets version` this config value is respected, but as `getDependentsGraph` is used here without the second `opts` argument, the "Package `a` must depend on the current version of `b`" ([`get-dependency-graph`](https://github.com/changesets/changesets/blob/main/packages/get-dependents-graph/src/get-dependency-graph.ts#L116-L122)) error will fire while the config is being read as part of [any CLI command](https://github.com/changesets/changesets/blob/main/packages/cli/src/run.ts#L44).

As far as I can tell, there are no other negative affects here, just noise created by the error.

### The solution

Pass the config value of `bumpVersionsWithWorkspaceProtocolOnly` to `getDependentsGraph` during `config.parse`.


packages/config/src/index.test.ts
```diff
@@ -138,6 +138,38 @@ it("should not fail when validating ignored packages when some package depends o
   expect(() => read(cwd, packages)).not.toThrow();
 });
 
+it("should pass bumpVersionsWithWorkspaceProtocolOnly option to getDependentsGraph during validation", async () => {
```
Andarist: this test passes even without your changes... so it doesn't actually test this change

.changeset/itchy-comics-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
+Reads the config value for `bumpVersionsWithWorkspaceProtocolOnly` during config parse to prevent noisy console errors.
```
Andarist:
```suggestion
Respect config value for `bumpVersionsWithWorkspaceProtocolOnly` when validating `ignore` option to prevent spurious console errors.
```

Issue #654, opened 2021-10-14
Allow passing a custom glob to find packages


adambrgmn:
## Affected Packages

- `@changesets/cli`

## Problem

I'm trying to advocate at work to use `changesets` as a tool to automate the release of packages that we have. Today the release process is very manual and cumbersome. But I have an issue – we work in a monorepo but we don't follow the setup of either `yarn`, `lerna`, `pnpm` or `bolt`. Instead we use [`nx`](https://nx.dev). It is a rather special setup, I realize that. But a tweak to the `changesets` cli would enable us to actually use it.

The thing with `nx` is that we have a single `package.json` in the root with all dependencies. Then we have placeholder `package.json`'s in the libraries that we want to publish, with only metadata such as name and version. And when packaging the packages into publishable versions they end up in a `dist`-folder, with a `package.json` and everything.

That means we would like to run `add` and `version` agains one set of packages. And then `publish` against what's in `dist`.

This doesn't work with `changesets` for two reasons:

- `@manypkg/get-packages` can't determine that we use an nx repo setup and aren't able to find our packages
- The packages we want to publish end up in a different directory than the source code; `libs/package-a` -> `dist/libs/package-a`

## Proposed solution

First I thought of making a PR against `@manypkg/get-packages`. But I realize that nx-repos function very differently than others which means it is not suitable for that project.

I have a working solution locally. The proposed solution is to be able to pass a `--packages=<glob>` flag to the `add`, `version`, `publish` and `status` cli commands.

Then the workflow for an nx monorepo would be something like this:

1. Changes done in package-a
2. Create changesets for package
   ```sh
   yarn changeset --packages="libs/*"
   ```
3. Create and merge PR into main
4. Changeset github action kicks in with this config:
   ```yml
   uses: changesets/action@master
   with:
     version: |
       yarn changeset version --packages="libs/*"
     publish: |
       yarn build
       yarn changeset publish --packages="dist/libs/*"
   ```
  - `yarn changeset version` will check for the source libs so that the changelog and everything ends up where it should
  - `yarn build` packages the module into the dist folder, with distributable files and a package.json and everything
  - then `yarn changeset publish` will publish the packages in the dist folder

As I said I have a rough solution locally and I would be happy to polish it and supply a PR if this is a feature that you are interested in.

adambrgmn:
[...]

PR #1557, opened 2025-01-12
[PR-RFC]Add ability to specify workspaces through config

### LINKED ISSUE #1144: Support for non-workspace packages
## Problem

I have a package in my monorepo which is not part of `npm workspace`. The published package in npm is very different from the local version as i only publish the `dist` directory and change the folder structure. So it is not possible to use the local symlink in the workspace. Thus i moved it out from the `packages` directory. 

```
.
├── package.json
├── packages
│   ├── pkg-1
│   └── pkg-2
└── libs
    └── pkg-3
```
Now the `npx changeset` command doest show that package for version update. Is there way to include that package in `changesets` without adding it to the npm workspace ?
Right now i am manually copying the package path in root package.json workspaces before changesets command and remove it after releasing the update.

lachieh:
I was going to do this as an issue first, but after doing some experimentation, I ended up with a solution that works for my initial use case so I figured it can be discussed as a PR instead.

This PR introduces a new `@changesets/get-packages` package that extends the functionality of `@manypkg/get-packages`. The key addition is support for defining additional workspaces through the changesets config file, which is useful in setups where certain packages (like examples or nested workspaces) can't be included in the root `package.json`. See the example below for more info.

### Proposal
The core proposal is to supplement workspace discovery by introducing a new configuration option `additionalWorkspaces` under the experimental flag/key. This allows monorepos to eep their existing workspace configurations in `package.json` and add supplementary workspace paths through changesets config

### Example

For my exact use case, I have a monorepo with a root `package.json` that defines a `packages` directory as a workspace. I also have an `examples` folder that contains additional packages that are not included in the root `package.json` (and can't be because they need to be installed completely isolated to the package root). I want to be able to version these packages with changesets, so I add the following to my changesets config file:

`.changeset/.config.json`
```json
{
    "additionalWorkspaces": ["examples/*"]
  }
}
```

This allows me to version the packages in the `examples` directory alongside the packages in the `packages` directory without the package manager trying to mess with their installation.

### Solution
The implementation wraps `@manypkg/get-packages` with additional functionality that:

1. Reads the changesets config file to discover additional workspace paths
2. Creates a temporary `package.json` and searches for matching packages
3. Merges the results from both standard and additional workspace scans based on the packages that are discovered

### Related Issues

- Closes #1144
- Related to #654 with some extra changes


lachieh:
@Andarist got any thoughts about this?

packages/get-packages/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# @changesets/changelog-git
```
Andarist, line 1:
```suggestion
# @changesets/get-packages
```

packages/get-packages/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export let getPackages = async (cwd: string): Promise<Packages> => {
+  const rootPackagesPromise = manyPkgGetPackages(cwd);
+
+  const workspaces = await loadAdditionalWorkspaces(cwd);
+  if (workspaces.length === 0) return rootPackagesPromise;
+
+  const { tmpDir, clean } = await makeTempPackageJson(cwd, workspaces);
+  const additional = await manyPkgGetPackages(tmpDir);
+  clean();
```
Andarist, line 82: This seems to be rather complicated for what it's doing. The current version of `@manypkg/get-packages` that we are using here returns a very simple javascript object that could be replicated manually here. We don't need to hack around this way just so we can call `manyPkgGetPackages`

That said... I'd like us to migrate to the new version of `@manypkg` and that doesn't return a POJO (or at least it doesn't return a POJO for everything, maybe a `Package` is still a POJO, I don't recall right now). I'm also considering if we shouldn't rethink `@manypkg` APIs though so perhaps that shouldn't even be a consideration for this pull request here.

packages/get-packages/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  Tool,
+} from "@manypkg/get-packages";
+import { mkdtemp, readJSON, remove, writeFile, realpath } from "fs-extra";
+import { join } from "path";
+import { tmpdir } from "os";
+import { WrittenConfig } from "@changesets/types";
+
+const UNSAFE_KEY = "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH";
+
+let loadAdditionalWorkspaces = async (cwd: string): Promise<string[]> => {
```
Andarist, line 14: I think `additionalWorkspaces` option introduces some unnecessary complexity. It would be much simpler if we'd just support a `workspaces` override in the config.

packages/get-packages/src/index.ts
lachieh: I found that while investigating this solution. Figured it would be a separate PR to update the manypkg version to the latest major. I actually got most of the way through the upgrade but decided it was too many changes unrelated to this specific feature.

I can submit that as a single PR if that's useful. Want me to give it a go?

packages/get-packages/src/index.ts
lachieh: That drastically changes this for sure. I'm happy to do that. My main concern here was that then I'd have to be certain that I'm keeping both the changesets config and the package.json workspaces in sync. Are there considerations to take there or is it fine to just have it be a hard override?

packages/get-packages/CHANGELOG.md
lachieh: If we go the workspaces override route, I'll likely just remove this package. Will adjust the name if it sticks around!

lachieh:
This is so far behind that it'd need to be redone and I'm not using it in the same way anymore so I'm going to close. Thanks!

PR #1560, opened 2025-01-16
Publish using yarn npm publish with Yarn PnP

chaance:
Yarn versions 2+ suggest using `yarn npm publish` to ensure packages are packed correctly before publishing. This PR addresses this by detecting Yarn version 2+ before setting the publish tool.

### DIFF
--- a/packages/cli/src/commands/publish/npm-utils.ts
+++ b/packages/cli/src/commands/publish/npm-utils.ts
@@ -10,6 +10,7 @@ import { askQuestion } from "../../utils/cli-utilities";
 import { isCI } from "ci-info";
 import { TwoFactorState } from "../../utils/types";
 import { getLastJsonObjectFromString } from "../../utils/getLastJsonObjectFromString";
+import type { SemVer } from "semver";
 
 interface PublishOptions {
   cwd: string;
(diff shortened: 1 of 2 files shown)


chaance:
Just noticed https://github.com/changesets/changesets/pull/674 which has been open for a few years, so not sure if there are any plans to finalize that. Feels like this could work as an interim fix?

efoken:
I used exactly this patch already in many monorepos, it works perfect. So it should really be merged ASAP

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -173,28 +199,37 @@ async function internalPublish(
...
-        );
+  let results: Awaited<ReturnType<typeof spawn>>;
+  if (publishTool.name === "pnpm") {
+    results = await spawn("pnpm", ["publish", "--json", ...publishFlags], {
+      env: Object.assign({}, process.env, envOverride),
+      cwd: opts.cwd,
+    });
+  } else if (publishTool.name === "yarn") {
+    results = await spawn("yarn", ["npm", "publish", ...publishFlags], {
+      env: Object.assign({}, process.env, envOverride),
```
marbemac, line 223: Isn't `cwd: opts.cwd,` needed here (similar to pnpm above)?

bluwy:
Fixed in #2097

PR #1577, opened 2025-02-06
Added a configuration option to specify which prerelease number to start with

LukasHechenberger:
In a company I work for people insist to start prereleases (`RC`s) with the number *1* instead of *0*.

In order to achieve this I implemented a new configuration option `pre.startWith` that defaults to *0* as to not break current behavior.

Let me know what you think about it!


LukasHechenberger:
Could someone please review this?

LukasHechenberger:
It's been three weeks... can someone please review this PR? 

@Andarist @bennypowers  (Sorry for spamming, idk who is in charge here)

LukasHechenberger:
@bluwy

docs/config-file-options.md
```diff
@@ -194,3 +194,15 @@ You can use the following placeholders for customizing the snapshot release vers
...
+
+Default value: `undefined`
+
+### `startWith` (optional number)
+
+Default value: `0`
+
+Use this number as a prerelease number when a package is first versioned as a prerelease.
+
+For example, a package with version `1.0.0` will be versioned as `1.0.0-RC.1` if `startWith` is set to `1` (and the tag is `RC`).
```
Andarist, line 208: This option being a number doesn't sit well with me. I don't see anybody using with anything else than 0 or 1. 

To be honest, I don't even see a big need for this option at all. Starting with `0` is a pretty common practice and how npm itself works (that can be verified by running, e.g., `npm version prepatch`). 

cc @bluwy, do u have any particular opinion about this?

LukasHechenberger: I personallly don't care either, but the team I'm currently setting up changesets for is used to name their first release candidate "RC 1". And I think that's pretty common, even [semver](https://semver.org) start with "-rc.1".

Of course, I could also run `changeset pre enter RC`, `changeset version`, remove the changesets from *pre.json*, reset the changelogs and run `changeset version` again, but that's pretty hacky in my opinion. Also, as it's fully optional, I don't see how it may hurt.

Andarist: > Of course, I could also run changeset pre enter RC, changeset version, remove the changesets from pre.json, reset the changelogs and run changeset version again, but that's pretty hacky in my opinion.

I wonder if some kind of a middleware wouldn't just serve this kind of customizability better.

> Also, as it's fully optional, I don't see how it may hurt.

Even optional API surface has some cost associated with it. Nothing comes free, I just don't want to expose toggles for every little behavior because it wouldn't be manageable to maintain it.

bluwy: I think starting with `.0` is also normal and I agree there's a cost in adding more options to this. The [regex given by semver.org](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) also supports `-pre.0`.

Personally I would lean on setting up a custom script to fix this if starting from `.1` is preferred, which could be used to update the versions in `package.json` and changelogs after `changeset version` is called.

LukasHechenberger: > The [regex given by semver.org](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) also supports -pre.0.

I actually meant that semver started releasing v2 with *2.0.0-rc.1*...

> Even optional API surface has some cost associated with it. Nothing comes free, I just don't want to expose toggles for every little behavior because it wouldn't be manageable to maintain it.

Alright, I'll check back with my team to decide how to proceed.

Thanks for the review!

Andarist: > Of course, I could also run changeset pre enter RC, changeset version, remove the changesets from pre.json, reset the changelogs and run changeset version again, but that's pretty hacky in my opinion. Also, as it's fully optional, I don't see how it may hurt.

I recognize how inconvenient this is. Maybe a different option would be to call `@changesets/assemble-release-plan` first, gather packages bumped to `.0` and just apply those `package.json#version` updates? My hunch is the next `changeset version` should bump them from `.0` to `.1`.

PR #1578, opened 2025-02-06
Remove `--snapshot` from `publish` command docs

ematipico:
I think there was a bug in the docs, or the CLI is incorrect: 

```bash
  Usage
    $ changeset [command]
  Commands
    init
    add [--empty] [--open]
    version [--ignore] [--snapshot <?name>] [--snapshot-prerelease-template <template>]
    publish [--tag <name>] [--otp <code>] [--no-git-tag]
    status [--since <branch>] [--verbose] [--output JSON_FILE.json]
    pre <enter|exit> <tag>
    tag
```


bluwy (approved):
Seems correct to me 👍 `snapshot` is only used and passed to the version command

https://github.com/changesets/changesets/blob/5c2bac5fd2c1c2904774c4378e9107dcf8313c7d/packages/cli/src/run.ts#L190-L194

PR #1579, opened 2025-02-10
fix(assemble-release-plan): errors caused by non-standard packages#835

aui:
If some internal packages in the workspace lack version numbers, it will cause changesets to crash.

I strictly check the input and throw error messages to let users know where they can improve their code to avoid confusion.

BUG：https://github.com/changesets/changesets/issues/835


bluwy:
The linked issue was fixed by #1589, so I'll assume that we don't need this fix now. I'll close the PR, but thanks for contributing!

PR #1585, opened 2025-02-12
Add no commit flag to CLI

milofultz:
Often, we have staged changes and then make a changeset. This rolls up the staged changes with the changeset and is quite annoying.

Adding this `--no-commit`/`-n` flag will allow a user to make a changeset and not commit the newly created changeset.


milofultz:
Not sure who would be the best for review, but I see that @bennypowers and @Andarist were recent contributors. Sorry if this is the wrong ping 🙇

Issue #1593, opened 2025-02-20
Changeses config prettier can't accept boolean value


tk-o:
Couple days ago, [Changesets Config schema included `prettier` setting](https://github.com/changesets/changesets/blame/1621d9d7e69ae0e26ced773bc1e6988a6cacb73a/packages/config/src/index.ts#L334):
```json
"prettier": {
  "type": "boolean",
  "description": "When false, Changesets won't format with Prettier",
  "default": true
},
```

However, validation process for this new `prettier` setting does not allow passing anything but `undefined`:
https://github.com/changesets/changesets/blob/84a4a1b1d399bfd0a58677b0182b9c053194febf/packages/config/src/index.ts#L334-L342

PR #1590, opened 2025-02-18
fix(config): validate prettier option

bennypowers:
fixes https://github.com/changesets/changesets/issues/1593

PR #1594, opened 2025-02-20
fix(config): allow booleans as `prettier` values

tk-o:
This PR updates validation during config file parsing to allow passing `boolean` type values.

Fixes
- #1593


tk-o:
Now I noticed another similar bugfix:
- #1590 

Feel free to pick either of them, and let's have the prettier off when it needs to be off 🚀

tk-o:
@Andarist any chance we could have this hotfix released soon? Cheers 🙃

PR #1608, opened 2025-03-12
Use package scope registry during npm info execution

onlywei:
Use the registry from `npm config` when publishing a package within a scope.


bluwy:
Fixed in #2064

PR #1609, opened 2025-03-14
Update @manypkg/get-packages to ^2.2.2

Pewtro:


Andarist:
This might not affect CLI consumers but for other packages this is a breaking change.

Pewtro:
Sure, bumped the changeset and added a link to v2.0.0 that brought along the breaking changes in the description

PR #1613, opened 2025-03-19
Remove node-fetch dependency

bluwy:
`fetch()` is builtin in node 18 now with the updated engine ranges

### DIFF
--- a/packages/get-github-info/src/index.ts
+++ b/packages/get-github-info/src/index.ts
@@ -1,5 +1,4 @@
 // @ts-ignore
-import fetch from "node-fetch";
 import DataLoader from "dataloader";
 
 const validRepoNameRegex = /^[\w.-]+\/[\w.-]+$/;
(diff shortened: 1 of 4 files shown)


packages/get-github-info/src/index.ts
```diff
@@ -1,5 +1,4 @@
 // @ts-ignore
```
Andarist, line 1:
```suggestion
```

packages/get-github-info/src/index.ts
```diff
@@ -1,5 +1,3 @@
-// @ts-ignore
-import fetch from "node-fetch";
```
Andarist, line 2: I hope we won't run into any experimental warnings when using builtin `fetch`

bluwy: I believe the warnings are only before node 18 which we no longer support in next. The [flag](https://nodejs.org/docs/v22.14.0/api/globals.html#fetch) was removed in node 18.

Andarist: I'm pretty sure I ran into experimental warnings for some features of "stable" `fetch` on node 18. It was really annoying as, in the end, it prevented me from using `fetch` - but we probably don't need anything non-basic here so it should be fine.

PR #1616, opened 2025-03-19
Remove term-size dependency

bluwy:
The package is [deprecated](https://www.npmjs.com/package/term-size) but we can also replace with a simple `process.stdout.rows` check


packages/cli/src/utils/cli-utilities.ts
```diff
@@ -1,5 +1,4 @@
 // @ts-ignore it's not worth writing a TS declaration file in this repo for a tiny module we use once like this
```
Andarist, line 1:
```suggestion
```

PR #1619, opened 2025-03-20
fix: use resolve instead of join for publishConfig.directory

manucorporat:
- This will support better paths like: `../../dist/package`


bluwy (approved):
We may also need a changeset for this

Issue #700, opened 2021-12-08
`changeset status` succeeds locally, fails in CI with a changeset file verifiably present


michaelwarren1106:
## Affected Packages
cli

## Problem
I have a monorepo of related packages and `yarn changeset status` succeeds locally, but fails in CI (azure devops). There is definitely a changeset file present in the ADO workspace.

Before I try 

```
yarn changeset status
```
I am testing using just 
```
ls -la
```
to make sure the files actually exist:

![image](https://user-images.githubusercontent.com/8377369/145289791-33b523d7-a54f-4a50-a2bf-f9e4b76cb4bb.png)

and in the `.changeset` folder:
![image](https://user-images.githubusercontent.com/8377369/145289860-4777ca8c-4caa-47f9-a535-754bccf8ecd8.png)

So i KNOW the changeset file exists. I generated it with a standard 

```
yarn changeset add
```

However, when I run 
```
yarn changeset status
```
right after those `ls` commands, i get:

![image](https://user-images.githubusercontent.com/8377369/145289972-858819b6-d2dc-486c-a8d4-2349862a2a9a.png)

I've also tried 

```
yarn changeset status --since="origin/main"
``` 
since `main` is my `baseBranch`.  When i run that, the error is different:

![image](https://user-images.githubusercontent.com/8377369/145290130-529ba07b-8948-468b-a2a2-fefa31a7886f.png)

Here is my config.json:

```
{
  "$schema": "https://unpkg.com/@changesets/config@1.6.3/schema.json",
  "changelog": "@changesets/cli/changelog",
  "commit": true,
  "linked": [],
  "access": "restricted",
  "baseBranch": "main",
  "updateInternalDependencies": "minor",
  "ignore": ["@<scope>/docs", "@<scope>/playroom", "@<scope>/private-playroom"]
}

```

This CI job is running ON a merge of the `main` branch. So the `--since=` I would think isnt necessary, but i can't run the status command without `origin/main` because for some reason changesets cant find the main branch?

Any idea what is happening here? I do want to check for changesets, even an empty one in my CI at least for the time being until we get used to our deployment cadences and figure out better how and when we need changesets.

Andarist:
[...]

PR #1620, opened 2025-03-20
fix: git diff no-relative

Netail:
If diff.relative in the git config has (accidentally) been set to true (In our case by GitHub actions/workflows), the new changesets are not correctly resolved. (as it won't include `.changeset/` in the path, thus not passing the filter) This should fix it.

RE: https://github.com/changesets/changesets/issues/700#issuecomment-2741593490

Friendly pings; @Andarist @bluwy :)

### DIFF
--- a/.changeset/famous-loops-repair.md
+++ b/.changeset/famous-loops-repair.md
@@ -1,0 +1,7 @@
+---
+"@changesets/cli": patch
+---
+
+author: @Netail
+
+Correctly fetch new changesets with since if the git option diff.relative has been set to true
--- a/.changeset/khaki-peaches-refuse.md
+++ b/.changeset/khaki-peaches-refuse.md
@@ -1,0 +1,7 @@
+---
+"@changesets/git": patch
+---
+
+author: @Netail
+
+Make sure that git diff always returns the paths relative to the git root in case diff.relative has been set to true in the git config
--- a/packages/git/src/index.ts
+++ b/packages/git/src/index.ts
@@ -195,7 +195,11 @@ export async function getChangedFilesSince({
 }): Promise<Array<string>> {
   const divergedAt = await getDivergedCommit(cwd, ref);
   // Now we can find which files we added
-  const cmd = await spawn("git", ["diff", "--name-only", divergedAt], { cwd });
+  const cmd = await spawn(
+    "git",
+    ["diff", "--name-only", "--no-relative", divergedAt],
+    { cwd }
+  );
   if (cmd.code !== 0) {
     throw new Error(
       `Failed to diff against ${divergedAt}. Is ${divergedAt} a valid ref?`
@@ -226,7 +230,7 @@ export async function getChangedChangesetFilesSinceRef({
     // Now we can find which files we added
     const cmd = await spawn(
       "git",
-      ["diff", "--name-only", "--diff-filter=d", divergedAt],
+      ["diff", "--name-only", "--diff-filter=d", "--no-relative", divergedAt],
       {
         cwd,
       }


.changeset/famous-loops-repair.md
bluwy, line 1: Not sure we need this changeset. When `@changesets/git` is released, the version range from the CLI should pick it up with an `npm update`.

bluwy (commented):
Seems fine to me. Should we also add `--no-relative` here too? https://github.com/changesets/changesets/blob/5c2bac5fd2c1c2904774c4378e9107dcf8313c7d/packages/git/src/index.ts#L198

.changeset/famous-loops-repair.md
Netail: To have some more specific entries in the CHANGELOG, idk something I had to do in a previous PR

Netail:
> Seems fine to me. Should we also add `--no-relative` here too?
> 
> https://github.com/changesets/changesets/blob/5c2bac5fd2c1c2904774c4378e9107dcf8313c7d/packages/git/src/index.ts#L198

As long as the cwd is always in the root, it should be fine. But with getChangedChangesetFilesSinceRef it first CDs into the .changesets directory, meaning the cwd is not in the root

.changeset/famous-loops-repair.md
bluwy: I'll leave @Andarist to decide perhaps. Usually I won't add another changeset but I'm fine either ways.

bluwy:
> As long as the cwd is always in the root, it should be fine. (I can add it just in case) But with getChangedChangesetFilesSinceRef it first CDs into the .changesets directory, meaning the cwd is not in the root

Ah I didn't realize that the `cwd` passed was the `.changeset` directory directly. I thought it didn't need to since there's already the `.changeset` regex next.

I wonder if we should not do that and let `cwd` be the actual current working directory, but I don't think we need to figure that now and I'm happy with the PR's fix for now.

Netail:
> > As long as the cwd is always in the root, it should be fine. (I can add it just in case) But with getChangedChangesetFilesSinceRef it first CDs into the .changesets directory, meaning the cwd is not in the root
> 
> Ah I didn't realize that the `cwd` passed was the `.changeset` directory directly. I thought it didn't need to since there's already the `.changeset` regex next.
> 
> I wonder if we should not do that and let `cwd` be the actual current working directory, but I don't think we need to figure that now and I'm happy with the PR's fix for now.

Oh that's true, as the cwd is in the .changeset directory already, we could technically get rid of the regex

Aubron:
Shouldn't this be --relative={relative root path from repo root}, rather than --no-relative?

With --no-relative, I still get repo relative roots:
```
aubron@127 % git diff --name-only --no-relative
libraries/javascript/{path}/.changeset/config.json
libraries/javascript/{path}/package.json
libraries/javascript/{path}/pnpm-lock.yaml
```

This still kills `status` in CI:

```
DEBUG: New changeset hashes: [
  "libraries/javascript/{path}/spotty-pumas-behave.md"
]
DEBUG: New hashes: [
  "javascript"
]
DEBUG: Filtered changeset files: []
DEBUG: Changesets: []
```

Netail:
Hi @Aubron

Let's continue the conversation on here, as it allows us more characters & code blocks :)

First of all, is this an issue you were experiencing before as well or only since 2.29.1?
Second, what command do you specifically run?

Aubron:
This is something I've seen in various old versions of the status cli, today was my opportunity to try and understand why it happens:

This happens when your repo includes your .changesets directory not in the root of the repo:

```
- libraries
  - my cool library
    - .changesets
```

* the `status` command gets your list of changed packages, and then reconciles that with your release plan
* This calls 'getChangesets', which (for me in this configuration) returns filename only paths:
```
DEBUG: Found changed files: [
  "README.md",
  "config.json",
  "funny-seas-cross.md",
  "spotty-pumas-behave.md"
]
```

* This calls `filterChangesetsSinceRef`, which returns relative paths to the repo root:
```
DEBUG: New changeset hashes: [
  "libraries/javascript/{path}/.changeset/spotty-pumas-behave.md"
]
```

this is then reduced to its first slice via `.split`, which for me, returns a partial path:
```
DEBUG: New hashes: [
  "javascript"
]
```

This causes the changeset resolution to fail, since it can't find the changesets in the set of hashes

And to confirm, I think our circumstances must be similar-but-not-the-same, and touching the same files.

Netail:
https://github.com/changesets/changesets/blob/5d0185771811680e77e3b1c0ccfe19f4b3f48247/packages/read/src/index.ts#L17

Ah yes, I might see what's wrong. The filter basically splits on "/" and takes item at position 1 (in your case "javascript"). Let me create a fix for you :)

Aubron:
For example, I have fixed this locally to the project by patching that c.split("/")[1] to instead be the last element in the split, but I assume that solution will break in some other configuration

Netail:
> For example, I have fixed this locally to the project by patching that c.split("/")[1] to instead be the last element in the split, but I assume that solution will break in some other configuration
> 
> EDIT: Thanks! Would be nice for someone who has just been touching this code to fix, I'm very worried about breaking others' configs

https://github.com/changesets/changesets/pull/1636 Should do it :)

aribakker:
Note that --no-relative was introduced in git 2.28.0 so this fails in earlier versions of git such as [2.25.1](https://git-scm.com/docs/git-diff/2.25.1) with the slightly misleading error 'Error: Failed to diff against {divergedAt}. Is {divergedAt} a valid ref?'

PR #1636, opened 2025-04-15
fix: correctly fetch changesets (nested changeset directory from git root)

Netail:
Brought up by @Aubron in https://github.com/changesets/changesets/pull/1620#issuecomment-2807668248 (but seperate from the issue fixed in that PR)

When the `.changeset` directory is located lower than the git root it wrongly parses the changeset filenames from the changeset paths. It always tries to get the second item in the array, which is the case for repos where the `.changeset` dir is located in the git root, but not when nested. It's safer to pop off the last item, as it's a file name (Which is filtered by `getChangedChangesetFilesSinceRef` already)

### DIFF
--- a/.changeset/big-maps-say.md
+++ b/.changeset/big-maps-say.md
@@ -1,0 +1,9 @@
+---
+"@changesets/cli": patch
+"@changesets/read": patch
+"@changesets/git": patch
+---
+
+author: @Netail
+
+Correctly resolve new changesets with since option when changesets directory is not directly in the git root
--- a/.changeset/big-maps-say.md
+++ b/.changeset/big-maps-say.md
@@ -6,4 +6,4 @@
 
 author: @Netail
 
-Correctly resolve new changesets with since option when changesets directory is not directly in the git root
+Correctly resolve new changesets with `since` option when the changesets directory is not directly in the git root
--- a/.changeset/big-maps-say.md
+++ b/.changeset/big-maps-say.md
@@ -6,4 +6,4 @@
 
 author: @Netail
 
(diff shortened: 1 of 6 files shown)


Andarist:
Could we add a test case for this? While at it, it would be great to add a test case for https://github.com/changesets/changesets/pull/1620 too

Aubron:
Thanks so much for this @Netail! I mentioned it in another thread, but I think I have wrestled against this issue multiple times when trying to integrate changesets with big monorepos, and its silent failures always eluded me - it will be good to know this particular bug is dead

Netail:
No problemo! Happy to help :)

.changeset/big-maps-say.md
```diff
@@ -0,0 +1,9 @@
+---
+"@changesets/cli": patch
+"@changesets/read": patch
+"@changesets/git": patch
+---
+
+author: @Netail
+
+Correctly resolve new changesets with since option when changesets directory is not directly in the git root
```
Andarist, line 9:
```suggestion
Correctly resolve new changesets with `since` option when the changesets directory is not directly in the git root
```

.changeset/big-maps-say.md
```diff
@@ -0,0 +1,9 @@
+---
+"@changesets/cli": patch
+"@changesets/read": patch
+"@changesets/git": patch
+---
+
+author: @Netail
+
+Correctly resolve new changesets with `since` option when the changesets directory is not directly in the git root
```
Andarist, line 9:
```suggestion
Correctly resolve new changesets with `since` option when the `.changeset` directory is not directly in the git root
```

Netail:
> Thank you! ❤️

You're welcome! 😄

PR #1639, opened 2025-04-17
Remove prettier prod dependency in favor of user-installed formatter

bluwy:
Uses https://github.com/JoshuaKGoldberg/formatly to auto-detect the project's formatter for formatting the changesets and changelogs. Introduces a new `format: "auto" | "prettier" | "biome" | "deno" | "dprint" | false` option to control the formatting, and removes the `prettier: boolean` option in favor of this new option.

Tests has also been updated. Because test fixtures are inside `/tmp` directories with no access to prettier, the test assertions are now all against un-formatted code and has been updated accordingly.

---

NOTES:

1. **On `prettier` option removal:** I didn't deprecate since it was introduced v2.28.0 (17 Feb 2025, not long ago), so I don't expect its usage to be widespread enough to need to maintain it for the next major, so I removed it directly.
2. **On formatting failure:** I can't seem to replicate cases where formatting fails, so I didn't add any warning/error logs for now, but could be something to follow-up if weird things can happen there.
3. **On `formatly` being new:** I think it has potential to be a bit more robust but it should relieve us from maintenance. One caveat I notice is that since it calls a child process, it can take around 300-800ms after confirming the changeset in the CLI.


packages/write/src/index.ts
```diff
@@ -34,8 +37,8 @@ async function writeChangeset(
     capitalize: false,
   });
 
-  const prettierInstance =
-    options?.prettier !== false ? getPrettierInstance(cwd) : undefined;
+  const formatter = await getFormatter(options?.format ?? "auto", cwd);
```
Andarist, line 40: how many of the supported formatters support formatting `.md` files?

bluwy: Ah good catch. Biome actually doesn't support it. The rest (prettier, deno, dprint) do. When detecting the formatter, should we prefer prettier in this case if both biome and prettier exist?

`formatly` currently detects and choose biome first, so we might need to workaround or fix that upstream 🤔

packages/apply-release-plan/src/index.ts
```diff
@@ -68,6 +54,23 @@ async function getCommitsThatAddChangesets(
...
+  cwd: string
+): Promise<Formatter> {
+  if (config === false) return async () => {};
+
+  const formatter =
+    config === "auto" ? (await resolveFormatter(cwd))?.name : config;
+  if (!formatter) return async () => {};
+
+  return async (filePath: string) => {
+    await formatly([filePath], { cwd, formatter });
```
JoshuaKGoldberg, line 70: > One caveat I notice is that since it calls a child process, it can take around 300-800ms after confirming the changeset in the CLI.

Filed: https://github.com/JoshuaKGoldberg/formatly/issues/114

JoshuaKGoldberg: YOLO, merged it 🚀 . Available as `formatly@0.3.0`.

bluwy:
I'm going to close this as it's probably easier to start over

beeequeue:
but using formatly is probably the way to do it at least, it seems like a very nice solution

PR #1652, opened 2025-04-26
Remove support for experimental useCalculatedVersionForSnapshots

bluwy:
It was deprecated in favor of `snapshot.useCalculatedVersion`


packages/assemble-release-plan/src/index.ts
```diff
@@ -139,9 +139,7 @@ function assembleReleasePlan(
         ...config,
         snapshot: {
           prereleaseTemplate: null,
-          useCalculatedVersion: (
-            config.___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH as any
-          ).useCalculatedVersionForSnapshots,
+          useCalculatedVersion: false,
```
Andarist, line 142: There is a TODO comment above this declaration that could be done now (in this PR) - if I'm reading things correctly

bluwy: IIUC it mentions changing the parameter type to be stricter so we don't have to refine it, so I think it would be a separate breaking change (if I understand correctly too).

I think it may be best tackled in another PR though which I can follow up later

Andarist: It's only refined to handle both the old and new formats of `snapshot.useCalculatedVersion`. Given this PR removed the legacy format - I'd say it would be appropriate to handle this TODO here

bluwy: I'm a bit confused but why didn't it refine if `config.snapshot` was passed then? From https://github.com/changesets/changesets/pull/914 it seems that it's to prevent a breaking change if `snapshot` wasn't pass with different versions installed.

The change for this is to make the `config.snapshot` non-optional, and the `snapshot` parameter stricter. And that feels like a different breaking change and out of scope of this PR which addresses `useCalculatedVersionForSnapshots` only.

bluwy:
@Andarist can we handle the TODO from your last review separately? I'd prefer to get this in first.

.changeset/tangy-buses-smoke.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/assemble-release-plan": major
```
Andarist, line 3:
```suggestion
"@changesets/cli": major
"@changesets/assemble-release-plan": major
```

Andarist (approved):

PR #1655, opened 2025-05-01
Update `@manypkg/get-packages` to v3

bluwy:
Some code added to handle the breaking change, mainly the `Package` now needs to specify `dir` and `relativeDir`. And the `tool` needs to be a tool from `@manypkg/tools` and not just a string.

There are a lot of line changes adding `package-lock.json` to the fixtures because that's the only way for it to detect as monorepo (npm). Previously a `"workspaces"` only field was sufficient. @Andarist I'm not sure this is the most ergonomic, shouldn't npm monorepos be detected as fallback without the need for `package-lock.json` to exist?

---

Unrelated but if I may give feedback on the new breaking APIs, I'm not a big fan of passing the `tool` object entirely from `@manypkg/tools`. It should've supported a simple string like before and a union of Tool object instead.

Also, I'm not entirely clear for `Packages` type, why the returned `rootPackage` could be undefined while `rootDir` is not? `rootPackage` shouldn't be possible to be undefined from what I can tell (at least for default tools).

Also, I'm not sure if it was necessarily to introduce both `dir` (absolute) and `relativeDir` (relative) in the object. Feels redundant.


package.json
```diff
@@ -33,12 +33,12 @@
     "Mateusz Burzyński <mateuszburzynski@gmail.com> (https://github.com/Andarist)"
   ],
   "license": "MIT",
-  "dependencies": {
+  "devDependencies": {
```
bluwy, line 36: Breaking change in `@manypkg/cli` v0.23 suggests using devDependencies instead

bluwy:
> That sounds good to me. I'll wait a couple of days before landing this to give @bluwy the chance to comment too.

I'll leave it to y'all to handle this as I don't have much time to follow-up on this lately.

PR #1659, opened 2025-05-06
Enable publishing with provenance

Andarist:


.github/workflows/changeset-version.yml
```diff
@@ -14,6 +14,11 @@ jobs:
     if: github.repository == 'changesets/changesets'
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    permissions:
+      contents: write # to create release (changesets/action)
+      issues: write # to post issue comments (changesets/action)
```
bluwy, line 19: I've checked some other changesets setup and they don't have this permission enabled. I think we can remove this line unless I've missed somewhere this is needed?
```suggestion
```

bluwy (approved):
A question about permissions but pre-approving this anyways

.github/workflows/changeset-version.yml
Andarist: I don't mind having this in already as I'd like to introduce issue comments at some point to the Changesets Action. But yeah, it isn't strictly needed *right now*

PR #1667, opened 2025-05-12
Fixed resolution of changelog/commit scripts paths

Andarist:


bluwy (approved):
Do we need a changeset since the bug before was already released?

bluwy:
I would prefer adding since it helps with debugging later, or just bookkeeping since we're not getting much users I guess 😅 But it's how I'm accustomed to using changesets from Svelte & SvelteKit side where they do the same.

In the final v3 release we could have a pass to remove these not-so-important changelogs though (I can help with that). Or maybe re-assess if including all the prerelease changelog in the stable entry is a good idea. Svelte/SvelteKit had to workaround it as they have like 300 prereleases choking changesets 😄

bluwy:
This needs a rebase, and I prefer a changeset too

Issue #1658, opened 2025-05-06
[Bug] Unexpected error during version command in prerelease mode with private packages


JasperJuergensen:
## Error

```
 InternalError: preVersion for demo does not exist when preState is defined
    at incrementVersion (<cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:152:13)
    at getNewVersion (<cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:468:10)
    at <cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:547:148
    at Array.map (<anonymous>)
    at Object.assembleReleasePlan [as default] (<cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:545:38)
    at version (<cwd>/node_modules/@changesets/cli/dist/changesets-cli.cjs.js:1174:60)
    at async run (<cwd>/node_modules/@changesets/cli/dist/changesets-cli.cjs.js:1339:11) {
  _error: Error
      at new ExtendableError (<cwd>/node_modules/extendable-error/bld/index.js:23:24)
      at new InternalError (<cwd>/node_modules/@changesets/errors/dist/changesets-errors.cjs.js:40:5)
      at incrementVersion (<cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:152:13)
      at getNewVersion (<cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:468:10)
      at <cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:547:148
      at Array.map (<anonymous>)
      at Object.assembleReleasePlan [as default] (<cwd>/node_modules/@changesets/assemble-release-plan/dist/changesets-assemble-release-plan.cjs.js:545:38)
      at version (<cwd>/node_modules/@changesets/cli/dist/changesets-cli.cjs.js:1174:60)
      at async run (<cwd>/node_modules/@changesets/cli/dist/changesets-cli.cjs.js:1339:11)
}
```

## Versions

- @changesets/cli@2.29.3
- node@v22.14.0

## Extra details

This bug has been introduced with #1589 in version 2.29.3.

[...]

PR #1589, opened 2025-02-17
Ignore packages without version during prerelease

remorses:
Fix #835

PR #1668, opened 2025-05-12
Fixed a crash in pre mode when trying to version private packages when tagging for private package is disabled

Andarist:
fixes https://github.com/changesets/changesets/issues/1658 (a regression from https://github.com/changesets/changesets/pull/1589 )


bluwy (approved):
LGTM 👍

PR #1673, opened 2025-05-14
docs: Align used function name with import

Netail:
Align used function name with import

### DIFF
--- a/packages/read/README.md
+++ b/packages/read/README.md
@@ -6,9 +6,9 @@
 Read in all changesets from a repository.
 
 ```js
-import getChangesets from "@changesets/read";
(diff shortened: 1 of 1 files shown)


packages/read/README.md
```diff
@@ -6,7 +6,7 @@
 Read in all changesets from a repository.
 
 ```js
-import read from "@changesets/read";
+import getChangesets from "@changesets/read";
 
 let changesets = await getChangesets(cwd);
```
Andarist, line 11:
```suggestion
import readChangesets from "@changesets/read";

let changesets = await readChangesets(cwd);
```

Netail: Makes sense, thanks!

PR #1675, opened 2025-05-19
fix: make the package manager name in error messages more accurate

btea:
When executing changeset, the error message shows the initialization command is `yarn changeset init`. I think it is more friendly to automatically match the package manager used by the current user and then give the corresponding prompt.


bluwy:
We've keep a simple `changeset init` command for now in main. I think we can revisit again if this is confusing and show the package manager command. Thanks for contributing!

Issue #1238, opened 2023-09-26
How to block major bump of linked package ?


FlorianBurgevin:
## Problem

Hello.
I'm using changeset with Turborepo to manage a Design System in a monorepo.

We can see my repo with two packages as an example:
pkg-A (version 0.5.0)
pkg-B (using the pkg-A with `^0.5.0` in package.json)

I modify my `pkg-A` with a breaking change, and I publish a changeset with major version for this one.

When I do the changeset version, it will update `pkg-A` with major (1.0.0) but also the package.json of `pkg-b` to use the new major version of `pkg-A`.

The problem is I juste updated `pkg-A` and I not managed the breaking changes in `pck-B`.

Is it the expected behaviour ? 
How can I block that ? I think I want to manually upgrade my second package later.

Thanks

R-Bower:
Maybe related: https://github.com/changesets/changesets/issues/1011

onlywei:
I'm having this issue as well. When I make a major version change to one of my packages, I don't want to update all the other projects in my monorepo that depend on it. Those other apps/packages will be able to continue using the existing version that has been published to my internal package registry.

In my opinion this is the whole point of versioning -- I want to be able to opt-in to the new major version LATER in future work.

I think there should be a config option that allows this.

beeequeue:
From what I understand, what you want is only possible if you have both the local workspace version _and_ published versions installed at the same time.

Would `bumpVersionsWithWorkspaceProtocolOnly` suffice? It makes Changesets only bump dependencies with `workspace:` version ranges.
It requires using [`linkWorkspacePackages: false`](https://pnpm.io/workspaces#linkworkspacepackages) with pnpm and `enableTransparentWorkspaces: false` with yarn.

I think using both local and published versions interchangeably without explicitly separating them (i.e. `workspace:`) sounds like a very fishy solution, with strange edge cases being able to pop up everywhere.

onlywei:
@beeequeue You make a good point, but there is also real pain in having to manually remove `workspace:` from every package that depends on the package being published when you suspect it may go out of semver range. It would require every contributor to manually search every usage of the package they are bumping to see if their new version would push any declaration out of semver range and manually removing the `workspace:` protocol.

It would be much nicer devex if changesets could automatically strip the `workspace:` protocol when bumping packages out of semver range.

And yes this does require having local workspace versions AND published versions at the same time, but publishing is sort of the whole point of changesets. I can't imagine anyone using changesets without publishing packages.

PR #1679, opened 2025-05-27
New config option: "requireSemverSatisfaction"

onlywei:
Addresses #1238 

This allows other internal packages to upgrade to breaking changes separately from the packages that make the breaking changes.


mojp-kassel (changes requested):
My github project have been hacked. Please return the all change in default. Thank you

beeequeue:
thanks for the PR!

i added a new comment in #1238, could you take a look at it?

PR #1690, opened 2025-06-16
Add config parameter to avoid adding user's references within changelog

kin0992:
Handle a configuration parameter to instruct changeset not to add the "Thanks @<author>" in the `CHANGELOG.md` file.


bluwy:
There is an existing PR at https://github.com/changesets/changesets/pull/1255. I prefer the code changes here, but since that PR was opened earlier (and I think has better docs), I'll move this change over to that PR and will mark you as a co-author there.

PR #1693, opened 2025-06-19
Fixed an issue with `workspace:^` and `workspace:~` dependency ranges not being semantically treated as, respectively, `^CURRENT_VERSION` and `~CURRENT_VERSION`

Andarist:


packages/cli/src/commands/version/version.test.ts
```diff
@@ -1053,7 +1053,7 @@ describe("workspace range", () => {
     ]);
   });
 
-  it("should bump dependent package when bumping a `workspace:^` dependency", async () => {
+  it("should not bump dependent package when patch bumping a `workspace:^` dependency", async () => {
```
Andarist, line 1056: This test had wrong expected results. It can easily be observed by introducing this change to it:
```diff
      "packages/pkg-a/package.json": JSON.stringify({
        name: "pkg-a",
        version: "1.0.0",
        dependencies: {
-          "pkg-b": "workspace:^",
+          "pkg-b": "^1.0.0",
        },
      }),
```

Those should behave exactly the same as `workspace:^` is just a sugar for the inlined "current version".

bluwy (approved):
Makes sense 👍

Fool256 (approved):

reabiliti:
@Andarist hi, We've tried setup new version and seems like now it's not follow docs - https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#updateinternaldependencies

We have

```
pkg-a @ version 1.0.0
pkg-b @ version 1.0.0
  depends on pkg-a at range `^1.0.0
```

When sets `"updateInternalDependencies": "patch",` in changeset config, patch and minor version of package pkg-a doesn't trigger new version for pkg-b

Seems like now updateInternalDependencies is useless?

bluwy:
From what I understand in the docs, the bump only happens when cutting a patch for both `pkg-a` and `pkg-b`, so you have to cut a patch in `pkg-b` too to trigger that.

It would be best to open a new issue though if what you described has worked in a previous changesets version before and this PR affected it.

PR #1701, opened 2025-07-08
docs(readme): updated leftover

mfranzke:
switched https://github.com/Noviny/changesets to https://github.com/changesets/changesets for the README.

There are still further old org names in CHANGELOG files, didn't touch those.


.changeset/kind-numbers-refuse.md
bluwy, line 1: I don't think we need a changeset here since it doesn't touch any code that requires a version bump. Other than that the changes looks good to me.

Issue #1710, opened 2025-08-07
Vulnerability detected for tmp package when using @changesets/cli


glad-adyen:
My project uses `v2.28.1` of `@changesets/cli`, and recently, a vulnerability was reported for the `tmp` package.
> tmp allows arbitrary temporary file / directory write via symbolic link `dir` parameter

`tmp@0.2.3` is vulnerable to an Arbitrary temporary file / directory write via symbolic link `dir` parameter.

**From my investigations, here is what I found:**
- The `tmp` package is a dependency of the `external-editor` package, which is a dependency of `@changesets/cli`.
- My project currently uses `v2.28.1` of `@changesets/cli`, but even updating to `v2.29.5` (the latest version at the time of opening this issue) still won't make this vulnerability go away _(reason in the next point)_.
- Both `v2.28.1` and `v2.29.5` of `@changesets/cli package` depend on the same version (`v3.1.0`) of the `external-editor` package, which is the latest version, released about 6+ years ago. This latest version of the `external-editor` package still uses a vulnerable version of `tmp` as a dependency (`v0.0.33`).

### Details

According to the documentation there are some conditions that must be held:

```
// https://github.com/raszi/node-tmp/blob/v0.2.3/README.md?plain=1#L41-L50

Other breaking changes, i.e.

- template must be relative to tmpdir
- name must be relative to tmpdir
- dir option must be relative to tmpdir //<-- this assumption can be bypassed using symlinks

are still in place.

In order to override the system's tmpdir, you will have to use the newly
introduced tmpdir option.


// https://github.com/raszi/node-tmp/blob/v0.2.3/README.md?plain=1#L375
* `dir`: the optional temporary directory that must be relative to the system's default temporary directory.
     absolute paths are fine as long as they point to a location under the system's default temporary directory.
     Any directories along the so specified path must exist, otherwise a ENOENT error will be thrown upon access, 
     as tmp will not check the availability of the path, nor will it establish the requested path for you.
```

Related issue: https://github.com/raszi/node-tmp/issues/207.

The issue occurs because `_resolvePath` does not properly handle symbolic link when resolving paths:
```js
// https://github.com/raszi/node-tmp/blob/v0.2.3/lib/tmp.js#L573-L579
function _resolvePath(name, tmpDir) {
  if (name.startsWith(tmpDir)) {
    return path.resolve(name);
  } else {
    return path.resolve(path.join(tmpDir, name));
  }
}
```

[...]

PR #1711, opened 2025-08-13
Switch to maintained fork of external-editor

benmccann:
Removes `tmp` as a dependency which fixes the vulnerability referred to in https://github.com/changesets/changesets/issues/1710

I sent this against the `next` branch. It can be cherry-picked against `main` if we care about pack-porting it or people who care about the vulnerability can switch to the 3.x branch


Andarist:
Agreed

PR #1714, opened 2025-08-19
Add --enable-pnpm-catalog flag to opt into pnpm workspace catalog detection

### LINKED ISSUE #1707: Changesets doesn't detect changes when updating dependency versions in pnpm workspace catalog
# Bug Report

## Environment
Changesets version: @changesets/cli@^2.27.7
Package manager: pnpm with workspaces and catalog feature
Node version: 20.19.1

## Current Behavior
When using pnpm workspaces with the catalog feature, updating a dependency version in the catalog section of pnpm-workspace.yaml is not detected as a change by changesets, even though this update affects all packages in the monorepo that reference that dependency via the catalog.

## Expected Behavior
Changesets should detect when catalog dependency versions are updated and prompt for changelog entries, since these changes effectively update dependencies across multiple packages in the workspace.

## Steps to Reproduce

1. Set up a monorepo with pnpm workspaces using the catalog feature
2. Have multiple packages that reference dependencies via the catalog (e.g., "dependency": "catalog:")
3. Update a dependency version in the catalog section of [pnpm-workspace.yaml](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
4. Run `npx changeset`
5. Observe that no packages have changed according to `changesets`

AFAIK, the issue is that `getChangedPackagesSinceRef` in `@changesets/git` only looks at changed files within packages (`getChangedFilesSince`). It would also need to look at changes in the catalog, and check if the updated external packages are used by a package from the monorepo.

cyril-sf:
When using pnpm workspaces with the catalog feature, updating a dependency version in the catalog section of pnpm-workspace.yaml is detected as a change by changesets.

Allow `changeset add` and `changeset status` to include updates from pnpm workspace catalogs when requested via the new flag.

Fixes #1707

### DIFF
--- a/.changeset/kind-radios-make.md
+++ b/.changeset/kind-radios-make.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+Add --enable-pnpm-catalog flag to opt into pnpm workspace catalog detection
--- a/.changeset/tender-socks-tie.md
+++ b/.changeset/tender-socks-tie.md
@@ -1,0 +1,5 @@
+---
+"@changesets/git": minor
+---
+
+Add support for catalog feature with pnpm workspace
--- a/docs/command-line-options.md
+++ b/docs/command-line-options.md
@@ -3,10 +3,10 @@
 The command line for changesets is the main way of interacting with it. There are 4 main commands. If you are looking for how we recommend you setup and manage changesets with the commands, check out our [intro to using changesets](./intro-to-using-changesets.md)
 
 - init
-- add [--empty][--open]
+- add [--empty][--open][--enable-pnpm-catalog]
 - version [--ignore, --snapshot]
 - publish [--otp=code, --tag]
-- status [--since=master --verbose --output=JSON_FILE.json]
+- status [--since=master --verbose --output=JSON_FILE.json --enable-pnpm-catalog]
 - pre [exit|enter {tag}]
 - tag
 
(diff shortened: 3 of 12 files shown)


chriswiggins:
Found out about pnpm catalogs yesterday (awesome feature!) and this is the only blocker stopping us from adopting them. Anything we can do to help with merging this PR?

elliott-with-the-longest-name-on-github:
I just wanted to respond to this as it bit us over at SvelteKit today:

> If you think about catalog: as being just a sugar over inlined versions from the catalog then Changesets would, traditionally, not bump your monorepo packages based on changes to external dependencies. There were requests to make that work though so perhaps this would be a good occasion to implement this as a flag if you have already implemented the core sinceRef~ logic for the catalog anyway.

This isn't really correct, at least not as I'm reading it. When you publish a package with `catalog:x` versions, the `catalog:x` in that package's `package.json` is replaced with the version from the workspace catalog. So, in effect, updating the version in the workspace catalog is the same as you manually going into each `package.json` that references that package using `catalog:x` and updating the version there. If you were to do that manually, you'd have to change the `package.json` in each package, which would show up as a change to changesets. Ergo, changing the catalog file should show up as a change for each package that references the updated dependency.

Hoping this can get merged at some point; we'd love to be able to use catalogs as our monorepo has quite a few packages and we're basically stuck only using it for dev dependencies.

Netail:
Perhaps a little more generic, to also support Yarn's catalogs?

Issue #779, opened 2022-03-14
Why are monorepo packages not taken into account for status and add commands?


codyzu:
## Problem

`changesets status` does not take into account which packages in a monorepo have or don't have changesets. Instead, it just checks "if any changeset exists".

Given a monorepo with packages `packA` and `packB`. If I make changes to both `packA` and `packB` _and_ add a changeset for only `packA` then `changesets status` executes successfully and does not detect that `packB` has changes.

I would expect `changesets status` to fail in this case and report that `packB` is missing a changeset.

Similarly, it would be helpful for the cli to take into account which packages are missing changesets. So given the above example, I would expect `changeset add` to indicate to the user that `packB` is missing a changeset.

While comparing changesets with [beachball](https://microsoft.github.io/beachball/) I noticed this key difference. I imagine this was a design decision and I am wondering why this is the case and if there are any plans to change this behavior?

Is this a constraint for allowing [empty changsets](https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md#blocking)? Seems like empty changesets could also indicate which package, without requiring a bump. beachball has the concept of a [none](https://github.com/microsoft/beachball/blob/master/docs/cli/change.md#walking-through-the-form) version bump.

Thanks for all of the great work and my apologies if this is already documented somewhere (could be nice to add something to the [decisions docs](https://github.com/changesets/changesets/blob/main/docs/decisions.md)).

Andarist:
> I would expect changeset status to fail in this case and report that packB is missing a changeset.

this behavior could be implemented as an additional flag to the status command. PRs are welcome

> Similarly, it would be helpful for the cli to take into account which packages are missing changesets. So given the above example, I would expect changeset add to indicate to the user that packB is missing a changeset.

interesting - could u propose CLI texts for this? How it would be printed to the user? I guess we could make a new group for such packages that would be listed at the top

> Seems like empty changesets could also indicate which package, without requiring a bump. beachball has the concept of a [none](https://github.com/microsoft/beachball/blob/master/docs/cli/change.md#walking-through-the-form) version bump.

We support this as well! :)

codyzu:
Great, thanks for the feedback... I'm going to check with the team to see if we have bandwidth to help. Based on stars and activity, changesets is more interesting, but we were looking for these slightly different features.

airtonix:
We ended up having to write a dangerjs check to detect if any monorepo package changes in a PR was not covered by any provided changesets.

bluwy:
[...]

PR #1721, opened 2025-09-03
Add --strict option to status command

jonatanschroeder:
Addresses the first part of #779. The `status --strict` option causes `status` to fail if any specific package is not included in the changeset. In essence, in monorepos, currently any changeset is accepted even if it doesn't address all changed packages. This option allows `status --strict` to be used, e.g., in CI for monorepos.

Quick note for information: while I have considered using the changesets bot instead of a status CI, the issue is that our repo includes multiple packages, and most PRs affect only ignored packages or private packages where a changeset is not needed, and the bot is very verbose in those cases. A CI is quiet when it needs to be.


jonatanschroeder:
I'm closing this because I no longer have interest in this feature, and it seems maintainers don't have any interest in reviewing it.

Issue #1724, opened 2025-09-08
Generating wrong changelog when there are special replacement patterns


colinaaa:
We have a changeset like this:

```md
---
"@lynx-js/react-alias-rsbuild-plugin": patch
---

Allow customization of the `react$` alias.
```

And the generated changelog (with `@changesets/changelog-git`) looks like this:

```md
## @lynx-js/react-alias-rsbuild-plugin@0.10.14

### Patch Changes

Allow customization of the `react# @lynx-js/react-alias-rsbuild-plugin alias.
```

See a real-world example at https://github.com/lynx-family/lynx-stack/pull/1665

## Details

After dive into the source code, this issue is caused by the `String.prototype.replace` method used in the `prependFile` function of `@changesets/apply-release-plan`:

https://github.com/changesets/changesets/blob/ca0be26207d74e40abe9edc82219a66a2d09853c/packages/apply-release-plan/src/index.ts#L328

There are [some special replacement patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement) for `String.prototype.replace`. E.g.: `$$`, `$&`, `$'`, `$1`, `$<name>`, etc.

PR #1725, opened 2025-09-08
fix(apply-release-plan): avoid using `String.prototype.replace`

colinaaa:
close: #1724

### DIFF
--- a/.changeset/ninety-crews-allow.md
+++ b/.changeset/ninety-crews-allow.md
@@ -2,4 +2,4 @@
 "@changesets/apply-release-plan": patch
 ---
 
-Fix generating wrong CHANGELOG when having special replacement patterns in changesets.
+Fix an issue that caused an incorrect `CHANGELOG` to be generated when a changeset contained a special string replacement pattern.
--- a/packages/cli/src/commands/version/version.test.ts
+++ b/packages/cli/src/commands/version/version.test.ts
@@ -523,7 +523,7 @@ Awesome feature, hidden behind a feature flag
     `);
   });
 
-  it("should allow special replacement patterns in the changelog", async () => {
+  it("should ignore special string replacement patterns in appended changesets", async () => {
     const spy = jest.spyOn(fs, "writeFile");
 
     const cwd = await testdir({


packages/apply-release-plan/src/index.ts
```diff
@@ -325,7 +325,8 @@ async function prependFile(
     );
     return;
   }
-  const newChangelog = fileData.replace("\n", data);
+  const [header, ...oldChangelogLines] = fileData.split("\n");
```
Andarist: Let's use `indexOf('\n')` + `.slice(...) + data + .slice(...)`

colinaaa: Thanks! I've changed the implementation. PTAL~

packages/apply-release-plan/src/index.ts
```diff
@@ -325,7 +325,18 @@ async function prependFile(
...
-  const newChangelog = fileData.replace("\n", data);
+  const firstNewlineIdx = fileData.indexOf("\n");
+  let newChangelog: string;
+  if (firstNewlineIdx === -1) {
+    // No newline found, treat the whole file as header
+    newChangelog = fileData + data;
+  } else {
+    newChangelog =
+      fileData.slice(0, firstNewlineIdx) +
+      "\n" +
```
Andarist: Doesn't this end up adding an extra newline that *wasn't* in the original `newChangelog`?

Andarist:
```suggestion
      fileData.slice(0, firstNewlineIdx) +
```

packages/apply-release-plan/src/index.ts
```diff
@@ -325,7 +325,17 @@ async function prependFile(
...
+  let newChangelog: string;
+  if (firstNewlineIdx === -1) {
+    // No newline found, treat the whole file as header
+    newChangelog = fileData + data;
+  } else {
+    newChangelog =
+      fileData.slice(0, firstNewlineIdx) +
+      data +
+      fileData.slice(firstNewlineIdx + 1);
+  }
```
Andarist:
```suggestion
  const index = fileData.indexOf("\n");
  const newChangelog =
    index === -1
      ? fileData + data // treat the whole file as header
      : fileData.slice(0, index) + data + fileData.slice(index + 1);
```

packages/cli/src/commands/version/version.test.ts
```diff
@@ -523,6 +523,56 @@ Awesome feature, hidden behind a feature flag
     `);
   });
 
+  it("should allow special replacement patterns in the changelog", async () => {
```
Andarist:
```suggestion
  it("should ignore special string replacement patterns in appended changesets", async () => {
```

.changeset/ninety-crews-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": patch
+---
+
+Fix generating wrong CHANGELOG when having special replacement patterns in changesets.
```
Andarist:
```suggestion
Fix an issue that caused an incorrect `CHANGELOG` to be generated when a changeset contained a special string replacement pattern.
```

PR #1733, opened 2025-10-07
chore: add more robust error handling to get-github-info

francinelucca:
Adds check for when response.json() fails after fetching GitHub information to provide more error information


bluwy:
This is handled by #1781. Sorry for missing this

PR #1737, opened 2025-10-12
update vitest

beeequeue:
this pr updates vitest to the latest version and does a bunch of small code improvements in the tests

- disables the vitest globals as stated in the TODO comment and imports everything manually
- removes all the `@ts-ignore`s and uses either `vi.mocked()` or type assertions instead
- replaces the custom date mock with vitest's built in date mocking
- ~~replaces the custom async `pathExists` functions with `fs.existsSync`~~
- increases timeouts on three tests that always timed out on my machine
- ~~removes `strip-ansi` in favor of `node:util#stripVTControlCharacters`, available since node 16~~
- ~~replaces `file-url` with `node:url#pathToFileUrl`, available since node 10~~

### DIFF
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -3059,8 +3059,8 @@ describe("apply release plan", () => {
     it("should delete an old format changeset if it is applied", async () => {
       const releasePlan = new FakeReleasePlan();
 
-      let changesetMDPath: string;
-      let changesetJSONPath: string;
+      let changesetMDPath!: string;
+      let changesetJSONPath!: string;
 
       const setupFunc = (tempDir: string) =>
         Promise.all(


packages/assemble-release-plan/src/increment.test.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
bluwy, line 1: It would be nice to followup in a PR to remove this comment by installing `vitest` as a devdep of this package (and others with this issue). It looks like some packages already has `vitest` as a devdep.

beeequeue: I think it would be better to update the eslint config to recognize top level dev deps in all workspaces

bluwy: If there's a way to exclude the lint for Vitest only, then I'm fine with that too. Otherwise if we have to disable it entirely I think it's better to add the devdep and call it a day.

beeequeue: I can do that, I was considering taking a crack at updating eslint and the config for it if you don't have any objections :)

Andarist: > It would be nice to followup in a PR to remove this comment by installing vitest as a devdep of this package (and others with this issue). It looks like some packages already has vitest as a devdep.

I would do exactly this, just keep it simple 😉 Having an extra dev dep is not a big deal and doesnt impose any maintenance burden but it's explicit and should work with just any package manager we could think of.

beeequeue: The main reason I wouldn't recommend installing dev tools in every workspaces that use them is that it generally makes the root workspace's dependencies pointless

It feels like adding vitest to the dev deps everywhere is a solution to a misconfigured eslint  rather than providing a benefit

PR #1744, opened 2025-10-15
Prettier v3

beeequeue:
Updates Prettier to v3.

This also means the minimum version of it as a peer dependency also gets bumped to the latest one. Maybe the peer dependency requirement should be changed to `^3`?

I reverted the updated `trailingCommas` setting back to the v2 value to minimize the amount of formatting changes as recommended by @bluwy, but if @Andarist agrees I can revert to the default v3 config and re-format the files according to that.
(Personally I would also set the line-width to 90-100 as well, for better readability)

### DIFF
--- a/packages/get-github-info/src/index.test.ts
+++ b/packages/get-github-info/src/index.test.ts
@@ -95,7 +95,7 @@ test("associated with multiple PRs with only one merged", async () => {
             },
           },
         },
-      })
+      }),
     );
 
   let result = await getInfo({ commit: "a085003", repo: "emotion-js/emotion" });
@@ -211,7 +211,7 @@ test("associated with multiple PRs with multiple merged gets the one that was me
             },
           },
         },
-      })
+      }),
     );
 
   let result = await getInfo({ commit: "a085003", repo: "emotion-js/emotion" });
@@ -291,7 +291,7 @@ test("gets the author of the associated pull request if it exists rather than th
             },
           },
         },
-      })
+      }),
     );
 
   let result = await getInfo({
@@ -339,9 +339,9 @@ test("throws error on missing repo name", async () => {
 
   await expect(async () =>
     // @ts-expect-error
-    getInfo(request)
+    getInfo(request),
   ).rejects.toThrowErrorMatchingInlineSnapshot(
-    `[Error: Please pass a GitHub repository in the form of userOrOrg/repoName to getInfo]`
+    `[Error: Please pass a GitHub repository in the form of userOrOrg/repoName to getInfo]`,
   );
 });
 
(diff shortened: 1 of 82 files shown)


packages/get-github-info/src/index.test.ts
```diff
@@ -15,35 +15,37 @@ test("associated with multiple PRs with only one merged", async () => {
     },
   })
     .post(apiPath, ({ query }) => {
-      expect(prettier.format(query, { parser: "graphql" }))
```
Andarist, line 27: those specific formatting calls here felt quite nice, would it be a problem to keep those?

package.json
```diff
@@ -72,7 +71,9 @@
       "typeModule": true
     }
   },
-  "prettier": {},
+  "prettier": {
+    "trailingComma": "es5"
```
Andarist: let's just use the default here if the default is `"all"`, if not then let's configure it as `"all"` and format everything

.changeset/soft-sloths-burn.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": major
+"@changesets/write": major
```
Andarist:
```suggestion
"@changesets/cli": minor
```

.changeset/soft-sloths-burn.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Bumped minimum Prettier version to 3.6.2
```
Andarist:
```suggestion
Bumped the default Prettier version used in the absence of the local installation to 3.x
```

packages/get-github-info/src/index.test.ts
beeequeue: it comes down to nock not supporting async functions in the mocking, i spent a bit of time trying to refactor it to still work but didn't manage to find a solution i liked

beeequeue: i gave it another go and now the snapshots are formatted again :)

.changeset/soft-sloths-burn.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": minor
+"@changesets/cli": minor
+---
+
+Bumped the default Prettier version used in the absence of the local installation to 3.6.2
```
bluwy:
```suggestion
Bumped the default Prettier version used in the absence of the local installation to v3
```

It's 3.7.4 now, but probably it's simpler to mention v3 instead.

beeequeue: This changeset would hopefully be removed anyways as #1639 should be merged before the full v3 release, should we maybe just not add it in the first place?

bluwy: That PR is still ways off so I think we would likely release this PR's changes first in one of the next prereleases maybe, so would be good to track in which prerelease we did this change even if it might not be relevant later in the final stable release.

Andarist (approved):

PR #1745, opened 2025-10-25
Fix prerelease detection by fetching packument

yonran:
Fix fetching of prerelease packages so that `changeset publish` does not try to overwrite existing package.

Previously, `changeset publish` would fetch the package using `npm info` (aka [`npm view`](https://docs.npmjs.com/cli/v8/commands/npm-view)) and then if the package did not exist, do the `npm publish`.

However, when fetching prereleased packages, `npm info` filters out any published prerelease package since it calls `semver.satisfies` without the `{includePrerelease: true}` option ([view.js](https://github.com/npm/cli/blob/v11.6.2/lib/commands/view.js#L176)).

Example reproduction:

```
yarn changeset pre enter PXP-2165-publish-to-github-packages
yarn changeset version
yarn changeset publish

yarn changeset publish
```

The second publish log contained:
```
🦋  warn Received 404 for npm info "@themaven-net/tempest-common"
🦋  info @themaven-net/tempest-common is being published because our local version (1.2.0-PXP-2165-publish-to-github-packages.5) has not been published on npm
🦋  info Publishing "@themaven-net/tempest-common" at "1.2.0-PXP-2165-publish-to-github-packages.5"
…
🦋  error npm notice Publishing to https://npm.pkg.github.com/ with tag PXP-2165-publish-to-github-packages and restricted access
🦋  error npm error code E409
🦋  error npm error 409 Conflict - PUT https://npm.pkg.github.com/@themaven-net%2ftempest-common - Cannot publish over existing version
```

Disclosure: the code is all generated by claude code (sonnet 4.5) except the first commit which was gpt-5-codex-medium. Take it with a grain of salt.

### DIFF
--- a/packages/cli/src/commands/publish/__tests__/npm-utils.test.ts
+++ b/packages/cli/src/commands/publish/__tests__/npm-utils.test.ts
@@ -6,11 +6,28 @@ jest.mock("pacote", () => ({
   packument: jest.fn(),
 }));
 
+const mockLoad = jest.fn().mockResolvedValue(undefined);
+const mockFlat = {};
+
+jest.mock("@npmcli/config", () => {
+  return class MockConfig {
+    load = mockLoad;
+    flat = mockFlat;
+  };
+});
+
+jest.mock("@npmcli/config/lib/definitions/index.js", () => ({
+  definitions: {},
+  flatten: jest.fn((obj: any) => obj),
+  shorthands: {},
+}));
+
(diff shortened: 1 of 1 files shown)


benmccann (changes requested):
This PR adds almost 200 dependencies. It would be nice to do as much as possible without so many dependencies. E.g. using the built-in `fetch` where possible

56 dependencies: https://npmgraph.js.org/?q=@npmcli/config
124 dependencies: https://npmgraph.js.org/?q=pacote

PR #1753, opened 2025-10-31
Added an automatic update for `package-lock.json` content when applying the release plan

### LINKED ISSUE #462: [NICE TO HAVE][FEATURE REQUEST] Option to ask Changesets actopm to update versions in the package-lock.json
Great tooling! Thank you for your work!

It would be nice if changeset action would provide an option if versions in package-lock.json should also be updated and commited (under the provided token). At the moment I run npm install and commit after the flow with changeset is done.

Thank you!

### LINKED ISSUE #1139: Version not updated in package-lock.json

Andarist:
fixes https://github.com/changesets/action/issues/462

fixes #1139

### DIFF
--- a/packages/apply-release-plan/package.json
+++ b/packages/apply-release-plan/package.json
@@ -27,6 +27,7 @@
     "@manypkg/get-packages": "^1.1.3",
     "detect-indent": "^6.0.0",
     "fs-extra": "^7.0.1",
+    "jsonc-parser": "^3.3.1",
     "lodash.startcase": "^4.4.0",
     "outdent": "^0.5.0",
     "prettier": "^2.7.1",
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -11,6 +11,7 @@ import {
 import { Packages } from "@manypkg/get-packages";
 import detectIndent from "detect-indent";
 import fs from "fs-extra";
+import * as jsonc from "jsonc-parser";
 import path from "path";
 import prettier from "prettier";
 import resolveFrom from "resolve-from";
@@ -184,6 +185,22 @@ export default async function applyReleasePlan(
     );
   }
 
+  const packageLockJsonPath = path.join(cwd, "package-lock.json");
+
+  if (await fs.pathExists(packageLockJsonPath)) {
+    const lock = await fs.readFile(packageLockJsonPath, "utf-8");
+    const edits = finalisedRelease.flatMap((release) =>
+      jsonc.modify(
+        lock,
+        ["packages", path.relative(cwd, release.dir), "version"],
+        release.packageJson.version,
+        {}
+      )
+    );
+    await fs.writeFile(packageLockJsonPath, jsonc.applyEdits(lock, edits));
+    touchedFiles.push(packageLockJsonPath);
+  }
+
   // We return the touched files to be committed in the cli
   return touchedFiles;
 }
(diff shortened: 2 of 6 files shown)


packages/apply-release-plan/package.json
```diff
@@ -27,6 +27,7 @@
     "@manypkg/get-packages": "^1.1.3",
     "detect-indent": "^6.0.0",
     "fs-extra": "^7.0.1",
+    "jsonc-parser": "^3.3.1",
```
Andarist, line 30: Yeah, I know... a new dependency. I feel though it's quite nice to only touch what we really have to when editing those JSON files. The goal is to keep the original formatting as close as possible to minimize the diffs and to avoid people having to run formatters right after invoking Changesets.

packages/apply-release-plan/src/index.ts
```diff
@@ -184,6 +185,22 @@ export default async function applyReleasePlan(
...
   }
 
+  const packageLockJsonPath = path.join(cwd, "package-lock.json");
+
+  if (await fs.pathExists(packageLockJsonPath)) {
+    const lock = await fs.readFile(packageLockJsonPath, "utf-8");
+    const edits = finalisedRelease.flatMap((release) =>
+      jsonc.modify(
+        lock,
+        ["packages", path.relative(cwd, release.dir), "version"],
```
Andarist, line 195: it's likely we also need to update dependencies in those packages... I'll test it out more in some repo before landing this PR

Andarist: I'll also add some tests for this whole branch in the repo here

jakebailey, line 195: npm has 3 lockfile versions, which you should probably handle. Thankfully they overlap, such that v1 and v3 differ, and v2 is just both together.

bluwy (commented):
I'm not a big fan of editing the lockfile this way, especially when there can be multiple versions of it and what if we need to handle for more package manager in the future. It'd be nice if Changesets starts moving towards being package manager agnostic so we don't have to maintain a lot of specific code. 

I think we should try to tackle this differently, maybe a post-step to automatically call `npm i` or document it?

Andarist:
> I'm not a big fan of editing the lockfile this way, especially when there can be multiple versions of it and what if we need to handle for more package manager in the future. 

That's why I have hesitated to do anything about it for so long.

> It'd be nice if Changesets starts moving towards being package manager agnostic so we don't have to maintain a lot of specific code.

In theory, yes - I agree. In practice, it feels like we already have some other code specific to package managers (even when publishing packages). And this would just make this more usable to npm users. 

> I think we should try to tackle this differently, maybe a post-step to automatically call npm i or document it?

As to the documentation, I think it would just be a chore for people to always have to learn about the need to do this when they are using npm. Ideally, things "should just work" and be user-friendly and I feel relying on the docs here wouldn't be that.

`npm i` also has some footguns. If called incorrectly, it could touch unrelated parts of the lockfile and, ideally, only the parts relevant to `changeset version` should change. Then if u'd have some accidental lockfile mismatch on CI it would still fail for legitimate reasons on `--frozen-lockfile` (or similar)

bluwy:
> In theory, yes - I agree. In practice, it feels like we already have some other code specific to package managers (even when publishing packages). And this would just make this more usable to npm users.

But do we want to move towards making things more agnostic? We do have the specific code for publishing packages now, but if we agree, I'd like to try to generalize it again. But if this PR goes through, it'll cause another breaking change, or at least subtle differences.

> `npm i` also has some footguns. If called incorrectly, it could touch unrelated parts of the lockfile and, ideally, only the parts relevant to `changeset version` should change. Then if u'd have some accidental lockfile mismatch on CI it would still fail for legitimate reasons on `--frozen-lockfile` (or similar)

If they're applying a release plan and the lockfile is unstable, I think there's a bigger issue there. If we update anyways, the user would still have to later run `npm i` or something to stabilize the lockfile. So FWICT we should be able to assume that it's stable.

bluwy:
There's an existing PR that does similar changes too: #1484 

And there's also an existing issue about this: #1139

Serhansolo:
Is there an ETA on this?

ktarmyshov:
> If they're applying a release plan and the lockfile is unstable, I think there's a bigger issue there. If we update anyways, the user would still have to later run `npm i` or something to stabilize the lockfile. So FWICT we should be able to assume that it's stable.

Hi, @bluwy!

Thanks again for the clarification.

In my case the lockfile is stable, but my CI workflows use `npm ci`. After `changeset version` updates versions in `package.json`, the lockfile becomes out of sync, so the generated `Version Packages` PR fails in CI unless `package-lock.json` is updated as part of the process.

I used the workaround suggested by @mikeybinns and called `npm install` from the `changeset:version` script (`"changeset:version": "changeset version && npm install && git add --all"`). In my case that seems to update only the relevant version entries in `package-lock.json` and solves the CI issue.

I understand the concern about increasing the package-manager-specific surface area, so I am not trying to push for a particular implementation. I mainly wanted to share that this npm + `npm ci` case does come up in practice, even with an otherwise stable lockfile.

I think it could be helpful to document the recommended workaround for npm users. Since this is typically wired directly into the `changeset:version` script, it might also help to mention that users may want to add a check ensuring that `npm install` only changes the expected version-related entries in `package-lock.json` before running `git add --all`.

Thank you!

Konstantin

PR #1761, opened 2025-11-11
Update vitest to v4

beeequeue:
This PR updates vitest to v4 and makes it use `rolldown-vite`.

I also removed the `@babel/cli` package as I can't find it being used anywhere.

### DIFF
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -2061,7 +2061,10 @@ describe("apply release plan", () => {
         {
           ...releasePlan.config,
           changelog: [
-            path.resolve(import.meta.dirname, "test-utils/simple-get-changelog-entry.ts"),
+            path.resolve(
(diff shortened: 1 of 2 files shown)


bluwy (approved):

package.json
```diff
@@ -33,8 +33,10 @@
     "Mateusz Burzyński <mateuszburzynski@gmail.com> (https://github.com/Andarist)"
   ],
   "license": "MIT",
+  "resolutions": {
+    "vite": "npm:rolldown-vite@7.2.10"
```
Andarist: q: why do we need this? has anything changed since opening the PR? can this now be dropped?

beeequeue: it can't be dropped until vitest updates to use vite 8 by default, but we can update it to use the `vite` v8 beta instead of `rolldown-vite`

and it isn't _required_ but it reduces the amount of dependencies and makes it faster if we use it :)

PR #1764, opened 2025-11-15
Replace `js-yaml` with ad-hoc implementation

### LINKED ISSUE #1762: Security Advisory: `js-yaml` Prototype Pollution (CVE) used via `@changesets/*` Dependencies
Hi maintainers 👋 

### Summary
A recently-flagged security advisory affects the `js-yaml` dependency used transitively within several `@changesets/*` packages. All `js-yaml` versions below `4.1.0` are affected:

[Prototype pollution in merge](https://github.com/nodeca/js-yaml/security/advisories/GHSA-mh29-5h37-fv8m)

The dependency chain (via `@changesets/cli@2.29.7`) looks like this:

```
@changesets/cli@2.29.7
  |-- @changesets/read@0.6.5
      |-- @changesets/parse@0.4.1
          |-- js-yaml@3.14.1
  |-- @manypkg/get-packages@1.1.3
      |-- read-yaml-file@1.1.0
          |-- js-yaml@3.14.1 (deduped)
```

### Attempted Mitigation

I've attempted to override `js-yaml` in our project to the patched version (4.1.1):

```jsonc
"overrides": {
  "js-yaml": "4.1.1"
},
```

…but this results in runtime failures when executing `changeset version`:

<details>
  <summary><strong>Error Output</strong></summary>

```less
npx changeset version

🦋  error Error: could not parse changeset - invalid frontmatter: ---

🦋  error     at Object.parseChangesetFile [as default] (.../node_modules/@changesets/parse/dist/changesets-parse.cjs.js:35:11)
🦋  error     at .../node_modules/@changesets/read/dist/changesets-read.cjs.js:148:71
🦋  error     at async Promise.all (index 0)
🦋  error     at async Object.getChangesets [as default] (.../node_modules/@changesets/read/dist/changesets-read.cjs.js:152:48)
🦋  error     at async version (.../node_modules/@changesets/cli/dist/changesets-cli.cjs.js:1155:34)
🦋  error     at async run (.../node_modules/@changesets/cli/dist/changesets-cli.cjs.js:1339:11)
```
</details>

It appears that `@changesets/parse` depends on YAML parsing behaviour that changed in `js-yaml@4.x`, making the override incompatible.

### Environment
- `@changesets/cli`: 2.29.7
- Node.js: 22.19.0

### Request
Would it be possible to update the relevant `@changesets/*` packages to use `js-yaml@^4.1.1`, or otherwise confirm whether upgrading is feasible?

Given this is a security advisory (prototype pollution in merge), having a path to a secure version would be very helpful for downstream consumers.

Thanks for maintaining this project 😄

marcalexiei:
- Closes #1762 

https://github.com/changesets/changesets/issues/1762#issuecomment-3536615180

> I also wonder if it's worth implementing a simple parser ourselves since we really only need to parse "<pkg-name>": "patch|minor|major".

I implemented a simple parsing logic using Regexp. 
Let me know if you prefer to stick with `js-yaml`, upgrading to v4 should not be a problem.


Andarist:
For now, I'd prefer to update the dependency instead to avoid some unexpected weird breaks.

bluwy:
Since this PR will validate the line, it'll also fix https://github.com/changesets/changesets/issues/1301. Unless we do the validation manually with js-yaml if we wanted to, depends on how we draw the line for breaking changes.

bluwy:
For me I'm fine with this loose regex approach. We can always make it strict if we find people hitting issues, or we need more flexibility down the road. 

At the moment I don't think it's worth investing in writing a compliant parser until we need it.

PR #1765, opened 2025-11-15
ci: account also `skipped` and `cancelled` job statuses in `ci-ok` job

marcalexiei:
small ci tweaks to `ci-ok` job:

- account for `skipped` and `cancelled` statuses of dependant jobs
- use GitHub Actions conditions instead of a shell script to `exit 1` when dependant jobs are not successful


bluwy (approved):

.github/workflows/ci.yml
```diff
@@ -61,14 +61,7 @@ jobs:
...
-        run: |
-          echo $FAILURE
-          if [ "$FAILURE" = "false" ]; then
-            exit 0
-          else
-            exit 1
-          fi
+      - name: Exit with error if some jobs are not successful
+        run: exit 1
+        if: ${{ always() && (contains(needs.*.result, 'failure') || contains(needs.*.result, 'skipped') || contains(needs.*.result, 'cancelled')) }}
```
Andarist, line 67: It doesn't quite matter in our case right now but shouldn't skipped jobs be OK? I'd consider skipping to be done to optional jobs that can safely be skipped without affecting the overall result

bluwy: Right now we want all jobs to not be skipped, but we could definitely tweak later if we want to allow some. I think it's good to be as strict as possible for now.

marcalexiei: From my perspective, `ci-ok` should include every job that is required to pass,
so treating a `skipped` status as invalid is correct.

If a required `ci-ok` job depends on an optional job, that optional job should be omitted
from the `needs` list when it is not executed. 
In other words, the `needs` key should be generated dynamically based on which optional jobs actually run.

Andarist: I'm not sure `needs` can be dynamic (Claude claims it cannot be). I can imagine a "docs-build" job that could be skipped when no `.md` files were touched in the PR or something. In such a scenario, if it got ran (not-skipped) then `needs` should still await its result and require its success - but if it got skipped then it's fine to proceed without it.

FWIW, at this point I'm just thinking out loud and discussing best practices etc ;p As I mentiond, the change is fine for now anyway

marcalexiei: > FWIW, at this point I'm just thinking out loud and discussing best practices etc ;p

No problem, I replied just to share the reason I proposed that code. I'm happy to discuss it!

---

> I'm not sure needs can be dynamic (Claude claims it cannot be)

I've used dynamically produced matrix from an earlier jobs but I actually never used in `needs` 😅. 
I just naively assumed that should work there but probably is not feasible since the runner must establish jobs relationships before running any job.

---

> I can imagine a "docs-build" job that could be skipped when no .md files were touched in the PR or something. In such a scenario, if it got ran (not-skipped) then needs should still await its result and require its success - but if it got skipped then it's fine to proceed without it.

Ah I see your point.
In that case probably something like this should do the trick (indented for readability)

```yml
runs-on: ubuntu-latest
if: always()
needs: [test, typecheck, lint, docs-build]

steps:
  - name: Exit with error if some jobs are not successful
    if: ${{ 
      always() &&
      (
        contains(needs.*.result, 'failure') ||
        contains(needs.*.result, 'cancelled') ||
        (
          contains(needs.*.result, 'skipped') &&
          needs.docs-build.result != 'skipped'
        )
      )
    }}
    run: exit 1
```

PR #1774, opened 2025-11-18
Support ESM for custom changelog and commit options

### LINKED ISSUE #1406: Changeset using its CJS files in a ESM repo
## Affected Packages

Noticed with `apply-release-plan`, because it does try to load some of my code (for the changelog functions).

## Problem

So, I recently updated [my repo](https://github.com/Siteimprove/alfa) from CJS to ESM (changing `type: "module"` in `package.json`, plus a bunch of TS settings). Among others, this caused my file who's providing [changelog functions](https://github.com/Siteimprove/alfa/blob/fcc0904a67ff42418bfaaa0d5891a24599ad266e/packages/alfa-toolchain/src/changeset/changelog-individual.ts) to also change.

This broke the `yarn changeset version` flow: [failing action](https://github.com/Siteimprove/alfa/actions/runs/9744310217/job/26889659339) with the message:
> error Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/alfa/alfa/packages/alfa-toolchain/dist/changeset/changelog-individual.js from /home/runner/work/alfa/alfa/node_modules/@changesets/apply-release-plan/dist/changesets-apply-release-plan.cjs.js not supported.

So, Changeset used its CJS `apply-release-plan/dist/changesets-apply-release-plan.cjs.js` file to load my changelog functions (pointed through [the "changelog" field of my config](https://github.com/Siteimprove/alfa/blob/main/.changeset/config.json#L4), and since it is a CJS module trying to `require` an ESM one, it broke.

I did [switch back this single file to CJS](https://github.com/Siteimprove/alfa/pull/1643) which resulting in [successful release](https://github.com/Siteimprove/alfa/actions/runs/9745645224) 🎉 

Now, it seems that changeset also provides a `apply-release-plan/dist/changesets-apply-release-plan.esm.js` file (OTOH, which still tries to `let possibleChangelogFunc = require(changelogPath);` on line 336).

So, I assume that changeset should pick its CJS or ESM version depending on the project, and I was hoping in my case it would automagically decide to use the ESM version and correctly loads the changelog functions I provide as ESM 🤔

### LINKED ISSUE #1510: [bug] @changesets/cli running cjs file, even though the package is a module
## Affected Packages

@changesets/cli

## Context

I'm trying to create a script that runs on `commit`. Following the documentation, I have added this to my changesets config.json:

{
  "commit": ["../scripts/commit.js", { "customOption": true }]
}

When using es6 syntax inside `commit.js`,  `pnpm changeset` fails because it is run in a `.cjs` file. Relevant output:

```shell
❯ pnpm changeset
🦋  What kind of change is this for rh-admin-utils? (current version is 2.3.5) · patch
🦋  Please enter a summary for this change (this will be in the changelogs).
🦋    (submit empty line to open external editor)
🦋  Summary · goo bar baz
🦋  
🦋  === Summary of changesets ===
🦋  patch:  rh-admin-utils
🦋  
🦋  Is this your desired changeset? (Y/n) · true
🦋  error Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/[...]/scripts/commit.js from /Users/[...]/node_modules/.pnpm/@changesets+cli@2.27.9/node_modules/@changesets/cli/dist/changesets-cli.cjs.js not supported.
```

I've seen that there exists an esm version just besides the `changesets-cli.cjs.js` version. My package has `{"type": "module"}`.

## Proposed solution

I don't know really :)

bluwy:
fix #1406 
fix #1510

Use `import()` instead of `require()` so it supports ESM too. Dynamic imports are supported since node 13.2 which I believe is already covered.

I also manually tested the return value of `import()` compared to `require()`, and added a nested `.default` check to maintain backwards compatibility. When we support ESM, we can decide whether we want to support `export const getReleaseLine = ...` for example so it doesn't have to be wrapped in a default-exported object. In this PR, I allowed it which required the nested `.default` check.

While this change is also indirectly fixed in `next` https://github.com/changesets/changesets/pull/1482, I figured it doesn't hurt to fix in main first.

### DIFF
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -223,6 +223,11 @@ async function getNewChangelogEntry(
   let possibleChangelogFunc = await import(changelogPath);
   if (possibleChangelogFunc.default) {
     possibleChangelogFunc = possibleChangelogFunc.default;
+
+    // Check nested default again in case it's CJS with `__esModule` interop
+    if (possibleChangelogFunc.default) {
+      possibleChangelogFunc = possibleChangelogFunc.default;
+    }
   }
   if (
     typeof possibleChangelogFunc.getReleaseLine === "function" &&
--- a/packages/cli/src/commit/getCommitFunctions.ts
+++ b/packages/cli/src/commit/getCommitFunctions.ts
@@ -17,6 +17,11 @@ export async function getCommitFunctions(
   let possibleCommitFunc = await import(commitPath);
   if (possibleCommitFunc.default) {
     possibleCommitFunc = possibleCommitFunc.default;
+
+    // Check nested default again in case it's CJS with `__esModule` interop
+    if (possibleCommitFunc.default) {
+      possibleCommitFunc = possibleCommitFunc.default;
+    }
   }
   if (
     typeof possibleCommitFunc.getAddMessage === "function" ||


reabiliti (approved):

reabiliti:
@Andarist hi, would you take a look? Needs this changes to use custom changeset generator in esm libs

PR #1783, opened 2025-11-20
Support GitHub URL environment variables

### LINKED ISSUE #1442: Custom GitHub URL
## Affected Packages

`pkg:@changesets/get-github-info`

## Problem

In my team, we have our own GitHub Enterprise instance, running on premises servers. This means that our repositories are **not** in github.com, and instead, we use another URL (like https://github.example.corp.com/) with a different API URL too (https://github.example.corp.com/api/v3/), and tokens generated from this instance (and not the original GitHub).

Changesets is not compatible with custom GitHub URL for now, but it would be great to be able to configure it.

Our workaround for now is to use `yarn patch` to edit all the references of `github.com` in the package `@changesets/get-github-info` into our instance URL, but this means that it is difficult to update changesets because we must change the patch everytime.

## Proposed solution

It would be great to have two new configuration properties in `.changeset/config.json`, one for the instance URL (like `https://github.example.corp.com/`) and another one for API access (`https://github.example.corp.com/api/v3` or `https://api.github.example.corp.com/`, depending on domain isolation set by the administrators of the instance).

These options should be optional, and default to the default GitHub instance.

mrginglymus:
- Fixes #1442

### DIFF
--- a/packages/get-github-info/README.md
+++ b/packages/get-github-info/README.md
@@ -58,6 +58,10 @@ You'll need to [get a GitHub personal access token](https://github.com/settings/
 GITHUB_TOKEN=token_here
 ```
 
+You can now bump your packages and changelogs with `changeset version` and it'll have the GitHub info. 🎉
+
+### GitHub Enterprise Server
+
 If you are using GitHub Enterprise Server, you can configure `@changesets/get-github-info` to point at it using the following
 environment variables:
 
@@ -66,7 +70,7 @@ GITHUB_SERVER_URL=https://github.example.com
 GITHUB_GRAPHQL_URL=https://github.example.com/api/graphql
 ```
 
-You can now bump your packages and changelogs with `changeset version` and it'll have the GitHub info. 🎉
+When using GitHub Actions, these environment variables will already have been set.
 
 ## API
 


packages/get-github-info/README.md
```diff
@@ -58,8 +58,18 @@ You'll need to [get a GitHub personal access token](https://github.com/settings/
...
+environment variables:
+
+```bash
+GITHUB_SERVER_URL=https://github.example.com
+GITHUB_GRAPHQL_URL=https://github.example.com/api/graphql
+```
+
 You can now bump your packages and changelogs with `changeset version` and it'll have the GitHub info. 🎉
 
+When used in GitHub Actions, all three environment variables will have been set already.
```
bluwy: Minor thing, `GITHUB_TOKEN` isn't set by default in GitHub actions, right now the actions documented setting `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`. Somewhat related I have a feature request to make it automatic though: https://github.com/changesets/action/issues/525

mrginglymus: Oh my bad...I guess I'm just used to putting that on all my actions files :D

packages/get-github-info/README.md
```diff
@@ -58,8 +58,18 @@ You'll need to [get a GitHub personal access token](https://github.com/settings/
...
+environment variables:
+
+```bash
+GITHUB_SERVER_URL=https://github.example.com
+GITHUB_GRAPHQL_URL=https://github.example.com/api/graphql
+```
+
 You can now bump your packages and changelogs with `changeset version` and it'll have the GitHub info. 🎉
 
+When used in GitHub Actions, the URL environment variables will have been set already.
```
Andarist: This feels redundant given we also have to handle defaults. The fact that they might be set just in case in GitHub Actions doesn't really change that much for the user
```suggestion
```

packages/get-github-info/src/index.ts
```diff
@@ -257,7 +267,7 @@ export async function getInfoFromPullRequest(request: {
       commit: commit
         ? `[\`${commit.abbreviatedOid.slice(0, 7)}\`](${commit.commitUrl})`
         : null,
-      pull: `[#${request.pull}](https://github.com/${request.repo}/pull/${request.pull})`,
+      pull: `[#${request.pull}](${data.url})`,
```
Andarist, line 270: fyi: I changed this to rely on the return value instead of creating the URL on our own here.

packages/get-github-info/README.md
mrginglymus: I added this to emphasise that this change will Just Work (tm) without any configuration required when running on a GitHub Enterprise Server

packages/get-github-info/README.md
```diff
@@ -58,6 +58,14 @@ You'll need to [get a GitHub personal access token](https://github.com/settings/
 GITHUB_TOKEN=token_here
 ```
 
+If you are using GitHub Enterprise Server, you can set configure `@changesets/get-github-info` to point at it using the following
```
mrginglymus:
```suggestion
If you are using GitHub Enterprise Server, you can configure `@changesets/get-github-info` to point at it using the following
```

packages/get-github-info/README.md
```diff
@@ -58,6 +58,14 @@ You'll need to [get a GitHub personal access token](https://github.com/settings/
...
 ```
 
+If you are using GitHub Enterprise Server, you can configure `@changesets/get-github-info` to point at it using the following
+environment variables:
+
+```bash
+GITHUB_SERVER_URL=https://github.example.com
+GITHUB_GRAPHQL_URL=https://github.example.com/api/graphql
+```
+
```
mrginglymus, line 72: I do think it's worth emphasising that these variables _do not_ need to be set when running on GitHub Actions.

packages/get-github-info/README.md
Andarist: Ah, that makes sense! Could you re-add this? I have not thought about this to be tightly-tied to this section on GitHub Enterprise. With that in mind, it makes sense to include this information

mrginglymus: Moved it to its own subsection of the readme to make it clearer

PR #1795, opened 2025-12-03
Update `@manypkg/get-packages` to `v2`

marcalexiei:
- Follow-up to **#1772**

`@manypkg/get-packages@1` depends on `read-yml-files@1`, which uses `js-yaml@3`.

Upgrading to **`@manypkg/get-packages@2.2.2`** removes this and other dependencies.

### Why not `@manypkg/get-packages@3`?

Version 3 is ESM-only. 
This repository still relies on Jest 29, which does not support ESM by default, so upgrading to v3 isn’t feasible yet.

---

I’m aware that a few changes may conflict with the `next` branch, 
but I believe it’s worthwhile to merge this into `main` to reduce unnecessary dependencies.

If you're happy with this PR, I can also prepare a follow-up to bring the same changes into `next`. 
Since `next` branch uses `vitest`, we could consider upgrading `@manypkg/get-packages` to v3 there.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -959,7 +959,7 @@
     "@babel/plugin-transform-modules-commonjs" "^7.26.3"
     "@babel/plugin-transform-typescript" "^7.27.0"
 
-"@babel/runtime@^7.5.5", "@babel/runtime@^7.7.7":
+"@babel/runtime@^7.7.7":
   version "7.27.0"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.27.0.tgz#fbee7cf97c709518ecc1f590984481d5460d4762"
   integrity sha512-VtPOkrdPHZsKc/clNqyi9WUA8TINkZ4cGk63UUE3u4pmB2k+ZMQRDuIOagv8UVd6j7k0T3+RRIb7beKTebNbcw==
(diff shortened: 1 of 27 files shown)


yarn.lock
```diff
@@ -5264,16 +5219,6 @@ react-is@^18.0.0:
   resolved "https://registry.yarnpkg.com/react-is/-/react-is-18.2.0.tgz#199431eeaaa2e09f86427efbb4f1473edb47609b"
   integrity sha512-xWGDIW6x921xtzPkhiULtthJHoJvBbF3q26fzloPCK0hsvxtPVelvftw3zjbHWSkR2km9Z+4uxbDDK/6Zw9B8w==
 
-read-yaml-file@^1.1.0:
```
marcalexiei, line 5267: @bluwy I'm moving the conversation here since #1772 is closed

> [I think we cant upgrade it in this major due to its breaking change.](https://github.com/changesets/changesets/pull/1772#issuecomment-3605423038)

From [`@manypkg/get-packages@2.0.0` changelog](https://github.com/Thinkmill/manypkg/releases/tag/%40manypkg%2Fget-packages%402.0.0)

There is a change in `getPackages` structure:

> The get-packages package now returns a slightly different structure.  The old tool string has been replaced with a tool object, using the new Tool interface provided by @manypkg/tools. Each Package now contains both the absolute directory and relative directory path. Last, the root package has been renamed rootPackage and is optional, to support monorepos that do not contain a root package.

I assumed that when working with a changeset there is always a root `package`, so I added a non-null assertion to avoid the type-checking error.

For the `tool` updates, I added access to the `type` property and used a simple `{ type: 'tool' }` object in the mock functions, along with a type cast.

marcalexiei:
@Andarist As I understand it, the original goal was for `assembleReleasePlan` to receive a serialisable structure. 
With the introduction of the new `tool` property in `manypkg@2` and `manypkg@3`, that is no longer possible, correct?

I reviewed the following packages:

- `assemble-release-plan`
- `apply-release-plan`

In most places, `@manypkg/get-packages` is used primarily to import the types that define the shape of the `packages` argument.

### Proposal

Introduce a new interface in `@changesets/types` that defines the expected structure for the `packages` parameter, and use that internally within the repository.

This would decouple the implementation from `@manypkg/get-packages` and allow consumers to construct the required structure independently, whether they rely on `@manypkg/get-packages` or not (E.g. the bot).

The structure would closely resemble `Packages` from `@manypkg/get-packages@1`.

### Rough Interface Sketch

```ts
interface ChangesetPackages {
  rootPackage?: {
    dir: string;
  };
  packages: Array<{
    dir: string;
    // from "@changesets/types"
    packageJson: PackageJSON;
  }>;
}
```

Would this direction be something you’d consider?

PR #1796, opened 2025-12-06
feat(cli): Allow monorepos where all packages are fixed to use the root-style-tag (#1683)

### LINKED ISSUE #1638: "Fixed' packages in mono repo should reuse single short-named tag/release
When `fixed` packages detected and the count is exactly match the mono packages could be published, then we should use single tag and release.

I'd like to raise a PR for it.

cc @Andarist

harrisonhoward:
# Summary
- Centralised the git tagging to a single location to improve reliabilty of the process.
- Added support for monorepos where all packages are fixed together to utilise the root-style-tag format (v0.0.0).
# Related issues
Closes #1638


packages/cli/src/commands/tag/index.ts
harrisonhoward, line 1: TODO: This is untested.

packages/cli/src/utils/getUntaggedPackages.ts
```diff
@@ -1,35 +1,43 @@
 import * as git from "@changesets/git";
+import { PackageJSON } from "@changesets/types";
 import { Package, Tool } from "@manypkg/get-packages";
-import { PublishedResult } from "../commands/publish/publishPackages";
+
+type UnpublishedPackage = {
```
harrisonhoward: I believe it's entire possible to undo this now. While `tagName` is used, it doesn't need to be anymore, I had originally planned for this function to handle the forced root, however that has changed.

harrisonhoward (commented):
Minor changes needed but the end result is the same.

ilfa:
Hey @beeequeue!

Any chance that this one will be in v3 in future?

PR #1800, opened 2025-12-11
Update GH Actions, Test Against Node 24

beeequeue:
Adds Node 24 to the testing matrix, along with adding `--ignore-scripts` to the yarn install command.

I also went ahead and updated all the versions of the actions and pinned them to their hashes, except for codecov since it has a breaking change.

### DIFF
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -23,6 +23,6 @@ runs:
       shell: bash
       run: yarn preconstruct dev
 
-    - name: Verify
+    - name: Manypkg check
       shell: bash
       run: yarn manypkg check


.github/workflows/ci.yml
```diff
@@ -32,18 +35,18 @@ jobs:
...
       # Run coverage only once on node 22 to avoid duplicate codecov uploads
       - name: Vitest tests with coverage
-        if: matrix.node_version == 22
+        if: matrix.node_version != 20
         run: yarn vitest --no-file-parallelism --coverage
 
       - name: Upload coverage
         uses: codecov/codecov-action@v4
-        if: matrix.node_version == 22
+        if: matrix.node_version == 24
```
bluwy, line 49: I think we should only replace the `22` before with `24` and keep the equality checks as before. The tests with coverage should only run in one matrix variant

beeequeue: i changed it since i dont really see the point in only running one with coverage when the next step only uploads one of them anyways

doing it this way gets it closer to a future point where the node 20 step can be dropped entirely and just have one test running step

bluwy: We only need one coverage result, why run tests with coverage multiple times? This isn't about node 20 not being able to run coverage tests, we specifically only want to pick one variant to run and collect coverage info.

beeequeue: Okay I'll change it to what you say.
Is there some special reason to not run coverage on every test job I don't know about?

perf-wise it's negligible

image

beeequeue: Okay I'll change it to what you say.
Is there some special reason to not run coverage on every year run I don't know about?

bluwy: I think it's only that we don't have to do more work when we can, and doing coverage collection only once isn't too hard.

.github/actions/ci-setup/action.yml
bluwy:
```suggestion
    default: 24
```

Probably should update here too

.github/workflows/ci.yml
```diff
@@ -32,18 +35,18 @@ jobs:
         run: git config --global user.email "you@example.com" && git config --global user.name "Your Name"
 
       - name: Vitest tests
-        if: matrix.node_version != 22
+        if: matrix.node_version != 24
         # @changesets/release-utils depend on `--experimental-strip-types` so we exclude those test here for simplicity
         run: yarn vitest --no-file-parallelism --exclude=packages/release-utils
 
       # Run coverage only once on node 22 to avoid duplicate codecov uploads
```
bluwy:
```suggestion
      # Run coverage only once on node 24 to avoid duplicate codecov uploads
```

.github/actions/ci-setup/action.yml
```diff
@@ -4,17 +4,25 @@ inputs:
...
       with:
         node-version: ${{ inputs.node-version }}
         cache: yarn
 
     - name: Install dependencies
       shell: bash
-      run: yarn install --frozen-lockfile
+      run: yarn install --frozen-lockfile --ignore-scripts
+
+    - name: Run build
```
Andarist:
```suggestion
    - name: Setup dev environment
```

.github/actions/ci-setup/action.yml
```diff
@@ -4,17 +4,25 @@ inputs:
...
     - name: Install dependencies
       shell: bash
-      run: yarn install --frozen-lockfile
+      run: yarn install --frozen-lockfile --ignore-scripts
+
+    - name: Setup dev environment
+      shell: bash
+      run: yarn preconstruct dev
+
+    - name: Verify
```
Andarist:
```suggestion
    - name: Manypkg check
```

PR #1806, opened 2025-12-18
Allow Changesets to be executed from non-root directories

### LINKED ISSUE #545: There is no .changeset folder
## Affected Packages

`pkg:cli`

## Problem

A typical configuration of a monorepo is that packages are two levels deep.

```
.
├── .changesets
└── packages/
    ├── package-a
    └── package-b
```

With the `changeset` command can be called in the root and in `./packages` but when the current working dir is deeper it errors with 

```
🦋  error There is no .changeset folder. 
```

## Proposed solution

I would expect the CLI to go up the directory tree until it find that `.changeset` folder.

### LINKED ISSUE #1824: enterPre() always creates pre.json at the top level of a monorepo
We only use changsets in one package (called core) in our monorepo.  When I run `npx changesets pre enter testingtesting` from that package it creates a pre.json in a folder at the top level of the repository, even though my changeset md files are in the `.changeset` folder in core and where `pre.json` is.  We recently changed to a monoreop architecture and everything worked fine before.

However, if I exit pre mode with `npx changsets pre exit` from that package it will change the `pre.json` in core.

### setting preStatePath
Looking through the code, it seems to be because in [enterPre](https://github.com/changesets/changesets/blob/27fd8f41dddafcc2e96e7df39dca04d92f916a0a/packages/pre/src/index.ts#L47) you set your `preStatePath` like this:

```
 let packages = await getPackages(cwd);
 let preStatePath = path.resolve(packages.root.dir, ".changeset", "pre.json");
```

However in exitPre you set `preStatePath` like this:

`let preStatePath = path.resolve(cwd, ".changeset", "pre.json");`

Which results in a very different file path to the `pre.json`

### get packages
getPackages always finds the root of the monorepo because of the way it calls getPackage from @manypkg/get-packages so I will never be able to enter pre mode in the core package.

`enterPre()` calls [getPackages()](https://github.com/Thinkmill/manypkg/blob/327e4ac7673134f4d49b8414679849daa2e2eaef/packages/get-packages/src/index.ts#L30) from `@manypkg/get-packages` which says it

> Given a starting folder, search that folder and its parents until a supported monorepo
> is found, and return the collection of packages and a corresponding monorepo `Tool`
> object.

Tat is because it calls [findRoot()](https://github.com/Thinkmill/manypkg/blob/327e4ac7673134f4d49b8414679849daa2e2eaef/packages/find-root/src/index.ts#L66) which:

> Given a starting folder, search that folder and its parents until a supported monorepo
> is found, and return a `MonorepoRoot`

It's possible I'm missing something and there is an easy fix, but I need to be able to enter pre mode in just one of my packages.

luisadame:
## Context

Allows changesets to be executed from deeply nested folders within a monorepo that has the .changeset folder at the root of it.

A common example of this is a monorepo with two packages: frontend and backend. These packages might have different scripts that people run within those folders. When they finish some piece of work, they have to change directories back to the root to create a changeset.

This change helps exactly with that, but also, with most of the changeset commands.

### Implementation details
I've added a new package to determine the root of the project: [@manypkg/find-root](https://github.com/Thinkmill/manypkg/tree/main/packages/find-root). This package gets dynamically imported when the current directory doesn't have a changeset folder. Then, it will try to determine if the root has a changeset folder or not. Otherwise the same behavior as always executes, which is throwing an error.

### Questions/Unknowns
The new package is an ES module and therefore I've had some issues using it here, in terms of typings, execution, and testing.
- The types can't be determined because the current TS module resolution can't work with ESM (as typescript compiler said when I run it).
- I initially imported the package at the top of the file, but that doesn't work well in CJS. However, the dynamic import seems to bypass that issue.
- Because it uses that dynamic import now the test suite has to run with a special node flag to enable that feature.

My question is, are we ok with this? I'm open to other ideas or suggestions.

fixes #545
fixes https://github.com/changesets/changesets/issues/1824


.changeset/warm-numbers-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+Changeset CLI can now be run from nested folders where the .changeset folder is not present
```
Andarist:
```suggestion
Changeset CLI can now be run from the nested directories in the project, where the `.changeset` directory has to be found in one of the parent directories
```

Andarist (approved):

xccxcs:
If the .changeset directory is included in the execution directory, it will still look up to the parent directory until the most root .changeset directory is found, which will cause the configuration of the .changeset folder in my subdirectory to be invalidated

Andarist:
@xccxcs please raise a new issue describing your directory structure in more detail

PR #1810, opened 2025-12-31
Replace `String.trimRight` with `String.trimEnd`

hirasso:
I was just researching how to create my own Changelog format. Stumbled upon `String.prototype.trimRight()`, wich is deprecated: https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/trimEnd

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ To make releasing easier, you can use [this changesets github action](https://gi
 - [verdaccio](https://verdaccio.org)
 - [Chakra UI](https://chakra-ui.com)
 - [Astro](https://astro.build)
+- [Biome](https://biomejs.dev)
 - [SvelteKit](https://kit.svelte.dev)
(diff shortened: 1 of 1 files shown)


.changeset/tidy-hoops-feel.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/changelog-github": patch
+"@changesets/get-github-info": patch
```
bluwy: Not quite sure we need this entry since it only affects the readme

hirasso: Should I do something here?

bluwy: I'm suggesting to remove the line:
```suggestion
```

This feels right to me, but if others think otherwise that we should keep I'm also fine hence I've approved it.

PR #1834, opened 2026-02-04
Export schema.json in `@changesets/config` package exports

RodrigoHamuy:
## Summary

Modern Node.js versions like `pnpm` with strict module resolution respect the exports field, which means the file is inaccessible even though it's physically in the package.

This mean projects using `pnpm` can't access locally installed  `../node_modules/@changesets/config/schema.json`.

This PR exports the `schema.json` file from the `@changesets/config` package so you can do that on `pnpm` too.

## Changes
- Added `./schema.json` to the `exports` field in `packages/config/package.json`
- `pnpm` projects can now reference the schema in their `.changeset/config.json` using:
  ```json
  {
    "$schema": "../node_modules/@changesets/config/schema.json"
  }
  ```

## Benefits
- Provides IDE autocomplete and validation for `.changeset/config.json` files
- Improves developer experience by catching configuration errors early
- Makes the schema discoverable and easily accessible to package consumers

### DIFF
--- a/.changeset/fix-config-schema-export.md
+++ b/.changeset/fix-config-schema-export.md
@@ -2,4 +2,4 @@
 "@changesets/config": patch
 ---
 
-Export `schema.json` in the package's `exports` field. Package managers with strict module resolution like `pnpm` respect the `exports` field, which previously made the schema file inaccessible. Users can now reference the schema in their `.changeset/config.json` using `"$schema": "../node_modules/@changesets/config/schema.json"`.
+Export `schema.json` in the package's `exports` field to make it accessible when using modern package managers.


.changeset/fix-config-schema-export.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
+Export `schema.json` in the package's `exports` field. Package managers with strict module resolution like `pnpm` respect the `exports` field, which previously made the schema file inaccessible. Users can now reference the schema in their `.changeset/config.json` using `"$schema": "../node_modules/@changesets/config/schema.json"`.
```
Andarist: Tbf, all package managers should now respect the `package.json#exports`:
```suggestion
Export `schema.json` in the package's `exports` field to make it accessible when using modern package managers.
```

PR #1838, opened 2026-02-05
Revert "Export schema.json in `@changesets/config` package exports (#1834)"

RodrigoHamuy:
As per comments https://github.com/changesets/changesets/pull/1834#issuecomment-3848062337

Reverted PR https://github.com/changesets/changesets/pull/1834 didn't make sense, sorry!


Andarist:
I think the `/schema.json` should be exposed through the public exports: https://github.com/changesets/changesets/pull/1834#issuecomment-3849283539

Issue #1167, opened 2023-05-31
Inconsistent handling of "ignore" packages


dipunm:
## Problem

When configuring a package as ignored by directly naming it, there are validation checks that lead to this error:

```
The package "packageA" depends on the ignored package "packageB", but "packageA" is not being ignored. Please add "PackageA" to the `ignore` option.
```

However, when configuring it with a wildcard, this error is no longer thrown.

I am unsure why this error was introduced exactly, however for my use case, I need to be able to ignore a package without getting this error. This is my use case:

I have a project with publishable packages, and a few private shared workspace packages that help define configurations for development purposes.

The advised way to skip these types of packages is to remove their `version` fields, however with `pnpm`, the version field is required when publishing.

This is an open issue [here](https://github.com/pnpm/pnpm/issues/5094)

## Proposed solution

My suggestion is to remove the error completely and to update the [docs](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages) to both document that wildcards are supported, and to remove the disclaimer that suggests that this should be a temporarily set field.

Alternatively, we should update the validation logic to detect packages captured by wildcards so that we don't have an inconsistent experience and the use case should be expanded upon so that it is clear why this error exists. If this is the option chosen, then we should also discuss the use case I gave above to find an alternative solution for it.

dipunm:
A followup, after configuring with wildcard, (I prefixed my config packages with "@config/" and used an ignore rule of "@config/*") `pnpm changeset add` allows me to create a changeset, but then `pnpm changeset version` errors with a similar but slightly different error:
```bash
The package "packageA" depends on the ignored package "@config/packageB", but "packageA" is not being ignored. Please pass "PackageA" to the `--ignore` flag.
```

NullVoxPopuli:
I've been running in to obnoxious behavior with `ignore` as well.

If I specify something in ignore:
- I don't care about devDependencies being published
- I don't care if things in `ignore` can't be found in my repo

I want an option to specify a list of projects to "just shut up about" :upside_down_face:

Issue #1093, opened 2023-02-18
@changesets/cli: Private package should be ignored from CHANGELOG.md


unional:
## Problem

https://github.com/unional/clibuilder/pull/459

In the above example, the `fixture-*` package are all private packages without `version` and with `private: true` in their package.json.

## Proposed solution

These packages should not be included in the CHANGELOG.md generation.

satya164:
Seems if I add the package names to ignore entry in the config then the changelogs aren't generated. Though I'd expect `private` to behave similarly since docs say to avoid `ignore` and use `private: true` instead.

e110c0:
I ran into a similar issue: even though an internal package in a mono repo is set `private: true`, I can select it when running `changeset`. If I add it to the ignore array, all works as expected, but according to the [documentation](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages) this should not be used permanently.

1oglop1:
This feature works but it may have ended up as a regression or went undocumented during

https://github.com/changesets/changesets/releases/tag/%40changesets%2Fcli%402.25.0

@Andarist Can you please confirm if this is just a documentation issue?

unional:
btw, if you try to use `ignore` to workaround this, it doesn't work.
I run into this:

```sh
🦋  error ValidationError: Some errors occurred when validating the changesets config:
🦋  error The package "clibuilder" depends on the ignored package "cjs-plugin", but "clibuilder" is not being ignored. Please add "clibuilder" to the `ignore` option.
```

Because I'm adding the test package (`cjs-plugin` in this case) as `devDependencies`.

enisdenjo:
Any news here? This issue can lead to congested changelogs when you have many private workspaces.

jakebailey:
If you set this in your config:

```json
  "privatePackages": {
    "tag": false,
    "version": false
  }
```

Tags, version bumps, and changelogs will all not be produced.

enisdenjo:
Oh nice! I reckon https://github.com/changesets/changesets/issues/1093#issuecomment-2254235349 can be an accepted answer and can therefore close this issue.

boyeln:
@jakebailey I'm not able to get that to work. We have a bunch of apps in our monorepo where all have `"private": true` in package.json. Setting the `"privatePackages"` option removes them from the list when using the changeset CLI, but they are still bumped and changelogs are generated when running the `version` command.

PR #1842, opened 2026-02-10
Allow versioned private packages to depend on skipped packages without requiring them to also be skipped

RodrigoHamuy:
Fixes https://github.com/changesets/changesets/issues/906
Fixes https://github.com/changesets/changesets/issues/1093
Fixes https://github.com/changesets/changesets/issues/1167

## Summary

Allow versioned private packages (apps) to depend on skipped/ignored packages without erroring.

Previously, `changesets version` would fail with "depends on the skipped package" even when the dependent was a private package that would never be published to npm. This fix applies to both the `--ignore` CLI flag (`@changesets/cli`) and the `ignore` config option (`@changesets/config`).

- Added test in `packages/cli/src/run.test.ts` for the CLI validation path
- Added test in `packages/config/src/index.test.ts` for the config validation path
- Existing tests for non-private packages still correctly enforce the validation

### DIFF
--- a/.changeset/private-deps-validation.md
+++ b/.changeset/private-deps-validation.md
@@ -1,6 +1,5 @@
 ---
 "@changesets/cli": patch
-"@changesets/config": patch
 ---
 
-Allow versioned private packages to depend on skipped packages without requiring them to also be skipped. Private packages are not published to npm, so it is safe for them to have dependencies on ignored or unversioned packages.
+Allow private packages to depend on skipped packages without requiring them to also be skipped. Private packages are not published to npm, so it is safe for them to have dependencies on ignored or unversioned packages.
--- a/.changeset/private-deps-config-validation.md
+++ b/.changeset/private-deps-config-validation.md
@@ -1,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
+Allow private packages to depend on skipped packages without requiring them to also be skipped. `devDependencies` on skipped packages no longer trigger a validation error in config parsing (aligning with the existing CLI behavior). The config validation for skipped-package dependents now also covers packages skipped via `privatePackages.version: false`, not just those in the `ignore` list.
--- a/.changeset/private-packages-shape-validation.md
+++ b/.changeset/private-packages-shape-validation.md
@@ -1,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
+Added shape validation for the `privatePackages` config option.


.changeset/private-deps-validation.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/cli": patch
+"@changesets/config": patch
+---
+
+Allow versioned private packages to depend on skipped packages without requiring them to also be skipped. Private packages are not published to npm, so it is safe for them to have dependencies on ignored or unversioned packages.
```
Andarist: @RodrigoHamuy any reason why do you think only **versioned** private packages should be exempt of this validation?

RodrigoHamuy: No strong feelings, always simpler if it's up the the user. You mean we should also allow versioned **public** packages to depend on skipped pckages, right? Happy to action if you prefer

RodrigoHamuy: Actually, I think as it is is correct. Publishing a dependency that depends on non-published libraries ain't gonna work.

Andarist: > You mean we should also allow versioned public packages to depend on skipped pckages, right? 

No, I meant that we perhaps should allow unversioned private packages to depend on skipped packages. Those parhaps shouldn't reach the relevant functions on many occassions but I couldn't see a strong reason for gating the new behavior on `config.privatePackages.version === true`

RodrigoHamuy: I think unversioned private packages were already implicitly exempt — since they never go through the versioning pipeline, they wouldn't trigger this validation in practice. So the net new behavior is really for **versioned** private packages, which is what the changeset describes.

Happy to update the wording if you feel it's misleading, but I think it's accurate as-is.

Andarist: I'll take care of the wording. That's not an issue. I just want to doublecheck the original intention of the PR. Notice that I already simplified the condition [here](https://github.com/changesets/changesets/pull/1842/changes/b4d847714f010d8729c481ad7d08f619e8970c7b#diff-df960c2b907455dd827b0ac9f55518091f3632fb6114b83465e909d07efdf97bL167-L170) - I'm just making sure it was the right call now.

RodrigoHamuy: @Andarist personally I think it was better as it was before, as only if `config.version` is set we consider this an `app` and not a package based on the docs: https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md

RodrigoHamuy: but if it works, great!! But you may end up with published package.json files that depend on things that aren't being published (even if to a private registry). Not sure if that's OK? Maybe it is if they are devDependencies?

PR #1860, opened 2026-02-28
fix(version): exit with code 1 when no unreleased changesets found

### LINKED ISSUE #1815: changeset version should return a non-zero error code if there are no changesets
In a repo with changesets configured but no changesets:

```sh
$ npx changeset version --snapshot canary
🦋  warn No unreleased changesets found, exiting.
$ echo $?
0
```

This makes it difficult to publish canary builds using changesets. For example the following CI job:

```sh
$ pnpm changeset version --snapshot canary
$ pnpm publish --tag canary
```

The expected behaviour is that we version our packages with a `-canary-<something>` suffix, and then publish packages with version `x.y.z.canary-<something>` to NPM

But in the case that there are no changesets, this will print the warning about no changesets, and then silently continue to publish our packages as `x.y.z` with the tag canary. This then causes issues if another CI job is intended to publish `x.y.z` with the tag latest, as that version is already published on npm.

I think the best solution would be for `changeset version` to return a non-zero exit code when there are no changesets.

mixelburg:
## Summary

Closes #1815

`changeset version` currently exits with code 0 when no unreleased changesets are found, even though it has done nothing. This makes CI pipelines unable to detect the no-op case, which can lead to subtle bugs — for example when using `--snapshot` mode:

```sh
# No changesets exist
$ npx changeset version --snapshot canary
🦋  warn No unreleased changesets found, exiting.
$ echo $?
0
# ⚠️ Publish still runs! Packages get published with wrong version tags
$ npm publish --tag canary
```

## Changes

- **`packages/cli/src/commands/version/index.ts`**: Replace `return` with `throw new ExitError(1)` when no unreleased changesets are found
- **`packages/cli/src/commands/version/version.test.ts`**: Update the no-changesets test to assert that an `ExitError` is thrown with code 1
- **`.changeset/fix-version-no-changesets-exit-code.md`**: Changeset describing the patch

## Migration

CI pipelines that currently call `changeset version` and ignore the exit code will need to handle the new exit code 1. A typical pattern would be:

```sh
changeset version --snapshot canary || exit 0  # ignore if no changesets
# or
if changeset version --snapshot canary; then
  npm publish --tag canary
fi
```


bluwy:
Should we move this to the next branch? I think this is an ok change.

.changeset/fix-version-no-changesets-exit-code.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": patch
```
Andarist:
```suggestion
"@changesets/cli": major
```

.changeset/fix-version-no-changesets-exit-code.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": major
+---
+
+`changeset version` now exits with code 1 when there are no unreleased changesets, instead of silently exiting with code 0. This makes it easier to detect in CI pipelines when a version step is a no-op — for example, to prevent accidentally publishing packages with incorrect version tags when using `--snapshot` mode.
```
beeequeue:
```suggestion
`changeset version` now exits with code 1 when there are no unreleased changesets, instead of silently exiting with code 0.

This makes it easier to detect when a version step is a no-op — for example, to prevent accidentally publishing packages with incorrect version tags when using `--snapshot` mode.
```

just a tiny phrasing and formatting change

beeequeue (changes requested):

PR #1871, opened 2026-03-04
Use `GITHUB_REPOSITORY` as the default repo when no `repo` option is configured in `changelog-github`

### LINKED ISSUE #1870: changelog-github - infer repo name from environment variables
Currently, when using `@changesets/changelog-github`, users have to always define the repo name as options: `["@changesets/changelog-github", { "repo": "org/repo" }]`

However, in GitHub Actions, the repo name is also available as an environment variable `GITHUB_REPOSITORY` (see https://docs.github.com/en/actions/reference/workflows-and-actions/variables).

It would simplify usage if the changelog function used that as a fallback, and only throw error if neither of them (env var or options.repo) are defined.

When running locally, this changelog generator already requires user to set up `.env` with `GITHUB_TOKEN` and some other env vars, so doesn't make it more complex IMO - also, I think this is mostly used in CI anyway.

------

Happy to send a PR implementing this.

mihkeleidast:
Fixes #1870.

Makes the changelog generator options object fully optional, as the env variable is defined in GitHub Actions.

Less configuring, simpler to use.


packages/changelog-github/src/index.test.ts
```diff
@@ -239,3 +239,16 @@ it("with multiple authors", async () => {
...
+  process.env.GITHUB_REPOSITORY = data.repo;
+  const [changeset, bump] = getChangeset("fixes #1234 and #5678", data.commit);
+  expect(await getReleaseLine(changeset, bump, null)).toMatchInlineSnapshot(`
+    "
+
+    - [#1613](https://github.com/emotion-js/emotion/pull/1613) [\`a085003\`](https://github.com/emotion-js/emotion/commit/a085003) Thanks [@Andarist](https://github.com/Andarist)! - something
+        fixes [#1234](https://github.com/emotion-js/emotion/issues/1234) and [#5678](https://github.com/emotion-js/emotion/issues/5678)"
+  `);
+
+  process.env.GITHUB_REPOSITORY = undefined;
```
Andarist: Please wrap this in try/catch so the env variable gets reverted correctly. Ideally, we should do this using a helper like this:
```ts
function setEnvironmentVariable(name: string, value: string | undefined) {
  const hadValue = Object.hasOwn(process.env, name);
  const originalValue = process.env[name];

  if (typeof value === "string") {
    process.env[name] = value;
  } else {
    delete process.env[name];
  }

  return () => {
    if (hadValue) {
      process.env[name] = originalValue;
    } else {
      delete process.env[name];
    }
  };
}
```

Feel free to add it to the test-utils

packages/changelog-github/src/index.ts
```diff
@@ -30,19 +38,15 @@ const changelogFunctions: ChangelogFunctions = {
     dependenciesUpdated,
     options
   ) => {
-    if (!options.repo) {
-      throw new Error(
-        'Please provide a repo to this changelog generator like this:\n"changelog": ["@changesets/changelog-github", { "repo": "org/repo" }]'
-      );
-    }
+    const { GITHUB_REPOSITORY } = readEnv(options);
```
Andarist: Let's not couple `options.repo` with env. Please introduce a new util called smth like `resolveRepository`

mihkeleidast: Done!

packages/changelog-github/src/index.test.ts
mihkeleidast: Should be done if I understood you correctly.

.changeset/big-lions-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/test-utils": patch
+---
+
+Add `setEnvironmentVariable` utility
```
Andarist: This is a private pkg so we just dont't care about changesets there ;p
```suggestion
---
---
```

Andarist (approved):

PR #1872, opened 2026-03-04
feat(types): add `Packages` to replace types from `@manypkg/get-packages`

marcalexiei:
- Followup of https://github.com/changesets/changesets/pull/1795

This PR introduces `ChangesetsPackages` types to reduce the number of package relying on `@manypkg/get-packages`

### DIFF
--- a/.changeset/green-bottles-leave.md
+++ b/.changeset/green-bottles-leave.md
@@ -1,0 +1,5 @@
+---
+"@changesets/types": minor
+---
+
+Export `Package` and `Packages` from `@changesets/types`. They are meant to be used instead of the types from `@manypkg/get-packages`.


packages/types/src/index.ts
```diff
@@ -175,3 +175,18 @@ export type PreState = {
   };
   changesets: string[];
 };
+
+export interface ChangesetsPackage {
```
Andarist: given we are in the scope of Changesets already, I'd prefer to remove the `Changesets` prefix from those types

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -63,17 +66,17 @@ export default function getDependencyGraph(
...
     string,
-    { pkg: Package; dependencies: Array<string> }
+    { pkg: ChangesetsPackage; dependencies: Array<string> }
   >();
   let valid = true;
 
-  const packagesByName: { [key: string]: Package } = {
-    [packages.root.packageJson.name]: packages.root,
+  const packagesByName: { [key: string]: ChangesetsPackage } = {
+    [packages.root!.packageJson.name]: packages.root!,
```
Andarist: isn't root required in the new type?

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -54,7 +57,7 @@ const getValidRange = (potentialRange: string) => {
 };
 
 export default function getDependencyGraph(
-  packages: Packages,
+  packages: Pick<ChangesetsPackages, "packages" | "root">,
```
Andarist: q: why do we use `Pick` here now?

packages/apply-release-plan/src/index.test.ts
```diff
@@ -2947,9 +2959,10 @@ describe("apply release plan", () => {
       await git.commit("first commit", tempDir);
 
       try {
+        const { root, packages, tool } = await getPackages(tempDir);
```
Andarist, line 2962: q: so now `tool` is just a string so we wrap it in `{ type: tool }` to conform to thy manypkg's v2 shape, right? once we migrate to it this value will change, we'll be able to just pass in `tool` directly because it will be an object with a `.type` property. On our end, in Changesets, we'll only require the `.type` to exist though

marcalexiei: Exactly. 
The idea is to use an object so it can be extended with additional fields in the future if needed. 

For now, the main benefit is that it keeps the type compatible with the one from `manypkg@v2`.

packages/get-dependents-graph/src/get-dependency-graph.ts
marcalexiei: Leftover, the not null assertion shouldn't be needed.

~~Maybe in the future I would consider turning on the [no-non-null-assertion](https://typescript-eslint.io/rules/no-non-null-assertion/) rule that reports similar situations.~~

EDIT: Maybe in the future I would consider turning on the [no-unnecessary-type-assertion](https://typescript-eslint.io/rules/no-unnecessary-type-assertion/) rule that reports similar situations.

packages/types/src/index.ts
marcalexiei: I prefixed the name with `Changeset` to avoid clashes with the ones coming from `manypkg`.
I can rename them now.

packages/get-dependents-graph/src/get-dependency-graph.ts
marcalexiei: I used a `Pick` since this package only accesses those two properties, so I thought it would be clearer to make that explicit.

packages/get-dependents-graph/src/get-dependency-graph.ts
beeequeue: https://github.com/changesets/changesets/pull/1876/changes#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R41
it's in progress :^)

Issue #1869, opened 2026-03-04
more ideas for cleaning up the project ahead of v3


beeequeue:
related to #665

the ideas ive looked at or started working on are
- migrate from yarn v1 to pnpm
  - the main part i would need help with for this is the publishing flow for the repository, since its very easy to miss some part of it
- build with [tsdown](https://tsdown.dev/) instead of preconstruct
- replace enquirer with @clack/prompts (#1861)
- replace @inquirer/external-editor with launch-editor
- run knip on the project
- replace micromatch with picomatch or zeptomatch
- replace spawndamnit with tinyexec

let me know which ones sound good and ill revive or start working on them :)

bluwy:
@beeequeue can you merge this into https://github.com/changesets/changesets/issues/1945? We can discuss some points in discord if there's some you didn't get consensus on yet.

beeequeue:
i added the pnpm migration to v3 as the final topic from here, tsdown will wait for v4 unless plans change

PR #1875, opened 2026-03-05
replace spawndamnit with tinyexec

beeequeue:
part of #1869 

replaces `spawndamnit` with the smaller `tinyexec` which will also dedupe much better as many other packages use it, as can even be seen in this repo! 😄

image

the only change code-wise is that `tinyexec` puts the node-specific options under `nodeOptions` instead of the root options objects:

```diff
-spawn("ls", { cwd: "./foo" })
+exec("ls", { nodeOptions: { cwd: "./foo" } })
```

### DIFF
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -124,10 +124,12 @@ async function testSetup(
     await git.commit("first commit", tempDir);
   }
 
+  const { root, packages, tool } = await getPackages(tempDir);
+
   return {
     changedFiles: await applyReleasePlan(
       releasePlan,
-      await getPackages(tempDir),
+      { root, packages, tool: { type: tool } },
       config,
       snapshot,
     ),
@@ -2950,9 +2952,10 @@ describe("apply release plan", () => {
       await git.commit("first commit", tempDir);
 
       try {
+        const { root, packages, tool } = await getPackages(tempDir);
         await applyReleasePlan(
           releasePlan.getReleasePlan(),
-          await getPackages(tempDir),
+          { root, packages, tool: { type: tool } },
           releasePlan.config,
         );
       } catch (e) {
@@ -3002,9 +3005,11 @@ describe("apply release plan", () => {
         await git.commit("first commit", tempDir);
 
         try {
+          const { root, packages, tool } = await getPackages(tempDir);
+
           await applyReleasePlan(
             releasePlan.getReleasePlan(),
-            await getPackages(tempDir),
+            { root, packages, tool: { type: tool } },
             {
               ...releasePlan.config,
               changelog: [
--- a/packages/release-utils/src/run.ts
+++ b/packages/release-utils/src/run.ts
@@ -1,11 +1,12 @@
-import path from "node:path";
-import { createRequire } from "node:module";
-import { exec } from "tinyexec";
-import { getPackages, type Package } from "@manypkg/get-packages";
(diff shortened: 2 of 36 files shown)


packages/apply-release-plan/src/index.test.ts
```diff
@@ -12,7 +12,7 @@ import { existsSync } from "node:fs";
 import fs from "node:fs/promises";
 import path from "path";
 import { outdent } from "outdent";
-import spawn from "spawndamnit";
+import { exec as spawn } from "tinyexec";
```
Andarist: let's just embrace `exec` ;p

beeequeue: alright, it was just easier to not find+replace 😄

beeequeue: do you have any clue why this is done in this file?

https://github.com/changesets/changesets/blob/next/packages/release-utils/src/utils.ts#L12

Andarist: Not really - it looks pretty standard though (assuming it works 😂). It might be replicating some (old?) code from `@actions/core` or similar github package but I'm not sure

packages/release-utils/src/utils.ts
```diff
@@ -92,78 +84,6 @@ export function getChangelogEntry(changelog: string, version: string) {
   };
 }
 
-const activeProcesses = new Set<ChildProcess>();
-
-onExit(() => {
-  for (let child of activeProcesses) {
-    child.kill(SIGTERM);
-  }
-});
```
Andarist: oh wow, this dates back to: https://github.com/changesets/changesets/commit/df424a4a09eea15b0fa9159ee0b98af0d95f58a7#diff-bc83e21a29c754b54d601fc5b2ab7be373aba7da931f30bf0b7a986dbf87defa

I have NO recollection of adding this 😅

packages/release-utils/src/run.ts
```diff
@@ -1,20 +1,17 @@
...
 import { readChangesetState } from "./readChangesetState.ts";
-import { createRequire } from "node:module";
-import path from "node:path";
 
 const require = createRequire(import.meta.url);
 
 type PublishOptions = {
-  script: string;
+  command: string;
+  args?: string[];
```
Andarist, line 15: this is a breaking change that should be called out in a changeset file

beeequeue: ah yes, in my mind it was internal-only but that's not the case :^)

packages/release-utils/src/utils.ts
```diff
@@ -92,78 +84,6 @@ export function getChangelogEntry(changelog: string, version: string) {
...
-
-export async function execWithOutput(
```
Andarist: techncially those 2 were public exports, I think it's fine to remove them - but we should call this out in a changeset file

packages/release-utils/src/utils.test.ts
```diff
@@ -72,35 +73,48 @@ let changelog = `# @keystone-alpha/email
...
-      private: false,
-    },
-    {
-      name: "c",
-      highestLevel: BumpLevels.major,
-      private: false,
-    },
-  ];
-  expect(things.sort(sortChangelogEntries)).toMatchSnapshot();
+describe("exec", () => {
```
Andarist: let's remove it - it won't bring a lot of value long-term

beeequeue: yes, it was for testing but i forgot to remove it

PR #1876, opened 2026-03-06
upgrade eslint to v10 and update config

beeequeue:
part of #1869 

- there are various rules marked as `TODO enable and fix errors` which are good to have but have too many errors and will be fixed in follow-up PRs
- i personally love sorted imports, but i know they're not everyone's cup of tea - should i add it in another PR?
- i will add all the formatting/fixing commits to the [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-23-ignore-commits-in-the-blame-view-beta/) file in the follow-up PRs.


Andarist:
> should prefer-const be enabled? i see let being used in most places so i assume its by choice, but could you explain why? @Andarist

This predates my involvement with the project ;p it was just a preference of some of the previous maintainers or smth.

packages/cli/src/index.ts
```diff
@@ -90,10 +89,7 @@ ${format("", err).replace(process.cwd(), "<cwd>")}
 
 ## Versions
 
-- @changesets/cli@${
```
beeequeue, line 101: this is something that `tsdown` could handle very well:

```js
import manifest from "@changesets/cli/package.json" with { type: "json }

export const foo = `@changesets/cli@${manifest.version}`
```

compiles to

```js
const manifest = {
  version: "0.0.0",
}

export const foo = `@changesets/cli@${manifest.version}`
```

or maybe even this, but probably only if minifying the output

```js
export const foo = `@changesets/cli@0.0.0`
```

beeequeue: ... this does of course assume one builds the package _after_ upping the versions and publishing

packages/get-github-info/src/index.ts
```diff
@@ -114,17 +114,23 @@ const GHDataLoader = new DataLoader(async (requests: RequestData[]) => {
       },
       body: JSON.stringify({ query: makeQuery(repos) }),
     });
-  } catch (e: any) {
+  } catch (e) {
     throw new Error(
-      `An error occurred when fetching data from GitHub\n${e.message}`,
+      `An error occurred when fetching data from GitHub\n${(e as Error).message}`,
+      { cause: e },
```
beeequeue, line 144: adding `cause`s here _should_ help with surfacing errors, but i'm not quite sure if they will actually show up anywhere.

scutuatua-crypto (approved):

Netail:
Could potentially also be Biome or OXC ;)

Andarist:
@Netail we plan to explore the switch once v3 gets released

bluwy (approved):

bluwy:
> * i will add all the formatting/fixing commits to the [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-23-ignore-commits-in-the-blame-view-beta/) file in the follow-up PRs.

Personally I think it's fine to not add it given the assorted changes here, plus it's not too noisy

beeequeue:
some of the fixes will be quite noisy i think, but i can do it only for those

bluwy (approved):

PR #1885, opened 2026-03-09
fix(config): do not error on unmatched ignore patterns

### LINKED ISSUE #1794: Changeset fails on ignore pattern when there's no matches
Prisma now generates their own nested subfolders which includes a `package.json` file. It's created via a `prisma generate` which we do on `postinstall`. Since they're automatically they are gitignored.

The issue is that we need the ignore locally but the changeset bot never runs an install so it never matches the ignore glob.  Here's my config:

## Config

```json
{
  "$schema": "https://unpkg.com/@changesets/config@3.1.1/schema.json",
  "commit": false,
  "privatePackages": {
    "version": true,
    "tag": true
  },
  "fixed": [],
  "linked": [],
  "access": "restricted",
  "baseBranch": "main",
  "ignore": ["prisma-client-*"]
}
```

## Error

Image

## Proposed Solution

The path(s) in the `ignore` config should NOT fail if they do not find a match.

mixelburg:
## Problem

When specifying a package name or glob expression in the `ignore` config option that doesn't match any workspace package, changesets throws a validation error:

```
The package or glob expression "generated-package-*" is specified in the `ignore` option but it is not found in the project.
```

This makes it impossible to use `ignore` patterns for packages that may not always be present, such as:
- Dynamically generated packages (e.g. Prisma client packages that only exist after `prisma generate`)
- Conditionally installed packages (present in some environments but not others)
- Future packages not yet created but already listed in config

## Solution

Remove the unmatched-pattern validation for the `ignore` option. Unlike `fixed` and `linked` groups — where all packages must exist to form a valid constraint group — ignoring a non-existent package is a safe no-op: the pattern simply matches nothing and has no effect on the release plan.

The `fixed` and `linked` group validations are unchanged, since those do require all referenced packages to exist.

## Changes

- Removes the unmatched-pattern error check in `@changesets/config`
- Updates tests: the two tests that previously expected errors for non-existent ignore packages now verify the config parses successfully with an empty `ignore` array

## Related

Closes #1794

### DIFF
--- a/packages/config/src/index.test.ts
+++ b/packages/config/src/index.test.ts
@@ -711,18 +711,15 @@ describe("parser errors", () => {
     `);
   });
   test("ignore package that does not exist", () => {
-    expect(() => unsafeParse({ ignore: ["pkg-a"] }, defaultPackages))
-      .toThrowErrorMatchingInlineSnapshot(`
-      "Some errors occurred when validating the changesets config:
-      The package or glob expression "pkg-a" is specified in the \`ignore\` option but it is not found in the project. You may have misspelled the package name or provided an invalid glob expression. Note that glob expressions must be defined according to https://www.npmjs.com/package/micromatch"
-    `);
+    // Should not throw - unmatched ignore patterns are allowed (e.g. for
+    // dynamically generated packages that may not always be present)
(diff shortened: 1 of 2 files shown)


mixelburg:
Closing due to inactivity. Over 3 months with no review engagement. Happy to reopen if needed.

PR #1888, opened 2026-03-12
Fix several `changeset version` issues with workspace protocol dependencies

### LINKED ISSUE #1886: `shouldUpdateDependencyBasedOnConfig` is called with `workspace:...` `depVersionRange`s
I'm using explicit `workspace:^1.2.3`-style versions for `peerDependencies` so as to comply with npm's recommendation to keep peerDependencies ranges as broad as possible - this way I can manually bump them when the change becomes significant.

However; here:

https://github.com/changesets/changesets/blob/3ab4d8928ce3c4d4f4e3d1b09be91ec0a1315bc0/packages/apply-release-plan/src/version-package.ts#L47-L65

The raw range (`workspace:^1.2.3`) is passed to `shouldUpdateDependencyBasedOnConfig`; which in turn passes it to `semver`:

https://github.com/changesets/changesets/blob/3ab4d8928ce3c4d4f4e3d1b09be91ec0a1315bc0/packages/apply-release-plan/src/utils.ts#L38-L42

Semver seems to always see this as "unsatisfied" (presumably because it's "invalid").

You have explicit checks later for `workspace:^`-style (implicit) ranges:

https://github.com/changesets/changesets/blob/3ab4d8928ce3c4d4f4e3d1b09be91ec0a1315bc0/packages/apply-release-plan/src/version-package.ts#L75-L88

so I don't think this is intentional.

I've worked around this by having `shouldUpdateDependencyBasedOnConfig` strip `workspace:` prefix before running semver comparison, but I'm not certain this is the right fix:

https://github.com/graphile/crystal/blob/2807c8e211b97302299029b88ecb5df7426994b4/.yarn/patches/%40changesets-apply-release-plan-npm-7.0.13-14603bf9e7.patch#L8-L16

mixelburg:
Fixes #1886

## Problem

When a package uses explicit -style versions (as opposed to the shorthand , , ), the `shouldUpdateDependencyBasedOnConfig` function passes the raw range string (including the `workspace:` prefix) to `semver.satisfies`. Semver can't parse the prefix and always returns `false`, making the function treat every explicit workspace range as "out of range" and forcing unnecessary version bumps.

For example, if `pkg-a` has `"pkg-b": "workspace:^1.0.0"` in `peerDependencies` and `pkg-b` bumps from `1.0.0` to `1.0.1`, changesets would incorrectly decide the range is violated and bump `pkg-a` as well — even with `onlyUpdatePeerDependentsWhenOutOfRange: true`.

## Fix

Strip the `workspace:` prefix from `depVersionRange` before calling `semverSatisfies`. The implicit workspace aliases (`^`, `~`, `*`) are not real semver ranges, so they're handled separately (the caller already `continue`s on them).

## Changes

- `packages/apply-release-plan/src/utils.ts`: strip prefix, guard implicit aliases
- `packages/apply-release-plan/src/utils.test.ts`: new unit tests for the fixed cases


packages/apply-release-plan/src/utils.test.ts
```diff
@@ -10,23 +10,23 @@ describe("shouldUpdateDependencyBasedOnConfig", () => {
     it("returns true when version leaves the semver range", () => {
       expect(
         shouldUpdateDependencyBasedOnConfig(
-          { version: "2.0.0", type: "major" },
+          { version: "2.0.0", oldVersion: "1.0.0", type: "major" },
```
Andarist: and for that reason, I'm really not a fan of unit tests for internal stuff that is not unlikely to change ;p

I think we should move those tests outside of here to `index.test.ts`. I'll handle this later today

Andarist (approved):

PR #1889, opened 2026-03-13
Error on unsupported flags for individual CLI commands

### LINKED ISSUE #1827: CLI silently ignores unknown flags and proceeds to execute  commands
### Current Behavior
The changesets CLI silently ignores unknown flags and proceeds to execute commands, which can lead to unintended actions.

### Expected Behavior
The CLI should either:
1. Show an error message for unknown flags
2. Display help information when invalid flags are used
3. At minimum, warn about unrecognized flags

### Steps to Reproduce
1. Run changeset publish --help (non-existent flag for this subcommand)
3. Observe that the command execute changeset publish normally instead of erroring

### Example Output
```bash
$ changeset publish --help  
# Proceeds to publish instead of showing help
```

### Impact
This behavior is dangerous because:
- Users expect --help to show documentation
- Users expect unknown flags to not be silently ignored
- Silent flag ignoring can lead to accidental publishes

### Environment
- @changesets/cli version: 2.29.8

### Suggested Fix
Add flag validation that errors on unknown flags, similar to how most CLI tools handle this scenario.

mixelburg:
Fixes #1827

Currently the CLI silently ignores any flag it doesn't recognise and proceeds to run the command. This can lead to confusing behaviour — e.g. a typo in `--otp` silently results in publishing without an OTP.

This PR adds a warning printed via `@changesets/logger` whenever `mri` surfaces a parsed flag that isn't in the known set:

```
$ changeset publish --otpp 123456
🦋 warn Unknown flag: --otpp
```

Multiple unknown flags are collected and shown together:

```
$ changeset add --emptty --opeen
🦋 warn Unknown flags: --emptty, --opeen
```

The known set includes all declared boolean/string flags, their alias canonical names, the deprecated v1 flags (`updateChangelog`, `isPublic`, `skipCI`, `commit`) that already have their own error path in `run.ts`, and the built-in `help`/`version` flags.

### DIFF
--- a/.changeset/flat-zoos-share.md
+++ b/.changeset/flat-zoos-share.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+Error on unsupported flags for individual CLI commands and print the matching command usage to make mistakes easier to spot.


packages/cli/src/index.ts
```diff
@@ -44,6 +44,45 @@ if (parsed.snapshot === "" && args[args.indexOf("--snapshot") + 1] !== "") {
...
+  // built-in mri flags
+  "help",
+  "version",
+]);
+
+const unknownFlags = Object.keys(parsed).filter(
+  (key) => key !== "_" && !knownFlags.has(key)
+);
+
+if (unknownFlags.length > 0) {
```
Andarist: ugh, I'd love to just migrate away from `mri` to something that would handle such things for us, something that would also provide us with better type-safety. 

@beeequeue would you maybe have any recommendations for an alternative?

beeequeue: i've used [`sade`](https://npmx.dev/package/sade) before which worked well and is teeny-tiny (44kb), but it isn't type safe. it could be used with a little manual parsing

there's also [`cleye`](https://npmx.dev/package/cleye) which i haven't used (yet?). it's bigger (166kb) and does proper type safety and parsing.

beeequeue: i can also recommend possibly changing the config parsing to use [`valibot`](https://valibot.dev/guides/introduction/#example), it would be easier to maintain in the long run, and if we make sure to bundle it it would also be very small.

valibot is 2MB to install since it has many types and checks, but [treeshakes](https://teardown.kelinci.dev/?q=npm%3Avalibot) extremely well.

PR #1890, opened 2026-03-16
feat(cli): non-interactive mode for add + new changed command

### LINKED ISSUE #979: [New Feature] support command way to add changelog

### LINKED ISSUE #1604: [Feature Request] add non-interactive mode for `changeset add`
Thanks for providing this awesome too! 

## Problem

We want to generate Changesets for upgrades of RenovateBot, but want to reuse the change detection logic of changesets. 

### Non - interactive mode

The most native solution and reusable one. 

```shell
changeset add --packages "*" --message "This is a change"
changeset add --packages "@foo/bar" --message "This is a change"
changeset add --packages "@foo/bar" --packages "@lore/ipsum"  --message "This is a change"
```

### Create a Github Action 

This is has been done in some manor by projects, but these are specific to their repo setups. https://github.com/backstage/actions/tree/main/renovate-changesets

### Creating a wrapper cli

This would be the chosen way to move forward if this feature is not desired to be in the core. 

## Implementation 

We are interested in contributing this feature.

brunorodmoreira:
## Summary

Adds non-interactive mode to `changeset add` and introduces a new `changeset changed` command, enabling **AI agents, CI pipelines, and automation scripts** to create changesets without interactive prompts.

### Motivation

The current `changeset add` flow requires interactive input for package selection, bump type, and summary — making it unusable for AI coding assistants (Claude Code, Cursor, GitHub Copilot, etc.), CI pipelines, and automation scripts. This has been requested across multiple issues (#979, #1604, [Discussion #1118](https://github.com/changesets/changesets/discussions/1118)).

As AI-assisted development becomes the norm, tools like changesets need a **programmatic interface** so agents can handle the full commit-and-changeset workflow autonomously.

### New flags for `changeset add`

| Flag | Alias | Type | Description |
|------|-------|------|-------------|
| `--package` | `-p` | `string` (repeatable) | Package name to include in the changeset. Use multiple times for multiple packages |
| `--type` | `-t` | `patch\|minor\|major` | Bump type applied to all selected packages |
| `--all` | — | `boolean` | Auto-select all changed packages (git-aware) |

Non-interactive mode activates when `--package` (or `--all`) **and** `--message` are both provided. When any required input is missing, it gracefully falls back to the interactive flow.

### New command: `changeset changed`

```bash
changeset changed [--since <branch>] [--json]
```

Lists packages that have changed since the base branch — the same detection logic used in the first step of the interactive `add` flow, now exposed as a standalone command. This enables a two-step workflow for AI agents: **discover → create**.

## Manual Testing

Tested against a real monorepo (private, 27+ packages across `apps/`, `packages/`, `libs/`).

### `changeset changed`

```bash
$ changeset changed
🦋  Changed packages since "develop":
🦋  @ollie-shop/functions
🦋  hub
```

```bash
$ changeset changed --json
🦋  [
🦋    {
🦋      "name": "@ollie-shop/functions",
🦋      "dir": "/path/to/packages/functions"
🦋    },
🦋    {
🦋      "name": "hub",
🦋      "dir": "/path/to/apps/hub"
🦋    }
🦋  ]
```

### Non-interactive `add` — single package

```bash
$ changeset add --package @ollie-shop/functions --type patch -m "test non-interactive add"
🦋  === Summary of changesets ===
🦋  patch:  @ollie-shop/functions
🦋  Changeset added and committed
🦋  info /path/to/.changeset/tiny-radios-hide.md
```

### Non-interactive `add` — multiple packages

```bash
$ changeset add --package @ollie-shop/functions --package hub --type minor -m "test multiple packages"
🦋  === Summary of changesets ===
🦋  minor:  @ollie-shop/functions, hub
🦋  Changeset added and committed
🦋  info /path/to/.changeset/soft-nights-make.md
```

### Error: unknown package

```bash
$ changeset add --package nonexistent --type patch -m "should fail"
🦋  error The package "nonexistent" was not found in the project. Available packages: admin, builder, ...
```

### Error: missing --type

```bash
$ changeset add --package hub -m "no type"
🦋  error --type is required when using non-interactive mode
```

## AI Agent Workflow Example

```bash
# Step 1: Agent discovers what changed
CHANGED=$(changeset changed --since main --json)

# Step 2: Agent reasons about bump types based on commit history/diff

# Step 3: Agent creates the changeset
changeset add \
  --package @acme/core \
  --package @acme/cli \
  --type minor \
  -m "Add streaming support to core and CLI"
```

## Test Plan

- [x] All existing `add` command tests pass (no regressions)
- [x] New tests in `run.test.ts` for `--package` flag normalization (single → array, multiple, undefined, `--all` passthrough)
- [x] Full CLI test suite passes
- [x] Manual end-to-end testing against real monorepo (ollie-shop, 27+ packages)

## Review Changes Applied

Addressed all feedback from @Andarist:

1. **Flattened nested `describe` blocks** — moved non-interactive tests from `add/__tests__/add.ts` to `run.test.ts`
2. **Removed colon syntax** (`pkg:type`) — sticking to `--type` for simplicity
3. **Renamed `--packages` → `--package`** (singular, repeatable via `mri`)
4. **Normalized `--package` early to `string[]`** in `run.ts`
5. **Changed `error` → `log`** for "no changed packages" (valid state, not an error)
6. **Added header with resolved `since` value** before listing changed packages

Closes #979
Closes #1604


packages/cli/src/commands/add/__tests__/add.ts
```diff
@@ -604,4 +604,296 @@ describe("Add command", () => {
       ]
     `);
   });
+
+  describe("non-interactive mode", () => {
```
Andarist: nested describe blocks are evil (in my opinion :P), I also think we should add new tests for this here: https://github.com/changesets/changesets/pull/1839/changes#diff-3ee601d2eea34a9a932aa60b25e4cda51265571ba64360a2de2282482d019943 (yes, i know, that code uses nested describe blocks 😉 )

packages/cli/src/commands/add/__tests__/add.ts
```diff
@@ -604,4 +604,296 @@ describe("Add command", () => {
...
+          name: "pkg-b",
+          version: "1.0.0",
+        }),
+      });
+
+      await addChangeset(
+        cwd,
+        {
+          empty: false,
+          packages: ["pkg-a:minor", "pkg-b:patch"],
```
Andarist: this is a new syntax... while it's technically possible (and sometimes more than desired!) to have different bump types within a single changesets file... i think we should aim for the streamlined scenario for now. Let's not add this new syntax and for now stick to `--type`

packages/cli/src/index.ts
```diff
@@ -17,12 +26,16 @@ const parsed = mri(args, {
     "tag",
     "snapshot",
     "snapshotPrereleaseTemplate",
+    "packages",
```
Andarist: `mri` does seem to support this:
```
> require('mri')(['--package', 'foo', '--package', 'bar'])
{ _: [], package: [ 'foo', 'bar' ] }
```

I think, it would be better to go with this variant over `--packages`

packages/cli/src/types.ts
```diff
@@ -13,6 +13,10 @@ export type CliOptions = {
   tag?: string;
   gitTag?: boolean;
   open?: boolean;
+  packages?: string | string[];
```
Andarist: we should normalize early to `string[]`

packages/cli/src/commands/changed/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (json) {
+    const output = changedPackages.map((pkg) => ({
+      name: pkg.packageJson.name,
+      dir: pkg.dir,
+    }));
+    log(JSON.stringify(output, null, 2));
+  } else {
+    if (changedPackages.length === 0) {
+      error("No changed packages found");
```
Andarist: it's a valid situation, not an error - we should just use log here

packages/cli/src/commands/changed/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    } else {
+      for (const pkg of changedPackages) {
+        log(pkg.packageJson.name);
```
Andarist, line 34: it would be good to add some header info before this loop that would also mention the resolved "since" value

samdcoding:
@brunorodmoreira would love to see this land 🥇

beeequeue:
this PR makes two separate modifications, non-interactive mode and a new command. could these be split up into separate PRs?

we have also made major changes to the CLI code in the `next` (v3) branch, which means upstreaming this would be quite difficult.

with these two points in mind, could you split the PR and base it on `next` instead?

---

i am also not sure about having a `changed` command, as it's something that could be handled by package managers or monorepo tooling instead.

for example, pnpm already supports filtering with git:
`pnpm ls --json --depth -1 --filter '...[git ref]'` returns all packages that have changed along with their dependents, see [Filtering](https://pnpm.io/filtering#--filter-since)

this supports everything pnpm related already, something Changesets fails in some regards (#1707)

brunorodmoreira:
Surething @beeequeue . I will rework it then and let you know.

beeequeue:
closing this based on previous comment

PR #1905, opened 2026-03-27
Add documentation for automating changesets when versioning apps

ThisIsMissEm:
(or for non-npm projects)

I just used changesets for versioning a swift package, with the GitHub Action and the GitHub changelog setup, and I was stuck for a moment where the release pull request would be created, but merging it would result in the following log output:

```
Run changesets/action@v1
setting git user
/usr/bin/git config user.name "github-actions[bot]"
/usr/bin/git config user.email "<omitted>"
setting GitHub credentials
No changesets found. Attempting to publish any unpublished packages to npm
No NPM_TOKEN found, but OIDC is available - using npm trusted publishing
```

And then no github release was actually created.

This was because I didn't supply the `publish` option to the GitHub Action, which needs to invoke `changeset publish`, because I was like "I don't have a publish command for swift, since there's no package registry", so I'd both omitted it initially, and then set `publish: /usr/bin/true` and both didn't have the expected results.

You can see the final configuration I'm using for publishing releases of swift packages here: https://github.com/germ-network/oauth4swift

### DIFF
--- a/docs/versioning-apps.md
+++ b/docs/versioning-apps.md
@@ -34,3 +34,7 @@ For example, if you have an app `A` that depends on a private library `B`, you c
 ```
 
 This works because `A` is private and will never be published to npm with a stale reference to `B`.
+
+## Automated Releases
+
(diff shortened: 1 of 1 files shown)


bluwy:
We're moving the docs to the new site and I've incorporated these changes into the new docs at https://github.com/changesets/changesets/commit/72289ecae42e5a6604c5739cae77eedce651cbfe (added as co-author). I'll go ahead and close this as we want to keep the old docs with little updates for now. Thanks for contributing!

PR #1912, opened 2026-03-30
fix(@changesets/read): ignore uppercase .md files in .changeset dir

### LINKED ISSUE #1906: [FEAT]: Allow ignoring non-changeset `.md` files in `.changeset/` DIR
### Summary

Right now, every `.md` file in `.changeset/` gets parsed as a changeset, except `README.md` and `config.json`. There's no way to put other markdown files in there without the parser choking on them.

This came up because I have an `AGENTS.md` in `.changeset/` for AI coding agents (Claude Code, Codex, etc.) that need project-specific instructions. The file has its own frontmatter format, so Changesets tries to parse it and fails:

```txt
Error: could not parse changeset - invalid frontmatter: Read AGENTS.md for project knowledge.
```

I can work around it by moving the file elsewhere or rename the file to just `AGENTS` / `CLAUDE`, but `.changeset/` is the natural place for instructions about how changesets work in this project.

A few ideas for how this could work:

1. A `ignoredFiles` array in `config.json` that accepts glob patterns
2. Skipping any `.md` file that starts with an uppercase name (convention-based, like `README.md` already is)
3. Only parsing files that match the changeset naming pattern (the random adjective-noun ones)
4. Auto ignore `CLAUDE.md` / `AGENTS.md` files.

Would any of these be worth pursuing?

mixelburg:
Fixes #1906

Currently any `.md` file in `.changeset/` that isn't `README.md` gets parsed as a changeset. This causes a cryptic parse error when you put docs like `AGENTS.md` or `CLAUDE.md` in there:

```
Error: could not parse changeset - invalid frontmatter: Read AGENTS.md for project knowledge.
```

The fix extends the existing `README.md` exclusion to cover all uppercase-starting filenames. The convention is already established (README.md) and natural — changeset IDs are always lowercase adjective-noun pairs.

Changed `!/^README\.md$/i.test(file)` to `!/^[A-Z]/.test(file)`, so `AGENTS.md`, `CLAUDE.md`, `CONTRIBUTING.md`, etc. are all silently skipped.

### DIFF
--- a/.changeset/ignore-uppercase-md-files.md
+++ b/.changeset/ignore-uppercase-md-files.md
@@ -1,0 +1,5 @@
+---
+"@changesets/read": patch
+---
+
+Uppercase `.md` files in the `.changeset/` directory (such as `AGENTS.md`, `CLAUDE.md`, `CONTRIBUTING.md`) are now silently ignored, matching the existing behavior for `README.md`. This allows keeping documentation or AI agent instructions inside `.changeset/` without causing parse errors.
--- a/packages/read/src/index.test.ts
+++ b/packages/read/src/index.test.ts
@@ -47,6 +47,27 @@ Nice simple summary
       },
     ]);
   });
+  it("should ignore uppercase .md files (AGENTS.md, CLAUDE.md, etc.)", async () => {
+    const cwd = await testdir({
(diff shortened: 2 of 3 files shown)


bluwy:
superseded by #2066

PR #1936, opened 2026-04-18
fix: use minor bump for 0.x packages when peer dependency causes major

### LINKED ISSUE #1887: Adding peer dependencies bumps 0.X.Y packages to 1.0.0
If I add a peer dependency to a package it automatically get's a `major` bump. However in `0.X.Y` versions this releases `1.0.0` which is not desirable as by convention `minor` versions of `0.X.Y` typically are allowed to contain breaking changes.

mixelburg:
### What changed

When a peer dependency change triggers a major bump for a dependent package with a 0.x version, the bump type is now  instead of . This avoids prematurely jumping from 0.x.y to 1.0.0.

### Why

By semver convention, 0.x packages are allowed to contain breaking changes in minor versions. A peer dependency change that would require a major bump in a stable package should only require a minor bump in a 0.x package. Before this fix, adding a peer dependency to a 0.x package would release 1.0.0, which is not desirable.

### How

In , the  check now additionally checks if the dependent package's current version starts with . If so,  is set to  instead of .

Closes #1887

### DIFF
--- a/packages/assemble-release-plan/src/determine-dependents.ts
+++ b/packages/assemble-release-plan/src/determine-dependents.ts
@@ -89,7 +89,13 @@ export default function determineDependents({
                     .onlyUpdatePeerDependentsWhenOutOfRange,
               })
             ) {
-              type = "major";
+              // For 0.x versions, peer dependency major bump should be minor
+              // since semver allows breaking changes in minor versions for 0.x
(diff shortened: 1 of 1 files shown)


mixelburg:
Closing due to inactivity. Over 2 months with no review.

Issue #1939, opened 2026-04-23
`--snapshot` flag no longer recognized in v2.31.0


rtkac:
🐛 Bug Report

When running the publish command with the `--snapshot` flag in @changeset/cli version 2.31.0, the CLI throws an error indicating that the flag is unknown.

🔍 Steps to Reproduce

Run the following command:
`changeset publish --tag next --no-git-tag --snapshot`

❗ Current Behavior

The command fails with:
```javascript
🦋  error Unknown flag for publish: --snapshot
🦋  error Usage: changeset publish [--tag <name>] [--otp <code>] [--no-git-tag]
```

✅ Expected Behavior

The --snapshot flag should be recognized and work as it did in previous versions.

🧪 Working Version

This command works as expected in:
`@changeset/cli@2.30.0`

💥 Broken Version

This issue occurs in:
`@changeset/cli@2.31.0`

Andarist:
To be fair, that flag *didn't* work in this context, right? So far we only have this single report about this being a problem and I assume you already fixed this on your side... so I'm inclined to close the issue as "working as intended".

rtkac:
Actually I had to revert it back to the previous version `2.30.0` to make it work. Could you please try `--snapshot` flag with `2.31.0` version? I don' think its "working as intended". Thank you 🙏

bluwy:
@rtkac Andarist means that `--snapshot` never worked on the `changeset publish` command, it's for `changeset version`. You can just remove that flag. We had discussed this and decided to not fix this.

Andarist:
If it doesn't work with `changeset version` though - please let us know, that would be a critical bug. I have just retested this manually and it does seem to work though.

Issue #1938, opened 2026-04-22
`changeset publish --no-git-tag` still creates git tags for private packages


MaximusHaximus:
Hello folks!! Thanks for the awesome toolset -- this is by far my favorite path for managing releases 💖.  I think I found a bug that I'd love to see fixed; I've worked around it with some hacky manual scripting but I think the functionality is just broken in its current form so thought I'd raise an issue here for feedback and hopefully resolution :).

## Summary

The `--no-git-tag` flag on `changeset publish` is advertised as suppressing git tag creation during publish, but it only suppresses tags for packages that were published to npm. When `config.privatePackages.tag: true`, `publish` continues to create annotated git tags for untagged private packages even when `--no-git-tag` is passed. The flag is non-functional for any repo that uses the built-in private-package tagging feature.

## Reproduction

Monorepo config:

```jsonc
// .changeset/config.json
{
  "privatePackages": { "version": true, "tag": true }
  // ...
}
```

A single private package at `0.1.0`, add a minor changeset, `changeset version`, then:

```
$ changeset publish --no-git-tag
🦋  No unpublished projects to publish
🦋  success found untagged projects:
🦋  pkg-a@0.2.0
$ git tag -l
pkg-a@0.2.0
```

Expected: no tags created.
Actual: `pkg-a@0.2.0` is created locally as an annotated tag.

## Root cause

In [`packages/cli/src/commands/publish/index.ts`](https://github.com/changesets/changesets/blob/%40changesets/cli%402.31.0/packages/cli/src/commands/publish/index.ts) at v2.31.0 (unchanged on `main`):

```ts
export default async function publish(
  cwd,
  { otp, tag, gitTag = true }: { ...; gitTag?: boolean },
  config
) {
  // ...
  const tagPrivatePackages = config.privatePackages && config.privatePackages.tag;
  // ...
  if (gitTag) {
    await tagPublish(tool, successfulNpmPublishes, cwd);   // public: guarded
  }
  // ...
  if (untaggedPrivatePackageReleases.length > 0) {
    await tagPublish(tool, untaggedPrivatePackageReleases, cwd);   // private: NOT guarded
  }
}
```

[...]

PR #1941, opened 2026-04-23
Fix publish flag handling: --snapshot rejected and --no-git-tag ignored for private packages

mixelburg:
## Summary

Two related fixes for the `changeset publish` command, both stemming from CLI flag handling issues.

### 1. `--snapshot` flag incorrectly rejected by `changeset publish` (#1939)

**Regression from #1889.** The unknown-flag validation added in 2.31.0 rejects `--snapshot` when passed to `changeset publish`, even though it was silently accepted in 2.30.0.

The flag is a no-op for publish, but users commonly pass shared flags when chaining commands:
```sh
changeset version --snapshot && changeset publish --snapshot
```

**Fix:** Destructure `snapshot` from flags in the publish case so it's not passed to `validateCommandFlags` as an unknown flag.

### 2. `--no-git-tag` still creates git tags for private packages (#1938)

When `privatePackages.tag: true`, `changeset publish --no-git-tag` still creates git tags for untagged private packages because the `gitTag` guard only wraps the public package tagging path, not the private one.

```ts
if (gitTag) {
  await tagPublish(tool, successfulNpmPublishes, cwd);   // public: guarded ✓
}
// ...
await tagPublish(tool, untaggedPrivatePackageReleases, cwd);  // private: NOT guarded ✗
```

**Fix:** Wrap the private package `tagPublish` call in the same `if (gitTag)` guard.

## Changes

- `packages/cli/src/run.ts`: Add `snapshot` to the publish case destructuring
- `packages/cli/src/commands/publish/index.ts`: Guard private package tagging with `if (gitTag)`
- Changesets for both fixes

## Tests

All existing tests pass (157 passed, 1 skipped).

### DIFF
--- a/packages/assemble-release-plan/src/determine-dependents.ts
+++ b/packages/assemble-release-plan/src/determine-dependents.ts
@@ -89,7 +89,13 @@ export default function determineDependents({
                     .onlyUpdatePeerDependentsWhenOutOfRange,
               })
             ) {
-              type = "major";
+              // For 0.x versions, peer dependency major bump should be minor
+              // since semver allows breaking changes in minor versions for 0.x
+              if (dependentPackage.packageJson.version.startsWith("0.")) {
+                type = "minor";
+              } else {
+                type = "major";
+              }
             } else if (
               (!releases.has(dependent) ||
                 releases.get(dependent)!.type === "none") &&
--- a/packages/assemble-release-plan/src/determine-dependents.ts
+++ b/packages/assemble-release-plan/src/determine-dependents.ts
@@ -100,7 +100,13 @@ export function determineDependents({
                     .onlyUpdatePeerDependentsWhenOutOfRange,
               })
             ) {
-              type = "major";
+              // For 0.x versions, peer dependency major bump should be minor
+              // since semver allows breaking changes in minor versions for 0.x
+              if (dependentPackage.packageJson.version.startsWith("0.")) {
+                type = "minor";
+              } else {
+                type = "major";
+              }
             } else if (
               (!releases.has(dependent) ||
                 releases.get(dependent)!.type === "none") &&
--- a/packages/cli/src/commands/publish/index.ts
+++ b/packages/cli/src/commands/publish/index.ts
@@ -107,7 +107,9 @@ export default async function publish(
     success("found untagged projects:");
     logReleases(untaggedPrivatePackageReleases);
 
-    await tagPublish(tool, untaggedPrivatePackageReleases, cwd);
+    if (gitTag) {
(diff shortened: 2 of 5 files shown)


packages/assemble-release-plan/src/determine-dependents.ts
```diff
@@ -100,7 +89,13 @@ export default function determineDependents({
...
               })
             ) {
-              type = "major";
+              // For 0.x versions, peer dependency major bump should be minor
+              // since semver allows breaking changes in minor versions for 0.x
+              if (dependentPackage.packageJson.version.startsWith("0.")) {
+                type = "minor";
+              } else {
+                type = "major";
+              }
```
Andarist, line 109: This is unrelated change, please revert it

packages/cli/src/run.ts
```diff
@@ -222,7 +222,7 @@ export async function run(
         return;
       }
       case "publish": {
-        const { otp, tag, gitTag, ...rest }: CliOptions = flags;
+        const { otp, tag, gitTag, snapshot, ...rest }: CliOptions = flags;
```
Andarist: I don't think we want to fix this, the usage was incorrect and accidentally accepted. So far we only have a single user report about this. So I'd revert this from here.

packages/cli/src/commands/publish/index.ts
```diff
@@ -107,7 +107,9 @@ export default async function publish(
     success("found untagged projects:");
     logReleases(untaggedPrivatePackageReleases);
 
-    await tagPublish(tool, untaggedPrivatePackageReleases, cwd);
+    if (gitTag) {
+      await tagPublish(tool, untaggedPrivatePackageReleases, cwd);
+    }
```
Andarist: I'm double-checking - this isn't related to a recent regression, right?

beeequeue:
this seems to be a bot opening PRs randomly without oversight

https://github.com/storybookjs/storybook/pull/34453#issuecomment-4184340711

image

Filipoliko:
Note that rejecting the `--snapshot` flag on `changeset publish` command was a breaking change released as a minor update.  Although it was a noop until now, it was a documented usage of snapshot feature until 27.5.2025 https://github.com/changesets/changesets/pull/1578

We ran into the same problem in our CIs, which broke after a package-lock.json update. Fix is quite simple (removing the --snapshot parameter from publish command), so we patched everything up now and moved on.

mixelburg:
Closing this PR as the maintainer has indicated the changes are not desired (the usage was intentionally unsupported). Thanks for the review and feedback!

PR #1947, opened 2026-04-30
Use parseEnv instead of dotenv and avoid side effects

bluwy:
https://nodejs.org/api/process.html#processloadenvfilepath exists since the recent required nodejs version bump. We no longer need to use and upgrade the dotenv dependency.

close #1931


.changeset/late-animals-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/changelog-github": patch
+---
+
+Use `loadEnvFile` instead of `dotenv` to load the `.env` file
```
beeequeue:
```suggestion
Replaced `dotenv` with `loadEnvFile` for loading `.env` files
```

i think generally changesets should use past tense, as they are a log of what _has changed_ in a package, compared to commit descriptions that use present tense as they describe what a commit _does_ or _will do_ to a repository

packages/changelog-github/src/index.ts
```diff
@@ -1,8 +1,15 @@
+import { loadEnvFile } from "node:process";
 import type { ChangelogFunctions } from "@changesets/types";
-import { config } from "dotenv";
 import { getInfo, getInfoFromPullRequest } from "@changesets/get-github-info";
 
-config();
+try {
+  loadEnvFile();
```
beeequeue: should we also add `empathic` to check for `.env` in any parent folders? i think that is the behavior of `dotenv` but am not sure

beeequeue (changes requested):

packages/changelog-github/src/index.ts
Andarist: looking at `dotenv`'s implementation, it doesnt search parent directories - so I think this is fine and would match preexisting behavior+expectations

beeequeue: then it's fine :)

_apparently i can't resolve my own comments :D_

packages/changelog-github/src/index.ts
```diff
@@ -1,8 +1,15 @@
+import { loadEnvFile } from "node:process";
```
Andarist: I have a semi-strong beef with unexpected side effects. This isn't new here, of course, but given we touch this now - what do you think about using `util.parseEnv` instead of this? It would be a couple of LOC more but I think it would be nice to avoid the side-effect and do this lazily.

beeequeue: that's true, it shouldn't effect the rest of the code

Andarist: I pushed out that change, let me know what do you think about this.

bluwy: Yeah I agree too. Personally I wouldn't want a package to implicitly read a `.env` file altogether though but we can think of the more breaking change in the future. The change now shouldn't be a big deal.

.changeset/late-animals-drive.md
bluwy: I've always used present tense here as they work like git commits to me. It depends how you read it but they could also describe what a specific version does. I don't think we need to standardize the format for now though.

bluwy:
@Andarist Should we cache and only load the env files once? From the code paths it looks like it could be called repeatedly

.changeset/late-animals-drive.md
beeequeue: it would be nice to have one before the v3 release to make it look nicer though :)

Andarist:
ye, that's a good idea

Andarist (approved):

.changeset/late-animals-drive.md
Andarist: we should go through the final changesets/changelog updates manually before releasing final v3 anyway, so we should be able to adjust the lanaguage then

packages/get-github-info/README.md
```diff
@@ -9,25 +9,23 @@
 
 > Note: This assumes you already have changesets setup.
 
-To use `@changesets/get-github-info`, you'll need to install it and you'll probably also want `dotenv` to provide a GitHub personal access token via a `.env` file.
+To use `@changesets/get-github-info`, you'll need to install it.
```
beeequeue: one of the statements of all time!

bluwy: Ha, let's get rid of the install part altogether then like the other readmes

beeequeue (approved):

packages/changelog-github/src/index.ts
Andarist: Yeah, I think `.env` support is kinda esoteric here. OTOH, it's pretty cheap so 🤷 but I doubt that many people rely on this. It was introduced when there was only one env variable supported too (`GITHUB_TOKEN`)

PR #1953, opened 2026-05-04
replace `micromatch` with `picomatch`

beeequeue:
currently stuck on the default config no longer parsing


packages/config/src/index.ts
```diff
@@ -12,7 +12,7 @@ import type {
   WrittenConfig,
 } from "@changesets/types";
 import { getPackages } from "@manypkg/get-packages";
-import micromatch from "micromatch";
+import picomatch from "picomatch";
```
Andarist, line 15: I wonder, how does it compare to the builtin? https://nodejs.org/docs/latest-v22.x/api/fs.html#fsglobpattern-options-callback

beeequeue: picomatch _wayyyy_ faster iirc as the [builtins use `minimatch`](https://github.com/nodejs/node/blob/59272da055883898bd40b9dcac57b98aae6e1a1b/lib/internal/fs/glob.js#L55-L57)

but i don't have any benchmarks saved so i'll try to find or make some

i also mentioned this on [bsky](https://bsky.app/profile/sxzz.dev/post/3mks4osp7ss2o) recently :^)

bluwy:
I updated `getUnmatchedPatterns` that wasn't migrated quite right, and I updated the matching to handle this specific behavior from micromatch:

```ts
import picomatch from 'picomatch'
import micromatch from 'micromatch'

const pkgs = ['pkg-a', 'pkg-b', '@scope/pkg-c', '@scope/pkg-d']
const globs = ['pkg-*', '!pkg-b', '@scope/*']

const matcher = picomatch(globs)
console.log('picomatch:', pkgs.filter((pkg) => matcher(pkg)))
// => [ 'pkg-a', 'pkg-b', '@scope/pkg-c', '@scope/pkg-d' ]

console.log('micromatch:', micromatch(pkgs, globs))
// => [ 'pkg-a', '@scope/pkg-c', '@scope/pkg-d' ]
```

The negation is order-sensitive. picomatch simply returns true if any pattern returns true.

packages/config/src/index.ts
bluwy: We need to use this instead so we don't touch the fs: https://nodejs.org/api/path.html#pathmatchesglobpath-pattern

Open to trying it though but we still kinda need a wrapper to handle array of files and patterns.

Andarist:
Hm, shouldn't we inherit picomatch semantics though? We do mention what matcher library gets used under the hood - so people can reason about their glob patterns etc. Diverging from that could feel surprising.

beeequeue:
i would prefer leaving the behavior to picomatch, less code for us to worry about

packages/config/src/index.ts
Andarist: It would be nice to know the benchmark results. However, that wouldn't really be a bottleneck in the case of Changesets - so we could just favor less dependencies over a more performans library here.

bluwy:
If we leave it entirely to picomatch, they'd have to craft every pattern to not include it, like `"pkg-!(a)"` instead to get the negation pattern work like before. I feel like this will be quite difficult to migrate though and not worth it. For reference, [here's how](https://github.com/micromatch/picomatch/blob/69a2d181b9c0c7bc5454eb2e188b4763427e9000/lib/picomatch.js#L32-L42) picomatch implements array support.

Should we consider documenting instead? I also think the link to picomatch also mostly refers to its pattern syntax than the array matching behaviour.

Andarist:
If we'd decide to use the builtin node API then this discussion would become obsolete :p I'm fine either way, so I'll leave you both to decide if we should use the node API or picomatch.

bluwy:
If `path.matchesGlob` works and is compatible, I think we should use it. I'm just a bit lazy to check and update it 😄

bluwy (approved):

Issue #1949, opened 2026-05-03
replace picocolors with node:util styleText


ghostdevv:
It looks like minimum node version for v3 will be 20.19 which means we could remove a dependency and instead use the built in [`node:util#styleText`](https://nodejs.org/api/util.html#utilstyletextformat-text-options) fn - there is only one bug that isn't patched until v22 which is that [nested styleText calls won't colour correctly](https://github.com/nodejs/node/issues/59035), but that's only issue for the codebase and _shouldn't_ effect consumers of changesets, so could still be worth it.

Happy to submit a PR if this is wanted!

Follow on from #1414

Andarist:
cc @beeequeue

beeequeue:
imo we can bump the minimum node version to 22 since 20 became EoL this weekend.

my opinion on `styleText` is mixed, and i think it should wait until after #1879, but my opinions are not strong enough to stop anyone from implementing it :)

it is built-in 👍
i strongly dislike the api 👎 (can be fixed with a small wrapper)
it's quite a bit slower 👎 (but not noticably iirc)

ghostdevv:
Happy to wait, and we can definitely add a small wrapper to make it more like picocolors. I didn't realise it was slower, that's a shame, but yea I don't think it's noticeable and so it's still likely worth it IMO.

bluwy:
Closed by https://github.com/changesets/changesets/pull/2004

PR #1954, opened 2026-05-04
bump lowest supported node version to v22, update vitest

beeequeue:
- sets the supported node versions to `^22 || >=24` as they are the non-EoL versions and are what we test for
- changes the scripts to use `node --run` where possible and reorders them
- updates vitest since i've gotten tired of it using multiple (^4.0.18 and ^4.1.2) :^)

part of #1949


.changeset/free-results-love.md
bluwy, line 1: I think we need a new changeset otherwise it wont trigger a re-release of the next versions. I also prefer if we can describe the changes between pre-releases so it's easier to audit later.

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        node_version: [20, 22, 24]
+        node_version: [22, 24, 25]
```
bluwy, line 24: I think we don't need to test 25 here. It's hard to guarantee odd versions to work.

packages/cli/package.json
```diff
@@ -73,9 +73,9 @@
     "@types/semver": "^7.7.1",
     "human-id": "^4.1.1",
     "outdent": "^0.8.0",
-    "vitest": "^4.1.2"
+    "vitest": "4.1.5"
```
bluwy: Looks like this got pinned somehow. I think we should keep the `^`. `@vitest/coverage-v8` is also pinned.

.changeset/free-results-love.md
beeequeue: true, we can remove the duplicate node version changesets later before releasing v3

.github/workflows/ci.yml
Andarist: The small benefit of testing 25 would be that we could file upstream issues if anything breaks. I'm fine either way.

.changeset/free-results-love.md
```diff
@@ -21,4 +21,4 @@
 "@changesets/pre": major
 ---
 
-Bumps minimum node version to `>=20.19.0`
+Bumped supported Node versions to `^22 || >=24`
```
Andarist: Hm, while we won't be super cautious about node v23... won't it just be kinda annoying for people on that version to have it outside of the supported range?

.github/actions/ci-setup/action.yml
```diff
@@ -4,7 +4,7 @@ inputs:
   node-version:
     description: "Node.js version"
     required: false
-    default: 24
+    default: 22
```
Andarist, line 7: The CI runs on already on 22 because we need to pass a whole matrix there. IIRC 24 comes with modern npm that supports trusted publishing, and that's why I put 24 as the default here (if my memory serves me well)

.github/workflows/ci.yml
beeequeue: i think it's fine if we think of the support policy as "non-EoL versions". then we will drop 25 as soon as 26 comes out so it won't be supported for too long

another small detail is that node is moving to a [once-per-year, every-version-LTS](https://nodejs.org/en/blog/announcements/evolving-the-nodejs-release-schedule) release schedule with v27, so we won't have to worry about it after v25 :)

Andarist: > another small detail is that node is moving to a [once-per-year, every-version-LTS](https://nodejs.org/en/blog/announcements/evolving-the-nodejs-release-schedule) release schedule with v27, so we won't have to worry about it after v25 :)

Yeah, this is quite nice.

packages/cli/package.json
beeequeue: this is a personal preference that snuck in, there's no reason for non-pinned dev deps (since they get pinned by the lockfile) so i never use ranges for them

maybe it should wait for a proper decision and switching everywhere though

.changeset/free-results-love.md
beeequeue: while it may be annoying i also don't think it should be supported for the same reason we don't support v20; it's EoL!

.github/actions/ci-setup/action.yml
beeequeue: i generally think one should run the CI on the minimum required node version to make sure it works properly.

we can manually install the latest npm version before publishing to fix this issue (`npm install -g npm`)

up to you

.changeset/free-results-love.md
Andarist: Ah, OK - I have not realized this one is already EOL. I'm fine with it then.

.github/actions/ci-setup/action.yml
Andarist: feel free to add that extra installation step before publish

.changeset/free-results-love.md
beeequeue: the rules are:
1. even versions are supported for multiple years.
2. odd versions are supported until there is a higher, even version

so 25 is going to be EoL when 26 releases, currently planned for the end of April :)

.github/actions/ci-setup/action.yml
beeequeue: turns out the publish workflow has [its own version specified](https://github.com/changesets/changesets/blob/aab48fc6c87bc4be24b6b15f2c1b96d8ac59fd37/.github/workflows/publish.yml#L31-L33) and i missed updating it, so we're already good :^)

.changeset/free-results-love-2.md
```diff
@@ -0,0 +1,24 @@
+---
+"@changesets/get-version-range-type": minor
```
Andarist: we can probably just mark all of those as major, that would at least ensure in v3 all of those will move out of the quirky 0.x versioning scheme

Andarist (approved):

packages/cli/package.json
bluwy: I would prefer we stick to the conventions of the codebase and discuss this separately.

bluwy (approved):

beeequeue:
i still can't merge :^)

PR #1956, opened 2026-05-04
Drop the legacy compatibility shim in `assembleReleasePlan` that accepted older `config` and `snapshot` argument shapes

Andarist:
followup to https://github.com/changesets/changesets/pull/1652

### DIFF
--- a/.changeset/green-pianos-sneeze.md
+++ b/.changeset/green-pianos-sneeze.md
@@ -1,0 +1,4 @@
+"@changesets/assemble-release-plan": major
+---
+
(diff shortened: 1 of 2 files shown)


beeequeue (approved):

.changeset/green-pianos-sneeze.md
```diff
@@ -1,4 +1,3 @@
-"@changesets/assemble-release-plan": major
----
```
beeequeue: this doesn't seem correct

Andarist: good catch, thanks!

beeequeue (approved):

PR #1957, opened 2026-05-04
Use fs-fixture for test fixtures

bluwy:
Compared to fixturez, it's up-to-date, much smaller, has types, and support the file tree syntax that we had to manually implement.

Also, I added a Vitest hook to cleanup the fixture directory when the test ends.


beeequeue:
funnily enough i spent the weekend making [tinyfixturez](https://npmx.dev/package/tinyfixturez) since i got annoyed at looking at the dependencies of fixturez 😆

i don't mind if we don't use it though, it was fun to make :^)

beeequeue (approved):

bluwy:
Ha it's good to have more options. Personally I've been using fs-fixture in my projects and happen to stumble on this in the repo

Andarist:
Nice!

PR #1958, opened 2026-05-04
migrate to pnpm

beeequeue:


.github/actions/ci-setup/action.yml
```diff
@@ -9,20 +9,33 @@ inputs:
...
+      shell: bash
+      run: echo "path=$(pnpm store path)" >> "$GITHUB_OUTPUT"
+
+    - name: Cache dependencies
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      with:
+        path: ${{ steps.cache.outputs.path }}
+        key: v1-pnpm-${{ runner.os }}-${{ hashFiles('**/pnpm-lock.yaml') }}
+        restore-keys: |
+          v1-pnpm-${{ runner.os }}-
```
bluwy: I believe setup-node already handles this?

beeequeue: sadly no, their implementation of the caching is terrible and will barely hit any runs :(

ghostdevv: > I believe setup-node already handles this?

yea! https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

ghostdevv: > sadly no, their implementation of the caching is terrible and will barely hit any runs :(

oh?

beeequeue: technically, we should have a time-dependant part of the cache key to make sure it refreshes every X time or the cache might balloon in size forever, but i can add that later

bluwy: > sadly no, their implementation of the caching is terrible and will barely hit any runs :(

Are you sure? I've not hit this before and seen that it works. I would prefer simplifying this setup.

beeequeue: their recovery keys only match if the [lockfile matches exactly](https://github.com/actions/setup-node/blob/main/src/cache-restore.ts#L43-L60), so as soon as any dependency changes the entire cache will miss - e.g. every dependabot or renovate PR

bluwy: Isn't that good so that the old unused dependencies will not be part of the cache anymore? Otherwise it might have a mix of old and new dependencies when we update deps, and it cumulates.

beeequeue: that's what i meant by "ballooning in size", which can be mitigated by adding something like the current month to the cache key

bluwy: I think we should leave it all to setup-node for now. There's tradeoffs with both techniques and I'd prefer the one built-in unless we feel the time-based cache is worth the effort.

beeequeue: all right 👍

bluwy (approved):

PR #1961, opened 2026-05-04
eslint update followup 1 & allow custom functions to be sync

beeequeue:
follow up to #1876 

the only functional change here is allowing sync functions in `CommitFunctions` and `ChangelogFunctions`.

async is still supported, but the default functions are now sync which could be a breaking change which is why i marked them as `major`

### DIFF
--- a/.changeset/some-papayas-bet.md
+++ b/.changeset/some-papayas-bet.md
@@ -1,0 +1,5 @@
+---
+"@changesets/changelog-git": major
+---
+
+`ChangelogFunctions` can now be both sync and async, and the `defaultChangelogFunctions` are now sync.
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -131,13 +131,13 @@ export type GetReleaseLine = (
   changeset: NewChangesetWithCommit,
   type: VersionType,
   changelogOpts: null | Record<string, any>,
-) => Promise<string>;
+) => string | Promise<string>;
 
 export type GetDependencyReleaseLine = (
   changesets: NewChangesetWithCommit[],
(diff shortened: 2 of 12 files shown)


packages/types/src/index.ts
```diff
@@ -131,13 +131,13 @@ export type GetReleaseLine = (
...
   changelogOpts: null | Record<string, any>,
-) => Promise<string>;
+) => string | Promise<string>;
 
 export type GetDependencyReleaseLine = (
   changesets: NewChangesetWithCommit[],
   dependenciesUpdated: ModCompWithPackage[],
   changelogOpts: any,
-) => Promise<string>;
+) => string | Promise<string>;
```
Andarist: We could use a local helper here
```ts
type MaybePromise<T> = Promise<T> | T
```

.changeset/some-papayas-bet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/changelog-git": major
+---
+
+`ChangelogFunctions` can now be both sync and async, and the `defaultChangelogFunctions` are now sync.
```
Andarist, line 5: "can now be both sync and async" has this truly changed? perhaps just in the types?

beeequeue: i would consider the types to be part of the API contract

Andarist: Fair, we could call out here that it's a type-only change though. Although, I guess TS is so prevalent that it doesn't matter ;p

Andarist (approved):
This has some CI issues - but the change is fine so I'm stamping it preemptively

bluwy (approved):

PR #1962, opened 2026-05-04
Replace release-utils markdown parsing with regex

bluwy:
And remove the `mdast-*` packages. This should also be more efficient as it doesn't have to parse or iterate through the entire changelog text.

It's probably bad rep to use regex to parse but I think this should work fine and could try.


beeequeue:
i would prefer using md4x if it doesn't have any major downsides, just to keep our code simpler

it would most likely be easier to work with if we add stuff like the semantic change types idea

bluwy:
It's really not any more complex, just different ways of extracting the data we want. If the semantic change idea is easier to work as an AST, then it can always resort to that later.

beeequeue (commented):
could you copy some (large?) changelogs from users of Changesets and use them as fixtures for some extra tests, just for extra safety?

also needs a rebase

beeequeue (approved):

bluwy:
I don't think it's great to keep large files for the tests, but I just manually tested with our changelog and Svelte's changelog and they all work correctly though

beeequeue:
we could zstd compress them :^)

but that's good enough for me

beeequeue (approved):

bluwy:
I'm going to merge this in. Happy to revert if we have concerns or if people are hitting issues with this.

PR #1964, opened 2026-05-04
fix: allow unmatched ignore globs

### LINKED ISSUE #1794: Changeset fails on ignore pattern when there's no matches

IgorGanapolsky:
Fixes #1794.

## Summary
- allow unmatched glob expressions in the `ignore` config option
- keep exact missing package names as validation errors
- add a regression test for an unmatched ignore glob expanding to an empty ignore list
- add a changeset for `@changesets/config`

This supports generated package folders such as `prisma-client-*` that may exist locally after install/postinstall but not in a release automation checkout.

## Verification
- `yarn jest packages/config/src/index.test.ts --runInBand`
- `yarn types:check`
- `yarn eslint packages/config/src/index.ts packages/config/src/index.test.ts --ext .ts`
- `yarn prettier --check packages/config/src/index.ts packages/config/src/index.test.ts .changeset/quiet-globs-ignore.md`

### DIFF
--- a/.changeset/quiet-globs-ignore.md
+++ b/.changeset/quiet-globs-ignore.md
@@ -1,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
+Allow unmatched glob patterns in the `ignore` config option.
--- a/packages/config/src/index.test.ts
+++ b/packages/config/src/index.test.ts
@@ -718,11 +718,9 @@ describe("parser errors", () => {
     `);
   });
   test("ignore package that does not exist (using glob expressions)", () => {
-    expect(() => unsafeParse({ ignore: ["pkg-*"] }, defaultPackages))
-      .toThrowErrorMatchingInlineSnapshot(`
-      "Some errors occurred when validating the changesets config:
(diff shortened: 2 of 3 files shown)


ssrisuwrrn455-maker:
https://github.com/changesets/changesets/blob/18e1661e3ad9bc4ab7bcd770ae172c560af44acc/packages/changelog-github/src/index.ts#L10

beeequeue:
thank you for the PR!

this is something we are considering for the next major version, and we have a branch (`next`) we are using to work on it.

could you re-open a PR against that branch instead?
it will require a few changes due us migrating from `micromatch` to `picomatch`, so let us know if you need help!

PR #1969, opened 2026-05-04
Add a named export that mirrors the current default export

### LINKED ISSUE #1951: Use named exports for all packages with a default export
Follow up to [Default exports vs named ones Discord thread](https://discord.com/channels/1499011163705573470/1499693371487289345)

We should move into using named exports across the whole codebase.

To avoid breaking consumers, we’ll keep the existing default exports for now but mark them as deprecated. 
These default exports will be removed in the next major release (v4).

> [!NOTE]
> CJS compatibility should not be a problem as changeless is a cli tool and usually is not bundled
>
> Adding ESM module log from a `require()` call for reference:
>
> `export { foo }` -> logs `{ foo: [Function: foo] }`
> `export default foo` -> logs `{ __esModule: true, default: [Function: foo] }`
> `export { foo }; export default foo` -> logs `{ __esModule: true, default: [Function: foo], foo: [Function: foo] }`

marcalexiei:
- Closes #1951

> [!NOTE]
> Internal `default`  export has been removed and only named exports are available


bluwy (approved):

marcalexiei:
Rebased due to conflicts coming from:
- #1958 
- #1963
- #1961

packages/cli/src/changelog.ts
```diff
@@ -1 +1,3 @@
+// todo understand purpose
```
Andarist: There is not much to it - it just reexports the default so everything stays exposed nicely and the default is also structured in the same way as a possible custom module (so it acts as a reference implementation).

Andarist:
```suggestion
```

Andarist (approved):

PR #1972, opened 2026-05-05
Restrict `publish` job to the `npm` env

Andarist:


.github/workflows/publish.yml
```diff
@@ -5,144 +5,54 @@ on:
...
-  #     - name: Add result reaction
-  #       if: always() && steps.report_in_progress.outcome == 'success'
-  #       run: |
-  #         gh api /repos/${{github.repository}}/issues/comments/${{github.event.comment.id}}/reactions \
-  #           -f content="${{ job.status == 'success' && 'rocket' || '-1' }}"
-  #       env:
-  #         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  publish:
+    name: Publish
+    if: needs.version.outputs.hasChangesets == 'false'
```
Andarist, line 39: This is a little bit tricky. The previous job has 2 purposes:
- mode selection (version vs publish)
- eagerly executing the version mode

I think a cleaner design would be to introduce `changesets/action/select-mode` subaction - but that has to be implemented separately.

.github/workflows/publish.yml
```diff
@@ -5,144 +5,52 @@ on:
...
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/ci-setup
         with:
           node-version: 24
 
-      - name: Create Release Pull Request or Publish to npm
+      - name: Publish to npm
         # https://github.com/changesets/action
         uses: changesets/action@6a0a831ff30acef54f2c6aa1cbbc1096b066edaf # v1
```
emmatown, line 53: why use the changesets action here it's effectively doing nothing, just do `yarn release` + probably pushing tags?

Andarist: Primarily for dogfooding purposes. It also ends up being responsible for:
- proper monorepo support
- tagging
- github releases

At the very least, the last thing can't be done inline as easily here.

beeequeue: i think this is what we're going to want to do in the long run, let the PMs do the publishing the way they need to, then they can also handle their own special cases like workspaces, catalogs, etc.

as for right now, i would want to test running `yarn publish` or `pnpm -r publish` to see what it looks like before using it here

beeequeue (commented):
needs to be rebased to use pnpm and `node --run`

Andarist:
@beeequeue this targets the `main` branch so `node --run` shouldn't be used here (yet), right?

.github/workflows/publish.yml
emmatown: > Primarily for dogfooding purposes. It also ends up being responsible for:
> - proper monorepo support
> - tagging
> - github releases

right, it might be good to move that to the cli or something but makes sense.

> as for right now, i would want to test running yarn publish or pnpm -r publish to see what it looks like before using it here

To be clear I wasn't talking about using yarn/pnpm publish, I still meant using `changeset publish` via  the `release` package.json script we have, not `yarn publish`, I was just talking about the use of the action here.

emmatown (approved):

beeequeue:
oh, my bad

PR #1975, opened 2026-05-05
eslint followup 3: imports

beeequeue:
let me know what you think of the sort order

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -77,7 +77,7 @@ export default defineConfig(
       "n/no-process-exit": "off",
       "n/no-unpublished-import": "off",
 
-      "n/prefer-node-protocol": "off", // TODO enable and fix errors
+      "n/prefer-node-protocol": "error",
       "n/no-unsupported-features/node-builtins": [
         "error",
         { ignores: ["fs/promises.cp", "import.meta.dirname"] },
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -98,6 +98,7 @@ export default defineConfig(
         {
           internalPattern: ["^[@~#]+/"],
           newlinesBetween: 0,
+          partitionByComment: "^\\[keep-order\\]",
           groups: [
             "builtin",
             "external",
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,10 +1,11 @@
-import { defineConfig } from "eslint/config";
 import js from "@eslint/js";
-import node from "eslint-plugin-n";
-import tseslint from "typescript-eslint";
 import vitest from "@vitest/eslint-plugin";
 import eslintConfigPrettier from "eslint-config-prettier/flat";
 import importLite from "eslint-plugin-import-lite";
+import node from "eslint-plugin-n";
+import perfectionist from "eslint-plugin-perfectionist";
+import { defineConfig } from "eslint/config";
+import tseslint from "typescript-eslint";
 
 export default defineConfig(
   {
@@ -21,6 +22,7 @@ export default defineConfig(
     plugins: {
       js,
       node,
+      perfectionist,
       tseslint,
       vitest,
     },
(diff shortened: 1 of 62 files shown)


scripts/test-utils/src/index.test.ts
```diff
@@ -1,7 +1,9 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { silenceLogsInBlock } from "./index.ts"; // this has to be imported before `@changesets/logger`
+
 import { log } from "@changesets/logger";
 
+import { silenceLogsInBlock } from "./index.ts"; // this has to be imported before `@changesets/logger`
```
Andarist: this got misplaced, according to the comment

beeequeue: thanks, I was looking for the problem before going to lunch

eslint.config.mjs
```diff
@@ -69,14 +77,31 @@ export default defineConfig(
...
       ],
 
       "import-lite/no-default-export": "error",
       "import-lite/no-mutable-exports": "off", // TODO enable and fix errors
+
+      "perfectionist/sort-imports": [
+        "error",
+        {
+          internalPattern: ["^[@~#]+[a-zA-Z0-9-]*/"],
+          newlinesBetween: 1,
```
bluwy: I personally prefer to not have a new line as it only takes up more space. Usually I'd use a new line two separate sorting groups, e.g. a CSS import needs to come first or something before the other imports. Kinda like https://perfectionist.dev/rules/sort-imports#partitionbynewline

Also, I'm not sure we need the `internalPattern`. I think it's ok to mix them.

beeequeue: i personally think it helps a lot with finding the imports i'm looking for when the lines are there, but i also like having a lot of newlines everywhere for the same reason so i am very biased

we don't have internal imports in this project, so we can probably remove the config for it here

scripts/test-utils/src/index.test.ts
beeequeue: man, having things depend on the order of imports sucks 😄

beeequeue: in some files importing `test-utils` _last_ is correct, in some it has to be _before_ the functions to test

eslint.config.mjs
bluwy: Let's see what @Andarist thinks then. If those two are addressed, the rest looks good to me. The import order here feels much natural than the default one.

scripts/test-utils/src/index.test.ts
beeequeue: i think most of the problems would be fixed on the Clack branch, since the errors are mostly about checking if things have been logged 😄

bluwy: In case you're still looking into this, I think setting https://perfectionist.dev/rules/sort-imports#partitionbynewline to true should fix it. Or `partitionByComment`. The former means that we would not have the new lines though 😄

eslint.config.mjs
Andarist: i think the newline between "groups" is redundant, especially if everything gets sorted automatically anyway

packages/get-dependents-graph/src/get-dependency-graph.ts
```diff
@@ -1,8 +1,8 @@
+import path from "node:path";
+import type { Package, Packages, PackageJSON } from "@changesets/types";
+import pc from "picocolors";
 // This is a modified version of the graph-getting in bolt
```
Andarist: this got misplaced - but we should just remove this comment at this point:

```suggestion
```

Andarist (approved):

bluwy (approved):
Do we need the `experimental.preParse` Vitest config?

beeequeue:
Not _really_ but I have been adding it in most branches while fixing tests, so I thought we may as well enable it.

It just makes it so adding .only makes vitest only run the file with it in it and skip the rest

PR #1976, opened 2026-05-05
eslint followup 4: eqeqeq & eq null

beeequeue:
- **enable rule without errors**
- **enable and fix `eqeqeq`**
- **prohibit strict nullish checks, checking against `undefined`**

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -40,7 +40,7 @@ export default defineConfig(
       parserOptions: { projectService: true },
     },
     rules: {
-      eqeqeq: ["off", "always", { null: "never", undefined: "never" }], // TODO enable and fix errors
+      eqeqeq: ["error", "always", { null: "never" }],
 
       "@typescript-eslint/consistent-type-exports": [
         "error",
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -40,7 +40,15 @@ export default defineConfig(
       parserOptions: { projectService: true },
     },
     rules: {
+      // enforce using `x == null` for nullish checks (no triple equals, no undefined)
       eqeqeq: ["error", "always", { null: "never" }],
+      "no-restricted-syntax": [
+        "error",
+        {
+          selector: "BinaryExpression:has(Identifier[name='undefined'])",
+          message: "Use `== null` instead of comparing with `undefined`.",
+        },
+      ],
 
       "@typescript-eslint/consistent-type-exports": [
         "error",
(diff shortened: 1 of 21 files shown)


eslint.config.mjs
```diff
@@ -40,7 +40,15 @@ export default defineConfig(
       parserOptions: { projectService: true },
     },
     rules: {
-      eqeqeq: ["off", "always", { null: "never", undefined: "never" }], // TODO enable and fix errors
+      // enforce using `x == null` for nullish checks (no triple equals, no undefined)
+      eqeqeq: ["error", "always", { null: "never" }],
+      "no-restricted-syntax": [
+        "error",
+        {
+          selector: "BinaryExpression:has(Identifier[name='undefined'])",
```
Andarist, line 48: I'm kinda in the camp of preferring `undefined` over `null` as it plays out better with default parameters. So my slight preference would be to prefer `== undefined`. I won't die on this hill though ;p

Andarist: Also, with type-aware linting - I'd probably be more than fine with just sticking to `===`. If a reference can only be one of those nullish values, handling both dilutes the purpose of the check etc.

bluwy (approved):
I'm personally indifferent if a project prefers strict `===` checks or loose `==` null/undefined checks as long as it's consistent. But since most projects I've worked on so far uses the loose pattern, and I also do this for my own projects, I'm approving it.

Also, I think the previous eslint setup uses `eqeqeq: ["error", "always"]`. If you set that, the errors only show up on recent code.

eslint.config.mjs
bluwy: If it's between `== null` and `== undefined`, I'd prefer the former. But I believe what's happening in this rule here is that it suggests `=== undefined` to `== null`, which means it's less granular.

From my [approval comment](https://github.com/changesets/changesets/pull/1976#pullrequestreview-4233890590), I'm indifferent about it, but the codebase is already sticking to everything `===` convention.

beeequeue: using `undefined` is not forbidden or discouraged, just comparing against it!

writing code that depends on a distinguishment between `undefined` and `null` makes it more complicated than using other ways to know whatever the two are supposed to indicate (i would consider this a code smell)

if code shouldn't depend on them having different meanings then stricly checking for them is pointless, and a loose check is better practice as it will catch both cases, lowering mental effort with no downside ("is it null-ish" vs. "could this value ever become null instead of undefined somehow")

PR #1977, opened 2026-05-05
migrate to tsdown

### LINKED ISSUE #1955: migrate from `preconstruct` to `tsdown`
[`preconstruct`](https://npmx.dev/package/@preconstruct/cli) has been unmaintained for a bit and [`tsdown`](https://npmx.dev/package/tsdown) is a fantastic build tool to replace it with!

would drop ~146 dependencies from the dev tree.

`tsdown` would allow us to remove the places we `require(package.json)` at runtime:
https://github.com/changesets/changesets/blob/aab48fc6c87bc4be24b6b15f2c1b96d8ac59fd37/packages/cli/src/index.ts#L71-L76

and use json imports which get inlined:

```ts
// src/index.ts

import pkgJson from "@changesets/cli/package.json" with { type: "json" }

console.log(pkgJson.version)

// dist/index.mjs

//#region package.json
var version = "0.0.0";
//#endregion
//#region src/index.ts
console.log(version)
//#endregion
```

---

i do also have a branch where i got it working if anyone wants to see what it would look like:
https://github.com/changesets/changesets/compare/next...beeequeue:changesets:tsdown

---

originally proposed in #1869

beeequeue:
closes #1955

- [x] verify that built types work correctly
  - `packages/changelog-git/dist/index.d.ts` seems to be missing the import of `ChangelogFunctions`?


package.json
```diff
@@ -34,39 +35,25 @@
...
-    "@babel/core": "^7.28.5",
-    "@babel/preset-env": "^7.28.5",
-    "@babel/preset-typescript": "^7.27.0",
     "@eslint/js": "10.0.1",
     "@manypkg/cli": "^0.25.1",
-    "@preconstruct/cli": "^2.8.12",
+    "@tsconfig/node22": "^22.0.5",
     "@types/node": "^24.12.2",
     "@types/semver": "^7.7.1",
+    "@typescript/native-preview": "7.0.0-dev.20260504.1",
```
bluwy: Is the dts generation very slow now? I'd prefer looking into isolatedDeclarations first, or discussing this separately before using tsgo.

packages/read/package.json
```diff
@@ -4,11 +4,15 @@
...
+    ".": "./dist/index.mjs",
     "./package.json": "./package.json"
   },
   "license": "MIT",
-  "repository": "https://github.com/changesets/changesets/tree/main/packages/read",
+  "repository": {
+    "type": "github",
+    "url": "changesets/changesets",
+    "directory": "packages/read"
+  },
```
bluwy, line 15: I believe this should be like
```suggestion
  "repository": {
    "type": "git",
    "url": "git+https://github.com/changesets/changesets.git",
    "directory": "packages/read"
  },
```

At least I've not seen this pattern documented before.

package.json
beeequeue: i tested tsgo when i saw that the ci times increased by 15-20s~ with the build step in front of each one.
i also started checking how much work it would be to use isolatedDeclarations so we can compare, but didnt't finish it yet

i reverted it to a draft while i looked into each possibility

packages/read/package.json
beeequeue: i _think_ i've used it at some point before and publint isn't complaining about it, which is why i used it instead of this

i'll change it to the git format though, it's known to work

Andarist: are we absolutely sure this will work with trusted publishing (it requires an npm-valid repository field)?

package.json
Andarist: I'm opposed to `isolatedDeclarations`. I don't think it's worth the effort (although it shouldn't be that big of a problem *here* as opposed to some TS apps). It imposes some weird restrictions and should only be considered when TS becomes a real bottleneck.

packages/read/package.json
beeequeue: the `git` type-format is known to work

package.json
bluwy: Does that mean if we have to choose between `isolatedDeclarations` or tsgo, you'd go with tsgo? Aside that I think it's strange tsdown is taking longer given both tool should be calling tsc the same way.

Also FWIW, while isolated declarations have some weird cases, I find it doesn't affect much of the DX after the migration. The friction is mostly during the migration.

Andarist: Are we forced to chose here? can't we stick right now to just TS-TS until tsgo gets its proper v7 release? but ye, if i'd have to chose i think it's better to go with tsgo cause it's already used by many and stable enough for the dev needs in this repo

> Also FWIW, while isolated declarations have some weird cases, I find it doesn't affect much of the DX after the migration. The friction is mostly during the migration.

inference is the king :p I don't want to annotate simple inference-based patterns. With ID even this becomes a problem:
```ts
declare function memoize<Args extends readonly unknown[], R>(
  cb: (...args: Args) => R,
): (...args: Args) => R;

export const getEnvFileEnv = memoize((path: string) => ({
  /* resolve object once */
}));
```

bluwy: We don't have to choose now but only wondering what's your stance on either choice.

vitest.config.mts
```diff
@@ -4,5 +4,6 @@ export default defineConfig({
   test: {
     experimental: { preParse: true },
     clearMocks: true,
+    globalSetup: ["./scripts/vitest-rebuild-hook.ts"],
```
bluwy: I really prefer we do not do this.

.github/workflows/ci.yml
```diff
@@ -15,8 +15,34 @@ permissions:
...
+        with:
+          download-dist: false
+
+      - name: Build artifacts
+        run: node --run build
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          compression-level: 4
```
bluwy:
```suggestion
          compression-level: 4
          retention-days: 1
```

I think we can set a short retention period so we don't store too much, and since we only need it for the current run.

vitest.config.mts
bluwy: I'm ok if we only do this behind a flag or different command, but I prefer to not have this by default. At least, let's see how the DX works out before making it the default.

.github/workflows/ci.yml
beeequeue: i'll also increase the compression level as there isn't very much data being compressed

vitest.config.mts
beeequeue: we could flip the default behavior so @Andarist can use `test:with-rebuilds`, what do you think?

Andarist: > I really prefer we do not do this.

What's the downside?

bluwy: I want to control the build myself, especially if I've already ran the build or in watch mode, I don't want it to build again twice. Plus, sometimes you don't need to rebuild everything if you change something small or unit tests.

beeequeue: technically we don't need to rebuild between _every_ run, only ones that touch the source files (vs. just changing a test file), and it does slow down re-running those tests by 4-6 seconds

it's also not very common practice to do it like this, so i think it's reasonable to have a specific command for it

Andarist: can we make it depend on an env variable? i could then opt into it myself and keep using a single simple test command

beeequeue: it already does! `test:with-rebuild` just runs `VITEST_AUTO_REBUILD=1 vitest` which the script checks for

beeequeue: removing this as per https://discord.com/channels/1499011163705573470/1501502984880918528/1501891316630687824

scripts/test-utils/package.json
```diff
@@ -5,11 +5,10 @@
   "type": "module",
   "description": "Utils for testing changesets packages",
   "exports": {
-    ".": "./dist/changesets-test-utils.js",
+    ".": "./dist/index.mjs",
```
Andarist: is this cutomizable? named files are nicer in the debugger as they display better in the call stack panel

beeequeue: if we change the entrypoint's name it will be reflected in the output

in my mind it's fine since i'm usually looking more at of the path anyways

beeequeue: we could change them in a follow up PR

packages/changelog-git/src/index.ts
```diff
@@ -45,10 +45,10 @@ const getDependencyReleaseLine = (
   return [...changesetLinks, ...updatedDependenciesList].join("\n");
 };
 
-const defaultChangelogFunctions = {
+const defaultChangelogFunctions: ChangelogFunctions = {
```
Andarist: nit: the other place got annotated as `Required<ChangelogFunctions>`, both should be typed consistently

scripts/test-utils/package.json
bluwy: If we want to stick with the previous naming convention, I'd prefer doing together here so we don't have a diff that goes back-and-forth. For me, I'm fine with the shorter name.

This bugs me more instead 😅 https://github.com/beeequeue/changesets/blob/421455b50f1a239b357cee24e2a775a51c4d2824/packages/cli/package.json#L23-L24

I'll leave @Andarist to decide though

.github/workflows/ci.yml
```diff
@@ -15,8 +15,37 @@ permissions:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - uses: ./.github/actions/ci-setup
+        with:
+          download-dist: false
+
+      - name: Build artifacts
+        run: node --run build
```
bluwy: Maybe we stick with `pnpm build` so it's consistent with the recent change.

bluwy (approved):

.changeset/config.json
```diff
@@ -6,6 +6,7 @@
   "baseBranch": "next",
   "commit": false,
   "access": "public",
+  "ignore": ["@changesets/test-utils"],
```
beeequeue, line 9: we don't need a changelog for `test-utils` since we don't publish it, and we can skip building and just use the `.ts` files directly since vitest/node can handle it nowadays

scripts/test-utils/package.json
beeequeue: > This bugs me more instead 😅 [beeequeue/changesets@`421455b`/packages/cli/package.json#L23-L24](https://github.com/beeequeue/changesets/blob/421455b50f1a239b357cee24e2a775a51c4d2824/packages/cli/package.json#L23-L24)

~~fixed... or is it?~~

nevermind it requires bigger changes :(

packages/cli/src/index.ts
```diff
@@ -1,13 +1,12 @@
-import { createRequire } from "node:module";
 import { format } from "node:util";
+// this requires that the package is built _after_ bumping versions before publishing
```
Andarist, line 2: good callout

packages/cli/package.json
```diff
@@ -15,9 +19,9 @@
   ],
   "type": "module",
   "exports": {
-    ".": "./dist/changesets-cli.js",
-    "./changelog": "./dist/changesets-cli-changelog.js",
-    "./commit": "./dist/changesets-cli-commit.js",
+    ".": "./dist/index.mjs",
```
Andarist, line 22: could we use `entryFileNames?: string | ChunkFileNamesFunction;` to give those better names (more similar to the old ones)?

package.json
```diff
@@ -50,24 +48,12 @@
     "eslint-plugin-n": "^17.24.0",
     "eslint-plugin-perfectionist": "^5.9.0",
     "prettier": "^3.7.4",
+    "publint": "^0.3.18",
```
Andarist, line 51: do we need this explicit dependency? doesn't this already come as part of the tsdown itself?

pnpm-workspace.yaml
```diff
@@ -8,4 +8,7 @@ allowBuilds:
 dedupePeers: true
 dedupePeerDependents: true
 
+overrides:
+  rolldown: 1.0.0
```
Andarist, line 12: I think the latest tsdown version already depends on this version so we could remove this override

tsconfig.json
```diff
@@ -1,13 +1,13 @@
...
-    "target": "esnext",
-    "module": "Node16",
-    "moduleResolution": "Node16",
     "allowImportingTsExtensions": true,
     "erasableSyntaxOnly": true,
-    "libReplacement": false,
+    "verbatimModuleSyntax": true,
     "strict": true,
     "noEmit": true,
+    "lib": ["es2025", "ESNext.Array", "ESNext.Collection", "ESNext.Iterator"],
```
Andarist: hm, what's in those `ESNext` libs? won't we risk publishing something out of our declared engines support by including those here?

tsconfig.json
```diff
@@ -1,13 +1,13 @@
...
     "types": ["node"]
-  }
+  },
+  "include": ["./", "./packages/*"]
```
Andarist: doesn't `"./"` already include "./packages/*"? ;p do we need this include declaration (we didn't have it before)?

package.json
beeequeue: it's an optional peer dep

pnpm-workspace.yaml
beeequeue: vitest doesn't, so this makes sure we don't intsall duplicate versions until they update

tsconfig.json
beeequeue: they're included by the `node22` preset so i assume they're already supported

pnpm-workspace.yaml
Andarist: Vite is already on 1.0.0 too - but it doesn't use `^` so this override is a better choice right now.

tsconfig.json
beeequeue: i'm not sure why i had to manually add them here, but if they're there i had no choice

Andarist (approved):

tsconfig.json
beeequeue: ah, it's because `node22` lists `lib: [es2024, ...]` and `Promise.withResolvers` is inside `es2025`

i narrowed it to `ES2025.Promise`

packages/cli/package.json
beeequeue: i think renaming the entrypoints would be a better solution, but `entryFileNames` are almost always not a good idea in my experience, they always have small quirks or details that make them a pain

Andarist: FWIW, in the Preconstruct/Rollup era we were using a pretty simple pattern:
https://github.com/preconstruct/preconstruct/blob/8a86cca691c1c1dbf353072853d6679b0636186d/packages/cli/src/build/config.ts#L101-L103

atahrijouti:
Amazing work, thank you 🙌🏼

PR #1983, opened 2026-05-06
Remove unused dotenv dependency

bluwy:
Looks like I forgot to remove the dependency after https://github.com/changesets/changesets/pull/1947


beeequeue (approved):
😄

PR #1984, opened 2026-05-06
Use Vite 8 for Vitest

bluwy:
So that it matches the docs branch that also uses Vite 8, and matches the tsdown branch that uses Rolldown. We should no longer have the Rollup and Esbuild binaries installed.

I re-triggered the update by
1. Setting `overrides[vitest>vite]` to `^8.0.0`.
2. Run `pnpm update vite`.
3. Set `autoInstallPeers: false` and `pnpm i` (removes outdated Vite 7 install)
4. Undo `autoInstallPeers` and `pnpm i` (bring back config but now only Vite 8 exists)
5. Undo overrides.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,9 +5,6 @@ settings:
   dedupePeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  vitest>vite: ^8.0.0
-
 importers:
 
   .:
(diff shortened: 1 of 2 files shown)


beeequeue (approved):
it must've gotten reverted to v7 in the pnpm migration, i remember making sure it was using v8 in the yarn lockfile...

Issue #1434, opened 2024-08-08
npm changeset version command formats the package.json


talkohavy:
## Problem

when running `npm changeset version`,  the `package.json` file is being formatted.
The formatting turns this:

```json
"files": ["dist"],
```

to this:

```json
"files": [
    "dist"
  ],
```
This formatting is **WRONG**.
A. Why does it need to happen?
B. What is causing the formatting?

I have no husky, I have no prettier. I am using VSCode, and the formatOnSave is disabled.
This is a very small project created today.

This is a HUGE ISSUE for me

with-heart:
B. That's how `JSON.stringify` stringifies JSON when using the `space` parameter.

A. If the `space` parameter isn't used, `JSON.stringify` stringifies JSON as a single line.

You can see the difference here: https://stackblitz.com/edit/vitejs-vite-f9dpnq?file=src%2Fmain.ts

moroshko:
Is there a way to tell changesets to run a custom npm script that formats the file? 
For example:
```
{
  "check": "biome check --write"
}
```
It would need to run after changesets updates `package.json` and before it creates a Pull Request.

talkohavy:
sadly, to this day of writing the comment, there isn't... 

I hate to be doing this in a another package's thread..., but I ended up switching to another tool on npm, called `lvlup`, which doesn't format your `package.json` at all, and only works for a single repo (not a monorepo with many packages) and just performs a simple find & replace.

Full Disclosure ⚠️
The tool barely has weekly downloads, 
as I am the developer/maintainer of it. I find myself often creating new packages and needed an appropriate tool.

Hugos68:
I have this problem too and it is frustrating, I hope some sort of post publish step in CI becomes available so I can run the formatter, it's annoying getting formatting issues in the `main` branch every time after publishing.

tk-o:
You can turn prettier auto-formatting off now with this release:
- https://github.com/changesets/changesets/pull/1588

Hugos68:
@tk-o Awesome, thank you!

with-heart:
@Andarist I think this can be closed now ❤

Hugos68:
@shrugs did you update your changeset package?

beeequeue:
@shrugs if you still have this issue let me know and i'll re-open the issue!

beeequeue:
actually sorry, this is related to the package.json writing which is unaffected by Prettier. i will create an umbrella issue for this later

PR #1985, opened 2026-05-07
Update detect-indent

bluwy:
Simple bump for now. May want to revisit altogether with https://github.com/changesets/changesets/issues/1434


beeequeue (approved):
i spent a little time trying to create a regex to rewrite the versions (much like how you did #1962) but got distracted after i ran into some issues

bluwy:
Yeah I imagine it could be done in a simple function too, but I haven't considered the possible edge-cases. Happy to review it as well if you have an implementation later though.

PR #1987, opened 2026-05-07
use native `Promise.withResolvers`, bump minim node version to 22.11

beeequeue:
- **use native `Promise.withResolvers()`**
- **up minimum node version to 22.11**


packages/cli/src/utils/createPromiseQueue.ts
```diff
@@ -1,18 +1,4 @@
...
-function withResolvers<T>(): PromiseWithResolvers<T> {
-  const rv = {} as PromiseWithResolvers<T>;
-  rv.promise = new Promise<T>((resolve, reject) => {
-    rv.resolve = resolve;
-    rv.reject = reject;
-  });
-  return rv;
-}
-
+// TODO [engine:node@>=23]: remove when supported by minimum node version
```
beeequeue: this comment is formatted to work with the [`unicorn/expiring-todo-comments`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md) eslint rule, but i don't want to add the entire plugin just for this rule

i was thinking of creating a plugin that just supplies this rule, does that sound like a good idea?

packages/cli/src/polyfills/promise-try.ts
```diff
@@ -0,0 +1,7 @@
+// TODO [engine:node@>=23]: remove when supported by minimum node version
+function promiseTry<T>(fn: () => Promise<T>): Promise<T> {
+  return new Promise<T>((resolve) => resolve(fn()));
+}
+
+// @ts-expect-error: overriding built-in
+Promise.try = promiseTry;
```
Andarist: I know it's in a CLI - but on principle I'd avoid this. We can do smth like this if we want to start using the real thing asap:
```ts
const promiseTry = Promise.try || function () { /* ponyfill */ }
```

packages/cli/src/utils/createPromiseQueue.ts
Andarist: I feel like we won't have a lot of such comments lingering anyway, so automation here might be just an overkill

packages/cli/src/polyfills/promise-try.ts
beeequeue: yeah i remembered that it's not good practice in packages that other people will use, i'm rewriting it as you said

packages/cli/src/utils/createPromiseQueue.ts
beeequeue: might be overkill, might be useful :)

the plugin would be very tiny since it would only be one rule

packages/cli/src/utils/createPromiseQueue.ts
```diff
@@ -1,21 +1,4 @@
...
-    rv.resolve = resolve;
-    rv.reject = reject;
-  });
-  return rv;
-}
-
-function promiseTry<T>(fn: () => Promise<T>): Promise<T> {
-  return new Promise<T>((resolve) => resolve(fn()));
-}
+import { promiseTry } from "../polyfills/promise-try.ts";
```
Andarist:
```suggestion
import { promiseTry } from "../ponyfills/promise-try.ts";
```

beeequeue: odd, i thought Webstorm fixed it when i renamed it

packages/cli/src/ponyfills/promise-try.ts
```diff
@@ -0,0 +1,5 @@
+// TODO [engine:node@>=23]: remove when supported by minimum node version
+const polyfill = <T>(fn: () => Promise<T> | T): Promise<T> =>
```
Andarist: nit: this won't have a good inferred `.name` :p

packages/cli/src/utils/createPromiseQueue.ts
Andarist: I'd recommend not spending too much time on custom ESLint setup given we want to try out oxlint once the dust settles - up to you though

beeequeue:
errors! yay!

packages/cli/src/ponyfills/promise-try.ts
```diff
@@ -0,0 +1,5 @@
+// TODO [engine:node@>=23]: remove when supported by minimum node version
+const promiseTryPolyfill = <T>(fn: () => Promise<T> | T): Promise<T> =>
+  new Promise((resolve) => resolve(fn()));
+
+export const promiseTry = Promise.try ?? promiseTryPolyfill;
```
Andarist: this will likely fix the CI issue
```suggestion
export const promiseTry = Promise.try.bind(Promise) ?? promiseTryPolyfill;
```

beeequeue: thanks, i was trying `.bind(this)` which didn't work 😄

packages/cli/src/utils/createPromiseQueue.ts
beeequeue: oxlint supports JS plugins now adays, so we could still use it with oxlint

bluwy (approved):

packages/cli/src/utils/createPromiseQueue.ts
bluwy: Yeah I think we can leave it as a comment now without needing to automate. We should look through the TODOs anyway when cutting majors.

beeequeue: Yes, the plugin can come later

PR #1988, opened 2026-05-07
replace `detect-indent` with regex replacement

### LINKED ISSUE #1434: npm changeset version command formats the package.json

### LINKED ISSUE #1591: The `changeset version` command keeps changing my package.json array formatting
When I run `changeset version`, I expect it to just update the "version" field of my `package.json` file. However, it seems to be doing a little bit more than that.

For example, in my `package.json` file I have this config:

```json
{
  "files": ["dist", "!dist/*.test.*"],
  ...
}
```

But the `changeset version` command transforms it into this:

```json
{
  "files": [
    "dist",
    "!dist/*.test.*"
  ],
  ...
}
```

This actually causes my format checker (biome) to fail in CI because it wants my package.json formatted in the first format, not the one that `changeset version` produced.

beeequeue:
closes #1434
closes #1591

the main benefit of doing this is that it maintains the formatting of the file

---

i guess i did something wrong in my rebase :/

### DIFF
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -3451,12 +3451,24 @@ describe("apply release plan", () => {
 });
 
 describe("updateVersionField", () => {
-  it("should update the version with a new value", () => {
-    const data = JSON.stringify({ version: "1.2.4" });
-    const value = "2.0.0";
+  const cases: Array<[from: string, to: string]> = [
+    ["1.0.0", "1.0.1"],
+    ["1.0.0", "1.1.0"],
+    ["1.0.0", "2.0.0"],
+    ["1.0.0", "2.0.0-beta.0"],
+    ["2.0.0-beta.0", "1.0.0"],
+    ["workspace:^", "1.0.0"],
+    ["1.0.0", "workspace:^"],
+    ["link:../foo", "1.0.0"],
+    ["1.0.0", "link:../foo"],
+    ["1.0.0", "npm:@foo/bar@1.2.3"],
+    ["npm:@foo/bar@1.2.3", "1.0.0"],
+  ];
+  it.each(cases)("should change the version from %s to %s", (from, to) => {
+    const data = JSON.stringify({ version: from });
 
-    expect(updateVersionField(data, value)).toEqual(
-      JSON.stringify({ version: value }),
+    expect(updateVersionField(data, to)).toEqual(
+      JSON.stringify({ version: to }),
     );
   });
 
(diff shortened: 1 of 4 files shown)


bluwy:
The CI fail aside, I'm ok with the direction here. I think obviously this could break down if there's multiple `"version"` in the package.json, but I've never really seen that. And if anything happens we can revisit again.

beeequeue:
multiple `version`s in a manifest should not be supported, and should probably throw an error if anything :^)

packages/apply-release-plan/src/index.test.ts
Andarist, line 1: We should make sure we handle updating the *last* version field in the package.json and that we handle weirdly formatted `"version": "0.0.0"` fields (like, throw some whitespace, including newlines etc between the key and the value etc)

packages/apply-release-plan/src/index.ts
```diff
@@ -363,16 +362,23 @@ async function updateChangelog(
...
-  pkgJsonPath: string,
-  pkgJson: any,
+// https://regexr.com/8lrc7
+const versionFieldRegex = /(\s*"version"\s*:\s*")[^"\\]*(")/;
+
+export function updateVersionField(
+  jsonString: string,
+  newVersion: string,
+): string {
+  return jsonString.replace(versionFieldRegex, `$1${newVersion}$2`);
```
Andarist, line 372:
```suggestion
  return jsonString.replaceAll(versionFieldRegex, `$1${newVersion}$2`);
```

packages/apply-release-plan/src/index.ts
```diff
@@ -153,10 +152,10 @@ export async function applyReleasePlan(
     config.prettier !== false ? getPrettierInstance(cwd) : undefined;
 
   for (const release of finalisedRelease) {
-    const { changelog, packageJson, dir, name } = release;
+    const { changelog, newVersion, dir, name } = release;
 
     const pkgJSONPath = path.resolve(dir, "package.json");
-    await updatePackageJson(pkgJSONPath, packageJson);
+    await updateVersionFieldInFile(pkgJSONPath, newVersion);
```
Andarist, line 158: the test are failing because we have to rewrite more than just the version field - the dependency ranges are being rewritten in all of this too and we need to target them within the correct dependency buckets (dependencies, devDependencies, peerDependencies etc)

beeequeue: Indeed, I think this is why I stopped working on this branch, as replacing a specific field that can appear more than once requires more complex logic

bluwy:
The best case might be to use something like this: https://www.npmjs.com/package/edit-json. It's hard to get rid of a dependency if we want to get this right.

PR #1989, opened 2026-05-08
Remove legacy changeset format

43081j:
Removes support for the old `changes.json` format.


packages/apply-release-plan/src/index.ts
beeequeue: it seems like you can also remove this, based on the comment

i don't really see why it was here in the first place, so maybe @Andarist can confirm?

bluwy:
@43081j are you still interested in updating this? If not, one of us could take over and update as well

packages/apply-release-plan/src/index.ts
Andarist: I think the whole `changesetFolderPath` targets the v1 changesets, so yes - this branch and that variable declaration can be removed

43081j:
done. let me know if the changeset should be a minor or not

.changeset/cool-places-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": minor
+---
+
+Remove support for legacy changeset formats.
```
Andarist:
```suggestion
Remove support for the legacy Changeset v1 format.
```

Andarist (approved):

PR #1990, opened 2026-05-08
chore: drop legacy default exports

43081j:
Removes all legacy default exports. These should already be unused in
the wild and our own usage (especially as of this change).

### DIFF
--- a/packages/assemble-release-plan/src/index.ts
+++ b/packages/assemble-release-plan/src/index.ts
@@ -371,7 +371,3 @@ function getPreInfo(
     preVersions,
   };
 }
-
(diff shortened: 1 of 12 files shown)


beeequeue (changes requested):
we're deprecating the default exports in v3 and removing them in v4!

bluwy:
Yeah we just deprecated them a few days ago. I'll close the PR and we can accept a new one when we're ready. Otherwise it'll easily conflict.

bluwy:
@43081j circling back on this, some of the refactors to use named imports instead makes sense to do now, if you'd like to submit a PR for that. We should still keep the deprecated default exports for now.

PR #1993, opened 2026-05-11
add pkg.pr.new

beeequeue:


.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,31 @@
+name: pkg-pr-new
+
+on:
+  # publish each pull request to next
+  pull_request:
+    branches: ["next"]
```
Andarist: let's add `main` here too - this PR could target the `main` branch instead of `next`

beeequeue: it's currently configured to use pnpm for publishing and packing which the main branch doesn't have

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,31 @@
...
+          persist-credentials: false
+
+      - uses: ./.github/actions/ci-setup
+        with:
+          node-version: 26
+
+      - name: Build
+        run: pnpm build
+
+      - run: pnpm exec pkg-pr-new publish --pnpm --packageManager=pnpm --commentWithDev --commentWithSha './packages/*'
```
Andarist: from the looks of it - this can be called locally too, right? it doesnt get any token from us here or anything. Pretty neat

bluwy: I think it's tied to the github app so you can only run via CI

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,39 @@
+name: pkg-pr-new
+
+on:
+  pull_request:
+    types: [labeled]
+  # publish each commit to main/next
+  #push:
+  #branches: ["main", "next"]
+  #tags: ["!**"]
```
Andarist: nit: can't those comments contain proper indentation?

Andarist (approved):

Edgarruiz8585:
- > by: Edgar Manuel Ruiz Arias @Edgarruuz8585
***@***.*** ***@***.***
***@***.*** ***@***.***

.github/workflows/pkg-pr-new.yml
beeequeue: do we want the on-push publishing? if not then we can just remove the comments

package.json
```diff
@@ -46,6 +46,7 @@
     "eslint-plugin-import-lite": "~0.6.0",
     "eslint-plugin-n": "^17.24.0",
     "oxfmt": "^0.49.0",
+    "pkg-pr-new": "0.0.71",
```
bluwy: Can we unpin this?

.github/workflows/pkg-pr-new.yml
Andarist: ye, removing would be even better :p

package.json
beeequeue: i think this one should be pinned since it isn't even `0.1` yet

bluwy: I mean `^0.0.71` theoretically works the same since when 72 comes it's breaking so it won't bump. Just I think would be nice if we have ^ consistently used.

beeequeue: i thought `0.0.x` counted as a minor+patch release?

bluwy: It also counts as a major/breaking release, so any bump on 0.0.X is considered breaking even if there's none. Semver states the breaking number to be the left-most non-zero number.

bluwy (approved):

PR #1999, opened 2026-05-12
chore: use named exports in our own files

43081j:
No changeset since its only tests affected.


beeequeue (approved):

packages/cli/src/commands/add/__tests__/add.test.ts
```diff
@@ -3,7 +3,7 @@ import { stripVTControlCharacters } from "node:util";
 import { defaultConfig } from "@changesets/config";
 import * as git from "@changesets/git";
 import * as logger from "@changesets/logger";
-import getChangesets from "@changesets/read";
+import { readChangesets as getChangesets } from "@changesets/read";
```
bluwy, line 5: Could we update the named imports to use the original name directly? e.g. `readChangesets` here. If possible if they're not shadowed.

beeequeue: i think we can do that over time, it would introduce many conflicts if we do it now

bluwy: I feel like we already have a lot of conflicts compared to main at this point with the packages we've migrated to.

beeequeue: not against main, but the other branches i started on for updating the commands and such... but i could deal with them

bluwy: I see. Let's merge it now then 👍

PR #2000, opened 2026-05-12
improve workflows, add zizmor linting

beeequeue:
- disables caching as discussed [in discord](https://discord.com/channels/1499011163705573470/1503666590707744778)
- adds [zizmor](https://docs.zizmor.sh/) for linting the workflows for issues
  - applies all the recommendations from it
- updates the codecov action to the latest version

will fix `next` when we merge `main` into it:

- `download-dist` should default to false

---

for whatever reason `package-manager-cache` does not override `cache` (despite being the more specific option), meaning logic for what we pass to the action has to be ridiculous

### DIFF
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -17,7 +17,8 @@ runs:
       uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version: ${{ inputs.node-version }}
-        cache: ${{ inputs.skip-cache != 'true' && 'yarn' }}
+        package-manager-cache: ${{ inputs.skip-cache != 'true' }}
+        cache: ${{ inputs.skip-cache != 'true' && 'yarn' || '' }}
 
     - name: Install dependencies
       shell: bash
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -4,7 +4,11 @@ inputs:
   node-version:
     description: "Node.js version"
     required: false
-    default: 20
+    default: "20"
+  skip-cache:
+    description: "Whether to skip the cache"
+    required: false
+    default: "false"
 
 runs:
   using: composite
@@ -13,7 +17,7 @@ runs:
       uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
       with:
         node-version: ${{ inputs.node-version }}
-        cache: yarn
+        cache: ${{ inputs.skip-cache != 'true' && 'yarn' }}
 
     - name: Install dependencies
       shell: bash
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -14,7 +14,7 @@ runs:
   using: composite
   steps:
     - name: Setup Node.js ${{ inputs.node-version }}
-      uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version: ${{ inputs.node-version }}
         cache: ${{ inputs.skip-cache != 'true' && 'yarn' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         run: yarn jest --ci --color --runInBand --coverage --reporters=default --reporters=jest-junit
 
       - name: Upload coverage
-        uses: codecov/codecov-action@b9fd7d16f6d7d1b5d2bec1a2887e65ceed900238 # v4
+        uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,8 @@ permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: false
-  queue: max
+  group: ${{ github.workflow }}-${{ github.sha }}
+  cancel-in-progress: true
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,13 +10,20 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
       - uses: ./.github/actions/ci-setup
 
       - name: Check Git version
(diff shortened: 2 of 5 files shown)


.github/actions/ci-setup/action.yml
```diff
@@ -4,16 +4,21 @@ inputs:
...
   using: composite
   steps:
     - name: Setup Node.js ${{ inputs.node-version }}
-      uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version: ${{ inputs.node-version }}
-        cache: yarn
+        package-manager-cache: ${{ inputs.skip-cache != 'true' }}
+        cache: ${{ inputs.skip-cache != 'true' && 'yarn' || '' }}
```
beeequeue, line 21: `cache` also does not support receivign `false` :/

.github/workflows/publish.yml
```diff
@@ -22,13 +24,18 @@ jobs:
       issues: write # to post issue comments (changesets/action)
       pull-requests: write # to create pull request (changesets/action)
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
```
beeequeue, line 30: i'm not sure if this step needs the credentials for committing or pushing, but if it does we should explicitly set `persist-credentials: true` to make it more visible

.github/workflows/ci.yml
```diff
@@ -10,13 +10,20 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
Andarist, line 15: I kinda like CI running on all~ commits - but it's not something I deeply care about so both ways of approaching this work for me.

beeequeue: we can change it to `cancel-in-progress: false` which queues them instead of cancelling, as long as it's explicitly declared

.github/workflows/publish.yml
Andarist: Hm, it might need this actually (I'm like 90% sure) - without it `changesets/action` won't be able to push

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor 🌈
+        uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
+        with:
+          persona: pedantic
+          annotations: true
+          advanced-security: false
```
Andarist, line 37: what are the usability diffs between both values?

.github/workflows/ci.yml
Andarist: it's more of a question if we need `concurrency` here at all or not

.github/workflows/zizmor.yml
beeequeue: it defaults to `advanced-security` which uses [GHAS](https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security) and is way less visible than `annotations` which just prints it to the logs with [annotation tags](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-creating-an-annotation-for-an-error)

.github/workflows/ci.yml
beeequeue: it's recommended by zizmor and i didn't want to disable it for this workflow specifically

.github/actions/ci-setup/action.yml
```diff
@@ -4,16 +4,21 @@ inputs:
...
 runs:
   using: composite
   steps:
     - name: Setup Node.js ${{ inputs.node-version }}
-      uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version: ${{ inputs.node-version }}
-        cache: yarn
+        package-manager-cache: ${{ inputs.skip-cache != 'true' }}
```
Andarist, line 20: hm, isn't this npm-specific option? see https://github.com/actions/setup-node/pull/1374/changes#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6

beeequeue: that just means the naming and behavior of the setting are even worse, but i would still keep it just for declaring intent :^)

.github/workflows/ci.yml
Andarist: i'm fine either way but just noting down this could be smth like:
```yml
concurrency:
  group: ${{ github.workflow }}-${{ github.sha }}
  cancel-in-progress: true
```

.github/workflows/publish.yml
```diff
@@ -6,7 +6,9 @@ on:
       - main
       - next
 
-concurrency: ${{ github.workflow }}-${{ github.ref }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
```
Andarist, line 11: TIL about this option but it feels like we should be using this too:

```suggestion
  cancel-in-progress: false
  queue: max
```

thoughts?

beeequeue: i definitely thought the default behavior was like `max`...

should we use it everywhere?

Andarist: I thiiink I just read the default is the other value (single?) but it might be good to doublecheck. I think it can be a sane default to use everywhere where we need the queuing behavior.

.github/workflows/ci.yml
```diff
@@ -10,13 +10,21 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: false
+  queue: max
```
Andarist: I have only meant that we should use `queue: max` where we actually want queuing. In here, we should stick to your previous version (`cancel-in-progress: false`) or just key the group by the `${{ github.workflow }}-${{ github.sha }}`

I'm fine with either - but this one should be executed ASAP, queueing those would get in the way

beeequeue: i thought queueing the ci steps was what you meant by "CI running on all~ commits" 🙈

Andarist: I just mentioned the old behavior - one that, for the CI job, was fine to me. If we'd want to maintain that then we should use a that different `group` key. Sorry for the confusion, the separate threads might have stepped on each other with some unspoken nuance ;p

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    paths: [.github/workflows/*]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: false
+  queue: max
```
Andarist: This is close to the CI workflow really, so it would probably be good to use the same approach here:

```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.sha }}
  cancel-in-progress: true
```

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,39 @@
...
+
+on:
+  push:
+    branches: [main, next]
+    paths: [.github/workflows/*]
+  pull_request:
+    paths: [.github/workflows/*]
+
+permissions:
+  contents: read
```
Andarist: I think the principle of using the root `: {}` and relying on explicit permissions on each job is quite nice:
```suggestion
permissions: {}
```

although ofc `contents: read` could be kinda treated as that `{}`, given this is the very basic permission. I'd just prefer to use a single flavor across all workflows - I'm fine with either

beeequeue: it's something i forgot to change after i copied it from my template project

Andarist (approved):

PR #2002, opened 2026-05-12
Load commands lazily

Andarist:


packages/cli/src/run.ts
```diff
@@ -43,7 +34,7 @@ export async function run(
 
   if (input[0] === "init") {
     validateCommandFlags("init", flags);
-    await init(packages.rootDir);
+    await (await import("./commands/init/index.ts")).init(packages.rootDir);
```
bluwy: I would probably do something like 

```ts
const { init } = await import("./commands/init/index.ts");
await init(packages.rootDir);
```

so it's a bit more readable, but just a nit.

bluwy (approved):

PR #2004, opened 2026-05-12
perf: replace picocolors with node:util styletext

### LINKED ISSUE #1949: replace picocolors with node:util styleText

ghostdevv:
Closes #1949 

I added a new `@changesets/color` internal package which is a small proxy based wrapper over styleText [as suggested by bq](https://github.com/changesets/changesets/issues/1949#issuecomment-4368932727) - it's a dev dependency so it can be bundled as it's small and otherwise would almost defeat the point.

### DIFF
--- a/.changeset/tough-balloons-buy.md
+++ b/.changeset/tough-balloons-buy.md
@@ -1,0 +1,8 @@
+---
+"@changesets/cli": patch
+"@changesets/get-dependents-graph": patch
+"@changesets/logger": patch
+"@changesets/read": patch
+---
+
+perf: replace picocolors with node:util styletext
--- a/.changeset/tough-balloons-buy.md
+++ b/.changeset/tough-balloons-buy.md
@@ -5,4 +5,4 @@
 "@changesets/read": patch
 ---
 
-perf: replace picocolors with node:util styletext
+Replace `picocolors` with `node:util`'s `styleText`
--- a/packages/cli/src/commands/add/createChangeset.ts
+++ b/packages/cli/src/commands/add/createChangeset.ts
@@ -1,7 +1,7 @@
+import color from "@changesets/color";
 import { ExitError } from "@changesets/errors";
 import type { Package, PackageJSON, Release } from "@changesets/types";
 import { log } from "@clack/prompts";
-import pc from "picocolors";
 import semverLt from "semver/functions/lt.js";
 import { askWithEditor } from "../../utils/askWithEditor.ts";
 import * as cli from "../../utils/cli-utilities.ts";
@@ -11,13 +11,13 @@ async function confirmMajorRelease({ name, version }: PackageJSON) {
   if (semverLt(version, "1.0.0")) {
     importantWarning(
       `
-The ${pc.red("major")} version of ${pc.blue(name)} will be its ${pc.red("first major release")} (1.0.0).
-If you are unsure if this is correct, contact the package's maintainers ${pc.red("before committing this changeset")}.   
+The ${color.red("major")} version of ${color.blue(name)} will be its ${color.red("first major release")} (1.0.0).
+If you are unsure if this is correct, contact the package's maintainers ${color.red("before committing this changeset")}.   
       `,
     );
 
     return cli.askConfirm(
-      `Are you sure you want to release the ${pc.red("first major version")} of ${name}?`,
+      `Are you sure you want to release the ${color.red("first major version")} of ${name}?`,
     );
   }
   return true;
@@ -73,7 +73,7 @@ function getPkgJsonsByName(packages: Array<Package>) {
 }
 
 function formatPkgNameAndVersion(pkgName: string, version: string) {
-  return `${pc.bold(pkgName)}@${pc.bold(version)}`;
+  return `${color.bold(pkgName)}@${color.bold(version)}`;
 }
 
 export async function createChangeset(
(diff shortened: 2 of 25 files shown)


bluwy (approved):

.changeset/tough-balloons-buy.md
```diff
@@ -0,0 +1,8 @@
+---
+"@changesets/cli": patch
+"@changesets/get-dependents-graph": patch
+"@changesets/logger": patch
+"@changesets/read": patch
+---
+
+perf: replace picocolors with node:util styletext
```
Andarist:
```suggestion
Replace `picocolors` with `node:util`'s `styleText`
```

packages/color/src/index.ts
```diff
@@ -0,0 +1,13 @@
+// supported and working as expected in Node 22.08+
+// https://nodejs.org/docs/latest-v22.x/api/util.html#utilstyletextformat-text-options
+// eslint-disable-next-line n/no-unsupported-features/node-builtins
```
Andarist, line 3: do we know why this is reported if it's not true? 🤔

bluwy: I believe it's because it detected `styleText` could be experimental then so it's "unsupported", even though the API is the same as the stabilized version. Personally I find this rule not always perfect.

Andarist (approved):
I'm fine with this, let's wait a moment before landing so @beeequeue has a chance to take a look at this too

packages/color/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+import { styleText } from "node:util";
+
+type Color = Exclude<Parameters<typeof styleText>[0], Array<unknown>>;
```
Andarist: it errors without this in the TS playground, perhaps that fetches some newer `@types/node` than the ones we have? the readonly variant below should exclude readonly and mutable arrays 
```suggestion
type Color = Exclude<Parameters<typeof styleText>[0], ReadonlyArray<unknown>>;
```
but perhaps this other version below would be more intentional about what it wants to pluck from the original type:
```suggestion
type Color = Extract<Parameters<typeof styleText>[0], string>;
```

packages/color/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+type ColorProxy = Record<Color, (text: string) => string>;
+
+export default new Proxy({} as ColorProxy, {
+  get(_, color: Color) {
+    return (text: string) => styleText(color, text);
```
Andarist: it could be nice to cache those:
```suggestion
  get(target, color: Color) {
    target[color] ??= (text) => styleText(color, text)
    return target[color];
```

packages/cli/src/commands/add/createChangeset.ts
```diff
@@ -11,13 +11,13 @@ async function confirmMajorRelease({ name, version }: PackageJSON) {
   if (semverLt(version, "1.0.0")) {
     importantWarning(
       `
-The ${pc.red("major")} version of ${pc.blue(name)} will be its ${pc.red("first major release")} (1.0.0).
-If you are unsure if this is correct, contact the package's maintainers ${pc.red("before committing this changeset")}.   
+The ${color.red("major")} version of ${color.blue(name)} will be its ${color.red("first major release")} (1.0.0).
```
beeequeue: i think i would prefer `c.x` over `color.x` since they're used so often, but it's not a dealbreaker to me :^)

packages/cli/src/commands/add/messages.ts
```diff
@@ -19,23 +19,23 @@ export function printConfirmationMessage(
   const minorReleases = getReleasesOfType("minor");
   const patchReleases = getReleasesOfType("patch");
 
-  let msg = pc.bold("Summary of changesets:");
+  let msg = color.bold("Summary of changesets:");
   if (majorReleases.length > 0) {
-    msg += `\n${pc.bold(pc.red("major"))}:  ${majorReleases.join(", ")}`;
+    msg += `\n${color.bold(color.red("major"))}:  ${majorReleases.join(", ")}`;
```
beeequeue: since we're using a proxy it would be nice to make it also support chaining (i.e. `color.bold.red()`!

bluwy: I think it's ok for now without chaining, otherwise it might also affect perf (if I'm imagining the code right).

beeequeue: yeah, it's fine without it.

beeequeue (approved):

packages/cli/src/commands/add/createChangeset.ts
ghostdevv: d7982d87d0c32c550d269c6f329a7ec2c73dba06

packages/color/src/index.ts
ghostdevv: Yea exactly, they don't mark it as stable until 22.13 but there doesn't appear to have been any changes to it between those versions. It's not flagged so works as expected in 22.08

ghostdevv:
Not sure why the test is failing :thinking: flake?

bluwy (approved):

Issue #1971, opened 2026-05-05
Tooling: migrate to `oxfmt` and `oxlint`


marcalexiei:
[Follow up of "Tooling: formatter and lint - migrate to oxc?" Discord thread](https://discord.com/channels/1499011163705573470/1500749127862980689)

We plan to migrate to the `oxc` toolchain, similar to what was done in the bot repo (https://github.com/changesets/bot/issues/99).

Given the scope of changes in v3, we’re deferring the `oxc` migration after the v3 release.

PR #2005, opened 2026-05-13
Format repo with oxfmt

bluwy:
part of https://github.com/changesets/changesets/issues/1971

We can remove both prettier and eslint-plugin-perfectionist. And since we've move to clack for logging, we don't need to force an ordering via comment.

I've also enabled `sortPackageJson` in the second commit.


packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1,7 +1,7 @@
 import { defaultConfig } from "@changesets/config";
 import { beforeEach, describe, expect, it } from "vitest";
-import { FakeFullState } from "./test-utils.ts";
 import { assembleReleasePlan } from "./index.ts";
+import { FakeFullState } from "./test-utils.ts";
```
bluwy, line 4: This seems to be the only difference with oxfmt's default ordering, which I think is acceptable.

.prettierignore
Andarist, line 1: don't we need a rule like this for oxfmt?

Andarist (approved):

.prettierignore
bluwy: It already detects from gitignore for the `dist`. And we don't have `__fixtures__` directory to ignore.

Andarist: cool 👍

Issue #1266, opened 2023-11-28
`pkg:cli` missing `./bin.js` in exports


ludovicm67:
## Affected Packages

- `pkg:cli`

## Problem

I'm getting this error in GitHub Actions after upgrading `@changesets/cli` to `2.27.0`:

```
Error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './bin.js' is not defined by "exports" in /home/runner/work/***/***/node_modules/@changesets/cli/package.json
Error: Package subpath './bin.js' is not defined by "exports" in /home/runner/work/***/***/node_modules/@changesets/cli/package.json
```

This is when `changeset tag` is called.

## Proposed solution

Add `./bin.js` in the exports in the `package.json` file.

alvesvaren:
I get the same error

Andarist:
Thanks for the info - I located the problem to be located [here](https://github.com/changesets/action/blob/2bb9bcbd6bf4996a55ce459a630a0aa699457f59/src/run.ts#L338).

ludovicm67:
Great, thanks for the quick fix! 👍

PR #2009, opened 2026-05-14
Refactor CLI to cac

bluwy:
Main changes:
1. Refactor each `src/commands/*` to independently resolve the required monorepo root and config etc. This makes calling the command function simple.
   
    _(This changes the command function signature. Let me know if you feel differently of how this is refactored. I went back-and-forth between `cwd` being its own param vs inside the options object.)_
4. `src/run.ts` removed. Move the CLI handling all to `src/index.ts`.
5. `src/run.tests.ts` removed. Move to respective commands tests.


beeequeue (approved):

packages/cli/src/index.ts
Andarist, line 1: not super important but we could maybe get rid of `packages/cli/bin.js`? it's just a small redundant indirection thrown into this mix

bluwy: I was hoping you'd say that 😄 didn't want to touch too much initially

Andarist: If we are going to change this then perhaps we should rename this file to `bin.ts` and reconfigure `tsdown` entrypoints? We don't really have a "traditional" index file here

packages/cli/src/index.ts
```diff
@@ -4,91 +4,117 @@ import manifest from "@changesets/cli/package.json" with { type: "json" };
...
+const cli = cac("changeset");
 
-if (parsed.help) {
-  const command = parsed._[0];
-  if (command && COMMAND_HELP[command]) {
-    console.log(`
-  Usage
-    $ changeset ${COMMAND_HELP[command]}
+cli.version(manifest.version);
+cli.help((sections) => {
```
Andarist, line 12: should we also support `changeset help` and `changeset help publish` styles? Currently those result in `CACError`

packages/cli/src/index.ts
bluwy: I noticed we still need to export at least a `bin.js` subpath because of https://github.com/changesets/changesets/issues/1266. How do we want to handle that? Should we keep compat like `exports["./bin.js"]: "./dist/bin.mjs"`?

Personally I'd remove it altogether and update `@changesets/action` accordingly to import from the `"."` entrypoint maybe? But it feels like a lot of decision of changes fitting of a separate PR.

packages/cli/src/index.ts
bluwy: I find it more common to support only `--help`. One way of doing it would be best I think. `changeset --help` also currently suggests using `--help` for the subcommands.

packages/cli/src/commands/version/index.ts
Andarist, line 39: For some reason, we end up here with:
```ts
pnpm run changeset version --snapshot 
```

In this situation, the `options` are getting parsed into:
```json
{
  "--": [
  ],
  snapshot: true,
  ignore: [
    "undefined",
  ],
}
```

1. That `ignore: ["undefined"]` is not added without `--snapshot`
2. But it's also added when using `pnpm run changeset version --snapshot foo`

Andarist (changes requested):
I have not finished reviewing yet, but this one definitely feels blocking

packages/cli/src/index.ts
Andarist: Good catch! Let's not mix this concern into this PR

packages/cli/src/index.ts
Andarist: Counterpoint: `git` supports both 😅 

We can certainly roll with just `--help`, it was just a thought I had - given the user doesn't have to know how to use this properly, supporting both just sounded like a more *help*ful option :p

packages/cli/src/commands/add/index.ts
```diff
@@ -13,27 +13,32 @@ import { getCommitFunctions } from "../../commit/getCommitFunctions.ts";
...
+  open?: boolean;
+  since?: string;
+  message?: string;
+}
+
+export async function add(options?: AddOptions): Promise<void> {
+  const cwd = options?.cwd ?? process.cwd();
+
   const packages = await getPackages(cwd);
+  await ensureChangesetFolder(packages.rootDir);
```
Andarist, line 32: basically all commands (but `init`) need this - maybe instead of doing `ensureChangesetFolder` it would be good to push this into `createCommand` factory? that could accept an option to disable this requirement (just to support `init`).

The `cwd` + `getPackages` resolution could also be shoved there to be shared across all commands (those are actually resolved by all of them)

Andarist: Similarly, the config is being resolved for all of the commands... minus `init` (duh) and 🥁 `pre`.

packages/cli/src/commands/add/index.ts
```diff
@@ -13,27 +13,32 @@ import { getCommitFunctions } from "../../commit/getCommitFunctions.ts";
 import * as cli from "../../utils/cli-utilities.ts";
 import { importantWarning } from "../../utils/cli-utilities.ts";
 import { getVersionableChangedPackages } from "../../utils/versionablePackages.ts";
+import { ensureChangesetFolder } from "../shared.ts";
 import { createChangeset } from "./createChangeset.ts";
 import { printConfirmationMessage } from "./messages.ts";
 
-export async function add(
-  cwd: string,
```
Andarist, line 20: >  I went back-and-forth between cwd being its own param vs inside the options object.)

yeaa, I can totally see going back-and-forth on it. I feeeel just *slightly* stronger about it being separate but at the end of the day, it doesn't matter all that much. It just feels like a totally different category than other options.

Pick your poison, I don't mind either really.

packages/cli/src/commands/version/index.ts
bluwy: This looks like a very strange cac bug. I think I can workaround for now by passing a custom type/validator

packages/cli/src/commands/add/index.ts
beeequeue: i vote for inside object ✋

packages/cli/src/commands/version/index.ts
bluwy: I pushed a fix with manual normalization instead. The `type` isn't very reliable to be used unfortunately

packages/cli/src/commands/add/index.ts
bluwy: I didn't put a lot of thought when I changed this, but I think my intuition is if this becomes a public API one day, then the object form would look more sensible as `cwd` can be inferred and compute for the root directory. Whereas if we made it as its own param, I'd imagine it'd be required `cwd: string` which might not make a lot of sense. Or if this would be the root directory directly.

So I think the object form may be a bit better.

packages/cli/src/commands/add/index.ts
bluwy: Maybe we can make a `resolveCommandContext(cwd, { ... })` function that returns the `cwd`, `packages`, `config` etc? I think it's nice if we can see the execution step-by-step instead of an abstraction.

But also I think the current step is fine for granular control, and repetition is sometimes fine, but it's not a strong pattern I'll insist on.

packages/cli/src/index.ts
bluwy: I think I'd prefer a single `--help` for now. As long as we document how it's used, I think users should then not get lost. Even though git supports it, it's still not very common for me 😅

packages/cli/src/commands/add/index.ts
```diff
@@ -53,30 +58,30 @@ No versionable packages found
...
+      summary: options?.message ?? "",
     };
   } else {
     let changedPackagesNames: string[] = [];
     try {
       changedPackagesNames = (
         await getVersionableChangedPackages(config, {
-          cwd,
-          ref: since,
+          cwd: packages.rootDir,
```
Andarist, line 75: I wooooonder... what do we really need `cwd` for? Perhaps we should just always default to `packages.rootDir`? `cwd` is useful for the initial resolution of `packages` but I wonder if it's useful for anything else.

Just food for thought here, we certainly don't have to touch this in this PR

Andarist: It might have not been clear what I was alluding to here, so let me clarify - perhaps we should rename things to use `rootDir` name over `cwd` so it's more apparent what's that value and what's its origin. And pass `packages.rootDir` to those few places in which we actually still use the input `cwd`

packages/cli/src/commands/status/index.ts
```diff
@@ -1,29 +1,37 @@
...
+  await ensureChangesetFolder(packages.rootDir);
+  const config = await read(packages.rootDir, packages);
+
+  const releasePlan = await getReleasePlan(
+    packages.rootDir,
+    options?.since,
+    config,
+  );
   const changedPackages = await getVersionableChangedPackages(config, {
     cwd,
```
Andarist: I doubt it actually makes any difference... but this would be consistent with the `add` command:
```suggestion
    cwd: packages.rootDir,
```

packages/cli/src/index.ts
```diff
@@ -4,91 +4,123 @@ import manifest from "@changesets/cli/package.json" with { type: "json" };
...
+  .option(
+    "--snapshot-prerelease-template <template>",
+    "Template for snapshot prerelease",
+  )
+  .action(async (options) => {
+    if (options.ignore) {
+      options.ignore = Array.isArray(options.ignore)
+        ? options.ignore
+        : [options.ignore];
+    }
```
Andarist: note to myself: it would be nice to report the original issue related to this workaround to `cac`

packages/cli/src/commands/add/index.ts
Andarist: yeah, such heloer would work too 👍

Andarist (approved):

packages/cli/src/commands/add/index.ts
beeequeue: it's what i did for my cleye and gunshi implementations as well

packages/cli/src/index.ts
Andarist, line 1: It would be nice if we could get a test for this file that would test the discovered `ignore` thing and perhaps support for mixed `snapshot` boolean/string support. But it's not super important

beeequeue: Agree, regression tests are awesome 😎

packages/cli/src/commands/add/index.ts
bluwy: Yeah I think we need to follow-up making `cwd` and `rootDir` separation clearer. It's a bit tricky for some `@changesets/*` packages because they're sometimes used directly and some people may expect those functions to auto-resolve to the root-dir as well. We should revisit for sure and decide how to handle this better.

packages/cli/src/commands/add/index.ts
bluwy: I tried to implement this but in the end I've written more code for the function than it's worth. I think we should stick to what we have now and revisit again once we get the changesets config refactor in and the rootDir sorted out so we get a better picture of the abstraction.

packages/cli/src/index.ts
bluwy: I ended up putting another fix where if a flag is passed multiple times (the option becomes an array), we pick the last value. And put these inside `normalizeOptions`. So even if this is fixed upstream, it might be better to stick with our little utility.

Note the flag array issue also happens in `main` with plain mri usage.

packages/cli/src/index.ts
bluwy: I wrote some tests initially, exporting the `cli` which exposes some things to test, but since a lot of the handlers are thin and we should be trusting cac's behaviour, the tests ended up not useful I think. Probably useful now with some quirks we found from cac though.

bluwy:
I'll merge this in now, but I'll follow-up on the tests later as I'm thinking of moving some things around and I don't want to touch to much to review again.

PR #2010, opened 2026-05-14
Update `pnpm` and harden its config

Andarist:


pnpm-workspace.yaml
```diff
@@ -2,13 +2,15 @@ packages:
   - packages/*
   - scripts/*
 
+overrides:
+  rolldown: 1.0.0
```
bluwy: IIRC if you run a pnpm command that edits this file, e.g. `pnpm approve-builds`, it would re-sort the keys alphabetically (except packages). So I think it's better to put this overrides at the bottom like before.

pnpm-workspace.yaml
```diff
@@ -2,13 +2,15 @@ packages:
...
+
 allowBuilds:
   esbuild: false
 
+blockExoticSubdeps: true
```
bluwy, line 8: This is already true by default FYI, I guess doesn't hurt to explicit specify though.

pnpm-workspace.yaml
```diff
@@ -2,13 +2,15 @@ packages:
...
 dedupePeers: true
 dedupePeerDependents: true
-
-overrides:
-  rolldown: 1.0.0
-
+minimumReleaseAge: 10080
```
bluwy, line 11: I feel like the default of 1 day is good enough though. If we're concerned because this is an important package, maybe 3 days instead?

bluwy (commented):
Left some of my opinions, but I'm generally ok if we ended up with the current config.

8bnns:
.

pnpm-workspace.yaml
beeequeue: we could do 1-3 days for the next branch and up it to 5-10 when we release v3?

pnpm-workspace.yaml
Andarist: It only has been made the default in v11 - somebody could still be running with v10, so it won't hurt to have this explicit

pnpm-workspace.yaml
Andarist: hm, i can't really find any info on this sorting and smth like `pnpm config set blockExoticSubdeps true --location=project`  just appends at the end. We can have this sorted alphabetically~, I just sorted it manually at the top as that felt like an option we might want to change from time-to-time (and I wanted to keep the simple key-value options together).

But either way, I'm fine with whatever here - so I just put it in the old place now

pnpm-workspace.yaml
Andarist: 7 days got recommended as a sensible default in the recent "guides" after the tanstack-related breach. We can always use `minimumReleaseAgeExclude` when needed and it's probably just better to be safe-than-sorry. I feel like any value here would be annoying when one wants to update a package - so we can just stick to that week as a sensible default. It's also how our dependabot is configured now:
https://github.com/changesets/changesets/blob/1bad6d243c6a626f6a5f516a62129b862ab07c1c/.github/dependabot.yml#L12-L13

bluwy (approved):

PR #2012, opened 2026-05-15
Include zizmor in CI workflow

Andarist:
I want to pull this into the CI workflow so it becomes part of the required checks (and so I don't have to configure another job to be required ;p)


beeequeue:
Why don't you want to have multiple required checks? :^)

Andarist:
> Why don't you want to have multiple required checks? :^)

Just for simplicity

bluwy:
I think I'd prefer keeping it separate so it's its own thing rather than mixed into our source code testing.

beeequeue (approved):

beeequeue:
> I think I'd prefer keeping it separate so it's its own thing rather than mixed into our source code testing.

same, but he convinced me in DMs :^)

PR #2015, opened 2026-05-15
rework config parsing, rewrite config tests

beeequeue:
this PR reworks the config package to use Valibot for config shape validation, and then a rules engine(-ish) for validating that the config is correct.

---

this PR does **NOT** remove the remaining `console.log`s from various packages, as i wanted to keep this PR slightly more focused


packages/config/scripts/generate-schema.ts
beeequeue, line 1: this file is outdated

packages/config/src/config.ts
```diff
@@ -0,0 +1,221 @@
...
+    v.description('Deprecated, just an alias for "restricted"'),
+  ),
+]);
+
+export const WrittenConfigSchema = v.object({
+  $schema: v.optional(v.string()),
+
+  baseBranch: rootKey(
+    v.string(),
+    "Determines the branch that Changesets uses when finding what packages have changed.",
```
beeequeue, line 48: some errors have `Invalid X:` in front which follows valibot's issue messages, so we'll probably want to use them everywhere for consistency

packages/config/src/config.ts
```diff
@@ -0,0 +1,221 @@
...
+  } else if (writtenConfig.commit === true) {
+    config.commit = ["@changesets/cli/commit", { skipCI: "version" }];
+  }
+
+  if (writtenConfig.ignore != null) {
+    config.ignore = globMatch(packageNames, writtenConfig.ignore);
+  }
+
+  // TODO consider enabling this by default in the next major version
+  // might be more context in here: https://github.com/changesets/changesets/pull/662
```
beeequeue, line 186: i copied this comment from the original code, no clue what it means still :^)

packages/config/src/rules.ts
```diff
@@ -0,0 +1,253 @@
...
+const noFixedAndLinkedPackages: Rule = ({ config, errors }) => {
+  if (config.fixed.length === 0 || config.linked.length === 0) return;
+
+  const allFixedPackages = new Set(config.fixed.flat());
+  const allLinkedPackages = new Set(config.linked.flat());
+
+  for (const fixedName of allFixedPackages) {
+    if (allLinkedPackages.has(fixedName)) {
+      errors.push(
+        `The package "${fixedName}" can be found in both fixed and linked groups. A package can only be either fixed or linked.`,
```
beeequeue:
```suggestion
        `Invalid group: The package "${fixedName}" can be found in both fixed and linked groups. A package can only be either fixed or linked.`,
```

packages/config/schema.json
beeequeue, line 1: i need to test this against some complex configs from users

packages/config/src/parse.ts
```diff
@@ -0,0 +1,78 @@
...
+  cwd?: string,
+  packages?: Packages,
+): Promise<ParseResult> {
+  cwd ??= process.cwd();
+  packages ??= await getPackages(cwd);
+
+  // parse
+  const json = await readConfigFile(packages.rootDir);
+
+  // normalize+validate
```
beeequeue:
```suggestion
  // read
  const json = await readConfigFile(packages.rootDir);

  // parse+normalize+validate
```

packages/config/src/parse.ts
```diff
@@ -0,0 +1,77 @@
...
+  cwd?: string,
+  packages?: Packages,
+): Promise<ParseResult> {
+  cwd ??= process.cwd();
+  packages ??= await getPackages(cwd);
+
+  // read
+  const json = await readConfigFile(packages.rootDir);
+
+  // prase+normalize+validate
```
Andarist:
```suggestion
  // parse+normalize+validate
```

packages/config/src/index.ts
```diff
@@ -1,630 +1,2 @@
...
-  },
-};
-
-export const defaultConfig = parse(defaultWrittenConfig, {
-  tool: { type: "root" },
-  rootDir: fakePackage.dir,
-  rootPackage: fakePackage,
-  packages: [fakePackage],
-});
+export { readAndValidateConfig } from "./parse.ts";
```
Andarist: Publicly, we are only exposing `readAndValidateConfig` here. I think the expectation of `read*` is to receive back a typed/valid config already and we don't have to be as verbose in the name.

pnpm-workspace.yaml
```diff
@@ -8,7 +8,8 @@ allowBuilds:
 blockExoticSubdeps: true
 dedupePeers: true
 dedupePeerDependents: true
-minimumReleaseAge: 10080
+
+minimumReleaseAge: 2440
```
Andarist: I'd prefer to bump it back:
```suggestion
minimumReleaseAge: 10080
```

tsdown.config.ts
```diff
@@ -5,9 +5,13 @@ const isCi = process.env.CI != null;
 export const baseConfig = defineConfig({
   entry: "src/index.ts",
   outDir: "dist",
+  hash: false,
```
Andarist, line 8: I kinda like the hashes - at least at the glance I can tell which file is an entry point and which one is a generated chunk :D

I'm fine either way though

.changeset/dumb-frogs-jump.md
```diff
@@ -0,0 +1,7 @@
+---
+"@changesets/config": minor
+---
+
+Added support for `access: "internal"`, used by the GitHub NPM Packages Registries.
```
Andarist: If possible, I'd prefer to pull this out from this PR. I don't fully understand the context behind this. I published some packages with Changesets to the GitHub Registry just recently and it worked OK. So it would be good to motivate this change and document the context behind introducing it.

pnpm-workspace.yaml
beeequeue: Yes, it's just so I could work in the repo

packages/get-release-plan/src/index.ts
Andarist, line 1: I feel this is kind of a driveby change here and similarly it would be good to move it out to a separate PR.

packages/types/src/index.ts
```diff
@@ -98,7 +99,7 @@ export type WrittenConfig = {
   commit?: boolean | readonly [string, any] | string;
   fixed?: Fixed;
   linked?: Linked;
-  access?: AccessType;
+  access?: AccessType | "private";
```
Andarist, line 101: what this new access is useful for? and why it's not part of the `AccessType` type?

packages/config/package.json
```diff
@@ -29,10 +28,15 @@
...
+    "@valibot/to-json-schema": "1.7.0",
     "outdent": "^0.8.0",
+    "valibot": "^1.4.0",
     "vitest": "^4.1.6"
   },
   "engines": {
     "node": "^22.11 || ^24 || >=26"
+  },
+  "inlinedDependencies": {
+    "valibot": "1.4.0"
```
Andarist, line 43: do we have to *also* declare this here even though it's already configured in the tsdown config? 🤔

packages/config/package.json
```diff
@@ -29,10 +28,15 @@
   "devDependencies": {
     "@changesets/test-utils": "workspace:*",
     "@types/picomatch": "4.0.3",
+    "@valibot/to-json-schema": "1.7.0",
```
Andarist: if we are bundling valibot itself, why we are not bundling this one?

beeequeue: it's notn a runtime dependency, we generate the json schema at some point (during the build?)

packages/config/package.json
beeequeue: it's a new [proposal](https://github.com/e18e/ecosystem-issues/issues/237)(-ish) by sxzz the maker of tsdown, to allow users of packages to know what dependencies are inlined

it gets added automatically by tsdown but i think we can remove it if we want to. i don't think it hurts though.

packages/types/src/index.ts
beeequeue: we accept `private` but it wasn't in this (`WrittenConfig`) type, then we change it to `restricted` when normalizing, so `Config` should not include it. both types use `AccessType`.

since we type check the config schema output against `WrittenConfig` they both need to accept `private` or we get an error.

packages/get-release-plan/src/index.ts
bluwy: Agreed. And I'm recently using `get-release-plan` too in actions because it has the better API ergonomics

packages/config/package.json
bluwy:
```suggestion
    "@valibot/to-json-schema": "^1.7.0",
```

packages/types/src/index.ts
bluwy: Should we consider removing `private` support entirely? It's already deprecated in main with a warning.

beeequeue: i'm okay with that, i just reimplemented the current support

packages/config/src/json-schema.ts
bluwy, line 1: Can we move this file to `scripts/generate-json-schema.ts`?

packages/config/src/config.ts
```diff
@@ -0,0 +1,221 @@
...
+const ChangelogSchema = v.union(
+  [
+    v.literal(false),
+    v.string(),
+    v.tuple([
+      v.string(),
+      // record or any?
+      // v.any(),
+      v.nullable(v.record(v.string(), v.unknown())),
+    ]),
```
bluwy: I think a record here is fine

packages/cli/package.json
```diff
@@ -31,7 +31,8 @@
     "./changelog": "./dist/changelog.mjs",
     "./commit": "./dist/commit/index.mjs",
     "./package.json": "./package.json",
-    "./bin.js": "./bin.js"
+    "./bin.js": "./bin.js",
+    "./config-schema.json": "./node_modules/@changesets/config/schema.json"
```
bluwy: Let's hold off this change for now

bluwy (commented):
Why do we manually implement a normalization and validation step (with rules), instead of using valibot's `v.transform` and `v.check`?

packages/config/src/config.ts
Andarist: From what I read, before the referenced PR, private packages wouldn't be git-tagged - despite being versioned. And this just mentions that we could reconsider changing the default (`version: true, tag: false`) to `version: true, tag: true`.

Note: `version: false, tag: true` sounds like an invalid combination

packages/types/src/index.ts
Andarist: if it's already deprecated, let's remove it here

packages/config/package.json
Andarist: it's fine - thanks for sharing the context

beeequeue:
We can't generate a json schema if there's any transforms. This is the best way to work around that issue I found

packages/cli/package.json
Andarist: I don't think would even work in some of the `node_modules` layouts. If we are going to do this then we'd have to just reexport this stuff properly (but given its `.json`... I'm not sure what would actually work here). It's a shame `package.json#exports` don't support `"./entry": "some-module"` when `package.json#imports` support that...

packages/config/src/json-schema.ts
Andarist: I'd also prefer this to be closer to `scripts` given this isn't meant to be a part of what we publish

packages/config/package.json
```diff
@@ -18,8 +18,10 @@
     "./package.json": "./package.json",
     "./schema.json": "./schema.json"
   },
+  "scripts": {
+    "generate-json-schema": "node scripts/generate-json-schema.ts"
```
Andarist, line 22: do we call this anywhere in an automated fashion? we should make sure the schema is up to date in our CI workflow and before publishing the package

packages/config/schema.json
```diff
@@ -1,32 +1,70 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
   "type": "object",
   "properties": {
-    "changelog": {
+    "$schema": {
+      "type": "string"
+    },
```
Andarist: this looks like a bug

packages/config/src/rules.ts
```diff
@@ -0,0 +1,231 @@
...
+  ctx: Omit<FullContext, "warnings" | "errors">,
+): Pick<FullContext, "warnings" | "errors"> {
+  const errors: string[] = [];
+  const warnings: string[] = [];
+
+  for (const rule of rules) {
+    rule({
+      ...ctx,
+      errors,
+      warnings,
```
Andarist: nit: `fullCtx` could be built once before the loop

packages/config/src/parse.ts
```diff
@@ -0,0 +1,77 @@
...
+import type { Config } from "@changesets/types";
+import { getPackages, type Packages } from "@manypkg/get-packages";
+import { type BaseIssue, getDotPath, safeParse } from "valibot";
+import { normalizeWrittenConfig, WrittenConfigSchema } from "./config.ts";
+import { validateConfigByRules } from "./rules.ts";
+
+type ParseResult =
+  | {
+      config: Config;
+      warnings: string[];
```
Andarist, line 12: note for myself and others: we have 3 warnings - 2 about unmatched groups (fixed/linked) and a single deprecation warning

packages/config/src/parse.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export function validateConfig(json: unknown, packages: Packages): ParseResult {
+  // parse (...and validate)
+  const writtenConfigResult = safeParse(WrittenConfigSchema, json);
+  if (!writtenConfigResult.success) {
+    return {
+      config: null,
+      warnings: [],
+      errors: flattenIssues(writtenConfigResult.issues),
```
Andarist, line 32: If I'm reading things correctly, those errors would have a path prefix but the errors/warnings coming from our custom validation rules wouldnt. It would be great to normalize this and have the path in both

packages/config/src/defaults.ts
```diff
@@ -0,0 +1,24 @@
+import type { Config, WrittenConfig } from "@changesets/types";
+import { parse } from "valibot";
+// this requires that the package is built _after_ bumping versions before publishing
+import manifest from "../package.json" with { type: "json" };
+import { normalizeWrittenConfig, WrittenConfigSchema } from "./config.ts";
+
+export const defaultWrittenConfig: WrittenConfig = {
+  ["$schema" as never]: `https://unpkg.com/@changesets/config@${manifest.version}/schema.json`,
```
Andarist, line 8: this `as never` here - smart!

packages/config/src/defaults.ts
```diff
@@ -0,0 +1,24 @@
...
+let defaultConfig: Config | undefined;
+export const getDefaultConfig = (): Config => {
+  if (defaultConfig == null) {
+    defaultConfig = {
+      ...normalizeWrittenConfig({
+        packageNames: [],
+        writtenConfig: parse(WrittenConfigSchema, defaultWrittenConfig),
+      }),
+    };
+  }
```
Andarist: nit:

```suggestion
  defaultConfig ??= {
    ...normalizeWrittenConfig({
      packageNames: [],
      writtenConfig: parse(WrittenConfigSchema, defaultWrittenConfig),
    }),
  };
```

packages/config/src/defaults.ts
```diff
@@ -0,0 +1,24 @@
...
+  access: "restricted",
+};
+
+let defaultConfig: Config | undefined;
+export const getDefaultConfig = (): Config => {
+  if (defaultConfig == null) {
+    defaultConfig = {
+      ...normalizeWrittenConfig({
+        packageNames: [],
+        writtenConfig: parse(WrittenConfigSchema, defaultWrittenConfig),
```
Andarist: so I guess, this is responsible for filling in the defaults? it could be good to add a comment mentioning that

packages/config/src/defaults.ts
```diff
@@ -0,0 +1,24 @@
+import type { Config, WrittenConfig } from "@changesets/types";
+import { parse } from "valibot";
+// this requires that the package is built _after_ bumping versions before publishing
```
Andarist, line 3: Which is basically the default when working with Changesets. I'm not sure if it's really worth mentioning here but I'm also OK with it

packages/config/scripts/json-schema.test.ts
Andarist, line 1: once we call the target script on CI, this will kinda become redundant - those tests are smoke at best anyway

packages/config/src/parse.ts
```diff
@@ -0,0 +1,77 @@
...
+import { validateConfigByRules } from "./rules.ts";
+
+type ParseResult =
+  | {
+      config: Config;
+      warnings: string[];
+      errors: [];
+    }
+  | {
+      config: null;
```
Andarist: personally, I prefer to standardize on `undefined` over `null`

packages/config/src/parse.ts
```diff
@@ -0,0 +1,77 @@
...
+import { getPackages, type Packages } from "@manypkg/get-packages";
+import { type BaseIssue, getDotPath, safeParse } from "valibot";
+import { normalizeWrittenConfig, WrittenConfigSchema } from "./config.ts";
+import { validateConfigByRules } from "./rules.ts";
+
+type ParseResult =
+  | {
+      config: Config;
+      warnings: string[];
+      errors: [];
```
Andarist: in this case we could just make it to return `errors: undefined`

packages/config/src/parse.ts
```diff
@@ -0,0 +1,77 @@
...
+    writtenConfig,
+  });
+  if (errors.length !== 0) {
+    return { config: null, warnings, errors };
+  }
+
+  return { config, warnings, errors: [] };
+}
+
+export async function readAndValidateConfig(
```
Andarist: publicly this is exported as `readConfig` now - so it would be good to just match that with here

packages/config/src/index.ts
Andarist, line 1: I feel like this should export `validateConfig` too. This might be helpful to the bot's code

packages/config/src/utils.ts
```diff
@@ -0,0 +1,42 @@
+import type { Config, WrittenConfig } from "@changesets/types";
+import type { Packages } from "@manypkg/get-packages";
+import picomatch from "picomatch";
+
+export type FullContext = {
```
Andarist: maybe `FullValidationContext`? or just `ValidationContext`?

packages/config/src/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  paths: readonly string[],
+  patterns?: readonly string[],
+): string[] {
+  if (!patterns) return paths as string[];
+
+  const matchers = patterns.map((p) => picomatch(p, undefined, true));
+  return paths.filter((path) => {
+    if (path.includes("\\")) {
+      path = path.replace(/\\/g, "/");
+    }
```
Andarist, line 23: q: have we supported this before?

packages/errors/src/index.ts
```diff
@@ -6,12 +6,10 @@ export class GitError extends Error {
   }
 }
 
-export class ValidationError extends Error {}
```
Andarist, line 9: just thinking out loud - we could support `.code`+`.message` on this and gather all of the issues in an `AggregateError`. Although, that's kinda noisy with repeated stack traces in my testing. 

But simplifying this - custom properties display nicely in node, smth like:
```ts
var e = new Error('elo'); e.issues = [{ path: 'foo', issue: 'msg' }, { path: 'bar', issue: 'other msg' }]; throw e
```

But anyway, given everything is kinda really originating in a single place anyway... the current design is good too.

packages/config/src/parse.test.ts
```diff
@@ -0,0 +1,726 @@
...
+  validateConfig,
+} from "./parse.ts";
+
+const omit = <Obj extends Record<string, unknown>>(
+  obj: Obj,
+  keys: (keyof Obj)[] | string[],
+): typeof keys extends keyof Obj ? Omit<Obj, typeof keys> : Obj =>
+  Object.fromEntries(
+    Object.entries(obj).filter(([k]) => !keys.includes(k)),
+  ) as never;
```
Andarist: We are only using this to strip away `$schema` so we could really just use a simpler:
```ts
function strip$Schema<T extends { $schema?: unknown }>({ $schema, ...value }) {
  return value;
}
```

But do we even care about stripping this? I've noticed we are doing this strip before some snapshots - but those could just easily contain this property in the snapshot.

packages/config/src/parse.test.ts
```diff
@@ -0,0 +1,726 @@
+import path from "node:path";
+import { testdir } from "@changesets/test-utils";
+import type { Config, WrittenConfig } from "@changesets/types";
+import { getPackages, type Packages } from "@manypkg/get-packages";
```
Andarist:
```suggestion
import type { Config, Packages, WrittenConfig } from "@changesets/types";
import { getPackages } from "@manypkg/get-packages";
```

packages/config/src/parse.test.ts
```diff
@@ -0,0 +1,726 @@
...
+  ({
+    rootPackage: {
+      packageJson: rootManifest,
+      dir: cwd,
+      relativeDir: "./",
+    },
+    rootDir: cwd,
+    packages: [],
+    tool: { type: "pnpm" } as never,
+  }) as Packages;
```
Andarist, line 156:
```suggestion
const getDefaultPackages = (cwd: string): Packages =>
  ({
    rootPackage: {
      packageJson: rootManifest,
      dir: cwd,
    },
    rootDir: cwd,
    packages: [],
    tool: { type: "pnpm" },
  });
```

packages/cli/src/utils/read-config.ts
```diff
@@ -0,0 +1,33 @@
...
+    messages.push(c.red(error));
+  }
+
+  const logFn = errors.length === 0 ? log.error : log.warn;
+  if (messages.length !== 0) {
+    logFn(`- ${messages.join("\n- ")}`);
+  }
+
+  if (errors.length !== 0) {
+    throw new ExitError(1, { cause: new Error(messages.join("\n")) });
```
Andarist: I thought this would print the errors twice but we'll ignore this `cause` here:
https://github.com/changesets/changesets/blob/bd2725600fb88769263072a89d200f365cf61cb9/packages/cli/src/index.ts#L45-L48

Attaching the `.cause` is nice - but perhaps this could actually throw that removed `ValidationError`? It would give it more context. 

And I'm nitpicky here... but this message should just use:

```suggestion
    throw new ExitError(1, { cause: new Error(errors.join("\n")) });
```

1. coloring shouldn't affect this
2. warnings are not the cause of the error and they are part of the `messages` now

packages/cli/src/utils/read-config.ts
```diff
@@ -0,0 +1,33 @@
...
+    messages.push(c.yellow(warning));
+  }
+  for (const error of errors) {
+    messages.push(c.red(error));
+  }
+
+  const logFn = errors.length === 0 ? log.error : log.warn;
+  if (messages.length !== 0) {
+    logFn(`- ${messages.join("\n- ")}`);
+  }
```
Andarist: I feel like both diagnostic "levels" (warn/error) should just use different log levels and not be put into the same category here

packages/cli/src/commands/status/index.ts
```diff
@@ -22,12 +24,16 @@ export async function status(options?: StatusOptions) {
 
   const packages = await getPackages(cwd);
   await ensureChangesetFolder(packages.rootDir);
-  const config = await read(packages.rootDir, packages);
 
-  const releasePlan = await getReleasePlan(
```
Andarist, line 27: could this switch back to `getReleasePlan` now?

packages/cli/package.json
beeequeue: it was just for testing it with a pkr.pr.new publish :^)

packages/config/package.json
beeequeue: we have a snapshot test in `json-schema.test.ts` that runs against the generated file. if it changes and isn't updated the snapshot will fail.

packages/config/src/json-schema.ts
beeequeue: this file is also used in the schema snapshot test, so i want to keep it here.

packages/config/src/rules.ts
beeequeue: good catch!

packages/config/src/defaults.ts
beeequeue: there was a reason for using a getter function instead of just building it on initialization, but i can't remember it anymore...

it could've been something to do with not automatically building it if it isn't needed

beeequeue: added a jsdoc comment to `normalizeWrittenConfig`

packages/config/scripts/json-schema.test.ts
beeequeue: i would argue that we shouldn't update the schema in CI, but let the snapshots tests handle it for us. then all changes are visible in PRs

packages/config/src/parse.ts
beeequeue: other tools (valibot 😋) use empty arrays to allow easily checking `.length` which i agree with. the checks then become `config == null` or `errors.length !== 0` depending on preference and typescript nuances

packages/config/src/parse.ts
beeequeue: i prefer the more explicit name here since we also have the `readConfigFile` function in the same file, which only reads the file.

packages/config/src/utils.ts
beeequeue: yes, minimatch automatically handled windows slashes while picomatch (and the glob standard) doesn't, so any path should technically be normalized to unix slashes before matching

packages/config/src/parse.test.ts
beeequeue: it now gets automatically omitted by valibot after removing `$schema` from `WrittenConfigSchema`

packages/cli/src/utils/read-config.ts
beeequeue: i would say that `ValidationError` could be useful **IF** we expected anyone to run the CLI programmatically, which we don't. 
since the CLI can only exit and print errors we should use `ExitError`.

---

the `.cause` shows up in vitest (iirc) so it's helpful for bubbling up the error without surfacing it in the CLI.

---

👍

packages/cli/src/utils/read-config.ts
beeequeue: this is a detail for how the output will look, we can polish it when we polish the rest of the CLI visuals

packages/cli/src/commands/status/index.ts
beeequeue: i still don't like how it's just a wrapper around `assembleReleasePlan`.

- its benefits are that it automatically fetches data we may not have available
- but we already have all the parts we need in the step, so we may as well just use the wrapped function.

packages/config/src/parse.ts
beeequeue: true, having visual snapshots would be great for validating this

packages/config/src/config.ts
beeequeue: this was already a warning before and has a rule now `noPrivateTagWithoutPrivateVersion`

packages/config/src/json-schema.ts
Andarist: hm, then perhaps that snapshot should also live in the scripts? that would be a little bit unusual perhaps but 🤷‍♂️ we could also not have that as a snapshot test but rather be it validated separately on CI

not a big deal though, it's also fine-ish as it is

packages/config/scripts/json-schema.test.ts
Andarist: Fwiw, I have not meant that we should call it to somehow commit it back automatically - but rather that calling it on CI could be followed by checking if the git state is pristine or not (so really, kinda like a snapshot test but outside of Vitest)

packages/config/src/parse.ts
Andarist: `errors.length !== 0` doesn't quite get you nice control flow narrowings: [TS playground](https://www.typescriptlang.org/play/?ts=6.0.3#code/C4TwDgpgBAwg9gOwGYEsDmUC8UDeAoKKMAQwGMBrYtCAZwC4oAlCUuAJwBMAeG4NlBGgA0UAK4JyCOAHcEAPgDceAL5K8oSFACCWAlAA+uPYUKtk6BvHNolJkxDZt29KAG0AurZPK9h-HdNEVDQGBFEAG3CvOwcnNhdefkEPaKhVPDwkcVJgFEQoYFpgAEYACmIGLQBKI0IUJChygDpY5ybwiEFgAAsoAEJMbAAGGv8THqdpKAQIKYBRR3Zm1vimgCs4AVKqqq8fQmIms2CmkgoqWiUfdXBoACFdQj9jE2OLWCD0VMIVl3EOCCoGYcVL7Ay1AJvELTCJRF4-RbxBiJARoFIvdKZbK5fKFXgAJlKACMGHdRnp6o0iS1ETRyXYJjJprMoAs4sSaXEaOtNghtrs9GDqVDTmRKNQaFc8EA)

packages/config/src/parse.ts
```diff
@@ -0,0 +1,77 @@
...
+      config: null;
+      warnings: string[];
+      errors: string[];
+    };
+
+function flattenIssues(issues: BaseIssue<any>[]): string[] {
+  return issues.map((issue) => `${getDotPath(issue)}: ${issue.message}`);
+}
+
+export async function readConfigFile(cwd: string): Promise<ParseResult> {
```
Andarist: Applying this suggestion can potentially introduce some type errors - but this should get fixed (or we could use `any` here):
```suggestion
export async function readConfigFile(cwd: string): Promise<unknown> {
```

packages/config/src/parse.ts
Andarist: do we even need `readConfigFile`? it seems like it could just be inline - it's not a public export of this package. It's used in 2 tests but I'd be fine with inlining this there too. It's a pretty small util anyway. 

I feel like a mismatch between the internal and public name is confusing and if possible we shouldn't introduce such mismatches

packages/config/src/parse.ts
beeequeue: i've inlined the function, so it's no longer applicable

packages/config/src/parse.ts
beeequeue: true, as i said `config == null` also works for that reason

packages/config/scripts/json-schema.test.ts
beeequeue: i don't really see the point in having a snapshot test outside of vitest vs. inside of it

packages/config/src/parse.ts
Andarist: ye, right - but it's not symmetric. Handling errors has to usually be done first and, personally, I prefer to enter such branch by checking some error-related thing (which is explicit and direct) over checking the "negative" case for the success branch.

Andarist (approved):

packages/config/src/json-schema.ts
bluwy: I still prefer keeping this in the scripts and configure it to work. This isn't published as source code, and we don't want to accidentally do so, so it shouldn't be in src.

packages/config/src/parse.ts
beeequeue: i prefer to never write `undefined` unless it's required (e.g. optional parameters), but we can do that if you feel strongly about it

packages/config/src/parse.ts
beeequeue: i get the reasoning behind it 👍 will change it

bluwy (approved):
Thanks for cleaning this up. I really appreciate it! This looks good to me now.

PR #2016, opened 2026-05-16
Call CI workflow before publish attempt

Andarist:
It always bothered me a little bit that the CI and publish workflows were disassociated from each other. This is mostly redundant now because only green PRs should end up on main/next and we have merge queues and all to verify that... but it's still kinda implicit.

Given the publish only happens from time to time, I think it would make sense to introduce this one extra re-verification step into the mix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,9 +47,6 @@ jobs:
   lint-workflows:
     name: Lint workflows
     runs-on: ubuntu-latest
-    permissions:
-      actions: read # only required in private repos
-      security-events: write # allow writing security events
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -100,7 +97,7 @@ jobs:
 
       - name: Upload coverage
         uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
-        if: matrix.node_version == 24
+        if: matrix.node_version == 24 && github.event_name != 'workflow_call'
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_call:
   pull_request:
   # merge queue is required so all commits on target branches trigger this workflow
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -46,9 +47,6 @@ jobs:
   lint-workflows:
     name: Lint workflows
     runs-on: ubuntu-latest
-    permissions:
-      actions: read # only required in private repos
-      security-events: write # allow writing security events
```
bluwy, line 51: Do we not want zizmor to write security events?

.github/workflows/publish.yml
```diff
@@ -43,10 +43,16 @@ jobs:
...
+  ci:
     if: needs.version.outputs.hasChangesets == 'false'
     needs: version
+    permissions:
+      contents: read
+    uses: ./.github/workflows/ci.yml
+
+  publish:
+    name: Publish
+    needs: ci
```
bluwy, line 55: Should we set up like this instead? https://stackoverflow.com/a/64733705

Otherwise `ci.yml` is going to be executed twice from a push event, and possibly cancel because of the `concurrency` setup.

.github/workflows/ci.yml
Andarist: @beeequeue configured zizmor to use annotations and not security events. But i guess in this mode we actually should use security events cause there is not much to annotate? Or at least those annotations wouldnt easily popup in a PR (because we are not running in a PR context here)

.github/workflows/publish.yml
Andarist: > Should we set up like this instead? https://stackoverflow.com/a/64733705

I'm not sure if this would work because we rely on `merge_group` now and they kinda fire *before* stuff lands on `main`. But either way, decoupling it like this would be a little bit implicit (or at least not as discoverable) - which would hurt DX a little bit because the changeset workflow would always be "delayed".

> Otherwise ci.yml is going to be executed twice from a push event, and possibly cancel because of the concurrency setup.

Good point about concurrency - I don't quite know how that would work in this special mode. This needs testing. 

I'm fine with the CI workflow executing more than once in those scenarios though because it would only happen pre-publish (but not pre-version)

bluwy: I think we'll have to figure out the concurrency issue. I also think this will trigger often than not though because we execute when there's no changesets, not only before publish.

If we can get both sorted out I'm ok with doing this.

PR #2020, opened 2026-05-18
Fix cac parsing behavior

bluwy:
I moved the CLI instance to its own `cli.ts` file with its own `cli.tests.ts` file testing the argument parsing and options passed to the underlying command functions.

The more test I add, I found the argument parsing by cac is kinda bad, so I have to fix them together, including:
- `pre enter` and `pre exit` commands don't work. You need to merge into a single `pre` command and handle the positional arguments manually. (Dk why it worked before for me)
- Parsed numbers are casted back as string. cac (or specifically mri) tries to be smart and auto-cast to numbers where possible, which we don't want.

Nice to have changes:
- Remove aliases and `--` in the options so it's more accurate of what we pass to the command functions

---

I'm not particularly happy we have to do these normalization. If there's a (small) CLI library one day that can solve this cleanly, I'm happy to switch too.


packages/cli/src/cli.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          "--snapshot",
+          "pr-123",
+          "--snapshot-prerelease-template",
+          "{tag}-{commit}",
+        ],
+        options: {
+          snapshot: "pr-123",
+          snapshotPrereleaseTemplate: "{tag}-{commit}",
+        },
+      },
```
Andarist, line 97:
```suggestion
      {
        args: [
          "--snapshot",
        ],
        options: {
          snapshot: true,
        },
      },
      {
        args: [
          "--snapshot",
          "pr-123",
          "--snapshot-prerelease-template",
          "{tag}-{commit}",
        ],
        options: {
          snapshot: "pr-123",
          snapshotPrereleaseTemplate: "{tag}-{commit}",
        },
      },
```

Andarist: the aboe is just an extra test case

packages/cli/src/cli.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    ],
+  },
+];
+
+for (const { command, fn, cases } of tests) {
+  describe(`changeset ${command}`, () => {
+    for (const { args, options } of cases) {
+      test(`${args.join(" ") || "<no args>"}`, async () => {
```
beeequeue, line 183: maybe this can use `test.each` instead of a for loop?

packages/cli/src/cli.ts
```diff
@@ -0,0 +1,149 @@
+// this requires that the package is built _after_ bumping versions before publishing
```
beeequeue:
```suggestion
// this json import requires that the package is built _after_ bumping versions before publishing
```

not relevant to this pr, but i want this comment i wrote to be more specific

packages/cli/src/cli.test.ts
bluwy: Yeah I tried quite hard using that but it's very difficult to understand the string substitution syntax so I made it a for loop.

beeequeue: you can use arrays-of-arrays and object arrays as well:
https://github.com/changesets/changesets/blob/be4d2b47be1550cebc221258bb079a400e025494/packages/config/src/parse.test.ts#L399

Andarist (approved):

packages/cli/src/cli.test.ts
bluwy: The `${args.join(" ") || "<no args>"}` is still a bit hard to do with that syntax.

PR #2021, opened 2026-05-18
Disable guide line and update intro

bluwy:
image

For the intro message, I use the same format as the help message.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -3,8 +3,10 @@ import { format } from "node:util";
 import manifest from "@changesets/cli/package.json" with { type: "json" };
 import c from "@changesets/color";
 import { ExitError, InternalError } from "@changesets/errors";
-import { intro, log, outro } from "@clack/prompts";
+import { intro, log, outro, updateSettings } from "@clack/prompts";
 import { cac } from "cac";
+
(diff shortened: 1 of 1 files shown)


packages/cli/src/index.ts
```diff
@@ -135,8 +137,8 @@ try {
   cli.parse(process.argv, { run: false });
 
   // Do not show intro for --help and --version, which have no command name
-  if (cli.matchedCommandName != null) {
-    intro("🦋");
+  if (cli.matchedCommand?.name != null) {
```
bluwy, line 140: For some reason `matchedCommandName` is undefined if running `changeset` (not `changeset add`)

Andarist (approved):

beeequeue:
i'd love if we could do some kind of snapshot visual testing for the commands so we can verify how everything looks when we do changes like this

i looked around and checked out james' new [ansivision](https://github.com/43081j/ansivision) project but didn't come up with anything yesterday

bluwy:
I think we can invoke the CLI through a child process and pipe the output as a simple snapshot if we want to.

beeequeue:
i think we could miss prompts or other inbetween states if we just pipe it, or the output could just be too jumbled, but it's worth a shot

PR #2026, opened 2026-05-22
Validate snapshot prerelease versions

### LINKED ISSUE #419: The value passed to --snapshot is not validated
## Problem

Make a snapshot release with a custom snapshot tag that is not a valid prerelease tag. For instance: `changesets version --snapshot a_b`

The versions will be update to `0.0.0-a_b-<timestamp>`. However, publishing will fail because versions cannot contain a `_` char.

## Proposed solution

There are two possible solutions:
1. fail with a meaningfull error when `--shapshot` gets an invalida prerelease tag (semver.valid may be used for validation)
1. try to make the passed value valid. Convert invlid chars to `-`, make all characters lower case, etc.

puneetdixit200:
## Summary
- validate the composed snapshot prerelease before returning snapshot release versions
- fail early when `--snapshot` tags or snapshot prerelease templates create invalid semver, such as tags containing underscores
- add a regression test and release changeset for `@changesets/assemble-release-plan`

Fixes #419

## Testing
- `YARN_CACHE_FOLDER=/tmp/changesets-yarn-cache yarn jest packages/assemble-release-plan/src/index.test.ts --runInBand`
- `env -u NO_COLOR TERM=xterm-256color FORCE_COLOR=1 YARN_CACHE_FOLDER=/tmp/changesets-yarn-cache yarn jest --ci --color --runInBand`
- `YARN_CACHE_FOLDER=/tmp/changesets-yarn-cache yarn types:check`
- `YARN_CACHE_FOLDER=/tmp/changesets-yarn-cache yarn lint`
- `YARN_CACHE_FOLDER=/tmp/changesets-yarn-cache yarn format`
- `YARN_CACHE_FOLDER=/tmp/changesets-yarn-cache yarn build`
- `git diff --check`

### DIFF
--- a/packages/assemble-release-plan/src/index.ts
+++ b/packages/assemble-release-plan/src/index.ts
@@ -51,12 +51,25 @@ function getSnapshotSuffix(
       .replace(/[^\d]/g, ""),
   };
 
+  if (
+    placeholderValues.tag !== undefined &&
+    !isValidPrereleasePart(placeholderValues.tag)
+  ) {
+    throw new Error(
+      `Failed to compose snapshot version: snapshot tag "${placeholderValues.tag}" is not a valid semver prerelease.`
+    );
+  }
+
   // We need a special handling because we need to handle a case where `--snapshot` is used without any template,
   // and the resulting version needs to be composed without a tag.
   if (!template) {
-    return [placeholderValues.tag, placeholderValues.datetime]
+    const snapshotSuffix = [placeholderValues.tag, placeholderValues.datetime]
       .filter(Boolean)
       .join("-");
+
+    validateSnapshotSuffixFromTemplate(snapshotSuffix);
+
+    return snapshotSuffix;
   }
 
   const placeholders = Object.keys(placeholderValues) as Array<
@@ -69,7 +82,7 @@ function getSnapshotSuffix(
     );
   }
 
-  return placeholders.reduce((prev, key) => {
+  const snapshotSuffix = placeholders.reduce((prev, key) => {
     return prev.replace(new RegExp(`\\{${key}\\}`, "g"), () => {
       const value = placeholderValues[key];
       if (value === undefined) {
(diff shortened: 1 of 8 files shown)


packages/assemble-release-plan/src/index.ts
```diff
@@ -250,6 +259,9 @@ function assembleReleasePlan(
       refinedConfig.snapshot.prereleaseTemplate,
       refinedSnapshot
     );
+  if (snapshotSuffix) {
+    validateSnapshotSuffix(snapshotSuffix);
```
Andarist: this validation should be moved to `getSnapshotSuffix`

packages/assemble-release-plan/src/index.ts
```diff
@@ -82,6 +83,14 @@ function getSnapshotSuffix(
   }, template);
 }
 
+function validateSnapshotSuffix(snapshotSuffix: string): void {
+  if (!semverValid(`0.0.0-${snapshotSuffix}`)) {
+    throw new Error(
+      `Failed to compose snapshot version: "${snapshotSuffix}" is not a valid semver prerelease. Check the --snapshot tag and snapshot.prereleaseTemplate values.`
```
Andarist: 1. we should prevalidate `tag` at the beginning of `getSnapshotSuffix` and raise a targeted error message
2. then before returning the value from `getSnapshotSuffix` we should validate the full value once again and raise a targeted message about `prereleaseTemplate` not being valid

Andarist (changes requested):

packages/assemble-release-plan/src/index.ts
```diff
@@ -82,6 +83,14 @@ function getSnapshotSuffix(
   }, template);
 }
 
+function validateSnapshotSuffix(snapshotSuffix: string): void {
```
beeequeue: i feel like this check should be done in the config validation and value parsing in the `pre enter` command rather than here

what do you think @Andarist?

packages/assemble-release-plan/src/index.ts
puneetdixit200: Addressed in 0b7b6e4 by moving the validation into `getSnapshotSuffix`; the snapshot tag is prevalidated there before composing the suffix.

packages/assemble-release-plan/src/index.ts
puneetdixit200: Addressed in 0b7b6e4: `tag` is validated first, and the full composed suffix is validated before returning so invalid `snapshot.prereleaseTemplate` output gets a targeted error.

packages/assemble-release-plan/src/index.ts
puneetdixit200: Added the earlier checks in 2d31322:

- `snapshot.prereleaseTemplate` is now validated while reading config when the static template shape cannot produce a semver prerelease.
- `changeset version --snapshot <tag>` now validates the tag in CLI parsing before calling `version()`.

I kept the final composed suffix check in `getSnapshotSuffix` because `{tag}` and `{commit}` are runtime values, so config validation cannot fully prove the final prerelease string.

PR #2033, opened 2026-05-24
Fix publish git tag to default true

bluwy:
No changesets since the PR that caused the issue hasn't been released yet.

I've updated tests to cover the CLI parsing case, but didn't add any tests for directly calling the publish command.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -97,7 +97,7 @@ cli
   .example("  $ changeset publish --tag beta")
   .option("--otp <code>", "One time password for npm publish")
   .option("--tag <name>", "Publish with the given npm dist-tag")
-  .option("--git-tag", "Create a git tag for the release")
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli.ts
```diff
@@ -97,7 +97,7 @@ cli
   .example("  $ changeset publish --tag beta")
   .option("--otp <code>", "One time password for npm publish")
   .option("--tag <name>", "Publish with the given npm dist-tag")
-  .option("--git-tag", "Create a git tag for the release")
+  .option("--git-tag", "Create a git tag for the release", { default: true })
```
Andarist, line 100: Was this the only option affected?

bluwy: Yeah it's the only one that's non-falsy by default FWICT

Andarist (approved):

PR #2034, opened 2026-05-26
Precompute `@changesets/config` defaultConfig

bluwy:
Helps a little bit in startup time by precomputing it in build time


bluwy:
Looks like we also need to set the define in the vitest config. I don't really like this so closing.

PR #2039, opened 2026-05-27
Use internal bot for versioning

Andarist:


.github/workflows/publish.yml
```diff
@@ -33,11 +34,19 @@ jobs:
...
+        with:
+          client-id: ${{ vars.APP_CLIENT_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
       - name: Create or update release pull request
         id: changesets
         # https://github.com/changesets/action
         uses: changesets/action@63a615b9cd06ba9a3e6d13796c7fbcb080a60a0b # v1.8.0
         with:
+          token: ${{ steps.app-token.outputs.token }}
```
Andarist: nowadays, we support `token` input and `GITHUB_TOKEN` env... I'm not sure which one I prefer

.github/workflows/publish.yml
```diff
@@ -33,11 +34,19 @@ jobs:
...
+          commitMode: github-api
```
Andarist: we don't require signed commits - so we might not need this special mode 🤔

.github/workflows/publish.yml
bluwy: It's `github-token` btw. I prefer `github-token` as the source of truth though because that's how the official actions handle it too. But because we need to set `GITHUB_TOKEN` for `changelog-github`, I think we should still respect the existing value if it was already set. Just we don't document about `GITHUB_TOKEN`.

.github/workflows/publish.yml
Andarist:
```suggestion
```

.github/workflows/publish.yml
Andarist:
```suggestion
          github-token: ${{ steps.app-token.outputs.token }}
```

Andarist: > It's github-token btw

oops ;p

bluwy (approved):

PR #2041, opened 2026-05-27
Tweak add command cli styles

bluwy:
It's a bit hard to fix the spacing issue without the guide, but at the meantime the guide does look better in an interactive flow, so re-enabling it for now.

image

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -6,14 +6,19 @@ import { ExitError, InternalError } from "@changesets/errors";
 import { intro, log, outro, updateSettings } from "@clack/prompts";
 import { cli } from "./cli.ts";
 
-updateSettings({ withGuide: false });
-
 try {
   cli.parse(process.argv, { run: false });
 
+  const commandName = cli.matchedCommand?.name;
+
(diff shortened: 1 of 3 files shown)


beeequeue (approved):

packages/cli/src/index.ts
```diff
@@ -6,14 +6,19 @@ import { ExitError, InternalError } from "@changesets/errors";
...
 
-updateSettings({ withGuide: false });
-
 try {
   cli.parse(process.argv, { run: false });
 
+  const commandName = cli.matchedCommand?.name;
+
+  // Disable clack guide for non-interactive commands
+  if (commandName !== "add") {
```
Andarist: Perhaps `add` should *reenable* this mode? There are requests to support all inputs to add as CLI options. So I imagine that when somebody calls `add` and supplies all required~ inputs, it won't actually end up entering the interactive mode and that shouldnt end up with the guide

bluwy: That makes sense. I've updated it to disable globally, then re-enable it again. We need to decide before the intro though as the guide applies to it too.

Andarist (approved):

PR #1979, opened 2026-05-06
Add new docs site

bluwy:
Ready to merge. It's at a state where most drastic changes are done, and changes later can be made as PRs.

Netlify may need to be updated later to use `main` as the target branch for deployments.

NO NEED TO NORMAL MERGE THIS. JUST SQUASH MERGE.

https://changesets.dev

PR #2042, opened 2026-05-27
docs: intructions to mitigate GitHub Actions permissions issues

soleuniverse101:
Added intructions to mitigate GitHub Actions permissions issues in the Automating Changesets page.

This is related to https://github.com/changesets/action/issues/106 and is basically what [this comment](https://github.com/changesets/action/issues/199#issuecomment-2012315190) suggests to do

### DIFF
--- a/docs/automating-changesets.md
+++ b/docs/automating-changesets.md
@@ -38,10 +38,37 @@ In some cases, you may _want_ to merge a change without doing any releases (such
 
 ## How do I run the version and publish commands?
 
+### Automatic setup
+
 We have a [github action](https://github.com/changesets/action) that
 
 - creates a `version` PR, then keeps it up to date, recreating it when merged. This PR always has an up-to-date run of `changeset version`
 - Optionally allows you to do releases when changes are merged to the base branch.
+
+#### GitHub permissions issues
+
+You may encounter issues of changesets/action jobs failing for no reason and if that's the case, you can try the following steps :
+
+1. Add writing permissions for the GitHub access token used by the job that triggers changesets/action :
+
+```yml
+# In .github/workflows/release.yml or similar
+
+jobs:
+  release:
+    # ...
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      # ...
+      - name: Create Release PR
+        uses: changesets/action@v1
+```
+
+2. In your GitHub repository's settings, go to `Code and automation > Actions > General` and enable the option `Allow GitHub Actions to create and approve pull requests` at the end.
+
+### Manual setup
 
 If you don't want to use this action, the manual workflow we recommend for running the `version` and `publish` commands is:
 


docs/automating-changesets.md
```diff
@@ -38,11 +38,38 @@ In some cases, you may _want_ to merge a change without doing any releases (such
...
+### Automatic setup
+
 We have a [github action](https://github.com/changesets/action) that
 
 - creates a `version` PR, then keeps it up to date, recreating it when merged. This PR always has an up-to-date run of `changeset version`
 - Optionally allows you to do releases when changes are merged to the base branch.
 
+#### GitHub permissions issues
+
+You may encounter issues of changesets/action jobs failing for no reason and if that's the case, you can try the following steps :
```
beeequeue, line 50: i think github responds with a proper error message, something like "Actions is not allowed to open PRs", but i can't find any examples of it in my logs.

i think if we can find an example of this message it would be better to refer to it in some way that can be searched for, rather than "for no reason" :^)

docs/automating-changesets.md
```diff
@@ -38,11 +38,38 @@ In some cases, you may _want_ to merge a change without doing any releases (such
...
+
+```yml
+# In .github/workflows/release.yml or similar
+
+jobs:
+  release:
+    # ...
+    permissions:
+      contents: write
+      pull-requests: write
```
beeequeue, line 62:
```suggestion
      contents: write # for changesets to push tags
      pull-requests: write # for changesets to create/update prs
```

adding comments help explain why the permissions are needed, and some tools like zizmor enforce having them with correct configuration!

docs/automating-changesets.md
```diff
@@ -38,11 +38,38 @@ In some cases, you may _want_ to merge a change without doing any releases (such
...
+    steps:
+      # ...
+      - name: Create Release PR
+        uses: changesets/action@v1
```
beeequeue, line 66: we should probably go through all the docs at some point and replace all action versions with pinned hashes, even in examples

beeequeue (changes requested):
we have a [branch](https://github.com/changesets/changesets/pull/1979) for the new [changesets website](https://changesets.dev) that will replace the current markdown documentation!

could you apply these changes to that and recreate or rebase this PR onto it?

docs/automating-changesets.md
soleuniverse101: It does! You're right that the message should be more precise.
I noticed that there are 2 messages that are displayed, one if permissions are missing (something like `remote: Permission to xxx.git denied to github-actions[bot]`), the other one (*GitHub Actions is not permitted to create or approve pull requests*) if PRs aren't allowed for Github Actions in repository settings.

docs/automating-changesets.md
soleuniverse101: Is that to prevent supply chain attacks ? I can add a patch on the PR but isn't there a way to do that more automatically ?

I'm not sure you're going to want to update that manually each time a new version goes out

soleuniverse101:
@beeequeue I appreciate your suggestions! I reopened a new PR (#2046) with some fixes, feel free to check it out.

Also could you double check the hash I added to specify the changesets/action version ? I took it from the latest release (1.8.0) but I want to be sure it's the correct one.

PR #2043, opened 2026-05-28
Remove unused code

bluwy:
detected by knip. I didn't set up knip entirely as I imagine we only need to run this occasionally.


Andarist (approved):
I'd be OK with adding knip to the CI

PR #2045, opened 2026-05-28
refactor `flattenReleases` for readability and performance

beeequeue:
done as part of the release plan tests cleanup

replaces `.filter().forEach()` loop with a single for-of for performance and readability, and restructures a little bit of code to make it more readable

### DIFF
--- a/packages/assemble-release-plan/src/flatten-releases.ts
+++ b/packages/assemble-release-plan/src/flatten-releases.ts
@@ -56,8 +56,6 @@ export function flattenReleases(
       }
 
       release.changesets.push(changeset.id);
-
-      releases.set(name, release);
     }
   }
 
--- a/packages/assemble-release-plan/src/flatten-releases.ts
+++ b/packages/assemble-release-plan/src/flatten-releases.ts
@@ -55,9 +55,6 @@ export function flattenReleases(
         release.type = type;
       }
 
-      // Check whether the bumpType will change
-      // If the bumpType has changed recalc newVersion
-      // push new changeset to releases
       release.changesets.push(changeset.id);
 
       releases.set(name, release);
--- a/packages/assemble-release-plan/src/flatten-releases.ts
+++ b/packages/assemble-release-plan/src/flatten-releases.ts
@@ -1,11 +1,28 @@
-// This function takes in changesets and returns one release per
-// package listed in the changesets
-
 import { shouldSkipPackage } from "@changesets/should-skip-package";
-import type { Config, NewChangeset, Package } from "@changesets/types";
+import type {
+  Config,
+  NewChangeset,
+  Package,
+  VersionType,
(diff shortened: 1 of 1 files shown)


packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -1,64 +1,77 @@
...
+        name,
+        `Couldn't find package named "${name}" listed in changeset "${changeset.id}"`,
+      );
+
       // Filter out skipped packages because they should not trigger a release
       // If their dependencies need updates, they will be added to releases by `determineDependents()` with release type `none`
-      .filter(({ name }) => {
-        const pkg = mapGetOrThrowInternal(
-          packagesByName,
+      const packageIsSkipped = shouldSkipPackage(pkg, {
```
Andarist: skipped? isSkipped? `packageIsSkipped` is OK - but kinda verbose and we don't conflict here with any other local names

packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -1,64 +1,77 @@
...
 import { mapGetOrThrowInternal } from "./utils.ts";
 
+const changeTypes = {
+  major: 3,
+  minor: 2,
+  patch: 1,
+  none: 0,
+} as const;
+
+const changeTypeIsBiggerThanPrevious = (
```
Andarist: nit: I think this could be easily inlined without sacrificing the readability

Andarist (approved):

packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -1,64 +1,68 @@
...
+      }
+
+      // if this changeset's type overrides the previous one
+      if (changeTypes[type] > changeTypes[release.type]) {
+        release.type = type;
+      }
+
+      // Check whether the bumpType will change
+      // If the bumpType has changed recalc newVersion
+      // push new changeset to releases
```
Andarist: this isn't a new "problem" - but this comment is slightly misplaced and coul be fixed here

beeequeue: where should they even go? i can't follow the logic before or after the changes 😄

Andarist: hm, you are right - I assumed it would just be combined with this new preceding comment but this is just too stale already. Let's just remove this comment altogether

packages/assemble-release-plan/src/flatten-releases.ts
```diff
@@ -1,64 +1,68 @@
...
+      release.changesets.push(changeset.id);
+
+      releases.set(name, release);
```
bluwy: I think this can be removed too? Doesn't seem to have any effect

PR #2046, opened 2026-05-28
docs: intructions to mitigate GitHub Actions permissions issues

soleuniverse101:
Added intructions to mitigate GitHub Actions permissions issues in the Automating Changesets page.

This is related to https://github.com/changesets/action/issues/106 and is basically what [this comment](https://github.com/changesets/action/issues/199#issuecomment-2012315190) suggests to do

### DIFF
--- a/docs/automating-changesets.md
+++ b/docs/automating-changesets.md
@@ -38,10 +38,37 @@ In some cases, you may _want_ to merge a change without doing any releases (such
 
 ## How do I run the version and publish commands?
 
+### Automatic setup
+
 We have a [github action](https://github.com/changesets/action) that
 
 - creates a `version` PR, then keeps it up to date, recreating it when merged. This PR always has an up-to-date run of `changeset version`
 - Optionally allows you to do releases when changes are merged to the base branch.
+
+#### GitHub permissions issues
+
+You may encounter issues of changesets/action jobs failing for no reason and if that's the case, you can try the following steps :
+
+1. Add writing permissions for the GitHub access token used by the job that triggers changesets/action :
+
+```yml
+# In .github/workflows/release.yml or similar
+
+jobs:
+  release:
+    # ...
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      # ...
+      - name: Create Release PR
+        uses: changesets/action@v1
+```
+
+2. In your GitHub repository's settings, go to `Code and automation > Actions > General` and enable the option `Allow GitHub Actions to create and approve pull requests` at the end.
+
+### Manual setup
 
 If you don't want to use this action, the manual workflow we recommend for running the `version` and `publish` commands is:
 
--- a/docs/automating-changesets.md
+++ b/docs/automating-changesets.md
@@ -47,7 +47,12 @@ We have a [github action](https://github.com/changesets/action) that
 
 #### GitHub permissions issues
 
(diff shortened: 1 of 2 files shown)


docs/automating-changesets.md
```diff
@@ -38,11 +38,43 @@ In some cases, you may _want_ to merge a change without doing any releases (such
...
+    permissions:
+      contents: write # For changesets to create/update PRs
+      pull-requests: write # For changesets to push tags
+    steps:
+      # ...
+      - name: Create Release PR
+        uses: changesets/action@63a615b9cd06ba9a3e6d13796c7fbcb080a60a0b # v1.8.0
+```
+
+2. In your GitHub repository's settings, go to `Code and automation > Actions > General` and enable the option `Allow GitHub Actions to create and approve pull requests` at the end.
```
bluwy, line 74: For this section, can we simplify as steps to setup permissions? i.e. the workflow permissions and repo settings. I don't think we need to mention the errors here because if we already document what they need to do, they shouldn't see it again.

bluwy (commented):
To make changes to the site, you should edit [this file](https://github.com/changesets/changesets/blob/next-docs/site/guide/automating.md) instead. The `docs/` folder here is kept for link compatibility.

docs/automating-changesets.md
beeequeue: i suggested adding the error messages in the previous PR, as they are very likely to be searched for when people encounter the issue

bluwy: If we want to, we can keep them in a info block so it's searchable, but I'd much prefer to keep the guide focused on direct/positive steps on setting it up. We can keep troubleshooting info on a different page or in the action readme. The page here also needs proper setup docs later too, taking some docs from the action repo.

Or it'd be even better if we document this in the actions repo first and then figure out how to adopt it here. There's already issues and PRs of this in the actions repo, and it has the more complete setup guide.

soleuniverse101: @bluwy So I guess I should commit the changes to the appropriate file while removing the list of errors ? What's an info block and how should I add the errors to it ?

bluwy: Yes I'd prefer moving it to the file that's inside `site/`. I think I'll update this page further later on for better action setup guide, so I can adapt your docs here regarding setting up proper permissions.

An info block is a [VitePress markdown feature](https://vitepress.dev/guide/markdown#custom-containers), like:

```md
::: tip Troubleshooting
You may ...
:::
```

You can put and explain the errors inside the block if you'd like, then we can figure out how to organize the page later on. You can check how the site renders with `pnpm site:dev`

soleuniverse101: Done! I moved the errors at the end of the section and they're hidden by default, see how that looks like :

image

I would also like to note that while I replaced the changeset action version by the hash of v1.8.0 (latest as of writing), it's unstainable if you update changesets/action regularly so you may wanna look at if VitePress has any way to automatically handle that (maybe a Vue component that gets updated by an action triggered at the release of new versions ?).

Finally, should I revert the changes already made to `docs/automating-changesets.md` or is the file fine as it is ?

beeequeue:
looks like this PR got closed because we merged the branch it was based on, sorry!

could you re-open the PR against `main`? if you do you can also remove the changes to `docs/` and only keep the ones to `site/`

soleuniverse101:
No problem! I reopened it on PR #2194

PR #2047, opened 2026-05-31
Replace `npm info` with package manager aware registry query

### LINKED ISSUE #965: yarn changeset publish is using npm info instead of yarn npm info
## Affected Packages
Private package

## Problem
When trying to do a `yarn changeset publish`, it first gets info about your packages, but it does that via `npm info`. I'd like that to be `yarn npm info`. This is the only thing my repo requires to keep an `.npmrc` file around, everything else uses `.yarnrc.yml`.

## Proposed solution
Allow configuration of `info` to use in .changeset/config.json
```
{
    "info": "yarn npm info"
}
```

zbrydon:
# Issue
Resolves #965

# Problem

The `getPackageInfo` function in the publish command always uses `npm info` to query package metadata, regardless of which package manager the project is configured to use.

Projects can declare their package manager using either the `packageManager` field or the `devEngines.packageManager` field in package.json:

```json
{
  "packageManager": "pnpm@10.12.1"
}
OR
{
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": ">=10",
      "onFail": "error"
    }
  }
}
```

The devEngines field is an [OpenJS Foundation standard](https://github.com/openjs-foundation/package-metadata-interoperability-working-group/blob/main/devengines-field-proposal.md) for declaring package manager and runtime requirements during development.

Because `npm info` is always used, projects that rely on pnpm or Yarn-specific registry configuration and authentication may fail or behave incorrectly. Both pnpm and Yarn provide their own info commands that respect their configured registries and credentials.

The publish workflow should use the project's configured package manager when querying package metadata.

# What changed

* Added `getInfoTool(cwd)` to detect the project's package manager using `detect({ cwd })` and map it to:

  * `npm`
  * `pnpm`
  * `yarn-classic`
  * `yarn-berry`

* Updated `getPackageInfo` to use the appropriate package manager command:

  * **npm / pnpm**

    ```sh
    <pm> info <pkg> --registry=<url> --json
    ```
  * **Yarn Berry**

    ```sh
    yarn npm info <pkg> --json
    ```

    Registry configuration is read from `.yarnrc.yml`.
  * **Yarn Classic**

    ```sh
    yarn info <pkg> --registry=<url> --json
    ```

    Handles Yarn v1's `{ type: "inspect", data }` response format and treats `{ type: "error" }` as a package-not-found result.

* Extended `isEmptyResult` to recognize Yarn Classic error responses, ensuring the existing two-query fallback and E404 normalization logic continue to work correctly.

* Threaded `cwd` through:

  ```
  publishPackages
    → getUnpublishedPackages
      → infoAllow404
        → getPackageInfo
  ```

  so package manager detection always runs relative to the project root.

# Testing

  New unit tests cover:
  - pnpm: verify pnpm info is called instead of npm info
  - yarn classic error response: verify it is treated as "not found" and falls through to E404
  - yarn classic success response: verify the { type: "inspect", data } wrapper is stripped before returning

### DIFF
--- a/packages/cli/src/commands/publish/__tests__/npm-utils.test.ts
+++ b/packages/cli/src/commands/publish/__tests__/npm-utils.test.ts
@@ -260,4 +260,67 @@ describe("getPackageInfo", () => {
 
     expect(result).toEqual(packageData);
   });
+
+  it("uses yarn npm info when packageManager field specifies yarn berry", async () => {
+    const cwd = await testdir({
+      "package.json": JSON.stringify({
+        private: true,
+        packageManager: "yarn@4.0.0",
+      }),
+    });
+
+    mockSpawn.mockImplementation(((cmd: string, args: string[]) =>
+      Promise.resolve(
+        cmd === "yarn" && args?.[0] === "npm" && args?.[1] === "info"
+          ? spawnResult(
+              JSON.stringify({
+                name: "pkg-a",
+                version: "1.0.0",
+                versions: ["1.0.0"],
+              })
+            )
+          : spawnResult("", 1)
+      )) as any);
+
+    const result = await getPackageInfo(
+      { name: "pkg-a", version: "1.0.0" },
+      cwd
+    );
+
+    expect(result).toMatchObject({ name: "pkg-a" });
+    expect(
+      mockSpawn.mock.calls.some(
+        ([cmd, args]) =>
+          cmd === "yarn" && args?.[0] === "npm" && args?.[1] === "info"
+      )
+    ).toBe(true);
+    expect(
+      mockSpawn.mock.calls.some(
+        ([cmd, args]) => cmd === "npm" && args?.[0] === "info"
+      )
+    ).toBe(false);
+  });
+
+  it("throws when yarn classic returns unparseable non-JSON output", async () => {
+    // isYarnClassicError's JSON.parse throws on non-JSON output, returning false
+    // from the catch block. The output is treated as non-empty, so jsonParse
+    // is called next and propagates the SyntaxError.
+    const cwd = await testdir({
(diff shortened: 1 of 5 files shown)


beeequeue:
thanks for the PR!

i have been considering rewriting the info fetching functionality to just use fetch to call the npm registry, as it should speed things up considerably to not have to start new processes for each call.

however that would require us being able to know which registry to call for each package which is not the simplest thing...

Andarist:
> i have been considering rewriting the info fetching functionality to just use fetch to call the npm registry,

I looked into this when working on https://github.com/changesets/changesets/pull/1845 and just gave up. We could do this but it would require us to reverse-engineer quite a bit around how the CLIs handle `.npmrc` and alikes

beeequeue:
Sounds like something a new package could do...

zbrydon:
> thanks for the PR!
> 
> i have been considering rewriting the info fetching functionality to just use fetch to call the npm registry, as it should speed things up considerably to not have to start new processes for each call.
> 
> however that would require us being able to know which registry to call for each package which is not the simplest thing...

mhmmm true that would be miles nicer that what I've whacked together 😅 

I wonder whether something like https://github.com/npm/npm-registry-fetch would solve the problem 🤔

Andarist:
@zbrydon we plan to release v3 by the end of the month. I'm refactoring this stuff as we speak - and while it might not address everything yet... any further work should target that refactor. I'll open some PRs over the weekend and ping you for review/input.

Andarist:
This is going to be superseded by https://github.com/changesets/changesets/pull/2097 where I'm working on a more holistic revamp of what package manager commands are invoked (and how).

PR #2049, opened 2026-06-01
`.editorconfig` and `tsconfig.json` tweaks

beeequeue:
- **improve which types are available**
- **add .editorconfig file**

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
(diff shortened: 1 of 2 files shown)


.editorconfig
beeequeue, line 1: my webstorm kept adding files with CLRF line endings, this fixes it.

bluwy (approved):

PR #2051, opened 2026-06-01
fix: should-skip-package doesn't respect glob patterns in config

hmnd:
With the config `["*", "@icewolf/*", "!@icewolf/other-package"]`, I got the error below:

```
🦋  error ValidationError: Some errors occurred when validating the changesets config:
🦋  error The package "@icewolf/public-package" depends on the skipped package "@icewolf/private-package", but "@icewolf/public-package" is not being skipped. Please add "@icewolf/public-package" to the `ignore` option.
```

I traced this to a bug in `@changesets/should-skip-packages`. Unlike `@changesets/config`'s parser, which supports globs via micromatch, `@changesets/should-skip-packages` treated the ignore list as literal strings (`ignore.includes(...)`). With this pr, that check now also uses micromatch as expected.

### DIFF
--- a/.changeset/wise-timers-count.md
+++ b/.changeset/wise-timers-count.md
@@ -1,0 +1,5 @@
+---
+"@changesets/should-skip-package": patch
+---
+
+Use micromatch when checking ignored packages so ignore globs are respected.
--- a/packages/config/src/index.test.ts
+++ b/packages/config/src/index.test.ts
@@ -752,6 +752,34 @@ describe("parser errors", () => {
     `);
   });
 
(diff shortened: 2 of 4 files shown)


beeequeue:
@hmnd does this still happen on the v3 versions? if it does, could you create a new PR based on the new `main` branch?

bluwy:
I'm going to close this as it should be fixed [here](https://github.com/changesets/changesets/blob/0061ef0454997469d9fba52a05036702e8357102/packages/config/src/config.ts#L181-L183) (config expanded and passed to skip checks) and there hasn't been an update

PR #2053, opened 2026-06-04
add new release plan tests for dependent version bumps

beeequeue:
this PR adds a new test generator for testing dependent version bumps as a first step towards changing the default behavior of bumping peerDependencies.

it does _not_ touch the rest of the tests in the file (something i started doing but ends up touching every test in the file in some way).

i will open a pr that renames and moves tests, and removes duplicate tests later.

<details>
<summary>it generates test output that looks like this:</summary>

image

</details>

---

i did ask an AI to help restructure the test generation i wrote to be more readable, and used its output to create the solution we have here. i'm still not perfectly happy with how large it is...

these tests _could_ go into their own file, but i don't want to restructure the tests too much in this PR. the file is already 1500~ lines long, so what's a few more? :^)

if you have any better ideas, feel free to bring them up!

### DIFF
--- a/packages/assemble-release-plan/src/index.test.ts
+++ b/packages/assemble-release-plan/src/index.test.ts
@@ -1,4 +1,8 @@
+import { randomUUID } from "node:crypto";
 import { defaultConfig } from "@changesets/config";
+import type { Config, VersionType } from "@changesets/types";
+import { defu } from "defu";
+import { inc } from "semver";
 import { beforeEach, describe, expect, it } from "vitest";
 import { assembleReleasePlan } from "./index.ts";
 import { FakeFullState } from "./test-utils.ts";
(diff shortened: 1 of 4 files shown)


packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+      },
+    },
+    overrides: {
+      peer: { minor: { "^": "1.0.0" } },
+    },
+  });
+
+  describe("workspace: protocol works the same as without it", () => {
+    describeDependentBumping("range only", {
+      renderRange: (range) => `workspace:${range !== "=" ? range : "*"}`,
```
Andarist, line 1833: I feel this would be easier to reason about
```suggestion
      // render workspace:*, workspace:^, workspace:~
      renderRange: (range) => `workspace:${range === "=" ? "*" : range}`,
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+    });
+
+    describeDependentBumping("range+version", {
+      renderRange: (range) => `workspace:${range !== "=" ? range : ""}1.0.0`,
```
Andarist, line 1838:
```suggestion
      // render workspace:1.0.0, workspace:^1.0.0, workspace:~1.0.0
      renderRange: (range) => `workspace:${range === "=" ? "" : range}1.0.0`,
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+    Record<VersionType, Record<Range, string>>
+  >;
+
+  // oxfmt-ignore
+  const baseExpectations: ExpectationTable = {
+    dep: {
+      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      patch: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.1" },
+      minor: { "^": "1.0.0", "~": "1.0.1", "=": "1.0.1" },
+      major: { "^": "1.0.1", "~": "1.0.1", "=": "1.0.1" },
```
Andarist, line 1592:
```suggestion
      // normal dependent has to be bumped only when dependency falls out of allowed range
      // otherwise, installing the latest versions of dependent and dependency would result in duplicate dependency package in the tree
      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
      patch: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.1" },
      minor: { "^": "1.0.0", "~": "1.0.1", "=": "1.0.1" },
      major: { "^": "1.0.1", "~": "1.0.1", "=": "1.0.1" },
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+    },
+    dev: {
+      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      patch: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      minor: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      major: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
```
Andarist, line 1599:
```suggestion
      // the dependent should stay untouched, given the dev dependency doesn't affect production installations
      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
      patch: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
      minor: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
      major: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+    },
+    peer: {
+      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      patch: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.1" },
+      minor: { "^": "2.0.0", "~": "2.0.0", "=": "2.0.0" },
+      major: { "^": "2.0.0", "~": "2.0.0", "=": "2.0.0" },
```
Andarist, line 1605: This feels inconsistent/wrong. I don't get why we get `patch: {"=": "1.0.1"}` but `minor: {"=": "2.0.0"}`. 

We mostly care for how all of this affects future production installations and if the previously used peer dependency is still likely to be compatible with the peer dependent. 

`=` doesn't allow anything but that specific peer dependency version. So does it really matter by how much (patch/minor/major) the peer dependency changed? The overall effect is the same...

I feel like `=` should have the same result here. Either 1.0.1 or 2.0.0 - but it should be the same.

Buuut - this is truly hard to reason about so maybe I'm missing something about this situation and this is currently, in fact, correct 🤷

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+      major: { "^": "1.0.1", "~": "1.0.1", "=": "1.0.1" },
+    },
+    dev: {
+      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      patch: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      minor: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+      major: { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
+    },
+    peer: {
+      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
```
Andarist, line 1602: we should put some extra comment explaining the reasoning behind the expected values here - but this is hard to put in words and, I feel, it also depends on resolution of https://github.com/changesets/changesets/pull/2053/changes#r3357645017 (because the comment should account for that)

Thinking out loud... in general, a peer dependency and peer dependent coexist together. The peer dependent declares a compatibility contract but it's like a more soft one than when dealing with regular dependencies. The package manager can be more lax and should rely more heavily on hoisting. 

The current default rules... just kinda assume smth like:
1. patch is a bug fix so OK - we don't need to touch anything
2. but ofc with `=` the peer dependent has to be touched at all times
3. peer dependencies are hard to get right, Changesets are automating things so the "safest" option is to just assume any change to peer dependency results in a new peer dependent's major

1 and 3 are still at odds with each other a little bit but it's meant to be a reasonable heuristic and that's it.

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+   * Invalid keys are caught by the type checker, so there's no runtime
+   * "no matching case" guard to maintain.
+   */
+  function applyOverrides(
+    cases: Case[],
+    overrides: DeepPartial<ExpectationTable>,
+  ): Case[] {
+    return cases.map((c) => {
+      const expected = overrides[c.dep]?.[c.bump]?.[c.range];
+      if (expected == null || expected === c.expected) return c;
```
Andarist: It feels like maybe `expected === c.expected` should result in an assertion failure? When reading concrete `overrides` it would be great to be able to treat them as "this is different from base" - by allowing `expected === c.expected` here we lose that.

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+
+  type ExpectationTable = Record<
+    DepKind,
+    Record<VersionType, Record<Range, string>>
+  >;
+
+  // oxfmt-ignore
+  const baseExpectations: ExpectationTable = {
+    dep: {
+      none:  { "^": "1.0.0", "~": "1.0.0", "=": "1.0.0" },
```
Andarist, line 1589: The expected value here is the value of the final dependent version - but it reads a little bit weirdly because visually it looks like dependency range (kinda): `"^": "1.0.0"`. So my brain reads it differently. 

It's definitely just a matter of getting used to it but I wonder if a different format wouldn't make this more obvious. Maybe even just replacing the version with none/patch/minor/major? Then a line would use `VersionType` both in the row's key and in the expected values...
```ts
minor: { "^": "none", "~": "patch", "=": "patch" },
```

Would this be more readable? I feel it's a slight improvement but it's not like I feel super strongly about it

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+  type DeepPartial<T> = T extends object
+    ? { [P in keyof T]?: DeepPartial<T[P]> }
+    : T;
+
+  /** Semver range prefix written on the dependency. */
+  type Range = "^" | "~" | "=";
+  /** Which field the dependency lives in. */
+  type DepKind = "dep" | "dev" | "peer";
+
+  const RANGES = ["^", "~", "="] as const satisfies readonly Range[];
```
Andarist, line 1560: Do we actually need `as const satisfies` combo for anything here?

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+    peer: (setup, range) =>
+      setup.updatePeerDependency("pkg-a", "pkg-a-b", range),
+  };
+
+  function runCase(c: Case, config: Config, renderRange: RangeRenderer) {
+    /*
+     * Set up the test "workspace":
+     *   - `pkg-a` depends on `pkg-a-b` via dependency kind `dep`
+     *     using the range produced by `renderRange(range)`
+     *   - `pkg-a-b` is bumped by `bump`
```
Andarist, line 1682: nit: I don't get what `-a-` in `pkg-a-b` gets us here :p `pkg-b` would work fine - or `pkg-dependent`+`pkg-dependency`

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+    const dependency = releases.find((r) => r.name === "pkg-a-b");
+    expect(dependency).toBeDefined();
+    expect(dependency!.newVersion).toEqual(
+      c.bump !== "none" ? inc(BASE_VERSION, c.bump) : BASE_VERSION,
+    );
+
+    // The dependent bumped (or didn't) as expected.
+    // Some cases don't bump the dependent so we assert it stayed put with the fallback value.
+    const dependent = releases.find((r) => r.name === "pkg-a");
+    expect(dependent?.newVersion ?? BASE_VERSION).toEqual(c.expected);
```
Andarist: hm, this kinda works but perhaps this would be more explicit?
```ts
if (c.expected === oldVersion) {
  expect(dependent).toEqual(undefined);
} else {
  expect(dependent!.newVersion).toEqual(c.expected);
}
```

packages/assemble-release-plan/src/index.test.ts
Andarist: But I guess the current format makes the whole rendering thing easier right now so maybe we shouldnt complicate that part more

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+  });
+
+  describeDependentBumping("onlyUpdatePeerDependentsWhenOutOfRange: true", {
+    config: {
+      ___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH: {
+        onlyUpdatePeerDependentsWhenOutOfRange: true,
+      },
+    },
+    overrides: {
+      peer: { minor: { "^": "1.0.0" } },
```
Andarist, line 1803: this is a... surprisingly small diff but it looks correct to me

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+    },
+  });
+
+  describe("workspace: protocol works the same as without it", () => {
+    describeDependentBumping("range only", {
```
Andarist: I feel smth like "modifier only" would be more accurate

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+  describeDependentBumping("default config");
+
+  describeDependentBumping("updateInternalDependents: always", {
+    config: {
+      ___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH: {
+        updateInternalDependents: "always",
+      },
+    },
+    overrides: {
+      dep: {
```
Andarist, line 1786:
```suggestion
      // this essentially makes all non-none bumps result in a patch bump of a dependent
      dep: {
```

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+        patch: { "^": "1.0.1", "~": "1.0.1" },
+        minor: { "^": "1.0.1" },
+      },
+      peer: {
+        patch: { "^": "1.0.1", "~": "1.0.1" },
```
Andarist, line 1791: I'm honestly not sure if this is correct. Peer dependents are just... special and weird. I think we should end up with some clear guidelines for peer dependents handling somewhere and then that would help us to address this concern and the otner ones. Right now it feels loosey goosey a little bit and we don't have a good point of reference to even judge if this is correct or not.

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+        });
+      }
+    });
+  }
+
+  // ---- Suites ----
+
+  describeDependentBumping("default config");
+
+  describeDependentBumping("updateInternalDependents: always", {
```
Andarist, line 1779: It would be interesting also to see the results for:
```
    ___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH: {
        onlyUpdatePeerDependentsWhenOutOfRange: true,
        updateInternalDependents: "always"
      },
```

I feel that would potentially show us just how loose goosey this stuff is :p

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -1544,10 +1548,249 @@ describe("bumping peerDeps", () => {
...
+
+  function describeDependentBumping(
+    name: string,
+    {
+      config = {},
+      overrides = {},
+      renderRange = defaultRange,
+    }: SuiteOptions = {},
+  ) {
+    const mergedConfig = defu(config, defaultConfig) as Config;
```
Andarist: This merging is pretty simple actually (with the potential to just become a one-level spread before we release v3):
```suggestion
    const mergedConfig = {
      ...defaultConfig,
      ...config,
      ___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH: {
        ...defaultConfig?.___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH,
        ...config?.___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH,
      }
    }
```

So perhaps we don't really need the `defu` dependency

packages/assemble-release-plan/src/index.test.ts
beeequeue: while these comments are good, it would be nice(r) to have a proper specification for the behavior.

then we could link each test case/expected result to a part of the spec, and check that each specification is followed via something like [tracey](https://tracey.bearcove.eu/)

i've wanted to write one for a while and refactoring these tests are also a way for me to learn how our current algorithm works

packages/assemble-release-plan/src/index.test.ts
beeequeue: it _is_ inconsistent and odd, which is part of why the peerDep bump is the most requested change for Changesets over the last 5 years 😄

beeequeue: but how it should behave is not relevant to this PR. these tests are what i wanted to have _before_ we start debating and changing it so we can properly see the behaviors before and after

packages/assemble-release-plan/src/index.test.ts
beeequeue: as i said in the comment above, i think we can skip explaining the peerDep behaviors for now as we will use these tests to change them

packages/assemble-release-plan/src/index.test.ts
beeequeue: i did have that behavior implemented at some point but removed it, can't remember why anymore

packages/assemble-release-plan/src/index.test.ts
beeequeue: i think we can leave it and change it if we find it hard to read in the future

packages/assemble-release-plan/src/index.test.ts
beeequeue: it makes sure the type is `["^", "~", "="]` while still validating that it only contains correct values

packages/assemble-release-plan/src/index.test.ts
beeequeue: i considered renaming all the packages from `pkg-*` to more intuitive names like `app`, `cli`, `shared`, `utils`, `components`, etc. but decided to leave it for the next PR where i refactor all the other tests.

i could do it for these new tests though, to see how it feels

packages/assemble-release-plan/src/index.test.ts
beeequeue: yes, being able to fix/improve peer dependent bumping is the whole point of adding these tests 😄

packages/assemble-release-plan/src/index.test.ts
beeequeue: i'd rather have it in case the config object becomes more complicated, it's only a dev dep and is very very tiny

bluwy: I also prefer not using defu, especially since we have control over the config passed here. Maybe later we can think about some mergeConfig API from `@changesets/config`

bluwy (approved):
Aside from the defu thing, which I don't think is a big blocker, this looks good to me. Let's try to get this in soon so we can unblock further work with fixing the peer bump issue

PR #2057, opened 2026-06-06
Setup backport workflow and sync next setup

bluwy:
Syncs the workflow setup from next (e.g. moves zizmor to ci.yml), and setup the backport branch (v2) to run later.


.github/workflows/ci.yml
```diff
@@ -2,10 +2,13 @@ name: CI
...
-  push:
-    branches: [main]
+  # merge queue is required so all commits on target branches trigger this workflow
+  # despite lack of the push event trigger here
   merge_group:
-    branches: [main]
+    branches:
+      - main
+      - next
+      - v2
```
Andarist:
```suggestion
      - maintenance/v2
```

.github/workflows/publish.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - next
+      - v2
```
Andarist:
```suggestion
      - maintenance/v2
```

.github/workflows/ci.yml
```diff
@@ -2,10 +2,13 @@ name: CI
...
-  push:
-    branches: [main]
+  # merge queue is required so all commits on target branches trigger this workflow
+  # despite lack of the push event trigger here
   merge_group:
-    branches: [main]
+    branches:
+      - main
+      - next
+      - "maintenance/v[0-9]+"
```
bluwy, line 11: I found that you can make the glob stricter like this, which is also [used in vite](https://github.com/vitejs/vite/blob/092320bfe115e4d9952a7d172bc40cd3ec14a4c3/.github/workflows/ci.yml#L24-L25)

Andarist: 🤪 
```suggestion
      - "maintenance/v[1-9][0-9]*"
```

Andarist (approved):

.github/workflows/ci.yml
bluwy: Hmm, in some world maybe someone would want to backport to a v0 version 😄 I think I slightly prefer the shorter version now purely because it's less ugly (not that it was pretty to begin with), but if you prefer the more explicit syntax I'm ok-ish too (can send a followup PR)

Issue #1959, opened 2026-05-04
Overriding the default changelog formatter


kbariotis:
I started investigating on how to extend the default changelog as I would like to include more information on each record, e.g. ticket ID which is part of the branch name, specific tags attached to the GH PR etc. 

I bumped into this [guide](https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md) which explains things clearly but I noticed that is incomplete and I wanted to explicitly ask.

Is writing a custom changelog formatter encouraged? Is it possible that something will break in the future that I should be aware?

Thank you everyone in advance and thank you for all the work you've been doing 👍 :)

beeequeue:
the changelog formatting has not changed in a while and should be fine to use!

we are not planning any changes for it for v3, and have not started doing any proper planning for v4.

kbariotis:
Thank you @beeequeue! That clears things up. 

One more question. What I'm trying to do basically is make the changelog the source of truth of all relevation information so I can on a later time, traverse it and take actions, like post to a channel of a newly created version or check what changes were included between some versions.

It would be great to have a CLI option to:
- `> changesets history @example/pg@0.2.1` --> lists all changes of that version
- `> changesets history --since @example/pg@0.1.4` --> lists all changes since that version

As far as I understand, traversing the changelog in such a way is not supported right now right? I will have to parse it manually.

Does it make sense as a use case? Thank you again :) Happy to contribute to it btw.

beeequeue:
we recently removed markdown parsing in #1962 since the library we used for it was gargantuan in comparison for what we used it for 😅 

in the future we could bring in something smaller to add history parsing, but that would have to be for v4

it in the meantime, you could write your own little tool/script and we could potentially upstream it later! just make sure to try and use better libraries than this project did :^)

beeequeue:
closing this for now, but feel free to ask more questions regarding the formatting here, and you can use the new issue for more discussion regarding reading/parsing the changelogs!

PR #1850, opened 2026-02-19
feat(changelog-github): linkify issue references in changelog entries

mixelburg:
## Problem

When changeset summaries contain issue references like `#1234`, they render as clickable links in GitHub PR descriptions but appear as plain text in the generated `CHANGELOG.md` (#538).

## Solution

Added a post-processing step in `getReleaseLine` that converts plain `#<number>` references into proper markdown links pointing to the repo's issues page. References that are already inside markdown links are skipped to avoid double-linkification.

The links use the configured `repo` option and `GITHUB_SERVER_URL` environment variable, consistent with how PR and commit links are generated.

## Tests

Added two tests:
- Verifies issue references are converted to links
- Verifies existing markdown links are not double-linkified

All existing tests continue to pass.

Fixes #538

PR #1830, opened 2026-02-01
docs(changelog-github): add README with installation and usage guide

njg7194:
## Summary

Add comprehensive documentation for `@changesets/changelog-github` package.

## Changes

Added `README.md` to `packages/changelog-github/` with:

- **Installation instructions** - npm, yarn, pnpm
- **Configuration** - How to set up in `.changeset/config.json`
- **GitHub token setup** - For private repos and avoiding rate limits
- **GitHub Enterprise support** - Via `GITHUB_SERVER_URL` environment variable
- **Changeset summary features** - Documentation for `pr:`, `commit:`, and `author:` metadata parsing
- **Example output** - Shows what the generated changelog looks like

## Why?

As requested in #377, there was no documentation for this package making it difficult for users to understand how to install and use it.

Closes #377

Issue #1821, opened 2026-01-19
[feat] Changeset file templates


Filipoliko:
After years of using changesets in our teams, some developers would still write very simplistic changeset files and they would omit some important information. We decided to enforce a fixed changeset file structure to make sure, that every changeset file answers a set of specific questions (What? Why? How?). This slightly backfired as changeset files may squash multi-line markdown text into a single line when displayed in preview mode. Again, we enforced more strict structure to keep the resulting changelog files readable. But developers keep making slight mistakes every now that lead to poorly readable markdown preview.

I would like to propose a new feature, that would enable using templates with changesets.

Example workflow:

We have `./.changesets/templates/default.md` with following content.

```markdown
<ticket>
- **What?** <Explain implementation details>
- **Why?** <Explain reasons, why do we need these changes>
- **How?** <Migration guide>
```

- Developer runs `changeset add --template ./.changesets/templates/default.md`
- CLI prompts about changed packages and developer selects some and their corresponding bump types
- CLI generates the file (**without asking for any additional user input**) with the content of the template.
```markdown
---
"my-package": patch
---

<ticket>
- **What?** <Explain implementation details>
- **Why?** <Explain reasons, why do we need these changes>
- **How?** <Migration guide>
```
- Developer edits the generated file to add missing information.
- Commit&Push

---

Is this something, you would be interested in? I can start a PR, if this sounds interesting as we would benefit from this a lot.

teemingc:
We'd also like this for Svelte. We currently use the fork at https://github.com/svitejs/changesets-changelog-github-compact for many of our repositories but it is no longer maintained by Dominik. If we had a template option, we could emit more compact changesets such as (summarised by Claude):

Original:
```
\n\n- [#123](url) [`abc1234`](url) Thanks [@user](url)! - message text
```

Compact fork:
```
\n- message text ([#123](url))
```

**Changes:**
1. Single `\n` instead of `\n\n` between entries
2. Message comes first; PR/commit link moved to end as `(link)` suffix
3. PR and commit not both shown — PR takes precedence
4. "Thanks @user!" attribution removed
5. Issue autolinking narrowed: only linkifies `#123` inside `(fix #123)` / `(fixes #123)` / `(see #123)`, not all bare `#123` occurrences

PR #2059, opened 2026-06-06
Add `template` option to `changelog-github`

jycouet:
Reproduces `@svitejs/changesets-changelog-github-compact` (unmaintained) without the dependency.

## Summary

Opt-in `template` option for `@changesets/changelog-github`. It lets it render compact changelog lines straight from `.changeset/config.json`. With `template` unset, output is byte-for-byte unchanged.

Reproduce the compact look:

```json
"changelog": ["@changesets/changelog-github", {
  "repo": "<org>/<repo>",
  "template": "\n- {summary} {ref}"
}]
```

Tokens render bare (you write the surrounding spaces; empties self-erase and trailing whitespace is trimmed):

- `{summary}` - first line, links every `#n`
- `{ref}` - PR-xor-commit, parenthesized
- `{pr}`, `{commit}` - the bare links
- `{authors}` - contributor links (honors `disableThanks`); write the prefix yourself, e.g. `Thanks {authors}!`

**Examples:** template/output pairs live in the package README, each locked by a test so they can't drift.

## Decisions

1. **`disableThanks` stays.** It is not deprecated and keeps working in template mode (`{authors}` honors it; you can also drop attribution by omitting `{authors}`). No back-compat break.
2. **This stays in `@changesets/changelog-github` for now.** Lifting templating into core (so every generator just *contributes variables* like `{pr}` / `{commit}` / `{authors}`) is a follow-up, not this PR. `buildReleaseLineTokens` is already shaped for that role and can move up later.

## Potentially relevant issues / PRs

Refs are intentionally in backticks so this draft creates no cross-links.

### Fully addressed

- `#538` - clickable issue links in CHANGELOG.md -> `{summary}`
- `#689` - dropping the "Thanks X" noise for solo maintainers -> omit `{authors}` / `disableThanks`
- `#1596` - customizing the per-release-line message -> `template`

### Partially addressed (done / not done)

- `#1821` - templates: DONE = output changelog-line template; NOT = input changeset-file template (`changeset add --template`)
- `#1178` - modify the changelog format: DONE = per-line formatting; NOT = whole-file grouping / headers
- `#1057` - omit lines: DONE = per-line pieces (Thanks / PR / commit); NOT = removing section headers like "Patch Changes"
- `#1635` - shorter commit hash: DONE = changelog `{commit}` / `{ref}` (7-char); NOT = the snapshot prerelease version string
- `#1959` - override the default changelog formatter: DONE = per-line tokens; NOT = arbitrary PR tags / extra per-record metadata
- `#1248` - username links: DONE = drop attribution via `{authors}` / `disableThanks`; NOT = emitting plain unlinked `@user` for GitHub's Contributors section

### Adjacent - not solved here

These need the headers/grouping or a deeper hook in `apply-release-plan`, not the changelog functions:

- `#1807` - changelog section titles
- `#995` - a `getChangelogEntry` hook for custom generators
- `#1204` - a contributors-acknowledgement section
- `#1332` - customizing `generateChangesForVersionTypeMarkdown`
- `#1127` / `#1123` - a `getVersionLine` hook (version header line)

### Related context

- `#1850` (merged) added `linkifyIssueRefs`; `{summary}` reuses it
- `#1255` (merged) added `disableThanks`; `{authors}` honors it (kept, see Decisions)
- `#1236` (merged) moved away from short commit IDs after a collision bug; this PR's `{commit}` / `{ref}` render 7-char hashes to match compact - flagging in case it needs revisiting
- `#1830` (open) adds a `changelog-github` README; if it lands, document these options there too (currently in `config-file-options.md`)

## Verification

`@changesets/changelog-github` suite 56/56, `types:check` / `lint` / `format` clean. Existing snapshots unchanged (the back-compat guarantee).


pomerantsev:
It would be great to see this merged. I'd love to do less manual updating of [my changelog entries](https://github.com/pomerantsev/accented/blob/main/packages/accented/CHANGELOG.md) for every release.

beeequeue (approved):

packages/changelog-github/src/index.ts
```diff
@@ -42,6 +54,57 @@ async function readEnv() {
...
+        `Unknown changelog template token "{${name}}". Valid tokens are: ${RELEASE_LINE_TOKENS.map(
+          (t) => `{${t}}`,
+        ).join(", ")}.`,
+      );
+    }
+    return tokens[name];
+  });
+}
+
+export function buildReleaseLineTokens(args: {
```
bluwy: Can we move these functions to a `render-template.ts` file and not export them from `index.ts`? I prefer to not make them public for now until we're sure of the template syntax.

packages/changelog-github/src/index.ts
```diff
@@ -42,6 +54,57 @@ async function readEnv() {
...
+    ? `(${links.pull})`
+    : links.commit
+      ? `(${links.commit})`
+      : "";
+  return {
+    // `{summary}` links every `#n`; `{summaryHints}` links only `(fix #n)` hints
+    summary: summaryLinked,
+    summaryHints,
+    ref,
+    pr: links.pull ?? "",
```
bluwy: I think we should name this to `pull` for now so it matches the internal name.

packages/changelog-github/src/index.ts
```diff
@@ -153,24 +216,36 @@ const changelogFunctions: ChangelogFunctions = {
             .join(", ")
         : links.user;
 
+    const linkOpts = { serverUrl: GITHUB_SERVER_URL, repo };
+    const summaryLinked = linkifyIssueRefs(firstLine, linkOpts);
+    const summaryHints = linkifyIssueHints(firstLine, linkOpts);
```
bluwy: Let's remove the `linkifyIssueHints` and `summaryHints` handling. The initial intention of `-compact` implementing this is because we dont autolink issues before, so it did a more reserved replacement.

Now that we implemented direct/more aggresive replacement and we haven't received much reports about it, let's keep using `linkifyIssueRefs` only instead.

docs/config-file-options.md
```diff
@@ -164,6 +164,48 @@ You would specify our github changelog generator with:
...
+Examples (for a change with PR `#123`, commit `abc1234`, author `@alice`, summary `fix the thing`):
+
+| `template`                                          | rendered line                                                                      |
+| --------------------------------------------------- | ---------------------------------------------------------------------------------- |
+| `"\n- {pr} {commit} Thanks {authors}! - {summary}"` | `- [#123](url) [abc1234](url) Thanks [@alice](url)! - fix the thing` (the default) |
+| `"\n- {summaryHints} {ref}"`                        | `- fix the thing ([#123](url))` (the compact form)                                 |
+| `"\n- {summary} (thanks {authors}!)"`               | `- fix the thing (thanks [@alice](url)!)`                                          |
+| `"\n- {summary} {pr}"`                              | `- fix the thing [#123](url)`                                                      |
+
+#### Replacing `@svitejs/changesets-changelog-github-compact`
```
bluwy: Let's remove this section.  Even though I use the package too, I don't think it's popular enough to require a migration guide here.

bluwy (changes requested):
I prefer if we mark or document the `template` option as experimental, and recommend pinning the version if used.

Right now there's things like:
1. If replacements like `{authors}` is empty, is going to leave an awkward `Thanks  !` rendered text. Similarly it doesn't handle conditional parenthesis and spaces.
2. `{ref}` has builtin parenthesis right now so it's easier to be compatible with `-compact`. I don't think it should have parenthesis builtin but we can leave it as is for now if we mark as experimental.
3. Eventually we should use `renderTemplate` for the default behaviour too. But it's not possible to handle the above right now without better templating syntax. So I'm ok with not doing this as well as long as it's experimental.

packages/changelog-github/src/index.ts
jycouet: yes, less intermediate things 👍

packages/changelog-github/src/index.test.ts
```diff
@@ -1,6 +1,11 @@
 import { parseChangesetFile as parse } from "@changesets/parse";
 import { describe, expect, it, test, vi } from "vitest";
 import changelogFunctions from "./index.ts";
+import {
+  renderTemplate,
+  buildReleaseLineTokens,
+  RELEASE_LINE_TOKENS,
+} from "./render-template.ts";
```
bluwy: Please move these tests to their own `render-template.test.ts` as well.

packages/changelog-github/src/index.ts
```diff
@@ -3,14 +3,17 @@ import path from "node:path";
 import util from "node:util";
 import { getInfo, getInfoFromPullRequest } from "@changesets/get-github-info";
 import type { ChangelogFunctions } from "@changesets/types";
+import { buildReleaseLineTokens, renderTemplate } from "./render-template.ts";
+
+const RX_ISSUE_REF = /\[.*?\]\(.*?\)|\B#([1-9]\d*)\b/g;
```
bluwy: Can this mean named `ISSUE_REF_REGEX` instead? It took me a while to figure out RX means regex.

packages/changelog-github/src/render-template.ts
```diff
@@ -0,0 +1,51 @@
+// NOTE: the `template` option is experimental and these helpers are intentionally
+// not exported from `index.ts`. The token syntax may change in a patch release.
+
+const RX_TOKEN = /\{(\w+)\}/g;
```
bluwy: Same for this name

bluwy (commented):
The base branch needs to be updated to `main` as well. Shouldn't have any big conflicts by directly switching.

docs/config-file-options.md
```diff
@@ -164,6 +164,31 @@ You would specify our github changelog generator with:
 
 If you want to disable thank you messages, add `"disableThanks": true` to the options.
 
+#### `template` (optional string, experimental)
```
bluwy: This should be documented in changelog-github actually, since the docs here talks about the `changelog` option generically. Can you create a very simple README in `changelog-github/README.md`, setup something like from `@changesets/get-version-range-type` and then copy this option docs to the readme?

Doesn't need to be perfect for now. We'll sort out the docs altogether again later.

trueberryless:
Do we need a next-docs PR for this?

bluwy:
Yeah we can do that after this is merged

bluwy (approved):

packages/changelog-github/README.md
```diff
@@ -0,0 +1,47 @@
+# @changesets/changelog-github
+
+[![Open on npmx.dev](https://npmx.dev/api/registry/badge/version/@changesets/changelog-github?name=true)](https://npmx.dev/package/@changesets/changelog-github)
+[![View changelog](https://npmx.dev/api/registry/badge/version/@changesets/cli?color=229fe4&value=View+changelog&label=+)](./CHANGELOG.md)
+
+A changelog entry generator for GitHub that links to commits, PRs and users.
```
trueberryless: Can we write `"authors"` in the README instead of `"users"`?

```suggestion
A changelog entry generator for GitHub that links to commits, PRs and authors.
```

I think it's okay that we leave `users` in the code, but to the public, `authors` makes more sense I think.

trueberryless (commented):
Sorry for the late review. I know you want to merge already. Just a small docs nitpick here:

jycouet:
@Andarist conflics fixed 👍
I updated the PR description also to reflect latest changes

Maybe you will want to link later some issues properly _(from PR description?)_ to notify them?

trueberryless (approved):
Amazing 😍
Everything LGTM 🚀 

Thanks @jycouet for your wonderful contribution. I'm looking forward to playing around with this option!

Issue #614, opened 2021-07-13
Add dry run option to CLI publish command


raypatterson:
## Affected Packages

* @changesets/cli

## Problem

* It's not possible to verify the end to end publishing configuration without publishing

## Proposed solution

```shell
[package manager] changeset publish --dry-run
```

Andarist:
This is a very good idea - I've actually lately created a duplicate of this issue: https://github.com/atlassian/changesets/issues/645

I probably won't have time to work on this at the moment though. An implementation provided by the community would be awesome.

dkozma:
@Andarist It looks like #669 can resolve this issue, and has been open for a few years, is there any way that can be merged in?

denbite:
just wanted to drop a note here since this PR has been open for quite a while and it doesn’t look like it’s getting merged anytime soon

for anyone looking for a workaround to simulate a dry run in the meantime, here’s a solution that’s worked for me

```bash
pnpm changeset status --output sim.json
```

this command will create a `sim.json` file like the following

```json
{
  "changesets": [
    {
      "releases": [
        {
          "name": "pkg-a",
          "type": "major"
        },
        {
          "name": "pkg-b",
          "type": "minor"
        }
      ],
      "summary": "Description",
      "id": "rotten-flies-butter"
    }
  ],
  "releases": [
    {
      "name": "pkg-a",
      "type": "major",
      "oldVersion": "1.3.1",
      "changesets": [
        "rotten-flies-butter"
      ],
      "newVersion": "2.0.0"
    },
    {
      "name": "pkg-b",
      "type": "minor",
      "oldVersion": "1.6.8",
      "changesets": [
        "rotten-cases-dance"
      ],
      "newVersion": "1.7.0"
    }
  ]
}
```

hope this helps someone 👌

raypatterson:
I've been using [verdaccio](https://verdaccio.org)

Andarist:
I'd like to understand the use case better before committing to this. What a dry run even is in this context? Could somebody write down what's the goal is and how it is different from calling `changeset version`?

raypatterson:
Thanks for the question and all the work on this tool @Andarist !

The dry run reports what _will_ happen without actually _doing_ anything.

Maybe this [PR](https://github.com/microsoft/beachball/pull/951) from the beachball tool could be a point of reference?

shakiba:
Use case: similar to npm publish dry-run, to validate what would be published without actually publishing.

Andarist:
[...]

PR #2061, opened 2026-06-06
Add `publish-plan` command

### LINKED ISSUE #200: Add way to output a 'release info' thing

Andarist:
This kinda addresses https://github.com/changesets/changesets/issues/614 (albeit in a different way)

closes https://github.com/changesets/changesets/issues/200

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -106,7 +106,7 @@ cli
 
 cli
   .command("publish-plan", "Show packages that are ready to publish or tag")
-  .option("-o, --output <file>", "Output the publish plan as JSON to a file")
+  .option("-o, --output <file>", "Output the publish plan as JSON to a file [experimental]")
   .action(async (options) => {
     normalizeOptions(options);
     const { publishPlan } = await import("./commands/publish-plan/index.ts");
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -105,6 +105,15 @@ cli
   });
 
 cli
+  .command("publish-plan", "Show packages that are ready to publish or tag")
+  .option("-o, --output <file>", "Output the publish plan as JSON to a file")
+  .action(async (options) => {
+    normalizeOptions(options);
+    const { publishPlan } = await import("./commands/publish-plan/index.ts");
+    await publishPlan(options);
+  });
+
+cli
   .command("status", "Show the changesets that currently exist")
   .example("  $ changeset status --verbose")
   .option("--since <branch>", "Show changesets since the provided git ref")
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -106,7 +106,10 @@ cli
 
 cli
   .command("publish-plan", "Show packages that are ready to publish or tag")
-  .option("-o, --output <file>", "Output the publish plan as JSON to a file [experimental]")
+  .option(
+    "-o, --output <file>",
+    "Output the publish plan as JSON to a file [experimental]",
+  )
   .action(async (options) => {
     normalizeOptions(options);
     const { publishPlan } = await import("./commands/publish-plan/index.ts");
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/publish-plan/__tests__/getPublishPlan.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      registry: "https://registry.npmjs.org",
+    });
+    mockedGit.tagExists.mockResolvedValue(false);
+    mockedGit.remoteTagExists.mockResolvedValue(false);
+
+    const result = await getPublishPlan(cwd, {
+      access: "restricted",
+      ignore: [],
+      privatePackages: { version: true, tag: true },
+    } as any);
```
beeequeue: you can use `satisfies Partial<Type> as any/never` here to still get type checking while also allowing an invalid object

packages/cli/src/commands/publish-plan/getPublishPlan.ts
```diff
@@ -0,0 +1,189 @@
...
+      }
+    } else {
+      alreadyPublishedCount++;
+    }
+  }
+
+  if (packagesToPublish.length !== 0) {
+    log.info(
+      `
+These packages will be published as they were not found on npm:
```
beeequeue:
```suggestion
These packages will be published as they were not found in the registry:
```

packages/cli/src/commands/publish/index.ts
```diff
@@ -68,36 +67,25 @@ To resolve this exit the pre mode by running ${c.cyan("changeset pre exit")}.
   }
 
   const config = await readConfig(packages);
-  const tagPrivatePackages =
-    config.privatePackages && config.privatePackages.tag;
+  const plan = await getPublishPlan(packages.rootDir, config, {
+    tag: releaseTag,
+  });
```
beeequeue, line 72: am i missing where `publish` re-uses plans created beforehand, or does it not do that (yet)?

Andarist: That’s implemented in: https://github.com/changesets/changesets/pull/2063

packages/cli/src/commands/publish-plan/getPublishPlan.ts
```diff
@@ -0,0 +1,189 @@
...
+          `${c.gray("└")} will be published to ${c.cyan("latest")} rather than ${c.cyan(preState.tag)} as it will be its first published version.`,
+        );
+      }
+    } else {
+      alreadyPublishedCount++;
+    }
+  }
+
+  if (packagesToPublish.length !== 0) {
+    log.info(
```
Andarist, line 124: should stuff like this be logged when `--output` gets used?

packages/cli/src/cli.ts
```diff
@@ -104,6 +104,15 @@ cli
     await publish(options);
   });
 
+cli
+  .command("publish-plan", "Show packages that are ready to publish or tag")
+  .option("-o, --output <file>", "Output the publish plan as JSON to a file")
```
bluwy:
```suggestion
  .option("-o, --output <file>", "Output the publish plan as JSON to a file [experimental]")
```

Can we mark like experimental, internal, or something here so we're free to change the output shape in the future?

bluwy (commented):
LGTM. Just one small nit/preference to keep the output as internal.

packages/cli/src/commands/publish-plan/getPublishPlan.ts
bluwy: I think we should keep it for now so it's consistent with `changeset status`, but we can revisit this later if it's better to remove.

bluwy (approved):

PR #2062, opened 2026-06-06
Add `pack` command

Andarist:


packages/cli/package.json
```diff
@@ -53,12 +53,14 @@
     "launch-editor": "^2.13.2",
     "package-manager-detector": "^1.6.0",
     "semver": "^7.8.0",
+    "tar-stream": "^3.2.0",
```
bluwy: We could try https://github.com/ayuhito/modern-tar instead which is smaller

packages/cli/src/commands/pack/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        const args =
+          publishTool.name === "pnpm"
+            ? ["pack", "--json", "--pack-destination", packagesDir]
+            : ["pack", publishDir, "--json", "--pack-destination", packagesDir];
+        const execCwd = publishTool.name === "pnpm" ? publishDir : pkg.dir;
+        const { exitCode, stdout, stderr } = await exec(
+          publishTool.name,
+          args,
+          { nodeOptions: { cwd: execCwd } },
+        );
```
bluwy, line 117: I have wrote an abstraction for this before if you want to reuse: https://github.com/publint/publint/tree/master/packages/pack#pack. Also ok to implement it here if you prefer too.

packages/cli/package.json
Andarist: I don't mind revisiting this choice. I mainly picked `tar-stream` because it is used by both pnpm and yarn berry.

packages/cli/src/commands/pack/index.ts
Andarist: I'd be OK with using something here but I think probably it's best to keep it in-house until the feature stabilizes.

Most notable diffs with your implementation:
- package manager support range. Changesets should improve its support wholesale - reusing something that handles package managers better just for `pack` would be weirdly inconsistent with how we handle package managers during `publish` etc
- in here, I've used `--json` to get clear access to the created filename. It seems your implementation tries to infer it from the file list in the destination dir.

packages/cli/src/commands/pack/index.ts
```diff
@@ -0,0 +1,200 @@
...
+
+        const publishTool = await getPublishTool(pkg.dir);
+        const publishDir = pkg.packageJson.publishConfig?.directory
+          ? path.resolve(pkg.dir, pkg.packageJson.publishConfig.directory)
+          : pkg.dir;
+        const args =
+          publishTool.name === "pnpm"
+            ? ["pack", "--json", "--pack-destination", packagesDir]
+            : ["pack", publishDir, "--json", "--pack-destination", packagesDir];
+        const execCwd = publishTool.name === "pnpm" ? publishDir : pkg.dir;
```
bluwy: Shouldn't `execCwd` here be always `pkg.dir` since `args` above handles the publishDir passing already? And because pnpm respects `publishConfig.directory` automatically.

packages/cli/src/commands/pack/index.ts
```diff
@@ -0,0 +1,200 @@
...
+        const { exitCode, stdout, stderr } = await exec(
+          publishTool.name,
+          args,
+          { nodeOptions: { cwd: execCwd } },
+        );
+
+        if (exitCode !== 0) {
```
bluwy, line 119: We could also pass `throwOnError` to tinyexec to error on non-0 code, but this is also fine I think if we want better error messages, if that's your intent.

packages/cli/src/commands/pack/index.ts
```diff
@@ -0,0 +1,200 @@
...
+        const tarballFilename = getTarballFilename(stdout.toString());
+
+        if (!tarballFilename) {
+          throw new Error(
+            `Failed to determine tarball filename for ${release.name}`,
+          );
+        }
+        const integrity = await getIntegrity(
+          path.join(packagesDir, tarballFilename),
+        );
```
bluwy, line 147: What's the integrity here for? If there's a malicious actor I suppose they can also update the integrity in publish-plan.json together. So is it only to prevent users from editing the tarball somehow? There's also nothing stopping them from editing the publish-plan.json though.

packages/cli/src/commands/pack/index.ts
```diff
@@ -0,0 +1,200 @@
...
+        tarball: packedRelease.tarball,
+      };
+    }),
+  );
+
+  await fs.writeFile(
+    path.join(outputDir, "publish-plan.json"),
+    JSON.stringify(
+      {
+        version: CURRENT_PUBLISH_PLAN_VERSION,
```
bluwy, line 188: I think it's ok to not version this as we should expect this to only be passed around between the same CLI version, especially considering that `publish-plan` marks the output file as experimental. But this can be discussed later.

bluwy (dismissed):

packages/cli/src/commands/pack/index.ts
Andarist: > but this is also fine I think if we want better error messages, if that's your intent.

yeah, that was the intention here

packages/cli/src/commands/pack/index.ts
Andarist: This is intended to be used as a consistency check - not a security measure. If you think this is totally redundant, I'd be OK removing this altogether

packages/cli/src/commands/pack/index.ts
Andarist: I agree that plan should only be produced and consumed by the matching CLI version. This is intended as a precaution and nothing more - it gives us means to throw a better~ error and that's it.

packages/cli/src/commands/pack/index.ts
Andarist: Good catch. I recall introducing `execCwd` *somewhere* on purpose (maybe not here) - but I can't find that right now anywhere and it's certainly not correct here.

Edgarruiz8585:
+ > - [x] ruizariasedgarmanuel@gmail.com

packages/cli/src/commands/pack/index.ts
bluwy: Hmm yeah let's keep it for now then for extra checks.

bluwy (approved):

PR #2063, opened 2026-06-06
Add `publish --from-pack-dir` packed artifact

Andarist:


packages/cli/src/cli.ts
```diff
@@ -97,6 +97,7 @@ cli
   .example("  $ changeset publish --tag beta")
   .option("--otp <code>", "One time password for npm publish")
   .option("--tag <name>", "Publish with the given npm dist-tag")
+  .option("--from <file>", "Publish from a packed artifact tarball")
```
beeequeue: it feels kind of bad to have to run it multiple times for each tarball, is there no better way?

glob matching would be a bit insecure, but maybe we could use a previously generated publish plan?

Andarist: With the current design this receives the tarball that contains packed packages + `publish-plan.json`. So you don't have to run this multiple times, you just hand off it a single artifact and it does the rest

beeequeue: i would prefer it accepting the plan file directly, having an opaque tarball feels bad. it would be better if it was a file with a custom extension, but still not _good_

Andarist: There was a desire to be able to publish packages from a different job than the packing one, smth like:
```yml
jobs:
  pack:
    runs-on: ubuntu-latest
    permissions:
      contents: read

    steps:
      - name: Pack packages
        uses: changesets/action/pack@v1
        with:
          destination-path: ./dist

      - name: Upload package artifact
        uses: actions/upload-artifact@v4
        with:
          name: packed-packages
          path: ./dist

  publish:
    needs: pack
    runs-on: ubuntu-latest
    permissions:
      contents: read
      id-token: write

    steps:
      - name: Download package artifact
        uses: actions/download-artifact@v4
        with:
          name: packed-packages
          path: ./dist

      - name: Publish packages
        run: changeset publish --from ./dist/changeset-pack.tgz
```

The main goal of that is to split permissions between the jobs. Also, at the end of the day - this is just a monorepo-aware version of `pm pack`+`pm publish` workflow. Any tarball is opaque~ but this is what package managers support natively.

beeequeue: i just don't see why we need to create and use a tarball for this in the changesets code, the upload-artifact action could do the exact same thing if we pass it the correct paths, and download-artifact automatically extracts it, allowing us to use the plan file directly

packages/cli/src/commands/publish/__tests__/releaseCommand.test.ts
bluwy, line 1: Should this be named `publishCommand.test.ts` or simply `index.test.ts` to reflect the file being tested?

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -252,23 +252,20 @@ async function internalPublish(
     // we specifically don't want any other output to interfere with the delegated auth flow
     const child =
       publishTool.name === "pnpm"
-        ? exec("pnpm", ["publish", ...publishFlags], {
+        ? exec("pnpm", ["publish", opts.target, ...publishFlags], {
```
bluwy: Could we do a `path.relative` for the `opts.target` here so that it's passed a relative path or empty string? It probably doesn't matter technically, but it feels better to me to call the command as if how one would call manually in practice.

packages/cli/src/commands/publish/index.ts
```diff
@@ -38,19 +42,30 @@ export interface PublishOptions {
...
     options?.tag && options.tag.length > 0 ? options.tag : undefined;
-  const preState = await readPreState(packages.rootDir);
+  const preState = !artifactDir
+    ? await readPreState(packages.rootDir)
+    : undefined;
+
+  if (artifactDir && releaseTag) {
+    log.error("Releasing under custom tag is not allowed in artifact mode.");
+    throw new ExitError(1);
+  }
```
bluwy, line 68: Just an observation, it doesn't look nice that the tag can be configured in different places now: publish-plan, pack, and publish. (Actually `pack` is missing a `tag` arg and I think it should need one)

We dont have to address this now, but I think we should figure something later so it's more predictable where tag should be passed to.

Andarist: Yep, this is what we have discussed on Discord recently. I want to remove `tag` and `access` from the publish plan entirely - and I'll try to drive towards that with follow up PRs

bluwy: Sounds good 👍 didn't connect this to the issue you discussed on discord, but if you have a plan for this then we can follow up later

packages/cli/src/commands/publish/npm-utils.ts
Andarist: sure, I pushed out such change

packages/cli/src/commands/publish/__tests__/releaseCommand.test.ts
Andarist: I agree, this was messy - I moved those tests

bluwy (approved):

PR #2065, opened 2026-06-06
Simplify auth state handling

Andarist:


packages/cli/src/commands/publish-plan/getPublishPlan.ts
```diff
@@ -67,33 +74,34 @@ export async function getUnpublishedPackages(
...
-            )
-          ) {
-            publishedState = "only-pre";
+
+          if (preState != null) {
+            if (
+              response.pkgInfo.versions &&
+              !response.pkgInfo.versions[
+                `${semverParse(pkg.packageJson.version)!.major}.${semverParse(pkg.packageJson.version)!.minor}.${semverParse(pkg.packageJson.version)!.patch}`
+              ]
```
beeequeue: could this be extracted to a named variable? i find it hard to see what it means

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -177,19 +115,22 @@ export function getPackageInfo(
           );
         }
       } else if (packageJson.publishConfig?.registry) {
+        // in pnpm `publishConfig.registry` is the only supported registry value and it's a strong publish-time override
+        // to emulate pnpm publish-time precedence when querying via pnpm info, we force it as a scoped override
```
beeequeue: technically, pnpm has 5 different options for this across 4 different files, (including `package.json`), but supporting all of those is way out of scope for us

but i do believe `publishConfig.registry` is prioritized highest as you wrote

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -259,50 +200,35 @@ type InternalPublishResult =
...
-  // Due to a super annoying issue in yarn, we have to manually override this env variable
-  // See: https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633
-  const envOverride = {
-    [scope ? `npm_config_${scope}:registry` : "npm_config_registry"]: registry,
-  };
-
-  if (requiresDelegatedAuth(twoFactorState)) {
+  if (process.stdin.isTTY && authState.shouldDelegate) {
     // it's not easily controllable but ideally no other work should happen until this is done
     // we specifically don't want any other output to interfere with the delegated auth flow
```
beeequeue, line 241: we could use `p-queue` or `ciorent` to guarantee this in the same thread at least

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -326,28 +252,20 @@ async function internalPublish(
   // in the delegated mode we don't need the json output
   // as we won't be handling the auth errors
```
beeequeue, line 275: should we split the two modes into functions? i think it would make the flow easier to follow

e.g. `publishInteractive`, `publishNonInteractive`?

packages/cli/src/utils/types.ts
```diff
@@ -1,5 +1,4 @@
-export type TwoFactorState = {
+export type AuthState = {
   token: string | undefined;
```
beeequeue: i think this should be called something like `otpCode` to distinguish it from npm tokens

packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -18,142 +19,130 @@ export type PublishedResult = {
...
-      packagesByName.get(release.name)!,
-      packages,
-      access,
-      artifactDir,
-      twoFactorState,
-    ),
-  );
-
-  if (!hasToDelegate && releases.length > 1) {
-    const p = progress({ max: releases.length });
```
beeequeue, line 99: it's a shame we lose the progress bar when not having to delegate 😞

packages/cli/src/utils/types.ts
```diff
@@ -1,5 +1,4 @@
...
-  isRequired: boolean;
-  allowConcurrency?: boolean;
+  shouldDelegate: boolean;
```
bluwy: Can we add a comment explaining what "delegate" means here?

packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -18,142 +19,130 @@ export type PublishedResult = {
...
+    if (artifactDir) {
+      target = resolve(artifactDir, release.tarball!.path);
+    } else if (publishTool.name === "pnpm") {
+      // pnpm supports `publishConfig.directory` natively. We have to let it resolve it on its own.
+      // Otherwise we'd risk it re-resolving from within the `publishConfig.directory` itself
+      // but original untouched relative paths in `publishConfig.directory` would not even point to correct locations anymore.
+      target = packagesByName.get(release.name)!.dir;
+    } else {
+      const pkg = packagesByName.get(release.name)!;
+      // npm, yarn classic and berry don't support `publishConfig.directory` natively.
```
bluwy, line 112: I'd much prefer we consider removing support of this as this feels non-standard, but we can discuss this later.

packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -18,142 +19,130 @@ export type PublishedResult = {
...
+        // it wouldn't actually be particularly workspaces-aware until npm 9 (until https://github.com/npm/cli/pull/4372).
+        // So we'd have to call npm from the root with a nested package target (essentially what we still do now)
+        // because .npmrc would resolve in respect to cwd and not the target package and that's what we wanted.
+        // Nowadays, this isn't as important as .npmrc lookup is workspace-aware and would find the root .npmrc just fine even if invoked from the package directory.
+        //
+        // However, there are still 2 reasons why we prefer to set it to the root:
+        // 1. it's a consistent approach that works across package managers and it's also a good location when publishing packed artifacts
+        // 2. it's important not to call npm from a directory that is an actual workspace (as per the workspaces configuration) and not from, for example, publishConfig.directory
+        //    because npm only resolves to the root's .npmrc for actual workspaces and not for arbitrary subdirectories of the root.
+        cwd: packages.rootDir,
```
bluwy: I would really prefer we call in the workspace directory instead. I don't think no1 is really important as it's a poor implementation from old npm versions. no2 I don't think is an issue if we use `pkg.dir` here?

Also typo?
```suggestion
        // However, there are still 2 reasons why we prefer to set it to the root:
        // 1. it's a consistent approach that works across package managers and it's also a good location when publishing packed artifacts
        // 2. it's important to call npm from a directory that is an actual workspace (as per the workspaces configuration) and not from, for example, publishConfig.directory
        //    because npm only resolves to the root's .npmrc for actual workspaces and not for arbitrary subdirectories of the root.
        cwd: packages.rootDir,
```

packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -18,142 +19,130 @@ export type PublishedResult = {
...
+      token: process.env.PNPM_CONFIG_OTP,
+      shouldDelegate: false,
+    };
+  }
+  if (process.env.NPM_CONFIG_OTP) {
+    return {
+      token: process.env.NPM_CONFIG_OTP,
+      shouldDelegate: false,
     };
   }
```
bluwy: Is `_CONFIG_OTP` some kind of pattern people use and supported by npm/pnpm CLIs?

packages/cli/package.json
```diff
@@ -51,7 +51,6 @@
     "cac": "^7.0.0",
     "import-meta-resolve": "^4.2.0",
     "launch-editor": "^2.13.2",
-    "package-manager-detector": "^1.6.0",
```
bluwy, line 55: Nice to see this can be refactored away!

packages/cli/src/commands/publish/publishPackages.ts
Andarist: yeah, that is s a thing in both - Yarn Berry also support a variable like this

packages/cli/src/commands/publish/npm-utils.ts
Andarist: We already have a local `createPromiseQueue` util. And we even wrap publishes with it ;p there is still some risk that some other async routine~ could log something while a single delegated publish is in flight though.

packages/cli/src/utils/types.ts
beeequeue: it could be named better as well, maybe something like `requiresUserPrompt` or `requiresInteractive`?

beeequeue (approved):
this is still way too complicated for my liking...

i would love to extract all the logic into PM-specific utilities to help organize it better, in a later PR

packages/cli/src/commands/publish/npm-utils.ts
Andarist: it's certainly complicated - but I just don't want to refactor this all too much right now

packages/cli/src/commands/publish/publishPackages.ts
Andarist: I agree, it would be good to get it back in some form later on. If it wouldn't already be fuzzy enough which registry we should query... `pnpm` doesn't even seem to have any command that we could use to even check for this (they don't have `pnpm profile` right now)

packages/cli/src/commands/publish/publishPackages.ts
Andarist: I looked into this. There are situations and tools that require supporting this. If we'd drop this, we would just break the compatibility with them and I don't think it's particularly worth it.

packages/cli/src/commands/publish/publishPackages.ts
Andarist: I juggled this so we call it with `cwd: pkg.dir`

PR #2066, opened 2026-06-07
Ignore agent files as changesets

### LINKED ISSUE #1906: [FEAT]: Allow ignoring non-changeset `.md` files in `.changeset/` DIR

bluwy:
closes #1906
supersedes and closes #1912

As mentioned in #1906, we hardcode the common list of files to ignore for now. We can likely expand on this non-breakingly later.

### DIFF
--- a/packages/read/src/index.ts
+++ b/packages/read/src/index.ts
@@ -6,7 +6,7 @@ import type { NewChangeset } from "@changesets/types";
 
 // Files in `.changeset` that should not be considered as changesets. We may
 // revisit this if it becomes difficult to maintain the common list of files.
-const ignoredMdFiles = ["README.md", "AGENTS.md", "CLAUDE.md", "GEMINI.md"];
+const ignoredMdFiles = [/^README\.md$/i, "AGENTS.md", "CLAUDE.md", "GEMINI.md"];
 
 async function filterChangesetsSinceRef(
   changesets: Array<string>,
@@ -51,7 +51,9 @@ export async function readChangesets(
     (file) =>
       !file.startsWith(".") &&
       file.endsWith(".md") &&
-      !ignoredMdFiles.includes(file),
+      !ignoredMdFiles.some((pattern) =>
+        typeof pattern === "string" ? pattern === file : pattern.test(file),
+      ),
   );
 
   const changesetContents = changesets.map(async (file) => {
--- a/packages/read/src/index.ts
+++ b/packages/read/src/index.ts
@@ -4,6 +4,10 @@ import * as git from "@changesets/git";
 import { parseChangesetFile } from "@changesets/parse";
 import type { NewChangeset } from "@changesets/types";
 
+// Files in `.changeset` that should not be considered as changesets. We may
+// revisit this if it becomes difficult to maintain the common list of files.
+const ignoredMdFiles = ["README.md", "AGENTS.md", "CLAUDE.md", "GEMINI.md"];
+
 async function filterChangesetsSinceRef(
   changesets: Array<string>,
   changesetBase: string,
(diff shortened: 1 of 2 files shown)


packages/read/src/index.ts
```diff
@@ -47,7 +51,7 @@ export async function readChangesets(
     (file) =>
       !file.startsWith(".") &&
       file.endsWith(".md") &&
-      !/^README\.md$/i.test(file),
+      !ignoredMdFiles.includes(file),
```
Andarist: we probably need to make a similar change [here](https://github.com/changesets/bot/blob/35e5d07accb3546312232a789de92153bad83ec2/get-changed-packages.ts#L121-L126) and [here](https://github.com/changesets/bot/blob/35e5d07accb3546312232a789de92153bad83ec2/index.ts#L111-L121) (or maybe only in one of those places, I'm surprised I have found 2 similar ones in the bot's code)

packages/read/src/index.ts
```diff
@@ -47,7 +51,7 @@ export async function readChangesets(
     (file) =>
       !file.startsWith(".") &&
       file.endsWith(".md") &&
-      !/^README\.md$/i.test(file),
```
Andarist, line 50: The previous check was case-insensitive, we should maintain that

bluwy: Ah my bad, will fix that

bluwy: Kept the readme case-insensitive. The rest aren't as they probably need to be all-caps by convention.

packages/read/src/index.ts
```diff
@@ -4,6 +4,10 @@ import * as git from "@changesets/git";
 import { parseChangesetFile } from "@changesets/parse";
 import type { NewChangeset } from "@changesets/types";
 
+// Files in `.changeset` that should not be considered as changesets. We may
+// revisit this if it becomes difficult to maintain the common list of files.
+const ignoredMdFiles = [/^README\.md$/i, "AGENTS.md", "CLAUDE.md", "GEMINI.md"];
```
Andarist, line 9: nit: I'd probably go with this:

```suggestion
const ignoredMdFiles = [/^README\.md$/i, /^(AGENTS|CLAUDE|GEMINI).md$];
```

This would allow us to avoid the string/regexp distinction at the consumer side

Andarist (approved):

packages/read/src/index.ts
bluwy: I think I prefer distinct items for each entries than combining, so it's a list of special files that we're maintaining. Otherwise we make each file a regex, but feels a bit off 🤔 The regex/string check isn't too bad for me.

bluwy: I'll merge this in but happy to discuss this still if you'd like

PR #2070, opened 2026-06-08
Preserve package.json formatting when update

### LINKED ISSUE #1434: npm changeset version command formats the package.json

bluwy:
fix #1434

Parse the package.json and edit it by the AST range to preserve formatting. It uses `jsonc-parser` to parse and apply edits.

### DIFF
--- a/packages/apply-release-plan/src/edit-json.test.ts
+++ b/packages/apply-release-plan/src/edit-json.test.ts
@@ -88,3 +88,20 @@ it("throws when a key path does not exist", () => {
     ]);
   }).toThrow('Key path "version" not found in JSON');
 });
+
+it("throws on malformed JSON", () => {
+  const jsons = [`{{{`, ``, `{//comment\n"version":"1.0.0"}`];
+
+  expect.assertions(jsons.length);
+
+  for (const json of jsons) {
+    expect(() => {
+      editJson(json, [
+        {
+          keys: ["version"],
+          value: "2.0.0",
+        },
+      ]);
+    }).toThrow(/Failed to parse JSON/);
+  }
+});
--- a/packages/apply-release-plan/src/edit-json.test.ts
+++ b/packages/apply-release-plan/src/edit-json.test.ts
@@ -63,6 +63,19 @@ it("preserves formatting", () => {
   );
 });
 
+it("preserves bizarre formatting", () => {
+  const json = `{\n\t\n\n\n"a"\n:\t\t\n{"b":{"c":"^1.0.0"}}\n\n\n,  \t  \n\t"name":"pkg-a"\n\n\n\n\n}\t\t\t\t`;
+
+  const result = editJson(json, [
+    {
+      keys: ["a", "b", "c"],
+      value: "^2.0.0",
+    },
+  ]);
+
+  expect(result).toEqual(json.replace(`"^1.0.0"`, `"^2.0.0"`));
+});
+
 it("throws when a key path does not exist", () => {
   const json = `{"name":"pkg-a"}`;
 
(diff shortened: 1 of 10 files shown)


packages/apply-release-plan/src/edit-json.ts
```diff
@@ -0,0 +1,47 @@
+import { parse, type CstValueNode } from "json-cst";
```
Andarist: how does this compare to `jsonc-parser`? I experimented with that one in: https://github.com/changesets/changesets/pull/1753

packages/apply-release-plan/src/edit-json.ts
```diff
@@ -0,0 +1,47 @@
...
+    const valueNode = getValueNode(parsed.root, op.keys);
+    if (!valueNode) {
+      throw new Error(`Key path "${op.keys.join(".")}" not found in JSON`);
+    }
+    nodesToUpdate.push({ node: valueNode, value: op.value });
+  }
+
+  // Update nodes in reverse order to avoid altering offsets of subsequent runs
+  nodesToUpdate.sort((a, b) => b.node.range.start - a.node.range.start);
+  for (const update of nodesToUpdate) {
```
Andarist: We should just use a regular for loop here, in the reverse direction ;p

packages/apply-release-plan/src/edit-json.ts
```diff
@@ -0,0 +1,47 @@
...
+ * must exist in the JSON for this implementation.
+ */
+export function editJson(
+  json: string,
+  operations: EditJsonOperation[],
+): string {
+  const parsed = parse(json);
+
+  const nodesToUpdate: { node: CstValueNode; value: unknown }[] = [];
+  for (const op of operations) {
```
Andarist: It feels a `.map` would be more idiomatic here

packages/apply-release-plan/src/edit-json.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      value: "^2.0.0",
+    },
+  ]);
+
+  expect(result).toMatchInlineSnapshot(
+    `"{"name":"pkg-a","version":"2.0.0-longer-than-before.0","dependencies":{"pkg-b":"^2.0.0"}}"`,
+  );
+});
+
+it("preserves formatting", () => {
```
Andarist, line 51: an extra test with some bizarre formatting would be good to have here

packages/apply-release-plan/src/edit-json.ts
Andarist: I specifically wonder how both would approach `package.json5`. We don't quite support it... but technically pnpm does 😂 (alongside beautiful `package.yaml`)

Andarist (commented):
I like it!

packages/apply-release-plan/src/edit-json.ts
bluwy: `jsonc-parser` seems to work fine too, like `@humanwhocodes/momoa`, it's only a bit large (200+ KB vs 31 KB). But I'm fine with anything if we prefer something more tested and popular.

I would choose to not support .json5 or .yaml until one day npm would also support, maybe 😄 Vite also intentionally not support it.

packages/apply-release-plan/src/edit-json.ts
bluwy: We can't exactly do that since the order of operations and order of the keys to edit could jump around

Andarist: Ah, ok - for some reason I have thought those were already pre-sorted, just in the ascending order.

beeequeue (approved):

Andarist (approved):

PR #2073, opened 2026-06-08
Show private in changeset add

### LINKED ISSUE #1706: Detect and show (private) packages when listing packages for `changeset add` command.
I acknowledge that there is this feature: https://github.com/changesets/changesets/issues/1383

My use case is different. We want to **typically** avoid bumping private package versions, however we also sometimes want to start publishing a private package.

A use case for this is when we are developing a new package in our monorepo. Sometimes a package needs to remain `private` in the `main` branch to ensure it does not get published, however when we are ready to publish that package, we create a changeset for that package.

As part of our automated tooling, we will remove the `private` field when a changeset is discovered for that package.

We also explored having teams remove the `private` field as part of their PR, however we have the following side effect:

1. Remove `private` field in PR
2. Add changeset for package in PR
3. Merge PR into `main`.

At this point, the release workflow will attempt to publish the package using an incorrect version number because we are still waiting for the `release` pull request to set the correct version number.

Effectively, what we want is this:

```
$ pnpm changeset add
◯ changed packages
  ◯ @org/package1
  ◯ @org/package2
  ◯ @org/package3 (private)
◯ unchanged packages
  ◯ @org/package4
  ◯ @org/package5 (private)
```

This way we can clearly identify private packages and exclude them. But in the situation where we WANT to start publishing this package, we include it.

Alternatively: 
```
$ pnpm changeset add
◯ changed packages
  ◯ @org/package1
  ◯ @org/package2
◯ changed private packages
  ◯ @org/package3
◯ unchanged packages
  ◯ @org/package4
  ◯ @org/package5 (private)
```
or: 
```
$ pnpm changeset add
◯ changed packages
  ◯ @org/package1
  ◯ @org/package2
◯ changed private packages
  ◯ @org/package3
◯ unchanged packages
  ◯ @org/package4
◯ unchanged private packages
  ◯ @org/package5
```

I would like to collect opinions and feedback before continuing and I am happy to contribute the feature after discussion.

bluwy:
close #1706

I kinda like the idea that when selecting packages, it'd show if it's private so they know if the changeset wouldn't trigger a publish of it. I'm also open to discuss if we really need this.

I also went ahead and refactored the packages question function a bit.

image

### DIFF
--- a/.changeset/wild-drinks-hang.md
+++ b/.changeset/wild-drinks-hang.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+Show if a package is private when selecting packages in `changeset add`
--- a/packages/cli/src/commands/add/__tests__/add.test.ts
+++ b/packages/cli/src/commands/add/__tests__/add.test.ts
@@ -447,8 +447,8 @@ describe("Add command", () => {
     expect(mockedUtils.askMultiselect).toHaveBeenCalledWith(
       expect.stringContaining("Which packages"),
       {
-        "changed packages": [{ value: "pkg-b" }],
-        "unchanged packages": [{ value: "pkg-a" }],
+        "changed packages": [{ label: "pkg-b", value: "pkg-b" }],
+        "unchanged packages": [{ label: "pkg-a", value: "pkg-a" }],
       },
       { required: true },
     );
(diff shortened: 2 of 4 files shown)


beeequeue (approved):
very nice 👍

Andarist (approved):

PR #2074, opened 2026-06-09
Set package manager in engines

bluwy:
Support package manager versions that are released shortly after node 22.11 (our lowest supported node version). According to its [release post](https://nodejs.org/en/blog/release/v22.11.0), it's on 2024-10-24.


.changeset/every-boats-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": major
+---
+
+Set supported package manager versions in `"engines"` field, including npm >=10.9.0, pnpm >=10.0.0, and yarn >=3.3.0.
```
beeequeue: this should probably describe what it means, rather what it does, i.e. "dropped support for x"

Andarist: the stated yarn version here doesn't match the one declared in `packages/cli/package.json`

bluwy:
```suggestion
Set supported package manager versions in `"engines"` field, including npm >=10.9.0, pnpm >=10.0.0, and yarn >=4.5.2.
```

bluwy: Dropping feels like a stronger word like it wouldn't work anymore (or likely wouldn't work). I followed the changeset for the nodejs version bump so we display what we support and anything else isn't guaranteed. Maybe it'll work but we won't be maintaining if anything fails later.

beeequeue (approved):
do we know what effects this will have if we publish it? will it warn or block on old installs?

Andarist (approved):

bluwy:
> do we know what effects this will have if we publish it? will it warn or block on old installs?

Copying and replying here for transparency:

> pnpm + yarn 1: fail
> npm: warn by default, can fail with a flag
> yarn berry: nothing

But this is only from a brief check

PR #2075, opened 2026-06-09
Change allowed publish pattern to `maintenance/v*` given rulesets behavior

Andarist:
I tested this with both patterns and a force push was only rejected with `maintenance/v*`


bluwy (approved):

bluwy:
We might need this change for `maintenance/v2` too

PR #2076, opened 2026-06-09
Configure Changesets for `maintenance/v2` line

Andarist:


package.json
```diff
@@ -15,7 +15,7 @@
     "changeset": "packages/cli/bin.js",
     "check-all": "yarn test && yarn types:check && yarn lint && yarn format",
     "version-packages": "changeset version && yarn format:fix",
-    "release": "yarn build && changeset publish"
+    "release": "yarn build && changeset publish --tag v2"
```
Andarist, line 18: This kinda should stay as `latest` for now - but I think there are rather slim chances we are going to release any new v2 release before we release v3

bluwy (approved):

PR #2082, opened 2026-06-09
refactor assemble-release-plan tests

beeequeue:
follow-up to #2053

this restructures and renames tests to make the file more readable.

a few tests have also been improved, e.g. asserting more data.

it also removes tests that are redundant after #2053. the coverage% is still the same when i ran it on my machine.

### DIFF
--- a/packages/assemble-release-plan/src/index.test.ts
+++ b/packages/assemble-release-plan/src/index.test.ts
@@ -6,7 +6,7 @@ import { beforeEach, describe, expect, it } from "vitest";
 import { assembleReleasePlan } from "./index.ts";
 import { FakeFullState } from "./test-utils.ts";
 
-describe("assemble-release-plan", () => {
+describe("assembleReleasePlan", () => {
   let setup: FakeFullState;
 
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


bluwy (approved):

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -235,359 +222,209 @@ describe("assemble-release-plan", () => {
...
+    ).toThrow(
+      "Found changeset small-dogs-sad for package pkg-z which is not in the workspace",
     );
+  });
 
-    expect(releases.length).toEqual(3);
-
-    expect(releases[0].name).toEqual("pkg-a");
-    expect(releases[0].newVersion).toEqual("2.0.0");
+  describe("link: protocol", () => {
```
Andarist, line 247: in general, I'd prefer to stay away from nested `describe` blocks - they are just notoriously hard to manage, both for humans and agents

beeequeue: maybe if one overuses `before*` and `after*` hooks in each layer of them, but i don't think using it as categories does anything but make it easier to find things

Andarist: The problem, from my experience, is with adding new tests. Often times they just land at the bottom of a file, in a kinda random block. And it’s just not discoverable well during review etc. So over time they lose their original meaning and the thing just becomes messy. Categorized tests are lovely but maintaining that takes a lot of discipline, from my experience.

PR #2083, opened 2026-06-10
Use yaml package to parse yaml

bluwy:
js-yaml recently got larger due to sourcemaps, so yaml is smaller now. yaml is also technically more spec-compliant, but not likely to affect us in practice anyways.

when yaml v3 releases, it'll also be much smaller

### DIFF
--- a/packages/parse/src/index.ts
+++ b/packages/parse/src/index.ts
@@ -1,6 +1,5 @@
 import type { Release, VersionType } from "@changesets/types";
-// eslint-disable-next-line e18e/ban-dependencies
-import yaml from "js-yaml";
+import * as yaml from "yaml";
 
 const mdRegex = /\s*---([^]*?)\n\s*---(\s*(?:\n|$)[^]*)/;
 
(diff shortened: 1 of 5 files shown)


packages/parse/src/index.ts
```diff
@@ -75,10 +74,9 @@ export function parseChangesetFile(contents: string): {
   const [, roughReleases, roughSummary] = execResult;
   const summary = roughSummary.trim();
 
-  let releases: Release[];
-  let yamlStuff: Record<string, VersionType> | undefined;
+  let yamlStuff: unknown;
```
bluwy, line 77: I use unknown here so the functions later can validate it better. Maybe we can also refactor to use valibot validate the object

packages/parse/src/index.ts
```diff
@@ -75,10 +74,9 @@ export function parseChangesetFile(contents: string): {
...
   try {
-    yamlStuff = yaml.load(roughReleases) as typeof yamlStuff;
+    yamlStuff = yaml.parse(roughReleases.trim());
```
bluwy: This trim is needed because the regex that extracts the frontmatter would including a trailing `\r`, which yaml does not treat as a new-line.

I could fix this by updating the regex to handle `\r\n`, or trim like this. I'm fine with any approach.

packages/parse/src/index.ts
Andarist: in here, it would probably be enough to just use plain JS checks to validate the parsed name+type are of correct types (if that's not guaranteed already)

packages/parse/src/index.ts
Andarist: I'd prefer updating the regex here but it's not a big deal either way. Do we have tests for the `\r` stuff?

Andarist (approved):

packages/parse/src/index.ts
bluwy: I'll update the regex then. Yeah it's because a test fail that I found this

beeequeue (approved):

PR #2086, opened 2026-06-11
update dependencies

beeequeue:
mainly just to update `@manypkg/tools` with the yaml replacment, but may as well do the rest while we're at it

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
         run: pnpm test --coverage
 
       - name: Upload coverage
-        uses: codecov/codecov-action@e79a6962e0d4c0c17b229090214935d2e33f8354 # v6.0.1
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
         if: matrix.node_version == 24
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
--- a/packages/changelog-github/package.json
+++ b/packages/changelog-github/package.json
@@ -18,8 +18,7 @@
     "@changesets/types": "workspace:^"
   },
   "devDependencies": {
-    "@changesets/parse": "workspace:*",
-    "vitest": "^4.1.6"
+    "@changesets/parse": "workspace:*"
   },
   "engines": {
     "node": "^22.11 || ^24 || >=26"
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,9 +2,10 @@ packages:
   - packages/*
   - scripts/*
 
-blockExoticSubdeps: true
 dedupePeers: true
 dedupePeerDependents: true
 minimumReleaseAge: 10080
+overrides:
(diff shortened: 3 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         run: pnpm test --coverage
 
       - name: Upload coverage
-        uses: codecov/codecov-action@e79a6962e0d4c0c17b229090214935d2e33f8354 # v6.0.1
```
beeequeue, line 103: no breaking changes here from what i can find

packages/changelog-github/package.json
```diff
@@ -18,8 +18,7 @@
     "@changesets/types": "workspace:^"
   },
   "devDependencies": {
-    "@changesets/parse": "workspace:*",
-    "vitest": "^4.1.6"
```
beeequeue, line 22: vitest is the only package we both register and use at the top-level, but also add to the workspaces' manifests.

i really don't see why we should have it, and then not have tsdsown, publint, typescript, etc. as well

pnpm-workspace.yaml
```diff
@@ -2,9 +2,12 @@ packages:
   - packages/*
   - scripts/*
 
-blockExoticSubdeps: true
```
bluwy, line 5: Why is this removed. Is rolldown using exotic specifiers or something?

pnpm-workspace.yaml
```diff
@@ -2,9 +2,12 @@ packages:
...
 dedupePeers: true
 dedupePeerDependents: true
 minimumReleaseAge: 10080
+minimumReleaseAgeExclude:
+  - "@manypkg/tools@2.1.2"
+overrides:
+  rolldown: 1.1.0
```
bluwy: Rolldown 1.1.0 still has some problems to work out even though it happen to work fine for us (https://www.github.com/vitejs/vite/pull/22593), so I think it's better to not force a version here. Let tsdown>rolldown and vitest>vite>rolldown resolve to a 1.0.x version for now

bluwy (commented):
Rest looks good to me except for some concerns in the pnpm config

pnpm-workspace.yaml
beeequeue: it already defaults to `true`

bluwy: Ah I think I also mentioned this before when @Andarist added this, and he wanted to be explicit in case a global config disable it or something

pnpm-workspace.yaml
bluwy: We can also pin to `1.0.x` for now, that's what I meant that it'll be safer since it's tested. I'm not sure pnpm dedupe could lower the existing 1.1.0 to 1.0.x due to from rolldown-plugin-dts for example.

beeequeue: i pinned it in the first place so we wouldn't have to install two versions of rolldown, so that works for me :^)

bluwy (approved):

PR #2087, opened 2026-06-11
feat: make init command interactive

### LINKED ISSUE #160: Improve init command
I'm thinking now that we have a JSON config, we could have the init command ask the user and write the config that they want.

The flow I'm thinking:

1. Do you want to generate changelogs? If so, which changelog entry generator would you like to use?
  - No, I don't want changelogs
  - Yes, I want changelogs with GitHub info (depends on #159)
  - Yes, I want changelogs with only commits
2. Would you like Changesets to commit changesets and version/changelog updates for you?
3. Should packages be published publicly or privately?

trueberryless:
#### Description

This PR adds support for making the `init` command interactive. 

The new command flow uses the `clack` utilities and has the following flow:

1. Should the GitHub integration be used for changelogs? 
   (Options: Yes / No, defaults to No)
   
   1a. [If Yes]: What is the GitHub repository? (e.g. org/repo)
       (Text input: requires a non-empty string)

2. Should changeset files and version bumps be automatically committed? 
   (Options: Yes / No, defaults to No)

3. Should packages be published publicly or privately by default? 
   (Options: Private, Public, defaults to Private since the `access` option defaults to `"restricted"`)

4. Which base branch should be used? 
   (Text input: defaults to "main" if left blank)

---

I have also added new test cases to confirm the behaviour and tested it manually after running `pnpm build`.

I also ran `pnpm check-all` before committing.
I hope targeting `main` is correct. Please let me know if `next` should be the branch target.

I did the docs in #2088 since it needs to target `next-docs`. To be honest, I am a little confused about the correct workflow 😅 

- [x] This PR includes AI generated code but it is all verified by me.

- Closes #160

### DIFF
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -21,32 +21,20 @@ export type AccessType = "public" | "restricted";
 async function getInteractiveConfig() {
   const config = { ...defaultWrittenConfig };
 
-  const changelogChoice = await cli.askList(
-    "Which changelog entry generator should be used?",
-    [
-      {
-        label: "Standard changelogs",
-        value: "@changesets/cli/changelog",
-      },
-      {
-        label: "Changelogs with GitHub info",
-        value: "@changesets/changelog-github",
-      },
-      {
-        label: "Changelogs with commit info",
-        value: "@changesets/changelog-git",
-      },
-    ],
+  const useGithub = await cli.askConfirm(
+    "Should the GitHub integration be used for changelogs?",
+    false,
   );
 
-  if (changelogChoice === "@changesets/changelog-github") {
+  if (useGithub) {
+    const changelogChoice = "@changesets/changelog-github";
     const repo = await cli.askQuestion(
       "What is the GitHub repository? (e.g. org/repo)",
       { placeholder: "org/repo", notEmpty: true },
     );
     config.changelog = [changelogChoice, { repo }];
   } else {
-    config.changelog = changelogChoice;
+    config.changelog = "@changesets/cli/changelog";
   }
 
   config.commit = await cli.askConfirm(
@@ -55,10 +43,10 @@ async function getInteractiveConfig() {
   );
 
   config.access = (await cli.askList(
-    "Should packages be published publicly or privately?",
+    "Should packages be published publicly or privately by default?",
     [
-      { label: "Privately", value: "restricted" },
-      { label: "Publicly", value: "public" },
+      { label: "Private", value: "restricted" },
+      { label: "Public", value: "public" },
     ],
   )) as AccessType | "private";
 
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/init/index.ts
```diff
@@ -6,57 +6,136 @@ import c from "@changesets/color";
...
+      {
+        label: "Changelogs with GitHub info",
+        value: "@changesets/changelog-github",
+      },
+      {
+        label: "Changelogs with commit info",
+        value: "@changesets/changelog-git",
+      },
+    ],
+  );
```
bluwy: The first and last options work the same. Can we ask about the git provider instead? GitHub or none?

packages/cli/src/commands/init/index.ts
```diff
@@ -6,57 +6,136 @@ import c from "@changesets/color";
...
+    "Should changeset files and version bumps be automatically committed?",
+    false,
+  );
+
+  config.access = (await cli.askList(
+    "Should packages be published publicly or privately?",
+    [
+      { label: "Privately", value: "restricted" },
+      { label: "Publicly", value: "public" },
+    ],
```
bluwy:
```suggestion
    "Should packages be published publicly or privately by default?",
    [
      { label: "Private", value: "restricted" },
      { label: "Public", value: "public" },
    ],
```

packages/cli/src/commands/init/index.ts
```diff
@@ -6,57 +6,136 @@ import c from "@changesets/color";
...
+      }
+
+      log.success(
+        `
+${c.green("Changesets")} initialization is complete.
+The ${c.blue(".changeset")} folder has been updated with:
+- ${c.blue("config.json")} (customized based on the provided preferences)
+- ${c.blue("README.md")} (a quick guide for using changesets)
+
+Versioning and publishing can now be managed!
```
bluwy:
```suggestion
- ${c.blue("README.md")} (a quick guide for using changesets)
```

I think we can skip the last line for now.

packages/cli/src/commands/init/index.ts
```diff
@@ -6,57 +6,136 @@ import c from "@changesets/color";
...
+${c.green("Changesets")} has already been initialized.
+Changeset commands can be run with no problems.
+        `.trim(),
       );
     }
   } else {
     await fs.cp(path.resolve(pkgPath, "./default-files"), changesetBase, {
       recursive: true,
     });
+
```
bluwy: Can you consolidate this else block with the above config handling. Since we're generating the README in both cases now, there's no need to split this up.

packages/cli/src/commands/init/index.ts
trueberryless: Thanks for the clarification. Addressed in [`c76ddc7` (this PR)](https://github.com/changesets/changesets/pull/2087/changes/c76ddc724d66c4efe31b4273e9e6c75718a523b7)

packages/cli/src/commands/init/index.ts
trueberryless: That's a nice refactor. Thanks for catching it 🙌 

Addressed in [`c76ddc7` (this PR)](https://github.com/changesets/changesets/pull/2087/changes/c76ddc724d66c4efe31b4273e9e6c75718a523b7)

packages/cli/src/commands/init/index.ts
```diff
@@ -6,58 +6,101 @@ import c from "@changesets/color";
...
 );
 
-const defaultConfig = `${JSON.stringify(defaultWrittenConfig, null, 2)}\n`;
-
 export interface InitOptions {
   cwd?: string;
 }
 
-export async function init(options?: InitOptions) {
+export type AccessType = "public" | "restricted";
```
bluwy: This can be imported from `@changesets/types`

bluwy (approved):

beeequeue (approved):
i don't really like the order of the properties in this generated config, but we can fix that later if we want to.

packages/cli/src/commands/init/index.ts
trueberryless: Nice catch. I also remove the additional `"private"` cast in [`a56d9ee` (this PR)](https://github.com/changesets/changesets/pull/2087/commits/a56d9ee448620ba0468972fa64c478e98a65abf7) since it is not a valid option in the [docs](https://changesets.dev/guide/config#access). I wonder why it is defined as a [valid config option](https://github.com/changesets/changesets/blob/bdf34317190db4a27394613db20527aaaf922060/packages/types/src/index.ts#L104) tho? 🤔

trueberryless:
> i don't really like the order of the properties in this generated config, but we can fix that later if we want to.

I played around with the idea before but thought it was unnecessary. Given that you mention it as well know, I decided to reimplement it, so the order of properties is now:

```ts
const priorityOrder = [
  "$schema",
  "changelog",
  "baseBranch",
  "commit",
  "access",
  "ignore",
  "fixed",
  "linked",
  "format",
  "updateInternalDependencies",
];
```

Feel free to suggest your desired sorting of properties (I just took the one that is used in current (`main`) `init` command.

packages/cli/src/commands/init/index.ts
bluwy: We should follow up removing `"private"`, it's compat for "restricted" and we've removed the runtime handling, but have missed removing from the types.

trueberryless: Thanks for the clarification. I removed it in [`61231f9` (this PR)](https://github.com/changesets/changesets/pull/2087/changes/61231f9998813a836d137b8203dee333bb7932c5)

packages/cli/src/commands/init/index.ts
```diff
@@ -4,60 +4,129 @@ import path from "node:path";
...
+    "changelog",
+    "baseBranch",
+    "commit",
+    "access",
+    "ignore",
+    "fixed",
+    "linked",
+    "format",
+    "updateInternalDependencies",
+  ];
```
beeequeue, line 70:
```suggestion
  const priorityOrder = [
    "$schema",
    "baseBranch",
    "access",
    "format",
    "changelog",
    "commit",
    "ignore",
    "fixed",
    "linked",
    "updateInternalDependencies",
  ];
```

bluwy:
Probably don't need to ask for the github repo anymore with #1871 but we can also follow up on that later

PR #2092, opened 2026-06-13
Do not bump ignored independent packages when exiting pre mode

### LINKED ISSUE #1543: changeset incorrectly bumps version after exit mode.
## Problem

I have a monorepo where sometimes I only want to release one package only. I want to start using changesets. My current monorepo has 2 packages where `packageA` has `1.6.0-alpha.1` and `packageB` has `1.8.1`. Ofc, I also want to be doing pre-releases. Let's say I want to release packageB in a pre-release mode.

1. I added changeset code.
2. I run `npx changeset pre enter alpha`.
3. For changeset action, I pass my command of `yarn run version`. Note that changeset action allows you to pass version and publish commands. 
4. `yarn run version` runs my code which basically does: `npx changeset version --ignore packageA`
5. The PR is created.

If you look at the PR, it contains the change of updating packageB's package.json from 1.8.1 to 1.8.1-alpha.0. So far so good. Assume that I decided to release this with non-prerelease. I changed my mind, so I re-run the workflow which basically exists the pre mode and updates the PR. Now, the PR contains packageB's package.json change from 1.8.1 to 1.9.0 (assuming I used `minor`). So far so good, but here is the problem. The PR now also contains `packageA`'s package.json version change from 1.6.0-alpha.1 to 1.7.0 which is NOT correct.

After looking into the src code, I found: https://github.com/changesets/changesets/blob/f0270f69793ddb6865f2510d334864b093cb10e3/packages/assemble-release-plan/src/index.ts#L218

There's a link on top of this line: https://github.com/changesets/changesets/pull/382#discussion_r434434182 and reading it makes sense why you're doing this, but the problem is that my packages don't depend on each other., but your discussion explains this when there's dependency. Since my packages don't depend on each other, your solution seems wrong(i.e doesn't handle my scenario).

Is this clear ?

chatman-media:
## Problem

When exiting pre mode, `assembleReleasePlan` gives a `patch` bump to *any* package whose recorded pre-version isn't `0` (i.e. it had a prerelease during the pre phase), so that packages that were prereleased but didn't otherwise get a bump still get released. The intent is documented [here](https://github.com/changesets/changesets/pull/382#discussion_r434434182).

However, this block does not consistently respect the `ignore` config (and `privatePackages.version`):

- For a package that already has a `none` release in the plan (e.g. an ignored *dev dependent*), the `else if` branch correctly checks `shouldSkipPackage` and skips it.
- For a package that has **no** existing release in the plan — e.g. an **ignored package that is independent** (not a dependent of anything being released) — the `if (!existingRelease)` branch creates a `patch` release **without** checking `shouldSkipPackage`.

As a result, an ignored, standalone package that happened to have a prerelease version gets bumped on `pre exit` even though the user explicitly ignored it (and passed e.g. `--ignore <pkg>`). This matches the report in #1543, where a package that other packages do not depend on is incorrectly version-bumped on pre exit.

Note that ignored packages are intentionally excluded from `preInfo.preVersions` (in `getPreInfo`), so `preVersions.get(name)` returns `undefined` for them, and `undefined !== 0` is `true` — which is exactly what slips them past the guard.

## Fix

Hoist the `shouldSkipPackage` check to the top of the condition so it guards **both** the "create new patch release" and the "upgrade existing none → patch" branches, and drop the now-redundant check from the `else if`. Behavior for non-ignored packages is unchanged.

## Tests

Added a test asserting that an ignored, independent package that had a prerelease is **not** bumped when exiting pre mode (it fails without the fix). The existing "should not bump ignored dev dependents when exiting pre-release mode" test continues to pass, and the full test suite is green.

Closes #1543


packages/assemble-release-plan/src/index.ts
```diff
@@ -184,10 +184,11 @@ export function assembleReleasePlan(
 
   if (preInfo?.state.mode === "exit") {
     for (const pkg of packages.packages) {
+      const preVersion = preInfo.preVersions.get(pkg.packageJson.name);
       // If a package had a prerelease, but didn't trigger a version bump in the regular release,
       // we want to give it a patch release.
       // Detailed explanation at https://github.com/changesets/changesets/pull/382#discussion_r434434182
-      if (preInfo.preVersions.get(pkg.packageJson.name) !== 0) {
+      if (preVersion != null && preVersion !== 0) {
```
bluwy, line 199: Changed to a check like this as `preInfo.preVersions` already does not included ignored packages and we can rely on that check

packages/assemble-release-plan/src/index.test.ts
```diff
@@ -868,6 +868,34 @@ describe("assembleReleasePlan", () => {
       expect(releases[1].newVersion).toEqual("1.0.1-next.0");
     });
 
+    it("should not bump ignored independent packages that had a prerelease when exiting pre-release mode", () => {
```
Andarist, line 871: Does this reported situation relate to a temporarily or permanently ignored package? If the first one, how do you expect to handle the ignored package's transition to the unignored state?

.changeset/spicy-zebras-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/assemble-release-plan": patch
+---
+
+Fixed a bug present in v2 where exiting prerelease mode could incorrectly patch-bump ignored or otherwise skipped packages.
```
Andarist:
```suggestion
Fixed a bug where exiting prerelease mode could incorrectly patch-bump ignored or otherwise skipped packages.
```

Andarist (approved):

Issue #1468, opened 2024-09-23
Changeset does not replace workspace version in monorepo when publish


melvinmcrn:
## Affected Packages
pkg:cli

## Problem
Using [bun](https://bun.sh/) as a package manager on a monorepo, changeset somehow does not replace the `workspace:*` version of a dependency. It used to work normally with pnpm though. Not sure is the problem really on the changeset.

j4w8n:
Would love to see this resolved; just ran across it in the last couple of days, while setting up a CI pipeline on GitHub.

macalinao:
Hi all, I was able to get Bun workspaces working with changesets.

I wrote a blog post here on my website on how to do this: https://ianm.com/posts/2025-08-18-setting-up-changesets-with-bun-workspaces

Dr-Nikson:
Honestly, it's a bit embarrassing that `bun` is still not properly supported. This feels like a pretty fundamental feature for a library of this scope.

PR #2097, opened 2026-06-14
Consistently use the repository's preferred package manager

### LINKED ISSUE #432: changeset publish does not replace workspace:* with current version in workspace before publishing

### LINKED ISSUE #598: [Feature Request] Support for Yarn 2 "workspace ranges"

### LINKED ISSUE #813: Yarn not used for publishing

### LINKED ISSUE #1411: [CLI]: yarn changeset publish is not attempting to publish with yarn
## Problem

I'm working on a yarn monorepo in which the packages must be published to a private registry. Initially when running `yarn changeset publish`, I noticed that it was ignoring the `npmPublishRegistry` field in my .`yarnrc.yml`. I got around the issue by adding the registry to `publishConfig` property of all the `package.json`s, but after successfully publishing the packages, I ran into major issue. Yarn replaces any references to the `workspace` protocol within the `package.json` with the correct version, but since the CLI seems to always publish with npm, this replacement never ocurrs, and thus a broken `package.json` is published, which cannot be installed with npm.

I haven't done a deep dive into the source code, but unless I'm misunderstanding something [getPublishTool](https://github.com/changesets/changesets/blob/main/packages/cli/src/commands/publish/npm-utils.ts#L48) only returns either `npm` or `pnpm`.

## Proposed solution

`getPublishTool` should also support yarn

### LINKED ISSUE #1454: Publishing with yarn doesn't work correctly

Andarist:
fixes https://github.com/changesets/changesets/issues/432
fixes https://github.com/changesets/changesets/issues/1454
fixes https://github.com/changesets/changesets/issues/598
fixes https://github.com/changesets/changesets/issues/813
fixes https://github.com/changesets/changesets/issues/1411


kwangure:
From what I can tell this does not support Vite Plus via `vp pm publish`. For example, I have Vite+ installed globally, but pnpm is not globally installed. Vite manages its own `pnpm` copies to do `vp install` etc. based on the package manager versions different projects specify. So even though, there's a pnpm-workspace.yaml in my project, I'd like changesets to use `vp` and VIte+'s package manager resolution algo. I imagine this'll be increasingly popular.

Andarist:
This is definitely outside the scope of this PR. Please open a new feature request so we could properly track this.

packages/cli/src/commands/__tests__/e2e-utils.ts
```diff
@@ -0,0 +1,447 @@
...
+    });
+    return cwd;
+  };
+}
+
+export const pmCases = [
+  {
+    name: "npm 10",
+    bins: { npm: "npm-10" },
+    gitdir: createNpmGitdir("npm@10.9.8"),
```
bluwy: Could these versions (e.g. the `10.9.8`) derive from the actual installed version? In case `npm-10` has a new patch release and we update in the package.json, then this seems to diverge.

bluwy (commented):
I can't say I understood every edge case and pm-specific handling, but I trust that this is handled well with the e2e tests 😄 And the logic seems good for me. 

Also, we need changesets here.

packages/cli/src/commands/__tests__/e2e-utils.ts
Andarist: yep, I wanted to do that at some point - just focused on other things to get this review-ready. I'll make this change in a moment

Andarist: done

packages/cli/src/commands/publish-plan/getPublishPlan.ts
```diff
@@ -117,6 +121,9 @@ export async function getUnpublishedPackages(
           if (
             preState != null &&
             response.pkgInfo.versions &&
+            // non-npm registries often don't auto-assign latest and when using those we don't have to care about only-pre case
+            // when the latest tag is not auto-assigned we can simply use the configured pre tag
+            response.pkgInfo["dist-tags"].latest &&
```
bluwy, line 126: Does this mean we support non-latest publishing? Should we mention in the changeset?

bluwy (approved):

packages/cli/src/commands/publish-plan/getPublishPlan.ts
Andarist: I'd like to revisit this behavior - it feels kinda surprising that prerelease packages can be published with the latest tag in prerelease mode (and not with the configured prerelease tag) and it all depends on the registry's state that the user isn't usually quite aware of. 

I added the changeset for this change.

jperezrealini:
Will these changes help solve this issue? https://github.com/changesets/changesets/issues/1468

Andarist:
@jperezrealini not yet - but we'll work on Bun support soon. This PR and https://github.com/changesets/changesets/pull/2155 prepare the groundwork for that

PR #2100, opened 2026-06-18
Chunk releases using topological sorting

### LINKED ISSUE #238: Packages should be published in topological order
## Affected Packages

I think `@changesets/assemble-release-plan` should do the sorting. 

## Problem

A broken release (for example stopped in the middle) might leave NPM registry in a broken state.

## Proposed solution

Packages should be published in topological order - dependencies before dependents.

I'm not sure what should be done about cycles.

Andarist:
fixes https://github.com/changesets/changesets/issues/238

### DIFF
--- a/.changeset/real-horns-shiver.md
+++ b/.changeset/real-horns-shiver.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+Order releases into dependency-aware chunks so packages are grouped in publish order.
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -48,6 +48,7 @@
     "@changesets/write": "workspace:^",
     "@clack/prompts": "^1.5.1",
     "@manypkg/get-packages": "^3.1.0",
+    "@pnpm/deps.graph-sequencer": "^1100.0.0",
     "cac": "^7.0.0",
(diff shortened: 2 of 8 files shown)


bluwy:
Should or can this be done in assemble-release-plan as mentioned in https://github.com/changesets/changesets/issues/238?

Also, should we allow publishing still with cycles? We can warn instead. We currently already publish non-topologically, and that has worked out fine for a long while, so publishing with a best-effort order is still fine I think.

Andarist:
> Should or can this be done in assemble-release-plan as mentioned in https://github.com/changesets/changesets/issues/238?

See https://github.com/changesets/changesets/issues/238#issuecomment-570415443

We could *also* sort in `assemble-release-plan` but that wouldn't really help with publishes interrupted in the middle etc. `assemble-release-plan` participates in `changeset version` and not `changeset publish`

> Also, should we allow publishing still with cycles? We can warn instead. We currently already publish non-topologically, and that has worked out fine for a long while, so publishing with a best-effort order is still fine I think.

I also considered adding an option to allow/disallow cycles - but I intentionally have started with the least permissive behavior here to hear out your thoughts on this.

I think warnings would be OK for now.

Andarist:
I made it to warn now

bluwy (approved):

PR #2109, opened 2026-06-23
Do not accumulate changesets in prerelease mode

### LINKED ISSUE #240: Exiting pre mode should prompt for changeset cleanup
## Problem

When in pre mode some generated changesets might only be relevant for that pre mode. For example when we introduce a new feature and later fix a bug in it. 

Ideally, a changeset describing that bug fix shouldn't be a part of post pre mode changelog entry.

## Proposed solution

I doubt there is really any automatic solution for this problem, but at the very least when exiting pre mode a user could be somehow prompted about this problem. Maybe each changeset could be interactively reviewed then (sorted from the oldest to the newest)?

### LINKED ISSUE #1050: Pre mode changeset will always be added into `pre.json` even if every package in that changeset is ignored
## Affected Packages

`@changesets/cli`

**Maybe** `@changesets/assemble-release-plan` `@changesets/read`

Noticed the result of [assembleReleasePlan()](https://github.com/changesets/changesets/blob/8447c677a37d1d23858059e0ce2b5cd98f4c24ae/packages/cli/src/commands/version/index.ts#L69) always hold the full set of pre mode changesets.

## Problem

Say we have two packages, `@foo/a` and `@foo/b`. Changesets is already in `pre` mode.

In `pre.json`:

```json
{
  "mode": "pre",
  "tag": "beta",
  "initialVersions": {
    "@foo/a": "1.0.0-beta.1",
    "@foo/b": "1.0.0-beta.1"
  },
  "changesets": []
}

```

Now we have a changeset  `fifty-balloons-taste.md` with the content:

```md
---
"@foo/a": minor
---

Some update
```

And run:

```bash
changeset version --ignore @foo/a
```

Then `fifty-balloons-taste.md` will be added to the `changesets` field in `pre.json`.

If we commit the change then run `changeset version`, the changeset of `@foo/a` takes no effect and looks like it's lost.

However, in normal release mode,  `fifty-balloons-taste.md` will not be removed after `changeset version --ignore @foo/a`; and if there's any changeset that includes the change of `@foo/b`, only that one will be removed and `@foo/b` will have version bump accordingly.

What happened in normal release mode is the desired behavior.

## Proposed solution

Align pre mode's behavior with normal mode.

### Additional context

We're trying to separate our packages into two groups and publish separately, since they may have different publish frequencies.

### LINKED ISSUE #1688: Exiting pre mode fails with 400+ changesets
Our repo has been in a pre-release state for some time and we're nearing the deadline, so we tried to exit pre and run version on the repo. Currently, we have 425 changesets pending. When trying to run `pnpm changeset version`

**Repo**: https://github.com/swisspost/design-system
**Steps to reproduce**:
1. Clone the repo
2. Run `pnpm install`
3. Run `pnpm changeset pre exit`
4. Run `pnpm changeset version`
5. GitHub API call errors with "Something went wrong while executing your query ..."

A call to the GitHub API is placed that fails on GitHub side. I've investigated a bit and saw that the query payload sent to the API is 338KB of GraphQL (?) like this:

```txt
a1dd4791446531a1d01a8bbb8722ef8caa5b3e85a: object(expression: "1dd4791446531a1d01a8bbb8722ef8caa5b3e85a") {
            ... on Commit {
            commitUrl
            associatedPullRequests(first: 50) {
              nodes {
                number
                url
                mergedAt
                author {
                  login
                  url
                }
              }
            }
            author {
              user {
                login
                url
              }
            }
          }}
```

Do you know if there's a limit on this API call payload and we're reaching it with this many changesets? Also, is there any way to work around this?

bluwy:
close https://github.com/changesets/changesets/issues/1688
close https://github.com/changesets/changesets/issues/240
fix #1050

No tracking of changesets in `pre.json`, and clear changesets during version in prerelease like normal. The final pre exit release will not contain changelog of all previous prerelease changelog. 

I'm still thinking whether to handle `pre.json` "changesets" still (filtering) so it's more compatible, but in this PR I took the more drastic approach for now and remove it entirely, and rely on `readPreState` to automatically apply the migration. The migration is mainly for those migrating from v2 while still in prerelease mode.

Any calls to `readPreState` will delete the `"changesets"` property and its referenced changeset files. This may be a little surprising.

### DIFF
--- a/.changeset/legal-papers-train.md
+++ b/.changeset/legal-papers-train.md
@@ -1,0 +1,14 @@
+---
+"@changesets/apply-release-plan": major
+"@changesets/assemble-release-plan": major
+"@changesets/cli": major
+"@changesets/pre": major
+"@changesets/release-utils": major
+"@changesets/types": major
+---
+
+When exiting prerelease mode, the final changelog will now not include all changesets made since `pre enter`. This change was made as often changes during prereleases are not relevant to the final release, and it's not possible to query the entire Git and GitHub metadata when the backlog of changesets become sufficiently large.
+
+As a result, the `pre.json` file also no longer tracks changesets added since `pre enter`, and `version` commands during prereleases will clear all changesets same as a normal release.
+
+If you're migrating to Changesets v3 during prerelease mode, this new behavior will be automatically applied the next time you run the `version` command. Specifically, the `pre.json` `"changesets"` field will be cleared, and all of its referenced changesets will be deleted in `.changeset/*.md`.
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -3273,7 +3273,6 @@ describe("apply release plan", () => {
           "pkg-a": "1.0.0",
           "pkg-b": "1.0.0",
         },
-        changesets: [],
       };
 
       const { tempDir } = await testSetup(
--- a/packages/apply-release-plan/src/index.ts
+++ b/packages/apply-release-plan/src/index.ts
@@ -160,40 +160,38 @@ export async function applyReleasePlan(
     await formatter(filesToFormat);
   }
 
-  if (releasePlan.preState == null || releasePlan.preState.mode === "exit") {
-    const changesetFolder = path.resolve(cwd, ".changeset");
-    await Promise.all(
-      changesets.map(async (changeset) => {
-        const changesetPath = path.resolve(
-          changesetFolder,
-          `${changeset.id}.md`,
-        );
+  await Promise.all(
+    changesets.map(async (changeset) => {
+      const changesetPath = path.resolve(
+        cwd,
+        ".changeset",
+        `${changeset.id}.md`,
+      );
+      if (
+        await fs.access(changesetPath).then(
+          () => true,
+          () => false,
+        )
+      ) {
+        // DO NOT remove changeset for skipped packages
+        // Mixed changeset that contains both skipped packages and not skipped packages are disallowed
+        // At this point, we know there is no such changeset, because otherwise the program would've already failed,
+        // so we just check if any skipped package exists in this changeset, and only remove it if none exists
+        // options to skip packages were added in v2, so we don't need to do it for v1 changesets
         if (
-          await fs.access(changesetPath).then(
-            () => true,
-            () => false,
+          !changeset.releases.some((release) =>
+            shouldSkipPackage(packagesByName.get(release.name)!, {
+              ignore: config.ignore,
+              allowPrivatePackages: config.privatePackages.version,
+            }),
           )
         ) {
-          // DO NOT remove changeset for skipped packages
-          // Mixed changeset that contains both skipped packages and not skipped packages are disallowed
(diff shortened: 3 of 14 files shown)


.changeset/legal-papers-train.md
```diff
@@ -0,0 +1,14 @@
+---
+"@changesets/apply-release-plan": major
+"@changesets/assemble-release-plan": major
+"@changesets/cli": major
+"@changesets/pre": major
+"@changesets/release-utils": major
+"@changesets/types": major
+---
+
+When exiting prerelease mode, the final changelog will now not include all changesets made since `pre enter`. This change was made as often changes during prereleases are not relevant to the final release, and it's not possible to query the entire Git and GitHub metadata when the backlog of changesets become sufficiently large.
```
Andarist, line 10: how are we going to help with `changeset pre exit` then?

bluwy: I don't quite understand the question. You mean how should the exited/stable release changelog should be? They could add changesets that summarizes the changes if they like just before the stable is released.

Andarist: 1. Re-adding changesets will screw up commit/PR auto-references generated by our changelog generators. 
2. It feels to me that we have many more meaningful changesets aggregated during pre mode than the temporary ones. So it feels that cleaning up temporary ones is less of a chore that having to re-add what is still relevant

beeequeue: i don't think we should remove the accumulating changesets.
while it does suck when it has been open a long time i also don't think a lot of people are prepared to re-add changesets when exiting pre-release mode, at least not now.

maybe an option for this behavior would be better?

bluwy: 1. I don't think it matters a lot but we can probably explore/expand on "none" changeset types that are added only for adding changelog, and we'd not link the commit/PR for it. They can also write a summary that links to the right PRs.
2. Well there's a tipping point in which adding vs removing becomes a chore or not, depending on the size, like #1688. But fundamentally for me, it's better to keep changesets as summaries that describes the next version. Lending it to also describe the stable version doesn't feel right by default.

bluwy: > while it does suck when it has been open a long time i also don't think a lot of people are prepared to re-add changesets when exiting pre-release mode, at least not now.

They don't have to add a changeset when exiting. If there's not many prereleases, users can also read through the changes that happen before instead of duplicating it. If they don't have the habit of making the changelog good anyways then it's better we don't generate potentially low quality changelogs either.

> maybe an option for this behavior would be better?

I prefer to not make an option for now. It also keeps our handling simpler.

Andarist: > They don't have to add a changeset when exiting. If there's not many prereleases, users can also read through the changes that happen before instead of duplicating it. If they don't have the habit of making the changelog good anyways then it's better we don't generate potentially low quality changelogs either.

This strategy doesn't quite work for GitHub releases that are derived from the latest `CHANGELOG.md` version entry.

bluwy: I don't think it doesn't work, they'd have to look for the changes themselves elsewhere from the UI and that's an ok tradeoff.

PR #2110, opened 2026-06-23
Remove initialVersions in pre.json

bluwy:
Unused since #454

### DIFF
--- a/packages/pre/src/index.ts
+++ b/packages/pre/src/index.ts
@@ -66,11 +66,7 @@ export async function enterPre(rootDir: string, tag: string) {
   const newPreState: PreState = {
     mode: "pre",
     tag,
-    initialVersions: {},
   };
-  for (const pkg of packages.packages) {
-    newPreState.initialVersions[pkg.packageJson.name] = pkg.packageJson.version;
-  }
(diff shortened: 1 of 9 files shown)


packages/pre/src/index.ts
```diff
@@ -66,33 +66,39 @@ export async function enterPre(rootDir: string, tag: string) {
...
+    for (const changesetId of preState.changesets) {
+      await fs
+        .rm(path.resolve(rootDir, ".changeset", changesetId + ".md"))
+        .catch(() => {
+          // It's possible to specify an unknown id before without errors, so we
+          // don't throw an error here if the file doesn't exist.
+        });
+    }
+    delete preState.changesets;
+  }
```
Andarist, line 95: isn't this just to handle some old versions too? perhaps this block could also be removed entirely?

Andarist (approved):

packages/pre/src/index.ts
Andarist: oh wait, this PR is stacked on top of https://github.com/changesets/changesets/pull/2109 (which I have not noticed initially)

PR #2112, opened 2026-06-23
Use split publishing workflow

Andarist:


.github/workflows/publish.yml
```diff
@@ -15,13 +15,40 @@ concurrency:
...
+        with:
+          skip-cache: true # avoid cache poisoning attacks
+
+      - name: Build artifacts
+        run: pnpm build
+
+      - name: Select mode
+        id: select-mode
+        # https://github.com/changesets/action
+        uses: changesets/action/select-mode@a45c4d594aa4e2c509dc14a9f2b3b67ba3780d0d # v1.9.0
```
Andarist: this has to be updated once we publish a new version of the action

.github/workflows/publish.yml
```diff
@@ -78,7 +135,8 @@ jobs:
 
       - name: Publish to npm
         # https://github.com/changesets/action
-        uses: changesets/action@a45c4d594aa4e2c509dc14a9f2b3b67ba3780d0d # v1.9.0
+        uses: changesets/action/publish@a45c4d594aa4e2c509dc14a9f2b3b67ba3780d0d # v1.9.0
         with:
-          # this expects you to have a script called release which does a build for your packages and calls changeset publish
-          publish: pnpm release
+          # # this expects you to have a script called release which does a build for your packages and calls changeset publish
+          # publish: pnpm release
```
Andarist: did we even need this? it just calls `changeset publish` - but i'm not quite sure how `PATH` behaves here, in this context

cc @beeequeue - maybe you'd know

beeequeue: it should not be required. `pnpm changeset` or `npm exec changeset` should work just as well.

Andarist: It's more of a question if this default branch will work correctly with our local files or not: https://github.com/changesets/action/blob/04905189325845f674e3d468085dbf44effdee85/src/run.ts#L110-L117

I think it will, but gotta be sure :p

bluwy (approved):

PR #2113, opened 2026-06-23
Fixed accidental success logs on failed npm publishes

Andarist:


packages/cli/src/commands/publish/publishPackages.ts
```diff
@@ -113,9 +113,11 @@ export async function publishPackages({
     return Promise.all(
       publishPromises.map(async (publishPromise) => {
         const result = await publishPromise;
-        log.success(
-          `Published ${c.blue(result.name)}@${c.green(result.version)}!`,
-        );
+        if (result.result === "published") {
```
Andarist, line 116: it feels we should be printing errors here too (or successes elsewere) - this logging is a little bit spread out through the code right now. But I don't have time to refactor this right now.

bluwy (approved):

PR #2117, opened 2026-06-24
Remove initialVersions in pre.json

bluwy:
#2110 but standalone to main

I've also ported the auto-migration on `readPreState`, but if we have concerns of this we can also not do this as leaving the property there as is is harmless. The next time they start prerelease with changesets v3, it won't be there anymore.


packages/pre/src/index.ts
```diff
@@ -62,11 +66,28 @@ export async function enterPre(rootDir: string, tag: string) {
...
+  },
+): Promise<PreState> {
+  if (preState.initialVersions == null) {
+    return preState;
+  }
+
+  // "initialVersions" is not used for a long time, so can directly delete
+  delete preState.initialVersions;
+
+  await outputFile(
```
Andarist, line 87: it would be nice to use the `editJson` here, but given this is like a one-time migration... we probably don't have to

Andarist (approved):

beeequeue (approved):

packages/pre/src/index.ts
bluwy: Yeah it'd be nice to reuse that, but we need to figure out how to share it and also needs an update to handle deletions. Let's see if people are getting formatting issues if we do something like this.

PR #2118, opened 2026-06-24
Improve default changelog new lines handling

### LINKED ISSUE #748: Release lines tend to be joined with too many new lines
`@changesets/changelog-github` prepends each release line with `\n\n`:
https://github.com/changesets/changesets/blob/d02f2b8223d518307f5d3fcc9cd9ca3a8d11fb6f/packages/changelog-github/src/index.ts#L118

And later on, concatenated release lines are prepended with `\n\n` too:
https://github.com/changesets/changesets/blob/d02f2b8223d518307f5d3fcc9cd9ca3a8d11fb6f/packages/apply-release-plan/src/get-changelog-entry.ts#L20

This creates `\n\n\n\n` at the beginning of the generated section for a particular bump type.

I've also noticed that a similar thing (`\n\n\n\n`) happens between release lines. 2 of them obviously get there from the first line of the code that has been linked here, the third one is coming from the second linked line (from the `join` call) and the fourth one is coming also from the first linked line - it's the one inserted between `firstLine` and `futureLines`. Note that there might not be any `futureLines` so this should only be added conditionally.

I didn't notice before because this gets formatted somewhat OK when we update CHANGELOGs **and** when we format the final write with Prettier. I think that we should revisit having prettier as a dependency in the future and if we'd ever get to that then this issue would be surfaced more often.

Currently, though, we do not format stuff when creating the initial CHANGELOG here:
https://github.com/changesets/changesets/blob/d02f2b8223d518307f5d3fcc9cd9ca3a8d11fb6f/packages/apply-release-plan/src/index.ts#L278

This should be fixed but I think that the original issue would still be in the code, while only being patched with a bandaid (the formatter).

bluwy:
fixes #748

Uses a slightly sophisticated handling when joining release lines to ensure there's a minimum of one new line and maximum of two new lines between the release lines.

One new line means there's only one `\n`, so there's no spacing in between. Two new lines is `\n\n` so there's a spacing in between.

Also ensure that headings such as `### Major changes` and `## 1.0.0` are followed with `\n\n`. This should look more proper to how most formatters format it.

## Context and thought process

I did this change in the `getChangelogEntry` side instead of the changelog generator side (e.g. `changelog-github`), because it's possible that some generators prefer lines to be followed together or with spacing in between. For example:

`changelog-git`:
```md
- change one
- change two
```

`changelog-github`:
```md
- [abc123] change one (#1234)

- [def456] change two (#5678)
```

But at the same time, we limit to a maximum of 2 new lines in between (one spacing) because there's not a lot of reason in practice to want more than that. So `getChangelogEntry` has a little opinion on that, but in return generators don't need to think too deeply on how new lines would be combined in the changelog. Maybe we want to discuss this magic a bit.

### DIFF
--- a/packages/apply-release-plan/src/get-changelog-entry.ts
+++ b/packages/apply-release-plan/src/get-changelog-entry.ts
@@ -12,17 +12,6 @@ type ChangelogLines = {
   patch: Array<Promise<string>>;
 };
 
-async function generateChangesForVersionTypeMarkdown(
-  obj: ChangelogLines,
-  type: keyof ChangelogLines,
-) {
-  let releaseLines = await Promise.all(obj[type]);
-  releaseLines = releaseLines.filter((x) => x);
-  if (releaseLines.length) {
-    return `### ${capitalize(type)} Changes\n\n${releaseLines.join("\n")}\n`;
-  }
-}
-
 // release is the package and version we are releasing
 export async function getChangelogEntry(
   cwd: string,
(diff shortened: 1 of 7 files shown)


packages/apply-release-plan/src/get-changelog-entry.ts
```diff
@@ -113,12 +102,48 @@ export async function getChangelogEntry(
...
+  // want the extra spacing after the heading.
+  let newLines = 2;
+
+  for (const line of releaseLines) {
+    // Factor in the starting new lines preferred by the release line
+    const startNewLinesCount = line.match(/^\n*/)?.[0].length ?? 0;
+    newLines += startNewLinesCount;
+
+    // Ensure a minimum of one new line and maximum of two new lines between release lines
+    const newLinesContent = "\n".repeat(Math.min(Math.max(newLines, 1), 2));
```
Andarist, line 140: `Math.clamp` - where are you when we need you?! 😢

Andarist (approved):

packages/apply-release-plan/src/get-changelog-entry.ts
bluwy: I was disappointed too 😔

PR #2116, opened 2026-06-24
Handle missing commit author data

bluwy:
fix #795 

Did a very simple fix only since the pr variant also handles similarly

https://github.com/changesets/changesets/blob/c0e6d14e22b81cbf7aefe9c30a7aac374891bf0c/packages/get-github-info/src/index.ts#L300-L301

Could look into better warnings or types later maybe.

PR #2120, opened 2026-06-24
Use `GITHUB_REPOSITORY` as the default repo when no `repo` option is configured in `changelog-github`

Andarist:
just a small potential quality of life improvement


beeequeue (approved):

bluwy (changes requested):
Can you address or comment on why we didn't go with https://github.com/changesets/changesets/pull/1871? And is there things there we can port here?

Andarist:
> Can you address or comment on why we didn't go with https://github.com/changesets/changesets/pull/1871? And is there things there we can port here?

No good reason! This was just something I thought of when reviewing https://github.com/changesets/changesets/pull/2116 and jumped on implementing this withou checking open PRs. I now landed that other PR after syncing it with this one. Thanks for catching this.

PR #2124, opened 2026-06-25
Add version providers with initial Ruby support

jbeckwith-oai:
First: thank you for Changesets. I personally really enjoy this project: it has a rare combination of being pragmatic, understandable, and flexible enough to fit many different release workflows. At OpenAI we use and encounter increasingly mixed-language repositories, and we would love to help expand Changesets' usefulness while preserving the package/release planning model that already works so well.

This PR is an exploratory but working step toward that: it adds a provider model for version-file updates, with the existing Node/npm behavior represented as the first provider and Ruby gems added as the first non-Node provider.

I am opening this as a draft because I would very much appreciate the maintainers' guidance on whether this shape feels aligned with the project before we go further. I am happy to reshape the API, naming, or provider boundaries based on your feedback.

## Motivation

Changesets already has a very good model for expressing release intent:

- a changeset names a package and bump type
- Changesets calculates the release plan
- `changeset version` applies the plan and consumes the changesets
- changelogs and release PRs communicate what will ship

The main limitation this PR tries to address is that applying a release plan currently assumes the version file is `package.json`. That works beautifully for npm packages, but it prevents the same release-planning model from being used naturally for packages whose canonical version lives elsewhere, such as Ruby gems, Python packages, Cargo crates, etc.

The goal here is not to make Changesets publish Ruby gems or Python packages. Publishing can stay external. The goal is narrower: let Changesets own the release-plan workflow and update the correct version files for a package.

## What This Adds

This PR introduces a `versionProvider` configuration option and provider interface.

The implemented providers are:

- `node`: preserves the existing package.json version/dependency update behavior.
- `ruby`: updates Ruby gem version files while allowing `package.json` to remain Changesets metadata.

For Ruby packages, the provider can update:

- `lib/<gem-name>/version.rb` or a configured/discovered `lib/**/version.rb`
- `<gem-name>.gemspec` or a configured/discovered root `.gemspec`
- `Gemfile.lock`

It also makes `changeset status` and `changeset version` read the current version from the configured provider before assembling/applying the release plan. That matters for non-Node packages because `package.json` may not contain the canonical package version.

## Configuration Shape

The new config is optional. If omitted, behavior remains the historical Node/npm behavior.

Examples:

```json
{
  "versionProvider": "ruby"
}
```

Or, for mixed repositories:

```json
{
  "versionProvider": {
    "default": "node",
    "packages": {
      "my-gem": {
        "type": "ruby",
        "gemName": "my-gem",
        "versionFile": "lib/my_gem/version.rb",
        "gemspec": "my-gem.gemspec",
        "gemfileLock": "Gemfile.lock"
      }
    }
  }
}
```

I intentionally avoided requiring users to write `"auto"` in their config. When the option is unset, Changesets behaves automatically using the default provider. The implementation still has internal resolution logic, but the public docs avoid making users spell out automatic behavior.

## Ruby Behavior

For a Ruby provider package, `changeset version` can now:

- read the current version from Ruby files
- compute the next version using the normal Changesets release plan
- update `VERSION = "..."`
- update `spec.version = "..."`
- update the package entry in `Gemfile.lock`
- write changelog entries
- consume the changeset files

It deliberately does not update `package.json` for Ruby packages. In this model, `package.json` is just the metadata file that lets Changesets discover and name the package; it is not treated as the canonical Ruby package version.

## Demo Evidence

I put together a standalone demo repository to prove the workflow end to end:

https://github.com/jbeckwith-oai/changesets-ruby-provider-demo

The generated release PR is here:

https://github.com/jbeckwith-oai/changesets-ruby-provider-demo/pull/1

That release PR demonstrates:

- multiple ordinary changesets merged to `main`
- a mixed minor + patch release plan
- generated changelog content
- Ruby version files moving from `0.2.1` to `0.3.0`
- no `package.json` version bump
- manual publishing expectations for the gem

One important note from the demo: `changesets/action@v1` currently builds its PR body by detecting `package.json` version changes after running `changeset version`. Since Ruby packages intentionally do not bump `package.json`, the stock action produced an empty `# Releases` section even though the CLI/provider behavior was correct. For the demo, I switched to explicit CLI orchestration:

- `changeset status --output`
- `changeset version`
- generate the PR body from the status/changelog data
- commit/push/create-or-update the release PR

I think a provider-aware follow-up to `changesets/action` would be useful, but I kept that out of this PR so we can discuss the core provider model first.

## Compatibility

This is intended to preserve existing behavior for current Changesets users:

- no config change is required for existing Node/npm projects
- the Node provider is the default
- dependency range updates remain in the Node provider
- existing `privatePackages` handling remains respected
- the new provider logic is contained behind the version-provider interface

## Tests And Validation

Added coverage includes:

- config parsing for string, object, package override, and provider-specific options
- JSON schema coverage for the new config
- Node provider package.json version/dependency behavior
- Ruby provider version file, gemspec, and Gemfile.lock updates
- Ruby provider path configuration and opt-out behavior
- Ruby provider prerelease conversion for Bundler lockfile format
- CLI `status` and `version` behavior when provider-managed versions differ from `package.json`
- docs for config and CLI behavior

Validation run locally on this branch:

- `pnpm vitest run packages/cli/src/commands/version/version.test.ts packages/apply-release-plan/src/index.test.ts packages/config/src/parse.test.ts packages/config/scripts/json-schema.test.ts`
  - 4 files passed
  - 200 tests passed
  - 1 existing todo
- `pnpm lint`
- `pnpm build`
- `pnpm types:check`
- `pnpm format`
- `git diff --check upstream/main...HEAD`

I also ran the full suite with `pnpm test run`. On my local machine, it produced baseline/environment failures unrelated to this change:

- `packages/cli/src/commands/publish/__tests__/npm-utils.test.ts` expected the default npm registry, but my OpenAI environment injects an internal socket-firewall npm registry. The same test passes when the npm registry env vars are cleared.
- `packages/git/src/index.test.ts` has three shallow-clone tests timing out at their 10s limit. I reproduced the same failures on an untouched `upstream/main` worktree after building it, so this appears to be local/baseline behavior rather than caused by this PR.

## Questions For Maintainers

The biggest things I would love feedback on:

- Is `versionProvider` the right conceptual boundary/name?
- Should the default provider resolution stay conservative as `node`, or should auto-detection be more prominent?
- Does keeping `package.json` as discovery metadata for non-Node packages feel acceptable?
- Should provider implementations live in `apply-release-plan`, or should the interface move somewhere more central?
- Would you prefer the Ruby provider to be in this PR as proof of the abstraction, or would you rather see the provider interface land first and Ruby support follow separately?
- Should provider-aware release PR rendering be considered part of Changesets core, `changesets/action`, or both?

Again, thank you for the project and for considering this direction. I would be very glad to iterate on the shape with you.

### DIFF
--- a/.changeset/fresh-ravens-provide.md
+++ b/.changeset/fresh-ravens-provide.md
@@ -1,0 +1,8 @@
+---
+"@changesets/apply-release-plan": minor
+"@changesets/cli": minor
+"@changesets/config": minor
+"@changesets/types": minor
+---
+
+Add version providers with initial Ruby support.
--- a/docs/command-line-options.md
+++ b/docs/command-line-options.md
@@ -79,7 +79,7 @@ changeset add --since=develop
 changeset version
 ```
 
-This is one of two commands responsible for releasing packages. The version command takes changesets that have been made and updates versions and dependencies of packages, as well as writing changelogs. It is responsible for all file changes to versions before publishing to npm.
+This is one of two commands responsible for releasing packages. The version command takes changesets that have been made and updates versions and dependencies of packages, as well as writing changelogs. By default it updates `package.json` files, and the [`versionProvider`](./config-file-options.md#versionprovider-node--ruby--object) config can keep other ecosystem-specific version files in sync before publishing.
 
 > We recommend making sure changes made from this command are merged back into the base branch before you run publish.
 
--- a/docs/config-file-options.md
+++ b/docs/config-file-options.md
@@ -244,6 +244,60 @@ When `tag` is set to `true`, Changesets will create a tag for private packages. 
 }
 ```
 
+## `versionProvider` (`"node"` | `"ruby"` | object)
+
+This option controls which provider reads and updates version files when `changeset version` applies a release plan. When unset, Changesets uses automatic provider selection.
+
+The `node` provider updates the package's `package.json` version and npm dependency ranges. This is the historical Changesets behavior.
+
+The `ruby` provider reads the current version from Ruby version files and updates them when they exist:
+
+- `lib/<gem-name>/version.rb` or a single discovered `lib/**/version.rb`
+- `<gem-name>.gemspec` or a single discovered root `*.gemspec`
+- `Gemfile.lock`
+
+When `versionProvider` is unset, Changesets selects the Ruby provider if it finds Ruby version files in the package directory. Otherwise it uses the Node provider. Package-specific entries in `packages` take precedence over the default provider.
+
+You can configure a single provider for every package:
+
+```json
+{
+  "versionProvider": "ruby"
+}
+```
+
+For mixed repositories, configure a default and package-specific overrides:
+
+```json
+{
+  "versionProvider": {
+    "packages": {
+      "my-gem": {
+        "type": "ruby",
+        "gemName": "my-gem",
+        "versionFile": "lib/my_gem/version.rb",
+        "gemspec": "my-gem.gemspec",
+        "gemfileLock": "Gemfile.lock"
+      }
+    }
+  }
+}
+```
+
+Set `versionFile`, `gemspec`, or `gemfileLock` to `false` to skip that Ruby file.
+
+### Ruby provider options
+
+- `type`: must be `"ruby"`.
+- `gemName`: the Ruby gem name used to find `<gem-name>.gemspec`, `lib/<gem-name>/version.rb`, and the `Gemfile.lock` entry. If omitted, Changesets derives it from the package name.
+- `versionFile`: path to a Ruby version file, relative to the package directory. If omitted, Changesets checks `lib/<gem-name>/version.rb`, `lib/<gem-name-with-underscores>/version.rb`, or a single discovered `lib/**/version.rb`.
+- `gemspec`: path to a `.gemspec` file, relative to the package directory. If omitted, Changesets checks `<gem-name>.gemspec`, `<gem-name-with-underscores>.gemspec`, or a single discovered root `.gemspec`.
+- `gemfileLock`: path to `Gemfile.lock`, relative to the package directory. If omitted, Changesets updates `Gemfile.lock` when it exists.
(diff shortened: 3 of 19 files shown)


Andarist:
We are working right now on v3 revamp (it should be released soon!). Having support for other publishing targets is something we'd like to support but it requires a deeper thought how to support this best. We definitely can't commit to it before we publish v3, this is going to be - hopefully - explored once we reach that milestone.

FWIW, there are tricky ways use Changesets with those other targets. For example, we release `changesets/action` itself using Changesets... even though it doesn't publish to npm. It requires a pinch of manual/custom scripting though.

jbeckwith-oai:
I completely understand!  Just so I can plan my own stuff here:  do you have a time horizon?  I am happy to fork and add support for all of this stuff we need in our ow, but I wanted to open a dialog here first.  Obviously we'd rather contribute upstream 🤗

Andarist:
Perhaps you could join [our Discord](https://chat.changesets.dev/) and start a thread about this?

beeequeue:
there's also [`sampo`](https://github.com/bruits/sampo) which also doesn't support Ruby, but is more focused on cross-language support

giancarlosisasi:
Hi there
Been following this thread and it got me thinking.

Changesets is honestly my favorite release tool, and I keep wishing I had the same workflow in other languages. So instead of one tool trying to cover every ecosystem, what if there were separate ports like changesets-python or changesets-ruby, living right here in the same changesets org alongside the JS package? Each language has its own particularities around packaging, registries, semver and version file formats, so bending a single abstraction around all of it feels like it'd get painful fast.

That way each port could stay close to its own ecosystem while keeping the changesets workflow everyone already loves, all under the same org. And realistically, the people to maintain them are probably already here in these threads asking for it.

beeequeue:
we will be looking into big rewrites for v4, including setting up for supporting more ecosystems via a plugin system :)

PR #2126, opened 2026-06-26
docs: update and enhance comparison to other tools

trueberryless:
#### Description

This PR updates the "What distinguishes this from…" section on the "Technical Decisions" page.

One reason I wanted to update it is that it doesn't mention "Release Please", which is also a popular versioning automation tool. I think it fits very well to this section. I created two new subheadings to specifically compare Semantic Release and Release Please individually.

There are also some other reasons to enhance the section - many small details that I wanted to improve. It's probably best to review it by looking at the diff, but to give some overview: 
- Structure the order list.
- Update the mention of `lerna semantic release`, which is unmaintained since 2019.
- Reference community-driven changeset ports to other languages.
- In general, just provide some more information that I would have found useful when first exploring all those options some years ago.

I'm open for suggestions, this is a draft of my vision of enhancement for this section. Feel free to suggest feedback 🫶

### DIFF
--- a/site/guide/technical-decisions.md
+++ b/site/guide/technical-decisions.md
@@ -37,7 +37,7 @@ There are two reasons we chose to do this. The first is so the changeset descrip
 
 While many versioning tools rely on conventional commits to determine releases, Changesets was built with a different philosophy, prioritizing monorepo management and flexible workflows over strict commit message parsing.
 
-1. Monorepo-first design: Changesets is built to manage complex workspace topologies. It allows you to group linked packages, define fixed packages, and explicitly declare how internal dependency bumps (like `peerDependencies`) cascade through your repository.
+1. Monorepo-first design: Changesets is built to manage complex workspace topologies. It allows you to group linked packages, define fixed packages, and explicitly declare how internal dependency bumps cascade through your repository.
 
 2. Intent-based files: Instead of parsing git commit messages, change intent is stored in dedicated Markdown files committed alongside your code. This ensures that release information is preserved regardless of how your git history is squashed or rewritten.
 
--- a/site/guide/technical-decisions.md
+++ b/site/guide/technical-decisions.md
@@ -33,21 +33,31 @@ All updating of dependencies is done as a patch bump. If you want to indicate a 
 
 There are two reasons we chose to do this. The first is so the changeset descriptions are editable after creation, and a user can go in and change this as they desire. The second is that it means we are unopinionated about your git workflows, with squashing and modifying commits being completely safe, without fear of breaking a release.
 
-## What distinguishes this from Semantic Release
+## What distinguishes this from other versioning tools
 
-If you have been looking at automating versioning previously, you may have come across [semantic release](https://github.com/semantic-release/semantic-release), or its monorepo equivalent [lerna semantic release](https://github.com/atlassian/lerna-semantic-release). It's good to understand how changesets operate differently.
+If you have been looking at automating versioning previously, you may have come across [Release Please](https://github.com/googleapis/release-please) or [Semantic Release](https://github.com/semantic-release/semantic-release). It's good to understand how changesets operate differently.
 
-1. Changesets are designed for monorepos first.
+1. Changesets are designed for monorepos first. Changesets allow you to group linked packages, define fixed packages, and explicitly declare whether an internal dependency bump should trigger a patch bump on the consuming package.
 
-This means we manage dependencies within the repository, which other tools do not do.
+2. We store our change intent in dedicated Markdown files alongside your code, rather than relying on strict git commit message parsing. See the above section on why we write files to disc.
 
-2. We commit our change information to the file system, instead of storing it in git.
+3. We use semver for specifying the change. When selecting the kind of change your package is, we do not specify any change types beyond `major`, `minor`, or `patch`. In comparison, [conventional commits](https://github.com/conventional-commits/conventionalcommits.org) specify the type of a commit (bug, feat) that gets converted to an appropriate semver type. This is a design decision on our part to push adding this information into the changeset description itself.
 
-See the above section on why we write files to disc.
+### Comparison to Semantic Release
 
-3. We use semver for specifying the change.
+[Semantic Release](https://github.com/semantic-release/semantic-release) focuses on publishing bug fixes and new features as soon as possible without release PRs. With the default CI for Semantic Release, conventional commits landing on the specified branch are published immediately.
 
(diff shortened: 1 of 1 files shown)


site/guide/technical-decisions.md
```diff
@@ -33,21 +33,31 @@ All updating of dependencies is done as a patch bump. If you want to indicate a
 
 There are two reasons we chose to do this. The first is so the changeset descriptions are editable after creation, and a user can go in and change this as they desire. The second is that it means we are unopinionated about your git workflows, with squashing and modifying commits being completely safe, without fear of breaking a release.
 
-## What distinguishes this from Semantic Release
+## What distinguishes this from other versioning tools
```
bluwy: Could we update this to "from conventional commits based versioning tools"? This might help address the main topic here without mentioning a specific name.

I think we'd also want to hold off documenting differences to names like Semantic Release and Release Pleas now because it can be a sensitive topic, so it's better for a core maintainer to draft the main differences between the tools later on.

site/guide/technical-decisions.md
```diff
@@ -33,21 +33,31 @@ All updating of dependencies is done as a patch bump. If you want to indicate a
...
 
-1. Changesets are designed for monorepos first.
+1. Changesets are designed for monorepos first. Changesets allow you to group linked packages, define fixed packages, and explicitly declare whether an internal dependency bump should trigger a patch bump on the consuming package.
 
-This means we manage dependencies within the repository, which other tools do not do.
+2. We store our change intent in dedicated Markdown files alongside your code, rather than relying on strict git commit message parsing. See the above section on why we write files to disc.
 
-2. We commit our change information to the file system, instead of storing it in git.
+3. We use semver for specifying the change. When selecting the kind of change your package is, we do not specify any change types beyond `major`, `minor`, or `patch`. In comparison, [conventional commits](https://github.com/conventional-commits/conventionalcommits.org) specify the type of a commit (bug, feat) that gets converted to an appropriate semver type. This is a design decision on our part to push adding this information into the changeset description itself.
```
bluwy: I think we should hold off this change for now, so only talk about using semver for changes. In the future we may actually want to support specifying changes by its type and then perform the correct semver bump. I believe there's open issues of this.

site/guide/technical-decisions.md
trueberryless: So should I just in general completely remove the references to Semantic Release and Release Please and just mention the general differences to conventional commit based tools for versioning?

IMHO this is great for a general picture, but my aim here also was to make the decision for the user who is unfamiliar with such tools and guide them to popular tools that exist. But I understand that it can be sensitive, that's why I tried to keep the phrasing objective as much as possible.

site/guide/technical-decisions.md
trueberryless: This info was actually not changed and more copied and little adapted from the original:

> When selecting the kind of change your package is, we do not specify any change types beyond `major`, `minor`, or `patch`. The semantic release allows you to specify a range of fields (bug-fix, feature) that it converts to an appropriate semver type. This is a design decision on our part to push adding this information into the changeset description itself.

However, I can adapt it if necessary. I just want to make one thing clear: I want documentation to represent the _current_ state of the project as much as possible, and because we currently do not support any semver type auto-bumps AFAIK, I am a little reluctant to removing this information. I mean, we can always adapt it later if we have support for it.

bluwy: Ah I see. Then ignore this then, I didn't realize it was there before.

site/guide/technical-decisions.md
bluwy: I think we can mention here "conventional commit based tools such as Semantic Release and Release Please", but I mean specifically the `# Comparison to Semantic Release` and `# Comparison to Release Please` sections (which seems new) covers the more specific differences which is the more sensitive part that I prefer to hold off.

site/guide/technical-decisions.md
trueberryless: Yeah, the git diff is not ideal with the changes, it doesn't recognise similar sentences when removing some newlines...

site/guide/technical-decisions.md
trueberryless: At first, I played around with just having one section, but it was too bloated. Therefore, I decided to split into these sections. But it's a good point that we probably shouldn't pick two other tools specifically. I can try to rewrite it so we just link to similar tools while keeping especially disadvantages in a more general section. 

One section/sentence I would like to keep tho is the newly added links to changeset-inspired adopted versioning tools for other languages. Wdyt?

bluwy: > One section/sentence I would like to keep tho is the newly added links to changeset-inspired adopted versioning tools for other languages. Wdyt?

Hmm I personally would also leave them out, otherwise it might be seen as tooling that we endorse, and we can't guarantee it's good to use and work the way we believe changesets should work.

trueberryless: Thanks for the detailed explanation!

I completely rephrase the section so it's more objective and doesn't specifically mention other tools. It now aims to provide an overview compared to conventional commit-based tools in general 👍

site/guide/technical-decisions.md
```diff
@@ -33,21 +33,15 @@ All updating of dependencies is done as a patch bump. If you want to indicate a
...
 
-## What distinguishes this from Semantic Release
+## How Changesets differs from conventional commit-based tools
 
-If you have been looking at automating versioning previously, you may have come across [semantic release](https://github.com/semantic-release/semantic-release), or its monorepo equivalent [lerna semantic release](https://github.com/atlassian/lerna-semantic-release). It's good to understand how changesets operate differently.
+While many versioning tools rely on conventional commits to determine releases, Changesets was built with a different philosophy, prioritizing monorepo management and flexible workflows over strict commit message parsing.
 
-1. Changesets are designed for monorepos first.
+1. Monorepo-first design: Changesets is built to manage complex workspace topologies. It allows you to group linked packages, define fixed packages, and explicitly declare how internal dependency bumps (like `peerDependencies`) cascade through your repository.
```
bluwy:
```suggestion
1. Monorepo-first design: Changesets is built to manage complex workspace topologies. It allows you to group linked packages, define fixed packages, and explicitly declare how internal dependency bumps cascade through your repository.
```

We don't have a perfect way to configure dep bumps for peer deps yet, other than the experimental options (which is still finicky). I think we should remove that part for now.

bluwy (commented):
Left a small comment but I like the new changes!

bluwy (approved):

PR #2127, opened 2026-06-26
docs: create page for GitHub generator package

trueberryless:
#### Description

This PR creates a new docs page for the official GitHub generator package from changesets. It also includes the documentation of the new experimental `template` option from https://github.com/changesets/changesets/pull/2059


site/guide/config.md
```diff
@@ -28,19 +28,25 @@ Set how the changelog for packages should be generated. The default changelog ge
 
 You can also specify a custom changelog generator by providing a string or a tuple with the module path and options. The module can be an npm package or a relative file path. Check out the [Customize Changelog Format](./customize-changelog-format.md) guide to learn how to write your own changelog generator.
 
-If a tuple (`[string, Record<string, any>]`) is set, you can pass options in the second item of the tuple to configure the changelog generator. For example, if you are using `@changesets/changelog-github`, you can pass the `repo` and `disableThanks` options like this:
+If a tuple (`[string, Record<string, any>]`) is set, you can pass options in the second item of the tuple to configure the changelog generator. For example, if you are using `@changesets/changelog-github`, you can pass the `repo`, `disableThanks`, and `template` options like this:
```
bluwy: Since we'll have docs for changelog-github directly, let's perhaps trim this example to only with `repo` so that it's simpler. We also want to avoid documenting using `template` here as using it requires pinning the version.

site/guide/config.md
```diff
@@ -28,19 +28,25 @@ Set how the changelog for packages should be generated. The default changelog ge
...
+      "disableThanks": true,
+      "template": "\n- {summary} {ref}"
+    }
   ]
 }
 ```
 
 Consult the documentation for the changelog generator you're using for more information on the options you can pass to it.
 
+Read more about the [`@changesets/changelog-github` generator](./github-changelog-generator.md).
```
bluwy: I _think_ we should refer this as "`@changesets/changelog-github` package" rather than "generator" for now, unless we refer to it as "changelog generator" entirely, but it makes it wordy, so we should use either one depending on the flow of the wording.

bluwy (commented):
Some general reviews for the wording before rebasing to next-docs

packages/changelog-github/README.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## Options
+
+The GitHub Changelog generator supports the following configuration options:
+
+### `repo`
+
+- **Required**
+- **Type:** `string`
```
bluwy, line 30: This should also be optional now as it can be specified via `GITHUB_REPOSITORY` which was supported recently. But must at least exists in one place, so we could document this as well.

site/guide/github-changelog-generator.md
```diff
@@ -0,0 +1,107 @@
...
+
+:::
+
+The GitHub Changelog generator requires a [`GITHUB_TOKEN`](https://github.com/settings/tokens/new?scopes=read:user,repo:status&description=changesets) with `read:user` and `repo:status` permissions. Add it to your repository's [GitHub Action secrets](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets) and your `.env` file:
+
+```bash
+GITHUB_TOKEN=token_here
+```
+
+Then you can use it in the [`changelog`](./config.md#changelog) option:
```
trueberryless: Re: Rebasing this onto the README.
Could we convert links like this to absolute links to docs site then?
e.g.: 

```suggestion
Then you can use it in the [`changelog`](https://changesets.dev/guide/config#changelog) option:
```

packages/changelog-github/README.md
```diff
@@ -3,47 +3,91 @@
...
+
+- **Type:** `boolean`
+- **Default:** `false`
+
+Set `"disableThanks": true` to drop the `"Thanks [@user]!"` attribution from each line.
 
-Set `"disableThanks": true` to drop the "Thanks [@user]!" attribution from each line.
+::: info
+It is recommended to not set `"disableThanks": true` when using `template`, as the `{authors}` token returns an empty string then, which could lead to unexpected results.
+:::
```
bluwy: This should be updated to use a GitHub supported syntax, like `> [!NOTE]` so that it renders correctly in both GitHub and VitePress.

packages/changelog-github/README.md
```diff
@@ -3,47 +3,91 @@
...
+Each piece of information is dynamically represented with a [token](#tokens). The above example uses this `template` syntax:
+
+```
+\n- {pull} {commit} Thanks {authors}! - {summary}
+```
 
-### `template` (optional string, experimental)
+::: tip
+Note that `\n- ` is included to generate each change as a bullet list item. However, you can also drop them and write all changes into a paragraph. Trailing spaces will be trimmed to avoid dangling spaces when the last token is empty.
+:::
```
bluwy: Same here

packages/changelog-github/README.md
```diff
@@ -3,47 +3,91 @@
...
+::: info
+It is recommended to not set `"disableThanks": true` when using `template`, as the `{authors}` token returns an empty string then, which could lead to unexpected results.
+:::
+
+### `template`
+
+- **Type:** `string`
+- **Default:** `"\n- {pull} {commit} Thanks {authors}! - {summary}"`
+
+> ⚠️ **Experimental.** The `template` option and its token syntax may change in any release, including a patch. If you rely on it, pin the exact `@changesets/changelog-github` version.
```
bluwy: Would also be nice if this part uses `> [!WARNING]`

packages/changelog-github/README.md
trueberryless: good call, thx

trueberryless:
```suggestion
> [!NOTE]
> It is recommended to not set `"disableThanks": true` when using `template`, as the `{authors}` token returns an empty string then, which could lead to unexpected results.
```

packages/changelog-github/README.md
trueberryless:
```suggestion
> [!WARNING]
> **Experimental.** The `template` option and its token syntax may change in any release, including a patch. If you rely on it, pin the exact `@changesets/changelog-github` version.
```

packages/changelog-github/README.md
trueberryless:
```suggestion
> [!NOTE]
> Note that `\n- ` is included to generate each change as a bullet list item. However, you can also drop them and write all changes into a paragraph. Trailing spaces will be trimmed to avoid dangling spaces when the last token is empty.
```

trueberryless:
Thanks for the review.

I decided to remove the warning sign emoji, as GitHub adds it in the title of the alert:

image

bluwy (approved):
Perfect. Thanks!

PR #2128, opened 2026-06-26
Rename `changeset tag` to `changeset git-tag`

Andarist:
`changeset publish` accepts both `--tag` and `--git-tag`. I think `changeset tag` was incorrectly named when it got introduced as that didn't match the preexisting `changeset publish` vocab.


trueberryless (approved):
LGTM 🎉 

Thanks for the cleanup!

bluwy:
I think it's fine keeping it as `tag` though. We don't need the commands to be related to npm, so tag being "tagged via git" is somewhat ok to me. So I think it's not worth breaking or deprecating for now, until we have some plans to create a different `changeset tag` command & behaviour.

Andarist:
I'm fine with keeping the name as-is - but I recall some users reporting in the past that `changeset tag` was misleading to them because it was different from what `changeset publish` calls a tag

bluwy:
I think after reviewing https://github.com/changesets/action/pull/680, I could kinda see the benefit of this. But if we're to land this, the `changeset tag` command should still work but log a deprecation warning instead.

trueberryless:
> I think after reviewing [changesets/action#680](https://github.com/changesets/action/pull/680), I could kinda see the benefit of this. But if we're to land this, the `changeset tag` command should still work but log a deprecation warning instead.

I mean to be fair, we also rename [some action inputs in that major bump](https://github.com/changesets/action/pull/668) without deprecation, so why not do it like that here?

Not to say, I do not like a slow migration. I can see that in the action two inputs for one setting at the same time are confusing, so it makes sense there. But theoretically we could do the same error with a "got renamed" message here as well...

Andarist:
I don't mind making the old `changeset tag` work with a deprecation warning - given we don't really have any other plans for `changeset tag`.

I pushed out a commit to that effect

packages/cli/src/cli.ts
```diff
@@ -147,11 +147,15 @@ cli
   });
 
 cli
-  .command("tag", "Create git tags for the current version of all packages")
+  .command("git-tag", "Create git tags for the current version of all packages")
+  .alias("tag")
```
Andarist, line 151: from what i can tell this alias doesn't appear in the `--help` and I couldn't think of any other place that could be somehow negatively affected by this

beeequeue (approved):
my initial reaction was that i didn't see the point in doing this, but in the end it's more specific which can only help.

packages/cli/src/cli.ts
bluwy: Yeah I think that's ok

bluwy:
> I mean to be fair, we also rename [some action inputs in that major bump](https://github.com/changesets/action/pull/668) without deprecation, so why not do it like that here?
> 
> Not to say, I do not like a slow migration. I can see that in the action two inputs for one setting at the same time are confusing, so it makes sense there. But theoretically we could do the same error with a "got renamed" message here as well...

For me the core commands here feel more important and general, where it can be used for many cases like setting up a custom bash script, child process, custom workflows etc that it's intricate. If it suddenly doesn't work, it'd have a larger impact compared to the breaking changes in the action.

But you're right, initially we're conservative with breaking changes, making the major mostly clean-up focused, but we've recently been making quite some breaking stuff already. But given compat support is only a few lines, I think we can swallow the cost for now.

PR #2129, opened 2026-06-26
Emit NDJSON `git-tag` events when `CHANGESETS_OUTPUT` is set

Andarist:
This is an alternative to the clunky system we were relying on until now. We had to temporarily bring back the old thing recently as it was removed because it was just a totally implicit way of handling things (see [this](https://github.com/changesets/changesets/pull/2091))

This will pair with https://github.com/changesets/action/pull/678


packages/cli/src/commands/publish/index.ts
```diff
@@ -92,6 +94,8 @@ To resolve this exit the pre mode by running ${c.cyan("changeset pre exit")}.
     return;
   }
 
+  await using reporter = await createOutputReport(options?.outputPath);
```
beeequeue: `using` is sadly only available in node 24.11+ :(
not sure if it's being transpiled right now, the tests should probably be failing...

packages/cli/src/cli.ts
```diff
@@ -46,6 +46,10 @@ function normalizeOptions(
       options[key] = String(options[key]);
     }
   }
+
+  if (outputPath && process.env.CHANGESETS_OUTPUT_PATH) {
+    options.outputPath = process.env.CHANGESETS_OUTPUT_PATH;
```
Andarist: 2 problems here...

1. we support `--output` + `--out-dir` (in different commands). It feels to me we should standardize on `--out-file` + `--out-dir` and then rename this variable to `CHANGESETS_OUT_FILE`
2. all pre-existing `--output`s spit out JSON. I like the idea of using NDJSON for those `git-tag` events. It seems just much easier to manage and extend. This would kinda mean we should refactor existing `--output` to also use the NDJSON format if we want to follow this path. This would also affect the recently added `publish-plan` (where a single JSON is kinda easier to manage but given we are primary producers/consumers of that publish plan... we can take it)

packages/cli/src/commands/publish/index.ts
Andarist: It gets transpiled fine by oxc though, see: https://github.com/oxc-project/oxc/blob/2a2d3b981f18a40b443ae6a1a9395cf42a192475/npm/runtime/src/helpers/esm/usingCtx.js

And `node@22.11` already has global `Symbol.dispose`/`Symbol.asyncDispose` available. It just doesn't support the `using` syntax itself (sadly)

beeequeue: seems like oxc can transpile it, yay!

packages/cli/src/cli.ts
Andarist: Although, after the second consideration... it wouldn't be good if `publish-plan` would return NDJSON. Or rather, returning that is fine... but we feed that plan back to `changeset pack` and it feels weird to make that accept NDJSON file. Or maybe not?

beeequeue: i think using different formats is fine.

beeequeue (approved):

packages/cli/src/cli.ts
```diff
@@ -18,7 +18,7 @@ cli.globalCommand.outputVersion = () => console.info(manifest.version);
 
 function normalizeOptions(
   options: Record<string, any>,
-  { array }: { array?: string[] } = {},
+  { array, outputPath }: { array?: string[]; outputPath?: boolean } = {},
```
bluwy: I'd prefer putting the `outputPath` handling in a separate function since this function is a general workaround to cac's quirks, i.e. something someone else could copy as is without including our business logic.

packages/cli/src/cli.ts
bluwy: 1. I don't think it should be `--out-file` and `--out-dir`. Both are for different things: output logs and command-specific artifact directory
2. I think mixing JSON and NDJSON is fine too. We can keep a rule where long-running tasks or "potential to fail mid-process but mid-output logs is still important" to use NDJSON, and the rest to use JSON? Then that justifies the split we have now.

Andarist: Ok, let's roll with the split logic - where some commands use NDJSON and some use JSON (a single JSON output is still valid NDJSON anyway 😉 ). 

> I don't think it should be --out-file and --out-dir. Both are for different things: output logs and command-specific artifact directory

What names do you suggest sticking to then?

bluwy: I think the current `--output` and `--out-dir` is fine.

Maybe if I'd start from scratch, I wouldn't do `--output` and support `--json` only (outputs to stdout) so it's more natural for script piping. It can be used for both json and ndjson output (or `--ndjson` if we really want to). And support an `CHANGESETS_JSON_LOG_PATH` env var that also allows writing the logs to a specific file (only env var and never a flag, so it works like debug logging). But that's just perhaps a different way to achieve the same thing and not worth it.

Andarist: I think it is still worth supporting those output files with a single env variable. 

I have just pushed support for this for `changeset status` and `changeset publish-plan` so a more full picture could be reviewed

bluwy:
The new changes also work for me. Is it too far to shorten the env var to `CHANGESETS_OUTPUT` to match `--output`? (Since it's not `--output-file`)

PR #2132, opened 2026-06-29
Fix pnpm 11 publish errors handling

### LINKED ISSUE #2099: changeset publish crashes on already-published packages under npm trusted publishing (OIDC)
### Summary

Under npm trusted publishing (OIDC), `changeset publish` runs with no auth token, because an `_authToken` in `.npmrc` makes npm skip its OIDC exchange. With no token, the pre-publish `npm info` check returns 404 for every private package, so changesets concludes nothing is published and attempts to publish every package, including unchanged ones whose current version is already on the registry. Those return `E403`, and instead of being skipped the run crashes:

```
🦋 error TypeError: Cannot read properties of undefined (reading 'includes')
🦋 error     at isAlreadyPublishedError (.../@changesets/cli/dist/changesets-cli.cjs.js:873:17)
🦋 error     at internalPublish (.../@changesets/cli/dist/changesets-cli.cjs.js:957:41)
```

### Root cause

In the `--json` publish path inside `internalPublish`:

```js
if (json.error.code === "E403" && isAlreadyPublishedError(json.error.summary)) { ... }
// isAlreadyPublishedError(output) => output.includes("cannot publish over the previously published version")
```

With `pnpm publish --json` (and bun), the error JSON has `error.code === "E403"` but no `error.summary`, so `isAlreadyPublishedError(undefined)` throws. npm does populate `summary`, so npm-based repos don't crash, but they still report the already-published package as `failed` rather than `skipped`, because the message ("...previously published versions: x.y.z") isn't always matched where the check looks. The publish tool is chosen by changesets via `package-manager-detector`, so which behaviour you get depends on the detected package manager rather than anything the user controls.

### When this happens

This only shows up with npm trusted publishing, which is still an uncommon setup, so most workflows never reach this path. But for a repo that does use it, the tokenless publish means the `npm info` precheck can never see private packages, so changesets re-attempts already-published versions on every release. Any release that leaves at least one package unchanged (its version already on npm) will hit the failing E403 path.

### Reproduction

1. pnpm (or bun) workspace with private scoped packages, publishing via npm trusted publishing (OIDC, no token present at publish time).
2. A release where at least one package is unchanged and its current version is already on npm.
3. `pnpm changeset publish` crashes on the first already-published package.

### Workaround

For anyone hitting this before it's fixed: we added a step that runs before `changeset publish`, while a read token is still available. It checks each workspace package's current version against the registry and sets `private: true` on the ones that are already published. changesets skips private packages, so `changeset publish` then only attempts the genuinely new versions. The read token is removed again before the publish step, so the actual publish still happens tokenless via OIDC. The package.json edits are made in the CI checkout and never committed. This sidesteps the crash entirely and has the side benefit of not re-attempting every unchanged package on each release.

### Proposed fix

Make `isAlreadyPublishedError` null-safe and have it check all the fields the message might land in across package managers:

```js
function isAlreadyPublishedError(...parts) {
  return parts.filter(Boolean).join("\n").includes("cannot publish over the previously published version");
}
// call: isAlreadyPublishedError(json.error.summary, json.error.detail, stderr.toString())
```

It would also help to document that trusted publishing leaves the `npm info` precheck unauthenticated, since the tokenless read is a consequence of the OIDC requirement.

Environment: `@changesets/cli@2.31.0`, npm trusted publishing (OIDC), pnpm 11.6 / bun 1.2.

cyphercodes:
## Summary

- Make already-published publish error detection tolerate missing JSON `error.summary`.
- Check `summary`, `detail`, `message`, and raw stdout/stderr for the already-published message.
- Add pnpm regression coverage for E403 JSON errors without a summary.

Fixes #2099

## Test plan

- `pnpm install --frozen-lockfile`
- `pnpm vitest run packages/cli/src/commands/publish/__tests__/index.test.ts packages/cli/src/commands/publish/__tests__/publishPackages.test.ts packages/cli/src/commands/publish/__tests__/npm-utils.test.ts`
- `pnpm lint -- packages/cli/src/commands/publish/npm-utils.ts packages/cli/src/commands/publish/__tests__/npm-utils.test.ts`
- `pnpm types:check`
- `pnpm format`
- `git diff --check`


packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -397,7 +399,13 @@ async function internalPublish(
     if (json?.error) {
       if (
         json.error.code === "E403" &&
-        isAlreadyPublishedError(json.error.summary)
+        isAlreadyPublishedError(
+          json.error.summary,
+          json.error.detail,
+          json.error.message,
+          stderr.toString(),
+          stdout.toString(),
```
Andarist: I'm not a fan of this. We should use targeted logic. We should probably hoist the `formatPublishError` call and use the normalized error throughout the caller's logic

Andarist (approved):

Andarist (approved):

PR #2135, opened 2026-06-30
feat(cli): allow comma-separated packages in --major/--minor/--patch

### LINKED ISSUE #2134: Allow comma-seperated packages designation in `--major`, `--minor`, and `--patch`
It would be much better if you don't have to repeat `--major`, `--minor`, or `--patch` multiple times by allowing the separation by commas (e.g. `--minor pkg1,pkg2 --patch pkg3,pkg4` instead of `--minor pkg1 --minor pkg2 --patch pkg3 --patch pkg4`).

_Originally posted by @tats-u in https://github.com/changesets/changesets/issues/1121#issuecomment-4842385981_

https://github.com/changesets/changesets/blob/ce2095d744ba03424641a8b7cf0a7cfee8bcc6f7/packages/cli/src/cli.test.ts#L69-L83

youdie006:
## Summary

Implements #2134 — allow comma-separated package lists in `--major`, `--minor`, and `--patch`.

`changeset add --minor pkg1,pkg2 --patch pkg3` now expands the comma-separated values, equivalent to `--minor pkg1 --minor pkg2 --patch pkg3`. Whitespace around names is trimmed and empty entries are dropped; single packages and repeated flags keep working as before.

- `normalizeOptions` in `packages/cli/src/cli.ts` flatMaps each value on commas.
- Updated the three option help strings to note `(comma-separated)`.
- Added a `cli.test.ts` case covering comma-split, whitespace-trim, and repeated-flag merge.

Fixes #2134


packages/cli/src/cli.ts
```diff
@@ -31,9 +31,17 @@ function normalizeOptions(
...
+    // Also expand comma-separated values so that e.g. `--minor pkg1,pkg2`
+    // is treated the same as `--minor pkg1 --minor pkg2`.
     if (array?.includes(key)) {
       const v = options[key];
-      options[key] = Array.isArray(v) ? v.map(String) : [String(v)];
+      const values = Array.isArray(v) ? v.map(String) : [String(v)];
+      options[key] = values.flatMap((value) =>
+        value
+          .split(",")
+          .map((part) => part.trim())
```
Andarist, line 48: let's throw in some simple tests for the trimming and `.filter(Boolean)` behaviors

packages/cli/src/cli.test.ts
```diff
@@ -82,6 +82,20 @@ const tests: CommandTest[] = [
...
         },
       },
+      {
+        args: [
+          "--minor",
+          "pkg-a, pkg-b",
+          "--patch",
+          "pkg-c,pkg-d",
+          "--patch",
+          "pkg-e",
```
Andarist, line 97: this combo is a little bit horrendous 😂 but I agree it should be supported

.changeset/comma-separated-bump-packages.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/cli": minor
+---
+
+Allow comma-separated package names in the `--major`, `--minor`, and `--patch` flags of the `add` command. For example, `--minor pkg-a,pkg-b` is now equivalent to `--minor pkg-a --minor pkg-b`.
```
Andarist: technically, this will also be supported by `--ignore` (and maybe some other flag too?), so we should mention this (+add tests for those other cases too)

bluwy:
@youdie006 Are you still interested in updating the PR?

packages/cli/src/cli.ts
youdie006: Added in f4fb72c: an `add` case with `--patch " pkg-a , , pkg-b , "` that exercises both the trimming and the `.filter(Boolean)` empty-entry dropping (expects `["pkg-a", "pkg-b"]`).

.changeset/comma-separated-bump-packages.md
youdie006: Good point - the same normalizeOptions path handles the version command's `--ignore` too. Updated the changeset to mention `--ignore` (and the trim/empty behavior), and added two `version` cases: `--ignore pkg-a,pkg-b` and a trim/empty variant.

packages/cli/src/cli.test.ts
youdie006: Kept that combined case and added narrower ones alongside it so each behavior (multi-flag, trimming, empty-drop, --ignore) is covered in isolation too.

youdie006:
@bluwy yes, still interested - sorry for the delay. Just pushed f4fb72c addressing @Andarist's review: tests for the trimming/`.filter(Boolean)` behavior, coverage for the version command's `--ignore` flag (same normalizeOptions path), and a changeset note mentioning it. All 27 cli tests pass locally.

bluwy (approved):
Thanks! LGTM 👍

Andarist (approved):

PR #2145, opened 2026-07-04
Remove Yarn Classic support

Andarist:


bluwy (commented):
Looks like this likely needs an update after the base branch is ready, so I'll hold off my approval. But for now the changes here all makes sense to me.

beeequeue (approved):

PR #1121, opened 2023-03-28
Added new `--major`, `--minor`, `--patch` flags to the `add` command

Sh031224:
## What is the Change?

Added new flag to add in @changeset/cli.

- `--major`
- `--minor`
- `--patch`

If the major, minor, and patch flags are present, no package is selected by user interaction.
Also, if there is a message flag, the summary can be injected as a flag without being input through interaction.

Reference: #1118

PR #2153, opened 2026-07-09
Add `--type` flag to `add` command

brunorodmoreira:
Reworks #1890 based on the feedback there: based on `next`, scoped down to only the `add` command (no `changed` command), and simplified to take just a bump type and a message instead of multiple `--package` flags.

The `add` command now supports a `--type` flag that provides the semver bump type (`patch`, `minor` or `major`) from the command line instead of prompting for it. The bump type is applied to all packages that have changed since the base branch (or the ref provided with `--since`). Combined with `--message`, the changeset is created without any prompts, so scripts, CI and AI agents can add changesets non-interactively:

```sh
changeset add --type patch --message 'Fix crash on startup'
```

Notes:

- `--type` composes with the existing flags like `--message` does: on its own it only replaces the package selection and bump type prompts, and the summary is still asked for. Single package repos just use their sole package.
- In `--type` mode, failing to detect the changed packages (or detecting none) is a hard error instead of the interactive flow's best-effort warning, since the result now decides what gets released.
- Detecting changed packages is now skipped in single package repos, where the result was never used. This also removes a spurious warning when running `changeset` in a non-git single package repo.
- While documenting `--type` I noticed the CLI README never documented `--since`, so I added it there too.

Tested with unit tests for all the new paths and end-to-end against a real monorepo (27+ packages), including the error paths and exit codes.

Refs #979, #1604

### DIFF
--- a/.changeset/tall-months-shout.md
+++ b/.changeset/tall-months-shout.md
@@ -1,0 +1,13 @@
+---
+"@changesets/cli": minor
+---
+
+Add `--type` flag to the `add` command
+
+The `add` command now supports a `--type` flag that provides the semver bump type (`patch`, `minor` or `major`) from the command line instead of prompting for it. The bump type is applied to all packages that have changed since the base branch (or the ref provided with `--since`).
+
+When combined with `--message`, the changeset is created without any prompts, which makes it possible to add changesets from scripts and other automation:
+
+```sh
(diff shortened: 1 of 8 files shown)


Andarist:
We went with this one instead https://github.com/changesets/changesets/pull/1121 - its `--patch`/`--minor`/`--major` approach fits the existing Changesets design better.

PR #2064, opened 2026-06-06
Use a more faithful registry selection for publish lookups

Andarist:

PR #2155, opened 2026-07-11
rework publish flow, start splitting out PM-specific logic

beeequeue:
todo: add long explanation of why splitting up this pr just didn't work

the unindented list of published packages is fixed by this PR: https://github.com/bombshell-dev/clack/pull/584

- [x] log publish errors somehow
- [x] make sure git tag behavior is consistent with `git-tag` command

---

recordings:

interactive, 2fa required: https://asciinema.org/a/43rIq4XG0hrapPpy


packages/cli/src/commands/publish/publishPackages.ts
beeequeue, line 1: this file is ready to be removed as well, but i kept it for now to keep the PR easier to review

packages/cli/src/lib/npm.ts
```diff
@@ -0,0 +1,192 @@
...
+        ...resultBase,
+        result: !interactive ? "published" : "published:interactive",
+      };
+    }
+
+    /* -- error handling -- */
+
+    if (interactive) {
+      return { ...resultBase, result: "failed", summary: stderr || stdout };
+    }
```
bluwy: Is this handling needed? Should it fall through to the `handlePublishError` below?

packages/cli/src/commands/publish/index.ts
```diff
@@ -95,103 +120,200 @@ To resolve this exit the pre mode by running ${c.cyan("changeset pre exit")}.
...
+      onResult: () => p.advance(1),
+    });
+    successfulNpmPublishes.push(
+      ...results.filter((p) => p.result === "published"),
     );
-    const unsuccessfulNpmPublishes = publishedPackages.filter(
-      (p) => p.result === "failed",
+    unsuccessfulNpmPublishes.push(
+      ...results.filter((p) => p.result === "failed"),
     );
```
bluwy: Should these be using `isPublishSuccessful` and `isPublishFailure`?

packages/cli/src/lib/npm.ts
beeequeue: indeed, it's a leftover from previous iterations

packages/cli/src/lib/pnpm.ts
```diff
@@ -0,0 +1,125 @@
...
+          env: sanitizeEnv(process.env),
+          cwd,
+        },
+      },
+    );
+    const resultBase = { name: release.name, version: release.version };
+    if (exitCode === 0) {
+      return {
+        ...resultBase,
+        result: !interactive ? "published" : "published:interactive",
```
bluwy: It looks like usually the consumer can already infer if it's published with interactive since it passed the interactive param, but I assume this is done like this for now so it's easier to handle the output directly than processing it again.

Just an observation and we can improve this later.

packages/cli/src/commands/git-tag/utils.ts
bluwy, line 1: I would prefer moving this file inside the `git-tag` command, like `packages/cli/src/commands/git-tag/utils.ts` or some other name so it stays consistent for now. `publish-plan` and `getPublishPlan.ts` is also in a similar boat.

We can look into re-organizing again if we want to separate commands and general functions, but I think the current of grouping only within command folders is fine.

bluwy (commented):
I left out some small nits as it's not worth addressing given the hot changes in this area recently, but I left one comment about the file organization below which I felt slightly strongly of to address/discuss.

Otherwise, the new publish separation looks great to me. I wouldn't put the otp re-prompt as a blocker for v3, but it'd be nice to address it when we have the time.

packages/cli/src/lib/npm.ts
```diff
@@ -91,7 +91,7 @@ ${json.error.summary ?? ""}
 ${json.error.detail ?? ""}
   `.trim();
 
-  if (json.error.code === "EOTP") {
+  if (json.error.code === "EOTP" && process.stdin.isTTY === true) {
```
beeequeue: i dont think this is correct. the publish function should always return the correct error code, and then the caller should handle it however is needed. otherwise we are splitting up the logic again

packages/cli/src/lib/pnpm.test.ts
```diff
@@ -72,34 +66,10 @@ describe("publishing", () => {
...
-      });
-
-      expect(result.result).toEqual("failed:needs-2fa");
-      expect((result as PublishResultFailed).summary).toEqual(
-        expect.stringMatching(
-          /(requires additional authentication)|(requires a one-time password)/,
-        ),
-      );
-    },
-  );
```
beeequeue: do we really not want this unit test?

packages/cli/src/lib/pnpm.ts
```diff
@@ -60,13 +60,13 @@ export const publish: PublishTool["publish"] = async ({
...
     const cwd = pkg.dir;
     const args: string[] = [
-      "--json",
       "--access",
       release.access,
       "--tag",
       release.tag,
       "--no-git-checks",
     ];
+    if (!interactive) args.unshift("--json");
```
beeequeue: i originally had this as well, removed it and i can't remember why

packages/cli/src/lib/pnpm.ts
```diff
@@ -47,6 +52,60 @@ function sanitizeEnv(env: NodeJS.ProcessEnv): NodeJS.ProcessEnv {
 
 export const name = "pnpm" satisfies PublishTool["name"];
 
+function parseInfoResult({
+  exitCode,
+  stdout,
+  stderr,
+}: import("tinyexec").Output):
+  | { pkgInfo: PackageInfo }
```
beeequeue: can be renamed to `info`

packages/cli/src/utils/package-manager-errors.ts
```diff
@@ -90,30 +30,22 @@ export function formatNpmPnpmJsonError(
...
-    // Note that the `--json` output is always printed at the end so this should work
-    const json =
-      getLastJsonObjectFromString(stderr) ||
-      getLastJsonObjectFromString(stdout);
-    if (json?.error) {
-      const jsonError = formatNpmPnpmJsonError(json.error);
-      if (jsonError) {
-        return jsonError;
-      }
+export function getNpmPnpmError({
```
beeequeue: these should be moved into `PublishTool`

packages/cli/src/lib/npm.ts
Andarist: you are right, I pushed out the fix for this

packages/cli/src/lib/pnpm.test.ts
Andarist: it feels to be tested more precisely by the e2e already. I don't think it's particularly important here what kind of an internal error code gets returned in this situation. The user-observable behavior is what matters. Happy to discuss this further though if you feel strongly that we should keep this

packages/cli/src/lib/pnpm.ts
Andarist: Without this, we'd print `--json` when delegating interactively to the underlying package manager. See this diff:
<details>
<summary>git diff</summary>

```diff
diff --git a/packages/cli/src/commands/publish/__tests__/__snapshots__/e2e.test.ts.snap b/packages/cli/src/commands/publish/__tests__/__snapshots__/e2e.test.ts.snap
index ca9b4336..a03b1b22 100644
--- a/packages/cli/src/commands/publish/__tests__/__snapshots__/e2e.test.ts.snap
+++ b/packages/cli/src/commands/publish/__tests__/__snapshots__/e2e.test.ts.snap
@@ -36,27 +36,36 @@ exports[`Publish command e2e > 'npm 10' > retries interactively after an OTP aut
 │  0 packages are already published.
 │
 ◒  Publishing packages◇  pkg-a requires 2FA verification to publish...
-npm notice
-npm notice package: pkg-a@1.0.0
-npm notice Tarball Contents
-npm notice 24B index.js
-npm notice 104B package.json
-npm notice Tarball Details
-npm notice name: pkg-a
-npm notice version: 1.0.0
-npm notice filename: pkg-a-1.0.0.tgz
-npm notice package size: 234 B
-npm notice unpacked size: 128 B
-npm notice shasum: 843d50e8a5f1a01846a86f4e30b58913b9403183
-npm notice integrity: sha512-IJNHXvYGzsReN[...]0lxUHcQEkBaTw==
-npm notice total files: 2
-npm notice
 npm notice Publishing to [registry-url]/ with tag latest and public access
 This operation requires a one-time password.
 Enter OTP: 654321
 
-
-+ pkg-a@1.0.0
+{
+  "pkg-a": {
+    "id": "pkg-a@1.0.0",
+    "name": "pkg-a",
+    "version": "1.0.0",
+    "size": 234,
+    "unpackedSize": 128,
+    "shasum": "843d50e8a5f1a01846a86f4e30b58913b9403183",
+    "integrity": "sha512-IJNHXvYGzsReNsSn5GVqnUDDOeDoG+pni3Yroo3ZbaB7TXimXtjdKKtJbir1nmuQjfSCwbWv00lxUHcQEkBaTw==",
+    "filename": "pkg-a-1.0.0.tgz",
+    "files": [
+      {
+        "path": "index.js",
+        "size": 24,
+        "mode": 420
+      },
+      {
+        "path": "package.json",
+        "size": 104,
+        "mode": 420
+      }
+    ],
+    "entryCount": 2,
+    "bundled": []
+  }
+}
 │
 ◆  Successfully published:
 │  pkg-a@1.0.0
@@ -151,26 +160,36 @@ exports[`Publish command e2e > 'npm 11' > retries interactively after an OTP aut
 │  0 packages are already published.
 │
 ◒  Publishing packages◇  pkg-a requires 2FA verification to publish...
-npm notice
-npm notice package: pkg-a@1.0.0
-npm notice Tarball Contents
-npm notice 24B index.js
-npm notice 104B package.json
-npm notice Tarball Details
-npm notice name: pkg-a
-npm notice version: 1.0.0
-npm notice filename: pkg-a-1.0.0.tgz
-npm notice package size: 234 B
-npm notice unpacked size: 128 B
-npm notice shasum: 843d50e8a5f1a01846a86f4e30b58913b9403183
-npm notice integrity: sha512-IJNHXvYGzsReN[...]0lxUHcQEkBaTw==
-npm notice total files: 2
-npm notice
 npm notice Publishing to [registry-url]/ with tag latest and public access
 This operation requires a one-time password.
 Enter OTP: 654321
 
-+ pkg-a@1.0.0
+{
+  "pkg-a": {
+    "id": "pkg-a@1.0.0",
+    "name": "pkg-a",
+    "version": "1.0.0",
+    "size": 234,
+    "unpackedSize": 128,
+    "shasum": "843d50e8a5f1a01846a86f4e30b58913b9403183",
+    "integrity": "sha512-IJNHXvYGzsReNsSn5GVqnUDDOeDoG+pni3Yroo3ZbaB7TXimXtjdKKtJbir1nmuQjfSCwbWv00lxUHcQEkBaTw==",
+    "filename": "pkg-a-1.0.0.tgz",
+    "files": [
+      {
+        "path": "index.js",
+        "size": 24,
+        "mode": 420
+      },
+      {
+        "path": "package.json",
+        "size": 104,
+        "mode": 420
+      }
+    ],
+    "entryCount": 2,
+    "bundled": []
+  }
+}
 │
 ◆  Successfully published:
 │  pkg-a@1.0.0
@@ -267,26 +286,36 @@ exports[`Publish command e2e > 'npm 12' > retries interactively after an OTP aut
 │  0 packages are already published.
 │
 ◒  Publishing packages◇  pkg-a requires 2FA verification to publish...
-npm notice
-npm notice package: pkg-a@1.0.0
-npm notice Tarball Contents
-npm notice 24B index.js
-npm notice 104B package.json
-npm notice Tarball Details
-npm notice name: pkg-a
-npm notice version: 1.0.0
-npm notice filename: pkg-a-1.0.0.tgz
-npm notice package size: 234 B
-npm notice unpacked size: 128 B
-npm notice shasum: 843d50e8a5f1a01846a86f4e30b58913b9403183
-npm notice integrity: sha512-IJNHXvYGzsReN[...]0lxUHcQEkBaTw==
-npm notice total files: 2
-npm notice
 npm notice Publishing to [registry-url]/ with tag latest and public access
 This operation requires a one-time password.
 Enter OTP: 654321
 
-+ pkg-a@1.0.0
+{
+  "pkg-a": {
+    "id": "pkg-a@1.0.0",
+    "name": "pkg-a",
+    "version": "1.0.0",
+    "size": 234,
+    "unpackedSize": 128,
+    "shasum": "843d50e8a5f1a01846a86f4e30b58913b9403183",
+    "integrity": "sha512-IJNHXvYGzsReNsSn5GVqnUDDOeDoG+pni3Yroo3ZbaB7TXimXtjdKKtJbir1nmuQjfSCwbWv00lxUHcQEkBaTw==",
+    "filename": "pkg-a-1.0.0.tgz",
+    "files": [
+      {
+        "path": "index.js",
+        "size": 24,
+        "mode": 420
+      },
+      {
+        "path": "package.json",
+        "size": 104,
+        "mode": 420
+      }
+    ],
+    "entryCount": 2,
+    "bundled": []
+  }
+}
 │
 ◆  Successfully published:
 │  pkg-a@1.0.0
diff --git a/packages/cli/src/lib/npm.ts b/packages/cli/src/lib/npm.ts
index 208afa8b..5789b5de 100644
--- a/packages/cli/src/lib/npm.ts
+++ b/packages/cli/src/lib/npm.ts
@@ -259,7 +259,7 @@ export const publish: PublishTool["publish"] = async ({
     const cwd = pkg.dir;
 
     const args: string[] = ["--access", release.access, "--tag", release.tag];
-    if (!interactive) args.unshift("--json");
+    args.unshift("--json");
     if (otpCode) args.push("--otp", otpCode);
     if (tarballPath) {
       args.unshift(path.relative(cwd, tarballPath));

```

</details>

packages/cli/src/lib/pnpm.test.ts
beeequeue: i do think that we should have unit tests for each result the `PublishTool`s can return, it feels wrong to skip one/some of them

packages/cli/src/lib/npm.ts
```diff
@@ -0,0 +1,377 @@
...
+    {
+      nodePath: false,
+      nodeOptions: { cwd: pkg.dir },
+    },
+  );
+  if (exitCode !== 0) {
+    return { error: getNpmError({ stderr, stdout }) };
+  }
+
+  const json = getLastJsonObjectFromString(stdout);
```
beeequeue, line 250: do we still need `getLastJsonObjectFromString`? i didn't encounter it myself yet, and it doesn't feel very good

packages/cli/src/lib/npm.ts
```diff
@@ -0,0 +1,377 @@
...
+      registryOverrides.push(
+        `--registry=${packageJson.publishConfig.registry}`,
+      );
+    }
+  } else if (packageJson.publishConfig?.registry) {
+    registryOverrides.push(`--registry=${packageJson.publishConfig.registry}`);
+  }
+
+  return registryOverrides;
+}
```
beeequeue, line 170: i think we should just let the CLI we're calling handle the resolution by themselves. re-building their config feels like a sisyphean task, especially when considering the other PMs (having to go through `pnpm-workspace.yml`, `~/.config/pnpm/config.yml`, `~/.config/pnpm/auth.ini`, etc.).

this has not been implemented for pnpm or yarn as far as i can see, so do we have to do it for npm?

if anything it should be left to a separate package that _just_ takes care of resolving registries and tokens for packages.

packages/cli/src/commands/__tests__/e2e-utils.ts
```diff
@@ -183,9 +183,25 @@ async function resolvePackageManagerSpec(
...
+) {
+  const tempDir = path.join(cwd, ".tmp");
+  await fs.mkdir(tempDir, { recursive: true });
   return {
+    // Exercise normal user behavior regardless of where the tests run.
+    // CI-specific tests can opt in through env when needed.
+    CI: undefined,
+    GITHUB_ACTIONS: undefined,
     ...env,
+    // pnpm packs the package into TMPDIR and runs npm from there. If that is
```
beeequeue: all pnpm versions or just 10?

packages/cli/src/commands/publish/index.ts
```diff
@@ -95,121 +168,244 @@ To resolve this exit the pre mode by running ${c.cyan("changeset pre exit")}.
...
+  );
+  const gitTagsToCreate: TagReleaseEntry[] = [];
 
-    const unpublishedPackages = chunk.filter(
-      (release) => release.kind === "publish",
-    );
-    const untaggedPrivatePackageReleases = chunk.filter(
-      (release) => release.kind === "tag-only",
+  let otpCode = publishTool.getOtpCode(options?.otp);
+  let sequential = process.stdin.isTTY;
```
beeequeue:
```suggestion
  let otpCode = publishTool.getOtpCode(options?.otp);
  let sequential = process.stdin.isTTY || otpCode != null;
```

if `otpCode` is provided, we should be able to use it to bulk publish immediately

packages/cli/src/commands/__tests__/e2e-utils.ts
Andarist: 10, only that one (from the ones we support) delegates to npm

packages/cli/src/lib/npm.ts
Andarist: I added this test with the specific intention to retest this: https://github.com/changesets/changesets/pull/2155/changes/071f06a441ae6cb91c881ba70913a5114bb706bc but I haven't checked yet if it's still needed or not.

One thing we can simplify is that we should only be looking at `stdout` - we don't need to handle `stderr`

packages/cli/src/commands/publish/index.ts
Andarist: good point

packages/cli/src/lib/npm.ts
Andarist: I have confirmed `getLastJsonObjectFromString` is still needed

packages/cli/src/lib/pnpm.test.ts
Andarist: brought them back

packages/cli/src/lib/npm.ts
Andarist: Note that this is not re-building any of those configs you mentioned. This only handles `packageJson.publishConfig.registry`.

I agree this is totally iffy and the role of this config option is underdocumented. Changing this behavior would be a breaking change now and whenever possible I was choosing to maintain backward compatibility.

But even ignoring backward compatibility with v2, I use those facts to determine this behavior in https://github.com/changesets/changesets/pull/2064 :
- it makes sense to look for the package in the registry used when publishing
- npm supports `packageJson.publishConfig.registry` as a strong override at publish time but that's not used at all when calling `npm info`. That's what led to this logic being implemented
- I'd support the same in pnpm because it also supports `packageJson.publishConfig.registry` but its own `pnpm publish --recursive` doesn't respect it (and they are doing similar preflight `info` calls to us there). So I decided to match their own behavior - but that behavior is questionable and could even be seen as a bug in their implementation.
- yarn doesn't support `packageJson.publishConfig.registry` **but** they have a clear model of supporting separate **read** and **write** registries. So for a single scope, you can set those to 2 different registries so their `info` and `publish` can naturally use different values.

packages/cli/src/lib/pnpm.ts
Andarist: I think this was a good suggestion and I pushed a change to that effect

packages/cli/src/lib/common.ts
```diff
@@ -0,0 +1,32 @@
...
+  return message.includes("cannot publish over the previously published");
+}
+
+export const isPublishSuccessful = (
+  result: PublishResult,
+): result is PublishResultSuccess => result.result === "published";
+
+export const isPublishFailure = (
+  result: PublishResult,
+): result is PublishResultFailed => result.result.startsWith("failed");
```
Andarist, line 32: I feel we should improve this. `2fa?: boolean` would be OK for the 2fa thing - but I'm not sure how to best denote the already published error. Perhaps we should just use `published: boolean; error?: { message?: string; code?: string }, someOtherField?: "2fa" | "republish_attempt"`.

OTOH, perhaps we should just not be so lenient with those already published cases and we should actually error and stop when that happens? I imagine some people could request `--tolerate-republish` then and then we'd still have to model this situation somehow.

beeequeue: I think we can keep it as is and rethink it for v4

packages/cli/src/lib/npm.ts
Andarist: I'm porting some of the comments lost in the refactor and I noticed now that npm actually uses `packageJson.publishConfig.registry` in its preflight info queries during publish. But bare `npm info` doesn't use that - so that's what we are matching here.

packages/cli/src/lib/common.ts
Andarist: Ye, this is just a nit - this isn't part of the public API anyway.

packages/cli/src/lib/npm.ts
beeequeue: > npm supports `packageJson.publishConfig.registry` as a strong override at publish time but that's not used at all when calling `npm info`.

in that case my comment is incorrect and irrelevant 👌😄

packages/cli/src/lib/npm.ts
```diff
@@ -0,0 +1,409 @@
...
+// we have this so that we can do try a publish again after a publish without
+// the call being wrapped in the npm request limit and causing the publishes to potentially never run
+export const publish: PublishTool["publish"] = async ({
+  pkg,
+  release,
+  tarballPath,
+  interactive,
+  otpCode,
+}: PublishOptions): Promise<PublishResult> =>
+  npmPublishQueue.add(async () => {
```
bluwy: I think we should move the queue handling to the callee in the future, but this is fine for now.

bluwy (approved):

packages/cli/src/commands/git-tag/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+  tagged: TagReleaseEntry[];
+  existing: TagReleaseEntry[];
+};
+
+export async function createGitTags(
+  opts: CreateGitTagsOptions,
+): Promise<CreateGitTagsResult> {
+  const plan =
+    opts.plan ??
+    opts.packages.packages.map((pkg) => ({
```
Andarist: This looks off to me. I see `shouldSkipPackage` filtering is done later in `splitByTagStatus` - but I feel it should be done immediately here

packages/cli/src/commands/publish/index.ts
```diff
@@ -95,121 +167,252 @@ To resolve this exit the pre mode by running ${c.cyan("changeset pre exit")}.
...
-    }
-  } else {
-    const tag = `v${packageReleases[0].version}`;
-    await git.tag(tag, cwd);
-    reporter?.write({
-      type: "git-tag",
-      tag,
-      packageName: packageReleases[0].name,
+  // finally, create git tags as plan instructs
+  const tagsToCreate = uniqBy(gitTagsToCreate, (r) =>
```
Andarist: Hm, I'm not sure what do we get from this `uniqBy` call here. Packages shouldn't be processed multiple times so we shouldn't need any deduplication mechanism like this

beeequeue: in my version we added a `git-tag` action after every successful publish, so we'd end up with duplicates if the plan contains both a tag and publish event for a package

packages/cli/src/commands/git-tag/utils.ts
beeequeue: optimally i think we should never end up with a skipped package in the publish flow, the plan should not include them

packages/cli/src/commands/publish/index.ts
Andarist: > so we'd end up with duplicates if the plan contains both a tag and publish event for a package

Generally speaking, the plan distinguishes today tag-only releases from publish ones - so we shouldnt end up with this problem.

beeequeue: I meant that in my version of the publish flow we manually added tag events to an array so we knew what to tag after successful publishing, I haven't checked if it's still needed in the new code

Andarist (approved):

PR #2159, opened 2026-07-13
Fixed handling of the npm 12 info output

ingvaldlorentzen:
In [npm v12 `npm view`/`npm info` always return an array](https://docs.npmjs.com/cli/v12/using-npm/changelog#:~:text=npm%20view%20%2D%2Djson%20now%20always%20returns%20an%20array).

This causes changesets to incorrectly interpret the packages as not published, and attempt to publish them, causing errors from the registry.
e.g.
image

This PR fixes this by detecting if the output is an array, and grabbing the last element if it is. This should keep compatibility with v11 and v12 behaviour. 

The "error" state is also changed, so it now detects if empty string and E404.

### DIFF
--- a/packages/cli/src/commands/publish/__tests__/publishPackages.test.ts
+++ b/packages/cli/src/commands/publish/__tests__/publishPackages.test.ts
@@ -109,6 +109,109 @@ describe("publishPackages", () => {
     ).toBe(false);
   });
 
+  it("detects already-published versions when npm 12 wraps info output in an array", async () => {
+    const cwd = await testdir({
+      "package.json": JSON.stringify({
+        private: true,
+        workspaces: ["packages/*"],
+      }),
+      "packages/pkg-a/package.json": JSON.stringify({
+        name: "pkg-a",
+        version: "1.0.0",
+      }),
+    });
+
+    mockSpawnImplementation((cmd, args) => {
+      if (cmd === "npm" && args?.[0] === "info") {
+        return spawnResult(
+          JSON.stringify([
+            {
+              name: "pkg-a",
+              version: "1.0.0",
+              versions: ["1.0.0"],
+            },
+          ])
+        );
+      }
+      return spawnResult("", 1);
+    });
+
+    const result = await publishPackages({
+      packages: (await getPackages(cwd)).packages,
+      access: "public",
+      preState: undefined,
+    });
+
+    expect(result).toEqual([]);
+    expect(
+      mockSpawn.mock.calls.some(
+        ([cmd, args]) => cmd === "npm" && args?.[0] === "publish"
+      )
+    ).toBe(false);
+  });
+
+  it("retries with an exact version when npm 12 reports no match for the bare query", async () => {
+    const cwd = await testdir({
+      "package.json": JSON.stringify({
+        private: true,
+        workspaces: ["packages/*"],
+      }),
+      "packages/pkg-a/package.json": JSON.stringify({
+        name: "pkg-a",
+        version: "1.0.0",
+      }),
+    });
+
+    mockSpawnImplementation((cmd, args) => {
+      if (cmd === "npm" && args?.[0] === "info") {
+        if (args?.[1] === "pkg-a") {
+          // npm 12 emits an E404 error instead of empty stdout when nothing
+          // matches `latest` (e.g. GitHub Packages without a `latest` dist-tag)
+          return spawnResult(
+            JSON.stringify({
+              error: {
+                code: "E404",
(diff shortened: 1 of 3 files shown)


Andarist:
This is already being fixed in https://github.com/changesets/changesets/pull/2097 but that's in the works and targets the upcoming (soon!) v3 release. 

Given this is a pretty big break for npm users, we can definitely accept a PR targeting v2 - but that should be opened against the [maintenance/v2](https://github.com/changesets/changesets/tree/maintenance/v2) branch.

packages/cli/src/commands/publish/npm-utils.ts
```diff
@@ -86,6 +86,30 @@ export function getPublishTool({ type }: Packages["tool"]): PublishTool {
...
 //   published with preState.tag rather than "latest".
+
+// npm 12 always wraps successful `npm info --json` output in an array.
+// Unwrap it so downstream consumers keep seeing a single manifest, if multiple versions matched, the last entry is
+// the highest one
+function normalizeInfoJson(parsed: any) {
+  if (!Array.isArray(parsed)) {
+    return parsed;
+  }
+  return parsed.at(-1) ?? { error: { code: "E404" } };
```
Andarist: I expect the `npm publish` to require the same kind of normalization. But that has to be rechecked

ingvaldlorentzen: I don' think so. I'm not very familiar with the `changesets` codebase, first time I'm poking around now, so I may very well be wrong and missing something.  
But it seems like the `npm publish` changes in v12 are more on making `npm pack` match `npm publish`. 

`npm publish --dry-run --json` -- npm 11:
```
{
  "@verify-scratch/pkg": {
    "id": "@verify-scratch/pkg@1.0.0",
```
`npm publish --dry-run --json` -- npm 12
```
{
  "@verify-scratch/pkg": {
    "id": "@verify-scratch/pkg@1.0.0",
```

`npm pack --dry-run --json` -- npm 11:

```
[
  {
    "id": "@verify-scratch/pkg@1.0.0",
...
```
`npm pack --dry-run --json` -- npm 12
```
{
  "@verify-scratch/pkg": {
    "id": "@verify-scratch/pkg@1.0.0",
...
```



If there is something I'm missing, and you can point me in the right direction, I'll gladly try to implement a fix, but I don't see any problems right now.

ingvaldlorentzen:
> This is already being fixed in #2097 but that's in the works and targets the upcoming (soon!) v3 release.

Awesome! Looking forward to check out v3!

> Given this is a pretty big break for npm users, we can definitely accept a PR targeting v2 - but that should be opened against the [maintenance/v2](https://github.com/changesets/changesets/tree/maintenance/v2) branch.

I've retargeted this PR to be based on and point towards `maintenance/v2` now!

packages/cli/src/commands/publish/npm-utils.ts
beeequeue: we have checked the publish output is still the same!

https://github.com/beeequeue/pm-publish-outputs/blob/main/src/__snapshots__/test.test.ts.snap#L169

packages/cli/src/commands/publish/__tests__/publishPackages.test.ts
```diff
@@ -109,6 +109,109 @@ describe("publishPackages", () => {
...
+        name: "pkg-a",
+        version: "1.0.0",
+      }),
+    });
+
+    mockSpawnImplementation((cmd, args) => {
+      if (cmd === "npm" && args?.[0] === "info") {
+        if (args?.[1] === "pkg-a") {
+          // npm 12 emits an E404 error instead of empty stdout when nothing
+          // matches `latest` (e.g. GitHub Packages without a `latest` dist-tag)
```
Andarist: tbh, I can't reproduce this - and the npm's source code seems to back that up:
https://github.com/npm/cli/blob/7b1f6c173d17b3bf30e45426f6df39473c6a1163/lib/commands/view.js#L189-L196

note: calling `npm info pkg-a` is like `npm info pkg-a@latest`, so even if that returns no matches, it still shouldn't return E404

Am I missing something? How have you observed this?

ingvaldlorentzen: No, you're right, I goofed.  
Had been testing against some actual versions when debuggin, and made an incorrect assumption, sorry about that.  
Just pushed a commit that removes this code and the related test, makes the PR quite a bit simpler, thanks for the feedback!

Andarist (approved):

PR #2160, opened 2026-07-14
add or update `files` in every package

beeequeue:
good practice to keep unwanted files out of the tarballs

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,9 +21,7 @@
   "files": [
     "default-files",
     "dist",
-    "bin.js",
-    "changelog",
-    "commit"
+    "bin.js"
   ],
   "type": "module",
   "exports": {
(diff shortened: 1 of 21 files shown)


packages/cli/package.json
```diff
@@ -19,11 +19,8 @@
     "changeset": "bin.js"
   },
   "files": [
-    "default-files",
```
Andarist, line 22: this one probably should be kept

beeequeue: The default files are always included in every package (manifest, readme) no matter what the files specify

Andarist: yes, but those `./default-files` are not `CHANGELOG.md`, `README.md`, etc. There is a literal `./default-files` directory in this package :p

beeequeue: right... i saw and recognized this at first then removed it on a second pass 🥲

Andarist (approved):

PR #2169, opened 2026-07-18
Run CI tests on Windows

Andarist:


packages/git/src/index.ts
```diff
@@ -188,15 +188,15 @@ export async function deepenCloneBy({ by, cwd }: { by: number; cwd: string }) {
 async function getRepoRoot({ cwd }: { cwd: string }) {
   const { stdout, exitCode, stderr } = await exec(
     "git",
-    ["rev-parse", "--show-toplevel"],
+    ["rev-parse", "--show-cdup"],
```
beeequeue, line 191: this looks unrelated?

packages/cli/src/commands/pack/__tests__/e2e.test.ts
```diff
@@ -12,7 +12,10 @@ import {
 
 function sanitizePackLog(message: unknown) {
   return stripVTControlCharacters(String(message))
+    .replaceAll("\r\n", "\n")
+    .replaceAll("\r", "\n")
```
bluwy: Does `\r\n` appear in CLI outputs on Windows? Or is it more of a defensive addition?

packages/cli/src/commands/publish/__tests__/e2e.test.ts
```diff
@@ -222,7 +222,15 @@ function createWebServer(handler: (request: Request) => Promise<Response>) {
 
 function sanitizePublishLog(message: unknown, registryUrl: string) {
   return stripVTControlCharacters(String(message))
+    .replaceAll("\r\n", "\n")
+    .replaceAll("\r", "\n")
```
bluwy: Here too

packages/release-utils/src/run.test.ts
```diff
@@ -20,6 +20,8 @@ const writeChangesets = (changesets: Changeset[], cwd: string) => {
   );
 };
 
+const normalizeNewlines = (value: string) => value.replaceAll("\r\n", "\n");
```
bluwy: Is this needed with the `.gitattributes` forcing to `lf` now?

bluwy (commented):
Some questions about the `\r\n` handling. I'd prefer if we only handle that if it could happen in practice otherwise it's unclear in the future if we consistently need it for any tests that has new lines.

beeequeue:
seeing so many "workaround"-style changes doesn't feel very good, but if it's what we have to do i guess we have to...

packages/git/src/index.ts
Andarist: this is related to this changeset: https://github.com/changesets/changesets/pull/2169/changes#diff-92e25cabce9250b89814c2afa36b1e22b0c838f55fb6f4e8f716241c8c30dcb8

packages/cli/src/commands/pack/__tests__/e2e.test.ts
Andarist: It's not purely defensive. `result.stdout` is the complete node-pty/ConPTY stream, including package-manager subprocesses, not only Changesets-authored output. And the extra standalone `\r` handles terminal progress redraws. I'll add comments about this

packages/release-utils/src/run.test.ts
Andarist: It was needed but now I added this:
> ".gitattributes": "* text=auto eol=lf\n",

to our `gitdir` util and that fixes it without us having to normalize in tests

bluwy (approved):

PR #2171, opened 2026-07-19
enable node compile cache

beeequeue:
https://nodejs-api-docs-tooling.vercel.app/module.html#module-compile-cache

### DIFF
--- a/packages/cli/bin.js
+++ b/packages/cli/bin.js
@@ -1,2 +1,9 @@
 #!/usr/bin/env node
-import "@changesets/cli";
+
+if (globalThis.process?.getBuiltinModule) {
+  const { enableCompileCache } =
+    globalThis.process.getBuiltinModule("node:module");
+  enableCompileCache();
+}
+
+await import("@changesets/cli");
--- a/.changeset/ready-cloths-hope.md
+++ b/.changeset/ready-cloths-hope.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/cli/bin.js
```diff
@@ -1,2 +1,9 @@
 #!/usr/bin/env node
-import "@changesets/cli";
+
+if (globalThis.process?.getBuiltinModule) {
+  const { enableCompileCache } =
+    globalThis.process.getBuiltinModule("node:module");
+  enableCompileCache();
```
Andarist, line 6: q: could this be enabled in `./index.ts`? or does this have to be enabled *before* loading the code? I just wonder cause it feels that perhaps at some point we should get rid of `./bin.js` but maybe this compilation makes it necessary?

Andarist (approved):

packages/cli/bin.js
beeequeue: it has to be enabled before importing the code that should be cached, it won't be cached if it's already loaded.

you can see this in [typescript](https://github.com/microsoft/TypeScript/blob/637d5746b70257028fb95aad32ddec6b26ab0a14/Herebyfile.mjs#L324-L342) as well

you can also enable it permanently for everything with `NODE_COMPILE_CACHE=/some/path/to/cache` and `NODE_COMPILE_CACHE_PORTABLE=1`

Andarist: ok, cool - then I feel less bad now about the `./bin.js`+`./index.js` split :p there is an actual reason for it now 👍

PR #2176, opened 2026-07-20
v3 migration guide

beeequeue:
i also added a script+ci job that checks how many of the changesets have been added to the migration guide, to help ensure we don't miss anything

preview deployment: https://deploy-preview-2176--changesets.netlify.app/guide/migration


beeequeue:
these changesets are currently `major`, but i'm not sure they have to be:

- `ready-rockets-boil`: removing `initialVersions` from `pre.json`
  - this is not breaking unless we crash if it's defined, but we say that it was unused so i doubt it.
- `some-papayas-bet`, `some-papayas-gamble`: changing the default formatting functions to sync instead of async.
  - going from async->sync is only breaking if you use any functions on promises, `await` still works.
- `wise-mirrors-fry`: manifests are formatted with `format` instead of Prettier
  - do we still do this now that we edit the files without modifying the formatting?

Andarist (approved):

bluwy (approved):

PR #2177, opened 2026-07-20
Avoid writing an `undefined` version when updating dependencies in unversioned private packages

Andarist:


bluwy:
Can we also update the types that newVersion and oldVersion can be nullable too?

Andarist:
@bluwy sure, fixed that

bluwy (approved):
Thanks for fixing! This looks good to me

PR #2178, opened 2026-07-20
test parallel workflow steps

beeequeue:


.github/workflows/ci.yml
```diff
@@ -138,17 +120,19 @@ jobs:
...
+            run: pnpm format
+          - name: Typecheck
+            run: pnpm types:check
 
   ci-ok:
     name: CI OK
     runs-on: ubuntu-latest
     if: always()
-    needs: [lint-workflows, build, test, typecheck, lint]
+    needs: [build, lint, lint-workflows]
```
Andarist, line 135: at this point, perhaps we shouldn't even be building in a separate job

beeequeue: Sadly we need to build in s separate step so we can use the matrix for testing

PR #2181, opened 2026-07-22
Add Bun support

Andarist:


packages/cli/src/commands/__tests__/e2e-utils.ts
```diff
@@ -228,9 +228,14 @@ export async function getPmBinPath(signal: AbortSignal, bins: PmBins) {
       root.path,
       isWindows ? `${command}.cmd` : command,
     );
-    const shim = isWindows
```
beeequeue, line 231: idea: we could depend on nushell for the end to end tests instead of relying on whatever shell is installed on the system

i've been using nushell as my login shell for the last 6 years and can highly recommend it for cross-platform use

packages/cli/src/lib/bun.ts
```diff
@@ -0,0 +1,205 @@
...
+    };
+    publishProcess.process?.stdout?.on("data", (chunk: Buffer | string) => {
+      observedStdout += chunk.toString();
+      detectWebAuth();
+    });
+    publishProcess.process?.stderr?.on("data", (chunk: Buffer | string) => {
+      observedStderr += chunk.toString();
+      detectWebAuth();
+    });
+  }
```
beeequeue, line 166: what is happening here? is this to handle a 2fa flow that they do not do well?

packages/cli/src/lib/bun.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+function getBunError(stdout: string, stderr: string): BunCommandError {
+  const output = stderr || stdout;
+  const statusCode = output.match(/(?:^|\n)(\d{3})(?: [^:\n]+)?:/)?.[1];
+  return {
+    code: statusCode ? `E${statusCode}` : undefined,
+    message: output.trim() || undefined,
+  };
+}
```
beeequeue, line 28: any chance to add these to the publish snapshot repo? it would make it easier to understand the parsing

PR #2183, opened 2026-07-23
Staged publishing

Andarist:


docs/command-line-options.md
```diff
@@ -5,7 +5,8 @@ The command line for changesets is the main way of interacting with it. There ar
 - init
 - add [--empty] [--open] [--since <ref>] [--message <text>]
 - version [--ignore, --snapshot]
-- publish [--otp=code, --tag]
+- publish [--otp=code, --tag, --stage, --no-stage]
```
Andarist, line 8:
```suggestion
- publish [--otp=code, --tag, --stage]
```

packages/cli/src/commands/publish/__tests__/e2e.test.ts
```diff
@@ -915,6 +915,107 @@ function createPmContext(
 
 describe("Publish command e2e", { tags: ["slow"] }, () => {
   describe.each(pmCases)("$name", (pm) => {
+    if (pm.name === "npm 12" || pm.name === "pnpm 11" || pm.name === "yarn 4") {
+      it("stages and approves an existing package version", async ({
+        signal,
+      }) => {
+        await using stack = new AbortableAsyncDisposableStack(signal);
+        const { pmBinPath } = stack.use(await getPmBinPath(signal, pm.bins));
```
beeequeue, line 923: i know we use it in the other tests as well, but why do we need AbortableAsyncDisposableStack? is it for the aborting part?

Andarist: Ye, it’s just so we can abort when the test times out

packages/cli/src/commands/publish-plan/__tests__/getPublishPlan.test.ts
```diff
@@ -69,6 +69,7 @@ describe("getPublishPlan", () => {
           version: "1.0.0",
           access: "restricted",
           tag: "latest",
+          isNew: true,
```
beeequeue, line 72: most of these changes seem like snapshot changes but are actually normal equal checks (also, they're not using `toStrictEqual`?)
they should probably be changed to `toMatchInlineSnapshot()`

packages/cli/src/commands/publish/__tests__/e2e.test.ts
beeequeue: can we make a vitest plugin that adds one to the test context, like `signal`?

packages/cli/src/commands/publish/__tests__/e2e.test.ts
```diff
@@ -915,6 +915,107 @@ function createPmContext(
...
+                  new Request(new URL("/pkg-a", request.url), {
+                    method: "PUT",
+                    headers: stagedRequest.headers,
+                    body: stagedRequest.body,
+                  }),
+                );
+              }
+              return undefined;
+            },
+          }),
```
beeequeue, line 966: i dont really like this api, should we instead use a real rest framework like polka or h3?

packages/cli/src/commands/publish/__tests__/index.test.ts
```diff
@@ -725,6 +729,196 @@ describe("Publish command", () => {
     expect(git.tag).not.toHaveBeenCalled();
   });
 
+  it("reports staged packages in topological order without tagging by default", async () => {
```
beeequeue, line 732: this smells like the code isn't split up properly, i.e. it should be testable with a small unit test and not require a full integration test

packages/cli/src/commands/publish/index.ts
```diff
@@ -392,6 +463,26 @@ Some packages failed to publish:
...
     );
+    if (stagedPublishing && successfulNpmPublishes.length > 0) {
+      const stageIds = successfulNpmPublishes
+        .filter(
+          (result): result is Extract<PublishResult, { result: "staged" }> =>
+            result.result === "staged",
+        )
+        .map((result) => result.stageId);
+      log.warn(
+        `Reject the successfully staged packages, then retry:\nchangeset stage reject ${stageIds.join(" ")}`,
```
beeequeue, line 474:
```suggestion
        `Reject the staged packages, then try publishing again:\nchangeset stage reject ${stageIds.join(" ")}`,
```

packages/cli/src/commands/publish/index.ts
```diff
@@ -392,6 +463,26 @@ Some packages failed to publish:
...
+    }
+  } else if (stagedPublishing && successfulNpmPublishes.length > 0) {
+    const stageIds = successfulNpmPublishes
+      .filter(
+        (result): result is Extract<PublishResult, { result: "staged" }> =>
+          result.result === "staged",
+      )
+      .map((result) => result.stageId);
+    log.info(`Approve the staged packages in this order:
+changeset stage approve ${stageIds.join(" ")}`);
```
beeequeue, line 485: this wording implies the user has to order packages manually, but then it presents a command to execute that should be ordered correctly

beeequeue (changes requested):
wip review

PR #2185, opened 2026-07-25
Cleanup some existing changeset files

Andarist:


.changeset/tender-mammals-win.md
bluwy, line 1: This can be kept, no?

bluwy: Nevermind this is covered in https://github.com/changesets/changesets/blob/main/.changeset/social-snails-battle.md

bluwy (approved):

PR #2186, opened 2026-07-25
No longer version private packages by default

Andarist:


packages/config/src/config.ts
```diff
@@ -182,16 +182,14 @@ export function normalizeWrittenConfig({
     config.ignore = globMatch(packageNames, writtenConfig.ignore);
   }
 
-  // TODO consider enabling this by default in the next major version
-  // might be more context in here: https://github.com/changesets/changesets/pull/662
   if (typeof writtenConfig.privatePackages !== "object") {
     config.privatePackages = {
-      version: writtenConfig.privatePackages ?? true,
+      version: writtenConfig.privatePackages ?? false,
```
bluwy, line 187:
```suggestion
      version: false,
```

I think we can make this false directly as `privatePackages` is typed as `object | false`

docs/versioning-apps.md
```diff
@@ -4,7 +4,7 @@ Changesets can also be used to manage application versions or non-npm packages (
 
 The only requirement is that the project has a package.json file to manage the versions and dependencies within the repo.
 
-To enable this feature set `privatePackages` to `{ version: true, tag: true }` in your `.changesets/config.json` file. By default changesets will only update the changelog and version (ie `{ version: true, tag: false }`).
+To enable this feature set `privatePackages` to `{ version: true, tag: true }` in your `.changesets/config.json` file. By default Changesets will not version or tag private packages (that is, `{ version: false, tag: false }`).
```
bluwy, line 7: FYI I think it's ok to not update the `docs/` too since we're moving to the site (I've added a warning at the top of these files in the docs branch). But we'll need to update these for the site later for sure.

bluwy (approved):
One small code nit. Thanks for handling this on behalf of my laziness 😄 

Also, maybe we should also consider supporting `privatePackages: true`

docs/versioning-apps.md
beeequeue: maybe we should just merge the next-docs branch at this point, to make it easier to edit?

packages/config/src/config.ts
Andarist: I made it to accept `privatePackages: true` now so now this has to stay in its current form

docs/config-file-options.md
```diff
@@ -227,9 +227,9 @@ Default value: `true`
 }
 ```
 
-## `privatePackages` (object or false)
+## `privatePackages` (object or boolean)
 
-This option is for setting how private packages should be handled. By default, Changesets will not version or tag private packages. You can configure this option to opt into either behavior.
+This option is for setting how private packages should be handled. By default, Changesets will not version or tag private packages. Set it to `true` to version private packages without tagging them, or use the object form to configure each behavior separately.
```
beeequeue: i think most people will assume `true` means setting both the children to true, just like how `false` works.

Andarist: Yep, i fixed that already

beeequeue (approved):

docs/versioning-apps.md
bluwy: There aren't much big changes needed for docs now, so yeah I think we can consider merging now

bluwy (approved):

PR #2190, opened 2026-07-27
Move versioned prerelease changesets to `.changeset/pre/` folder

### LINKED ISSUE #1688: Exiting pre mode fails with 400+ changesets

bluwy:
_This is just an experiment/idea I had. No need to push this for v3 if we don't feel like it._

---

From added changeset:

> Move versioned prerelease changesets to `.changeset/pre/` folder instead of accumulating in the root and tracking the versioned changeset ids in the `.changeset/pre.json` file. Existing `pre.json` will auto-migrate to this new structure on the next run of `changeset version` or when calling `changeset status`.
>
> This change allows easier management of versioned prerelease changesets (for the final stable release) and current queued changesets (for the next prerelease). Changesets in `.changeset/pre/` can be edited or deleted depending if it's still relevant for the final stable release of a package. There's no need to synchronize the changeset ids in `pre.json` if certain changesets are deleted.

---

I think this should close https://github.com/changesets/changesets/issues/1688. While not directly addressing its issue, this PR should make it easy to resolve the problem by going through all the versioned prerelease changesets and only keep the ones that matter. If they don't care about a final changelog, they can simply delete all the changesets in `.changeset/pre/`


packages/git/src/index.ts
```diff
@@ -85,6 +85,7 @@ export async function getCommitsThatAddFiles(
             [
               "log",
               "--diff-filter=A",
+              "--follow",
```
bluwy, line 88: This has its own changeset as it affects outside of this PR. If the changeset file was renamed, previously we would return the commit that renamed it, `--follow` fixes this by checking the parent for possible renames and then looking further for the actual commit that added it. While this is technically more work done for git, I noticed for about 93 changesets, it only slows down from 300ms to 350ms, so not horribly slow, but it still depends on the state of the repo and parent commit

beeequeue:
i really like this idea!

i have a growing concern that running these kinds of migrations when needed can be surprising to users, and having a `migrate` command that does it all in one go would be more fitting

looking at the [current migration guide](https://deploy-preview-2176--changesets.netlify.app/guide/migrating-to-v3), there are a few things we could put in it (e.g. `prettier: true` -> `format: "prettier"`)

bluwy:
Yeah a migrate command would be nice to have 👍

packages/apply-release-plan/src/index.ts
```diff
@@ -189,40 +182,55 @@ export async function applyReleasePlan(
...
+            touchedFiles.push(changesetPath);
           }
         }
-      }),
-    );
-  }
+      }
+    }),
+  );
+  // }
```
Andarist:
```suggestion
```

packages/assemble-release-plan/src/index.ts
```diff
@@ -274,10 +274,7 @@ function getRelevantChangesets(
   }
 
   if (preState && preState.mode !== "exit") {
-    const usedChangesetIds = new Set(preState.changesets);
-    return changesets.filter(
-      (changeset) => !usedChangesetIds.has(changeset.id),
-    );
+    return changesets.filter((changeset) => !changeset.id.startsWith("pre/"));
```
Andarist, line 277: gotta make sure this `pre/` prefix gets normalized on Windows

packages/git/src/index.ts
Andarist: could we add a test case for this?

Andarist (approved):
I like the idea behind this - I had very similar thoughts about this when cleaning up those old changesets in the other PR

packages/assemble-release-plan/src/index.ts
bluwy: Yeah right now `readChangesets` will ensure this starts with `pre/` as it reads the files in the directory. I think it's easier to handle the a forward-slash always as the changeset "id".

packages/git/src/index.ts
bluwy: Done 👍

beeequeue:
while this change helps with #1688 i would prefer to close the issue by fixing chunking so the requests don't fail.

bluwy:
We already chunk right? With the maxBatchSize config for dataloader before

beeequeue:
oh yeah, then we should verify the fix somehow and close the issue...

beeequeue (approved):

PR #2198, opened 2026-07-30
Clarify ignore publish caveats

### LINKED ISSUE #1534: Unclear documentation
## Affected Packages
Docs

4mthxmas20:
## Summary
- clarify that `--ignore` also fails when an ignored package needs dependency updates during publish
- mirror that explanation in the legacy docs and the current site CLI guide

Fixes #1534

## Testing
- `pnpm format`
- `pnpm build`
- `pnpm site:build`
- `git diff --check`

Note: `pnpm lint` currently fails on existing unrelated TypeScript lint errors in packages outside this docs-only change.


bluwy (approved):
We no longer need to keep the `docs/` up-to-date, but this is also fine for now

PR #2214, opened 2026-08-04
Slightly speed up Windows CI with `runner.temp`

bluwy:


bluwy:
Seems like it is kinda faster, around 15-30s faster on Windows

.github/workflows/ci.yml
```diff
@@ -88,10 +88,18 @@ jobs:
         run: git config --global user.email "you@example.com" && git config --global user.name "Your Name"
 
       - name: Vitest tests
-        if: matrix.node_version != 24 || matrix.os != 'ubuntu-latest'
+        if: matrix.os != 'windows-latest' && !(matrix.node_version == 24 && matrix.os == 'ubuntu-latest')
         run: pnpm test
 
-      # Run coverage only once on node 24 to avoid duplicate codecov uploads
+      # For Windows, run fixtures in the runner tempdir instead of os tempdir so it runs faster
```
Andarist: why is this faster? how you have learned about this? 😅

.github/workflows/ci.yml
```diff
@@ -88,10 +88,18 @@ jobs:
...
+      - name: Vitest tests on Windows
+        if: matrix.os == 'windows-latest'
+        run: pnpm test
+        env:
+          TEMP: ${{ runner.temp }}
+          TMP: ${{ runner.temp }}
```
Andarist, line 101: couldn't we assign those conditionally instead of creating a second conditional step?

.github/workflows/ci.yml
bluwy: I wrote a comment about this at https://github.com/publint/publint/blob/d228c3c1580014bc621b48dda132cd00360e9e45/packages/publint/tests/playground.test.js#L293-L296 before and recently just read about it again 😄

.github/workflows/ci.yml
bluwy: We could maybe but I didnt figure the syntax for it

Andarist: Perhaps this would work?
```yml
env:
  TEMP: ${{ matrix.os == 'windows-latest' && runner.temp || env.TEMP }}
  TMP:  ${{ matrix.os == 'windows-latest' && runner.temp || env.TMP }}
```

.github/workflows/ci.yml
Andarist: It would be good to include a verbatim reference here to those lines in publint ;p

.github/workflows/ci.yml
bluwy: That doesn't work. I tried some ways like setting `""` or `null` as fallback, but those would also force to an empty value when `$TEMP` should be its original value.

Alternatively we can always set this env for the other OSes, but I think the most robust/simpler way is this current step separation.

Andarist (approved):

PR #2216, opened 2026-08-05
Remove stray console.log

bluwy:
Oops

Do we need a changeset? Likely we can bundle this together in the next stable release so we don't have to delete it from `.changeset/pre` again


beeequeue (approved):

Andarist:
> Do we need a changeset?

it's not important enough, let's skip it

PR #2217, opened 2026-08-05
Remove old atlaskit changelogs

bluwy:
I was trying to link to the anticipated `3.0.0` section in the changelog, but it's currently shadowed by the `3.0.0` version from `@atlaskit/build-releases` section. I think we can remove it since it's very old now with broken links.


beeequeue (approved):

beeequeue:
we could probably remove some pre-v2 changelogs as well if we want to reduce the size of the file further in the future

Issue #2195, opened 2026-07-29
FYI: npm publish-time malware scanning makes the `npm info` probe / already-published path a routine case


trivikr:
Heads-up rather than a bug report. I don't have a reproduction, and part of this depends on registry behavior npm hasn't documented yet. Filing so it's on the radar.

### What changed

npm has started [scanning packages at publish time](https://github.blog/changelog/2026-07-28-npm-publish-time-malware-scanning-and-dual-use-metadata/). A newly published version is scanned before it becomes installable, so there is now a window — typically around five minutes, longer at peak — between a successful `npm publish` and the version being available. A publish can also be held for manual review, or blocked.

The changelog says `npm dist-tag` keeps working during the window while `npm deprecate` and `npm unpublish` do not until the version becomes available. It does not say what a registry *read* returns during the window.

### Where this touches publish

The probe is [`info` in `lib/npm.ts`](https://github.com/changesets/changesets/blob/1b9687fae5f3c7a837e69f9f0313a177b9e2c4c8/packages/cli/src/lib/npm.ts#L198-L224): a bare [`npm info <name> --json`](https://github.com/changesets/changesets/blob/1b9687fae5f3c7a837e69f9f0313a177b9e2c4c8/packages/cli/src/lib/npm.ts#L201), with an [exact `name@version` query](https://github.com/changesets/changesets/blob/1b9687fae5f3c7a837e69f9f0313a177b9e2c4c8/packages/cli/src/lib/npm.ts#L209-L217) as a fallback when the bare one comes back empty. [`getPublishPlan`](https://github.com/changesets/changesets/blob/1b9687fae5f3c7a837e69f9f0313a177b9e2c4c8/packages/cli/src/commands/publish-plan/getPublishPlan.ts#L171) then decides with `!publishedVersions.includes(localVersion)`.

On npmjs.org the bare query succeeds, so the exact fallback doesn't run and `publishedVersions` comes from the packument's `versions`. If a pending-scan version is absent from `versions`, the plan reads it as never published.

[...]

PR #2220, opened 2026-08-07
fix(cli): detect already-published errors outside `error.summary`

KealanAU:
Follow-up to #2099. The publish-tool refactor fixed the crash, but the npm path still only matches the already-published message in `error.summary`, so an `E403` that puts it elsewhere fails the release instead of being skipped.

- `isAlreadyPublishedError` takes any number of candidate strings and ignores nullish ones.
- `handlePublishError` also matches `error.detail` and the process output.
- A payload that fails `isNpmPublishError` (no `summary`) is now classified from the process output rather than falling straight to `failed`.
- Existing snapshots are untouched; one test covers the unrecognised-payload branch.

Context on why that branch matters in practice: #2195.

### DIFF
--- a/packages/cli/src/lib/common.ts
+++ b/packages/cli/src/lib/common.ts
@@ -25,11 +25,10 @@ export const npmPublishQueue = createPromiseQueue(
 export function isAlreadyPublishedError(
   ...messages: (string | undefined)[]
 ): boolean {
-  return messages
-    .filter((message) => message != null)
-    .some((message) =>
-      message.includes("cannot publish over the previously published"),
-    );
+  return messages.some(
+    (message) =>
+      message?.includes("cannot publish over the previously published") ?? false,
+  );
 }
 
 export const isPublishSuccessful = (
--- a/packages/cli/src/lib/common.ts
+++ b/packages/cli/src/lib/common.ts
@@ -17,10 +17,19 @@ export const npmPublishQueue = createPromiseQueue(
 
 /*
  * We check `npm info` before publishing but it can return stale data at times
- * so we need to gracefully handle this situation
+ * so we need to gracefully handle this situation.
+ *
+ * Which field carries the message varies by package manager and version, so
+ * callers pass every candidate and we match across all of them.
  */
-export function isAlreadyPublishedError(message: string) {
-  return message.includes("cannot publish over the previously published");
+export function isAlreadyPublishedError(
+  ...messages: (string | undefined)[]
+): boolean {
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/common.ts
```diff
@@ -17,10 +17,19 @@ export const npmPublishQueue = createPromiseQueue(
...
-export function isAlreadyPublishedError(message: string) {
-  return message.includes("cannot publish over the previously published");
+export function isAlreadyPublishedError(
+  ...messages: (string | undefined)[]
+): boolean {
+  return messages
+    .filter((message) => message != null)
+    .some((message) =>
+      message.includes("cannot publish over the previously published"),
+    );
```
beeequeue:
```suggestion
    .some((message) =>
      message?.includes("cannot publish over the previously published") ?? false,
    );
```

packages/cli/src/lib/npm.test.ts
```diff
@@ -113,6 +113,16 @@ describe("publishing", () => {
     },
   );
 
+  it("should handle an already-published error the JSON payload doesn't describe", () => {
+    const result = npm.handlePublishError(
+      { name: "@test/package", version: "1.0.0" },
+      { error: { code: "E403" } },
+      "npm error code E403\nnpm error 403 Forbidden - PUT https://registry.npmjs.org/@test%2fpackage - You cannot publish over the previously published versions: 1.0.0.",
+    );
```
beeequeue, line 121: is this exactly what the errors looks like?
which npm version produces this?
is it different in v11 vs. v12?
what do the errors look like with yarn and pnpm?
can it be reproduced easily? we have [a repo](https://github.com/beeequeue/pm-publish-outputs) with snapshots of publishing errors, it would be nice to add this case there

Andarist: Yeah, an exact reproduction scenario of this problem would be much appreciated.

KealanAU:
@beeequeue @Andarist 

Sorry for the round trip and waste of both of your time, I was tracking a bug in my own repo, claude code concluded it was upstream here and opened this on auto. I've since traced it properly.

The answer to your questions: no, that fixture isn't real. The actual case was pnpm 11 + changesets ≤2.31.x republishing an existing version pnpm returns {"error":{"code":"E403","message":"...cannot publish over the previously published versions: X."}}, no summary, which is what crashed 2.31.0 in #2099.  

So no action from you two is needed. 

Closing.

PR #2222, opened 2026-08-07
Fix linked and fixed glob patterns expansion

itsjxck:
Fixes loss of glob pattern support for `fixed` and `linked` configuration fields.

### DIFF
--- a/packages/config/src/rules.ts
+++ b/packages/config/src/rules.ts
@@ -36,13 +36,13 @@ const fixedGroupsExist: Rule = ({
   }
 };
 
-const noDuplicateFixedPackages: Rule = ({ writtenConfig, errors }) => {
-  if (!writtenConfig.fixed || writtenConfig.fixed.length === 0) return;
+const noDuplicateFixedPackages: Rule = ({ config, errors }) => {
+  if (config.fixed.length === 0) return;
 
   const foundNames = new Set<string>();
   const duplicatedNames = new Set<string>();
 
-  for (const fixedGroup of writtenConfig.fixed) {
+  for (const fixedGroup of config.fixed) {
     for (const name of fixedGroup) {
       if (foundNames.has(name)) {
         duplicatedNames.add(name);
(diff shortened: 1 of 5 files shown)


itsjxck:
I threw this up while I was digging into some more - the test cases here, they expect the result to be the glob pattern, but in every place I can see that consumes the returned config it expects the actual package names. Is there something I'm missing here or are the tests wrong and should be updated?

itsjxck:
I've update the validation rules for finding duplicate packages/globs to work against the `writtenConfig` and updated the tests to correctly validate the glob matching returns resulting package names and not the raw glob pattern

Andarist:
>  Is there something I'm missing here or are the tests wrong and should be updated?

It might be a genuine bug with the wrong expectations. I'd recommend cross-referencing with our v2 branch: https://github.com/changesets/changesets/tree/maintenance/v2

itsjxck:
Yeah looks like this just didn't survive the rework, `v2` does expand the glob patterns and the docs also do still say glob patterns are supported.

packages/config/src/rules.ts
```diff
@@ -36,13 +36,13 @@ const fixedGroupsExist: Rule = ({
...
-const noDuplicateFixedPackages: Rule = ({ config, errors }) => {
-  if (config.fixed.length === 0) return;
+const noDuplicateFixedPackages: Rule = ({ writtenConfig, errors }) => {
+  if (!writtenConfig.fixed || writtenConfig.fixed.length === 0) return;
 
   const foundNames = new Set<string>();
   const duplicatedNames = new Set<string>();
 
-  for (const fixedGroup of config.fixed) {
+  for (const fixedGroup of writtenConfig.fixed) {
```
Andarist, line 45: hm, the written config has unresolved values. I feel the goal here is to validate the resolved groups so we should be using `config` here. Maybe I'm missing something here. Could you explain the results you see when using both of them?

beeequeue: the main question that needs answering is whether the users of `.linked` and `.fixed` expect glob patterns or a list of packages, if it's the latter this is the correct fix

packages/config/src/rules.ts
```diff
@@ -76,13 +76,13 @@ const linkedGroupsExist: Rule = ({
   }
 };
 
-const noDuplicateLinkedPackages: Rule = ({ config, errors }) => {
-  if (config.linked.length === 0) return;
+const noDuplicateLinkedPackages: Rule = ({ writtenConfig, errors }) => {
+  if (!writtenConfig.linked || writtenConfig.linked.length === 0) return;
```
beeequeue, line 80: we should default the value to `[]` instead so we can avoid checks like this

packages/config/src/rules.ts
Andarist: I think this PR is right in the sense we should expand globs (that's already what's done for `.ignore`).

I'm adding extra test coverage + double-checking how those rules here are written. Gonna push out final version in the morning.

Andarist (approved):

PR #2237, opened 2026-08-11
Fix blog image layout shift

bluwy:
I noticed the images causes layout shift because the width and height attributes aren't set. This sets them.

Also, fixed the alt text for VPImage in object form.

### DIFF
--- a/site/blog/announcing-changesets-v3.md
+++ b/site/blog/announcing-changesets-v3.md
@@ -3,7 +3,6 @@ import { VPImage } from "vitepress/theme"
 import DiscordLogo from "~icons/logos/discord-icon"
 import OCLogo from "~icons/logos/opencollective"
 import Socials from "../.vitepress/theme/components/Socials.vue"
-import cliExampleImage from "../assets/cli-example.webp"
 import packThenPublishImage from "../assets/pack-then-publish.excalidraw.svg"
 import packThenPublishLightImage from "../assets/pack-then-publish-light.excalidraw.svg"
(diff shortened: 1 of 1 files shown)


site/blog/announcing-changesets-v3.md
```diff
@@ -102,11 +103,7 @@ Changesets now uses [`cac`](https://npmx.dev/cac) for CLI arguments and help mes
 
 <!-- https://lexidraw.app/#atproto=did:plc:skqg5gindwkuzjmjub6db6yn,3mskqppq5nk2u -->
 
-<VPImage
-  loading="lazy"
-  alt="cli flow example"
-  :image="cliExampleImage"
-/>
+![cli flow example](../assets/cli-example.webp)
```
bluwy, line 106: VitePress can auto-calculate the width and height using this markdown form, so converted it this way here.

I need to investigate later is there's a better way to support light-dark images without using VPImage so we can get this automatic size calculation as well.

Andarist (approved):

site/blog/announcing-changesets-v3.md
beeequeue: It doesn't add the lazy property iirc when I added it, even with the vitepress setting

beeequeue: But that should be fine, it's a small-ish image

bluwy: Setting `markdown.image.lazyLoad` seems to add for the markdown images, we can add it though

PR #2240, opened 2026-08-11
Skip release PR in PR status workflow example

marcalexiei:
The example comments on every pull request, including the release PR where `changeset version` has consumed all changesets, so the comment always reports that none is present.

This adds an `if` condition to the `pr-status` job skipping `changeset-release/*`
branches, matching what the Changesets GitHub Bot does internally.

### DIFF
--- a/site/guide/_snippets/automating-non-blocking.yaml
+++ b/site/guide/_snippets/automating-non-blocking.yaml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   pr-status:
     # recommended: skip the release pull request, as it never has changesets
-    if: ${{ !startsWith(github.event.pull_request.head.ref, 'changeset-release/') }}
+    if: ${{ !startsWith(github.head_ref, 'changeset-release/') }}
     runs-on: ubuntu-slim
     permissions:
       contents: read # to check out repo (actions/checkout)
--- a/site/guide/_snippets/automating-non-blocking.yaml
+++ b/site/guide/_snippets/automating-non-blocking.yaml
@@ -11,6 +11,8 @@ concurrency:
 
 jobs:
   pr-status:
+    # recommended: skip the release pull request, as it never has changesets
+    if: ${{ !startsWith(github.event.pull_request.head.ref, 'changeset-release/') }}
     runs-on: ubuntu-slim
     permissions:
       contents: read # to check out repo (actions/checkout)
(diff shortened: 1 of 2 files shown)


beeequeue (approved):
very nice improvement!

site/guide/_snippets/automating-non-blocking.yaml
```diff
@@ -11,6 +11,8 @@ concurrency:
 
 jobs:
   pr-status:
+    # recommended: skip the release pull request, as it never has changesets
+    if: ${{ !startsWith(github.event.pull_request.head.ref, 'changeset-release/') }}
```
beeequeue:
```suggestion
    if: ${{ !startsWith(github.head_ref, 'changeset-release/') }}
```

slightly shorter version

beeequeue (commented):
should we also add this to the `pr-comment` step?

site/guide/_snippets/automating-non-blocking.yaml
marcalexiei: Applied the `github.head_ref` suggestion, thanks.

For `pr-comment`: it is not needed.
It's a separate job with `needs: pr-status`, and GitHub skips dependents when a needs job is skipped (unless they override with `always()`).

Also combined the if in the `pull_request` note above the snippet.

site/guide/automating.md
```diff
@@ -38,14 +38,16 @@ You can also use the `pull_request` event if you prefer to lock permissions down
...
+    if: github.event.pull_request.head.repo.full_name == github.repository && !startsWith(github.head_ref, 'changeset-release/')
     # ...
 ```
 
 :::
 
 <<< ./_snippets/automating-non-blocking.yaml [.github/workflows/comment-changesets-pr-status.yml]
 
+The `if` condition on the `pr-status` job skips the release pull request opened by the Changesets GitHub Action, which never has changesets.
+
```
bluwy: Let's skip this as it's already mentioned in the workflow
```suggestion
```

site/guide/_snippets/automating-non-blocking.yaml
```diff
@@ -11,6 +11,8 @@ concurrency:
 
 jobs:
   pr-status:
+    # recommended: skip the release pull request, as it never has changesets
```
bluwy:
```suggestion
    # recommended: skip the release pull request as it never has changesets
```

bluwy (approved):

PR #2246, opened 2026-08-13
Infer the default branch during changeset init

### LINKED ISSUE #807: Determine baseBranch on init
## Affected Packages

`pkg:@changesets/cli`

## Problem

Running `changes init` creates a config with `"baseBranch": "main"` regardless of what the default branch actually is

## Proposed solution

It would be nice if the settings determined the default branch and initialised the settings file accordingly

Boulea7:
Closes #807.

When `changeset init` asks for a base branch, it now reads the symbolic `origin/HEAD` and verifies that its target ref exists. It continues using `main` if Git is unavailable or the ref cannot be resolved. A branch entered at the prompt still takes precedence.

Focused tests cover non-`main`, missing, dangling, non-origin, unavailable-Git, and explicit-input cases. The focused 49 tests, build, typecheck, lint, and format checks pass.

The full suite reports 915 passing tests and one unchanged `packages/write` formatter-fixture failure: `pnpm exec oxfmt` exits with status 1 inside its isolated temporary workspace.

### DIFF
--- a/.changeset/quiet-cats-remember.md
+++ b/.changeset/quiet-cats-remember.md
@@ -1,0 +1,6 @@
+---
+"@changesets/cli": minor
+"@changesets/git": minor
+---
+
+Use the branch referenced by `origin/HEAD` as the default base branch when initializing Changesets, while continuing to fall back to `main` when it cannot be resolved.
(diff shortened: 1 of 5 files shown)


beeequeue:
I'm not sure about this approach, there's no guarantee that the user will be on the default branch when they run changeset init and it would be confusing if we suggest the wrong default value.

Andarist:
Checking `origin/HEAD` should work OK (if done appropriately) regardless of your current branch. So the core of the idea seems to be fine to me - but I have not yet looked into the content of the PR.

PR #2249, opened 2026-08-13
chore(deps): replace semver with verkit

gameroman:
Context: https://github.com/e18e/ecosystem-issues/issues/277

https://npmx.dev/compare?packages=semver,verkit


beeequeue (approved):
i'm okay with this, with a lonely concern that it isn't deduped as well as semver is (yet?)

but i want to hear what you two think as well @bluwy @Andarist

beeequeue:
could you rebase and update the package?

PR #2258, opened 2026-08-21
docs: improve migration section about action

trueberryless:
This PR adds more documentation to the migration guide regarding the changes of the `changesets/action`. I chose to keep it simple and not mention any sub-actions in the code blocks, but just link to the docs for [Trusted Publishing](https://changesets.dev/guide/automating#trusted-publishing) as then it is recommended to split up the action.

Moreover, I added a section for updating parameters with "common gotcha" notes below and then linked to the release and the repo for full available parameters and breaking changes.

Preview link: https://deploy-preview-2258--changesets.netlify.app/guide/migration#github-actions-changes

Any suggestions are welcome!

PR #2259, opened 2026-08-22
Follow up migration guide changes

bluwy:
From #2258

Correct action input naming. Mention publish-script. No need to mention github token as it has been optional for a while. No need to mention the typo bug.

### DIFF
--- a/site/guide/migration.md
+++ b/site/guide/migration.md
@@ -256,9 +256,9 @@ While the existing action should work as before, the v2 action exposes more sub-
 
 For instance, if you are using [npm trusted publishing (provenance)](https://docs.npmjs.com/trusted-publishers), it is recommended to migrate to these sub-actions to tighten publish permissions. Check the new [Automating Changesets guide](./automating.md) for the new setup recommendations.
 
-### Update parameters
+### Update inputs
 
-If you passed any parameters or a GitHub token, make sure to update them accordingly:
+If you passed any inputs, make sure to update them accordingly:
 
 <!-- prettier-ignore -->
 ```yaml [.github/workflows/publish.yml]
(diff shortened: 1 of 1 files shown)


beeequeue (approved):

trueberryless:
I deliberately chose to include the `GITHUB_TOKEN` change as I have seen that it is used very frequently and people should probably know that the correct way to pass it is not via env variables anymore...

Apart from that, great changes ✅

bluwy:
I think if it were to show the github-token changes, it should be about custom GitHub tokens. If using the default, you don't have to set anything. And it's also a bit rare to need to set a custom GitHub token now. So with this section covering the common changes needed, I think it's fine to omit that.

Plus, since we direct users to a more complete workflow setup, they no longer have mention of the old GITHUB_TOKEN convention, so I think this would also slowly disappear itself.

trueberryless:
Do I also not need a custom token anymore if I set `push-with-git-cli: true` now?
Because if this needs a custom token, shouldn't we mention it in the paragraph after the code block which still mentions this addition?

bluwy:
You don't need a custom token for that option. It'll use the git authentication set up by actions/checkout IIRC

trueberryless:
Ah okay, then it comes down to whether the action would still work if you set the `GITHUB_TOKEN` as an env var. Because if this change is breaking and not documented in the migration guide, I believe there will be some users that might miss it and wonder why the changeset commits are authored by another bot. That's why I am in favor of documenting it. But feel free to merge this if you think otherwise 👍

bluwy:
If you passed a custom token to GITHUB_TOKEN env var, there's an error now that it doesn't match the `github-token` input (which would have the default token). So this should be caught but perhaps only when it errors.

I'll merge this for now but we can continue on following up on it if needed.

PR #2270, opened 2026-08-25
Add review message on pre exit

### LINKED ISSUE #240: Exiting pre mode should prompt for changeset cleanup

bluwy:
closes #240

Add a simple message on `pre exit` to ask to review the changesets in `.changeset/pre`.

NOTE: I updated the string formatting in the second commit, as the review message (added in the first commit) made the formatting a bit awkward. But I'm happy to revert this if we feel it's unnecessary for now.

### DIFF
--- a/.changeset/sunny-turtles-glow.md
+++ b/.changeset/sunny-turtles-glow.md
@@ -1,0 +1,5 @@
+---
+"@changesets/cli": patch
+---
+
+Add review message on pre exit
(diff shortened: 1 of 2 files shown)


beeequeue (approved):
it's a good guide for new users and reminder for old ones

Andarist (approved):

PR #2276, opened 2026-08-29
Fix pnpm 10 publishing with npm 12

### LINKED ISSUE #2274: pnpm.ts's `parseInfoResult` doesn't unwrap npm 12's array-wrapped
## Summary

`changeset publish` crashes with `pnpm` as the package manager, in any environment where the underlying `npm` binary is npm 12+ (e.g. after a `npm install -g npm@latest` step in CI, which is common right before a release job):

```
TypeError: Cannot read properties of undefined (reading 'includes')
at getUnpublishedPackages (.../@changesets/cli/dist/getPublishPlan.mjs:613:26)
at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
at async getPublishPlan (.../getPublishPlan.mjs:665:19)
at async publish (.../publish.mjs:70:96)
```

## Root cause

On pnpm ≤10, `pnpm info`/`pnpm view` is not a native pnpm command — it's passed straight through to the system's `npm` binary (see [`pnpm/src/pnpm.ts`](https://github.com/pnpm/pnpm/blob/v10.33.0/pnpm/src/pnpm.ts#L10-L47), the `passThruToNpm()` case list includes `info`/`view`). So `pnpm info <pkg> --json` on pnpm 10.x is really just `npm view <pkg> --json` under the hood, using whatever `npm` is currently resolvable on `PATH` (`packages exec/run-npm/src/index.ts`: `const npm = npmPath ?? 'npm'`).

npm 12.0.0 changed `npm view <pkg> --json` to always wrap its result in an array (`[ {...} ]`) instead of returning a bare object — this is exactly what #2164 reported and was fixed for, in the pre-v3 `npm-utils.ts`'s `getPackageInfo()`.

That fix correctly carried over into the new `npm.ts` after the publish-flow split in #2155 (`parseInfoResult$1`/the npm tool's `parseInfoResult` explicitly does `Array.isArray(parsed) ? { info: parsed[0] } : { info: parsed }`), but the equivalent guard was **not** added to the new `pnpm.ts`'s `parseInfoResult`:

```ts
// packages/cli/src/lib/pnpm.ts
function parseInfoResult({ exitCode, stdout, stderr }) {
  if (exitCode !== 0) return { error: getPnpmError(stderr, stdout) }
  return stdout ? { info: JSON.parse(stdout) } : undefined   // <-- no array unwrap
}
```

Since pnpm ≤10 delegates to the same underlying `npm view`, it hits the exact same npm-12 array-wrapping — but because `pnpm.ts`'s parser stores the array as-is in `response.info`, `response.info.versions` is `undefined`, and `getUnpublishedPackages` crashes calling `.includes()` on it in `getPublishPlan.ts`.

## Reproduction

* `@changesets/cli` 3.0.1
* Any pnpm ≤10 (tested 10.33.0)
* npm 12.0.0+ resolvable on `PATH` (e.g. right after `npm install -g npm@latest`)
* Publish a package that's already on the registry (or run `changeset publish` with no changesets) → crash instead of "already published" / "no unpublished projects" message.

Andarist:
fixes https://github.com/changesets/changesets/issues/2274

### DIFF
--- a/packages/cli/src/lib/pnpm.ts
+++ b/packages/cli/src/lib/pnpm.ts
@@ -3,7 +3,7 @@ import { exec } from "tinyexec";
 import { getLastJsonObjectFromString } from "../utils/getLastJsonObjectFromString.ts";
 import { isAlreadyPublishedError } from "./common.ts";
 import * as npm from "./npm.ts";
-import type { PackageInfo, PublishResult, PublishTool } from "./types.ts";
+import type { PublishResult, PublishTool } from "./types.ts";
 
 export type PnpmPublish2faRequiredError = {
   code: "ERR_PNPM_OTP_NON_INTERACTIVE";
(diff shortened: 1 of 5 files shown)


packages/cli/src/lib/pnpm.ts
```diff
@@ -49,34 +49,37 @@ type PnpmCommandError = {
...
+    error &&
+    typeof error === "object" &&
+    !Array.isArray(error) &&
+    typeof error.code === "string" &&
+    typeof error.message === "string"
+  ) {
+    return { code: error.code, message: error.message };
   }
-  return { message: stderr || stdout || undefined };
 }
```
beeequeue, line 64: this should probably use an early `null` return instead of a nested happy-path

bluwy (approved):

bluwy:
I think eventually we should have each tool handle the major version internally itself, but this is also fine for now

Andarist:
> I think eventually we should have each tool handle the major version internally itself, but this is also fine for now

yeah, I mainly went with this approach to avoid refactoring things. Those tools are not stateful right now and I didn't want to recheck the version on each "method" invocation

PR #2277, opened 2026-08-30
fix(git): wrong changeset file commit

### LINKED ISSUE #2272: User misattributed in CHANGELOG using `@changesets/changelog-github`
I just noticed a changeset being misattributed to the wrong user in a recent release.

Here’s the release PR: https://github.com/withastro/starlight/pull/4161 (I manually edited to correct the issue)

It included the following (in PR description and the generated CHANGELOG file):

> - [#4156](https://github.com/withastro/starlight/pull/4156) [`57af92d`](https://github.com/withastro/starlight/commit/57af92d91f9a9d0936a06efb4387898aded62aa9) Thanks [@pyxelr](https://github.com/pyxelr)! - Adds 1 new icon: `hypothesis`
>
> - [#4156](https://github.com/withastro/starlight/pull/4156) [`57af92d`](https://github.com/withastro/starlight/commit/57af92d91f9a9d0936a06efb4387898aded62aa9) Thanks [@pyxelr](https://github.com/pyxelr)! - Adds 1 new icon: `npmx`

However, the second of these changesets belongs to a different [PR](https://github.com/withastro/starlight/pull/4159), [commit SHA](https://github.com/withastro/starlight/commit/a0e0ba263c93bab214222b798b52a759fb799ba7) and user.

Our changesets config file: https://github.com/withastro/starlight/blob/a0e0ba263c93bab214222b798b52a759fb799ba7/.changeset/config.json#L1-L16

And versions in use:

```json
    "@changesets/changelog-github": "^1.0.0",
    "@changesets/cli": "^3.0.1",
```

Thanks as always for working on Changesets 💙

trueberryless:
#### Description

This PR fixes a regression added [here](https://github.com/changesets/changesets/pull/2190#discussion_r3654687068), where similar changeset files can lead to wrong commit SHA lookups if the file content is very similar.

The fix adds the `C` (copy) git type to the [`--diff-filter` option](https://git-scm.com/docs/git-log#Documentation/git-log.txt---diff-filterACDMRTUXB), which means that two changeset files will not be mixed up by git just because they have similar content.

- Fixes #2272

### DIFF
--- a/packages/git/src/index.test.ts
+++ b/packages/git/src/index.test.ts
@@ -430,7 +430,10 @@ describe("git", { tags: ["slow"] }, () => {
         await outputFile(path.join(cwd, "a.js"), 'export default "Andarist"');
         const originalCommit = await addFileAndCommit("a.js", cwd);
 
-        await outputFile(path.join(cwd, "a.js"), 'export default "Andarist tweaked"');
+        await outputFile(
+          path.join(cwd, "a.js"),
+          'export default "Andarist tweaked"',
+        );
         await add("a.js", cwd);
         await commit("tweak a.js", cwd);
 
@@ -438,7 +441,9 @@ describe("git", { tags: ["slow"] }, () => {
         await renameFileAndCommit("a.js", "pre/a.js", cwd);
 
         const clone = await shallowClone(cwd, 1);
-        const commits = await getCommitsThatAddFiles(["pre/a.js"], { cwd: clone });
+        const commits = await getCommitsThatAddFiles(["pre/a.js"], {
+          cwd: clone,
+        });
         expect(commits).toEqual([originalCommit]);
       });
 
@@ -450,12 +455,17 @@ describe("git", { tags: ["slow"] }, () => {
         await fs.mkdir(path.join(cwd, "pre"));
         await renameFileAndCommit("a.js", "pre/a.js", cwd);
 
-        await outputFile(path.join(cwd, "pre/a.js"), 'export default "bluwy tweaked"');
+        await outputFile(
+          path.join(cwd, "pre/a.js"),
+          'export default "bluwy tweaked"',
+        );
         await add("pre/a.js", cwd);
         await commit("tweak pre/a.js", cwd);
 
         const clone = await shallowClone(cwd, 1);
-        const commits = await getCommitsThatAddFiles(["pre/a.js"], { cwd: clone });
+        const commits = await getCommitsThatAddFiles(["pre/a.js"], {
+          cwd: clone,
+        });
         expect(commits).toEqual([originalCommit]);
       });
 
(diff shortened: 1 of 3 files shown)


packages/git/src/index.ts
```diff
@@ -84,10 +84,11 @@ export async function getCommitsThatAddFiles(
             "git",
             [
               "log",
-              "--diff-filter=A",
+              "--diff-filter=AC",
```
beeequeue, line 90: can we add a short comment with an explanation of what AC is and why we use it, along with a link to the git docs?

trueberryless: Thanks for the great suggestion! I added two comments for both tricky flags 👍

bluwy:
> This PR also fixes another issue by adding the `--` separator to the `git log` command to tell git to look for the history of a deleted file (because before this command runs, the `version` command has already deleted the changeset files in the checkout project).

Can you elaborate on this? We should be [capturing the commits first](https://github.com/changesets/changesets/blob/0061ef0454997469d9fba52a05036702e8357102/packages/apply-release-plan/src/index.ts#L287-L290) before[ deleting the files](https://github.com/changesets/changesets/blob/0061ef0454997469d9fba52a05036702e8357102/packages/apply-release-plan/src/index.ts#L226-L227). With the `--` flag it looks like it can match files not in the current branch, which we only want to match on the current branch.

trueberryless:
> Can you elaborate on this? We should be [capturing the commits first](https://github.com/changesets/changesets/blob/0061ef0454997469d9fba52a05036702e8357102/packages/apply-release-plan/src/index.ts#L287-L290) before[ deleting the files](https://github.com/changesets/changesets/blob/0061ef0454997469d9fba52a05036702e8357102/packages/apply-release-plan/src/index.ts#L226-L227).

Ah true, I confused the method order. Nonetheless, I strongly recommend using the `--` separator to prevent confusion with other parameters, as [the docs](https://git-scm.com/docs/git-log#Documentation/git-log.txt---path) mention.

> With the `--` flag it looks like it can match files not in the current branch, which we only want to match on the current branch.

Where exactly does it mention this?

bluwy:
> Where exactly does it mention this?

I tested it manually. If you delete `README.md` for example, if you do `git log README.md`, it fails. But if `git log -- README.md` it passes.

When it fails, it also logs this which kinda explains it ("path not in the working tree"):

```
fatal: ambiguous argument 'README.md': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

trueberryless:
> I tested it manually. If you delete `README.md` for example, if you do `git log README.md`, it fails. But if `git log -- README.md` it passes.
> 
> When it fails, it also logs this which kinda explains it ("path not in the working tree"):
> 
> ```
> fatal: ambiguous argument 'README.md': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'
> ```

Gotcha. I think the error happens because git doesn't fully understand that we want to pass the path argument (as the file doesn't exist). It interpretes our path as a revision, but this is no revision. So adding `--` makes sure that git always knows it's a path.

I initially included it because I also thought that the command would fail with the error you showed. However, since we get the commits before deleting, I think it isn't strictly necessarily to include the `--` separator currently. It is just more fragile if we don't include it. 

So IMHO we should keep the `--` separator, but im also open to removing it 👍

bluwy:
I'd prefer removing `--` so the command only works on paths in the current working tree. If you prefer to use `--`, we need a flag or something that makes sure to only return if the path exists in the current working tree. If there's no way to do that, we should remove `--`.

trueberryless:
I removed it for now because if you use `--`, git searches the commits even when the file was deleted (outside of current working tree). However, one potential problem without the separator is: If a branch exists with the exact name of the path we pass, git would show the history of the branch instead of the path. This is very unlikely I think because I never saw some branch name that includes `.md`. Nonetheless, it's worth considering. But removing it is fine for me, so this PR focuses on the actual bug only 👍

bluwy (approved):

packages/git/src/index.test.ts
```diff
@@ -424,6 +424,80 @@ describe("git", { tags: ["slow"] }, () => {
         const commits = await getCommitsThatAddFiles(["b.js"], { cwd: clone });
         expect(commits).toEqual([originalCommit]);
       });
+
+      it("reads the SHA of a file-add when tweaked then moved to pre", async () => {
```
Andarist, line 428: I love those tests ❤️

PR #2281, opened 2026-09-03
Upgrade to Vitest 5

Andarist:


packages/cli/src/commands/publish/__tests__/__snapshots__/e2e.test.ts.snap
```diff
@@ -169,19 +157,7 @@ pkg-a@1.0.0
 "
 `;
 
-exports[`Publish command e2e > 'npm 11' > publishes a new version of a package 1`] = `
```
Andarist, line 172: those were obsolete/stale - and Vitest 4 didn't fail on those

beeequeue (approved):

PR #2282, opened 2026-09-06
Report a package in multiple fixed or linked groups only once

MaxFreedomPollard:
A project with three or more `fixed` or `linked` groups gets the "defined in multiple groups" config error printed once for every group that follows the duplicate. With `fixed: [["pkg-a"], ["pkg-a"], ["pkg-b"]]` the `pkg-a` message is printed twice, and adding a fourth group prints it three times.

The cause is in `packages/config/src/rules.ts`. `noDuplicateFixedPackages` (line 39) and `noDuplicateLinkedPackages` (line 79) collect duplicates into a `duplicatedNames` set, but the `errors.push(...Array.from(duplicatedNames, ...))` call sits inside the loop over the groups, at line 52 and line 92. Every iteration re-emits a message for every name already in the set, so a name first seen twice while processing group 2 of 4 gets pushed by groups 2, 3 and 4. The two existing cases in `packages/config/src/parse.test.ts` use two groups, `fixed: [["pkg-a"], ["pkg-a"]]` and `linked: [["pkg-a"], ["pkg-a"]]`, where the duplicate is in the last group and no later iteration exists to repeat it, and the table harness compares expected substrings positionally without checking how many errors came back, so neither case could catch this.

The fix moves both pushes out of the loop. Each duplicated package now produces exactly one error regardless of how many groups follow it. The set of names collected, the message text, and the early return when the group list is empty are all unchanged.

Two regression tests are added to the `rules` block of `packages/config/src/parse.test.ts`, one for `fixed` and one for `linked`, each asserting exactly one error for `[["pkg-a"], ["pkg-a"], ["pkg-b"]]`. Against unmodified `rules.ts` both fail with `AssertionError: expected [ …(2) ] to have a length of 1 but got 2`. With the fix, `pnpm vitest run packages/config` reports `Test Files 2 passed (2)` and `Tests 65 passed (65)`.

Verified on macOS with Node 22.23.2 and pnpm 11.13.1. `pnpm test run --maxWorkers=3` reports `Test Files 41 passed (41)` and `Tests 916 passed | 2 skipped | 2 todo (920)` in 130.94s, with no failures. `pnpm types:check`, `pnpm lint` and `pnpm format` each exit 0, with `oxfmt --check` reporting "All matched files use the correct format" across 298 files.

Changeset included, patch bump for `@changesets/config`.

### DIFF
--- a/.changeset/soft-books-invent.md
+++ b/.changeset/soft-books-invent.md
@@ -1,0 +1,5 @@
+---
+"@changesets/config": patch
+---
+
+Report a package that is defined in multiple `fixed` or `linked` groups only once, instead of once for every group that follows the duplicate.
--- a/packages/config/src/parse.test.ts
+++ b/packages/config/src/parse.test.ts
@@ -624,6 +624,44 @@ describe("validateConfig", () => {
 
   // any tests that don't fit nicely into the cases above
   describe("rules", () => {
+    describe("noDuplicateFixedPackages", () => {
+      it("should report a duplicated package once when more groups follow it", async () => {
+        const config = { fixed: [["pkg-a"], ["pkg-a"], ["pkg-b"]] };
+
+        const cwd = await testdir({
+          ".changeset/config.json": JSON.stringify(config),
+          "package.json": JSON.stringify(rootManifest),
+        });
+        const packages = withPackages(cwd, ["pkg-a", "pkg-b"]);
+
+        const result = validateConfig(config, packages);
+
+        expect(result.errors).toHaveLength(1);
+        expect(result.errors![0]).toContain(
+          'fixed: Invalid group: The package or glob "pkg-a" is defined in multiple groups',
+        );
+      });
+    });
+
+    describe("noDuplicateLinkedPackages", () => {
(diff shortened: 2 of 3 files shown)


43081j (approved):
good catch!

beeequeue (approved):
while i do not appreciate fully automated prs like this, we don't have a policy against it (yet?) and the change makes sense.

@bluwy @Andarist

MaxFreedomPollard:
How is it fully automated?

MaxFreedomPollard:
Not that it's a bad thing, I'm just wondering why you say that. I don't mind either way, it's your repo and I respect it oc

bluwy:
Your PR description is overly verbose, we do not need to know the number of tests passed or what tests you've added. We can see on the GitHub UI and diff.

You've also opened 88 PRs in the last 2 days.

These kinds of contribution patterns just don't give confidence to any OSS maintainers.

MaxFreedomPollard:
I included that info so it can viewed at a glance. It's your preference, and it's now noted.

As for the large amount of PRs recently - I've only made one PR in this repository. What I do in other repositories, many of which are my own, is not your concern.

PR #2289, opened 2026-09-06
test: add happy path tests for `add`

43081j:
The tests were covering error cases and what not, but not the happy
path.

This just adds a test for each of major/minor/patch that asserts a
summary is seen, no prompt calls, etc.

### DIFF
--- a/packages/cli/src/commands/add/__tests__/add.test.ts
+++ b/packages/cli/src/commands/add/__tests__/add.test.ts
@@ -691,6 +691,100 @@ describe("Add command", () => {
     );
   });
 
+  describe("release type flags", () => {
+    let cwd: string;
+
+    beforeEach(async () => {
+      cwd = await testdir({
+        "package.json": JSON.stringify({
+          private: true,
+          name: "root-pkg",
+          workspaces: ["packages/*"],
+        }),
+        "package-lock.json": "",
+        "packages/pkg-a/package.json": JSON.stringify({
+          name: "pkg-a",
+          version: "1.0.0",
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/add/__tests__/add.test.ts
```diff
@@ -691,6 +691,100 @@ describe("Add command", () => {
     );
   });
 
+  describe("release type flags", () => {
+    let cwd: string;
+
+    beforeEach(async () => {
+      cwd = await testdir({
```
Andarist: nit: I'd prefer not to rely on `beforeEach`. Either inline this or extract the `testdir`'s to a top level constant (or create a short util to call testdir for u with the constant fresh argument)

packages/cli/src/commands/add/__tests__/add.test.ts
```diff
@@ -691,6 +691,100 @@ describe("Add command", () => {
     );
   });
 
+  describe("release type flags", () => {
```
Andarist: nit: I'd also prefer not to add nested describe blocks. In my experience, they always end up containing unrelated stuff cause people (and agents) dont pay attention where a new test should actualle be appended to)

43081j: this is pretty much here just to hold the `beforeEach` so both nits are the same issue 👍 

ill take a look

bluwy (approved):

Andarist (approved):

PR #2295, opened 2026-09-08
Insert new version below CHANGELOG intro

benjie:
Currently when you version, content is added directly below the file header (or at the very top of the file if the first line is a version header). This change instead adds the content directly above the first version header if one exists (otherwise previous position is maintained) allowing the CHANGELOG.md file to contain introductory content.

### DIFF
--- a/.changeset/real-comics-punch.md
+++ b/.changeset/real-comics-punch.md
@@ -1,0 +1,5 @@
+---
+"@changesets/apply-release-plan": minor
+---
+
+New version information will now be added to `CHANGELOG.md` above the first version header (if any) allowing intro content in `CHANGELOG.md` to be maintained at the top.
--- a/.changeset/real-comics-punch.md
+++ b/.changeset/real-comics-punch.md
@@ -1,5 +1,5 @@
 ---
-"@changesets/apply-release-plan": minor
+"@changesets/apply-release-plan": patch
 ---
 
 New version information will now be added to `CHANGELOG.md` above the first version header (if any) allowing intro content in `CHANGELOG.md` to be maintained at the top.
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -2154,6 +2154,57 @@ describe("apply release plan", () => {
       `);
     });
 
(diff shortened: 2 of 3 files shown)


packages/apply-release-plan/src/index.test.ts
```diff
@@ -2154,6 +2154,57 @@ describe("apply release plan", () => {
...
+
+        ## 1.1.0
+
+        ### Minor Changes
+
+        - Hey, let's have fun with testing!
+
+        ## 1.0.0
+
+        - Fixed some bug"
```
benjie, line 2204: This demonstrates the change - the "## Overview" is now maintained and "## 1.1.0" gets added afterwards, rather than adding "## 1.1.0" above the "## Overview" as currently occurs.

packages/apply-release-plan/src/index.ts
```diff
@@ -354,11 +354,13 @@ async function updateChangelog(
...
   // Note: we also need to handle prerelease versions here but that's already covered by the regex.
-  const isVersionHeading = /^#{1,6}\s+\d+\.\d+/.test(fileData);
+  const firstVersionHeaderIndex = fileData.search(/^#{1,6}\s+\d+\.\d+/m);
 
   let newChangelog: string;
-  if (isVersionHeading) {
-    newChangelog = templateString.trimStart() + fileData;
+  if (firstVersionHeaderIndex >= 0) {
+    const prefix = fileData.substring(0, firstVersionHeaderIndex);
+    const suffix = fileData.substring(firstVersionHeaderIndex);
```
bluwy:
```suggestion
    const prefix = fileData.slice(0, firstVersionHeaderIndex);
    const suffix = fileData.slice(firstVersionHeaderIndex);
```

nit: I think slice matches the pattern we use in the other code.

.changeset/real-comics-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@changesets/apply-release-plan": minor
```
bluwy:
```suggestion
"@changesets/apply-release-plan": patch
```

I believe this is a bug fix

bluwy (approved):
Looks great. Thanks!

PR #2297, opened 2026-09-10
Preserve bounds when updating dependency ranges

### LINKED ISSUE #2271: changeset version drops the upper bound of a compound (>=x <y) dependency range, silently
## Summary

`changeset version` rewrites a compound (`>=x.y.z <a.b.c`) internal dependency range down to a bare `>=x.y.z` — silently dropping the upper bound — whenever `updateInternalDependencies` touches that dependency. This is worse than not rewriting at all: the resulting range has no ceiling, so it will admit any future major of that dependency, silently, forever, until someone notices and hand-restores the ceiling.

Reproduced on both `@changesets/cli@2.31.1` and the current latest, `3.0.1`.

## Minimal reproduction

```
packages/pkg-a/package.json:
{ "name": "pkg-a", "version": "1.0.0" }

packages/pkg-b/package.json:
{
  "name": "pkg-b",
  "version": "1.0.0",
  "dependencies": { "pkg-a": ">=1.0.0 <2.0.0" }
}

.changeset/config.json:
{ "updateInternalDependencies": "patch", "fixed": [], "linked": [], "ignore": [] }

.changeset/repro.md:
---
"pkg-a": major
---
trigger a major bump on pkg-a
```

```
npm install   # link the workspace (needed on 3.0.1; 2.31.1 doesn't require this step)
npx @changesets/cli version
```

**Expected:** `pkg-b`'s dependency on `pkg-a` is updated to a range that (a) admits the new `2.0.0` and (b) still has *some* upper bound reflecting that `pkg-b` was only ever verified against `pkg-a` 1.x, e.g. `>=1.0.0 <3.0.0` or at minimum unchanged-shape reasoning applied to the new floor.

**Actual**, on both 2.31.1 and 3.0.1:

```diff
-  "dependencies": { "pkg-a": ">=1.0.0 <2.0.0" }
+  "dependencies": { "pkg-a": ">=2.0.0" }
```

The ceiling is gone. `pkg-b` will now silently accept `pkg-a@3.0.0`, `4.0.0`, or any future major, with zero further signal — no CI failure, no lint warning, nothing. Since the whole reason a package would use a compound range instead of a plain caret is to declare compatibility with more than one existing major (e.g. spanning both a pre-1.0 and a post-1.0 generation of a dependency during a migration window), losing the ceiling defeats the specific reason that range shape was chosen.

## Why this is worse than "unsupported input"

A tool that refuses to touch a range shape it doesn't understand would be a minor inconvenience (a manual fix once). What actually happens is worse: it *does* rewrite it, to something *looks* plausible (a valid, still-passing semver range) but is silently wrong, in the least safe direction (no ceiling at all rather than too tight a ceiling). Nothing in `changeset version`'s output flags that a compound range collapsed — the run reports success ("All files have been updated") with no warning that a dependency's declared compatibility window just got wider than anyone verified.

## Real-world impact (two independent occurrences, same class)

This was hit twice in the same week in a real monorepo, on two separate releases:

1. A tool package depended on six sibling packages via `>=x.y.z <1.0.0` ranges (deliberately spanning a pre-1.0 window). When one of those six went to `1.0.0`, all six ranges were rewritten to bare `>=1.0.0`, dropping every ceiling in one pass — even the five that hadn't changed version at all in that release.
2. The following release, after those ceilings were hand-restored to `>=x.y.z <2.0.0`/`<3.0.0`, a second package crossed another major. The *same six ranges* were clobbered *again*, this time down to a bare floor with no ceiling whatsoever (e.g. `>=0.7.0 <3.0.0` → `>=2.0.0`).

Both times, `npm install`/CI kept working — nothing failed loudly, because a floor-only range still legitimately resolves the currently-published version. The defect is invisible until a *future* major actually publishes and gets silently accepted.

## Suggested fix directions (any of these would help)

- Detect a range that doesn't match the simple `^`/`~`/exact shapes changesets already knows how to bump, and leave it untouched with a printed warning, rather than silently rewriting it.
- If rewriting a compound range is intended behavior, preserve the shape: bump the floor to the new version and shift the ceiling by the same relative amount (e.g. floor's major +1) rather than dropping the ceiling clause entirely.
- At minimum, log a warning line during `changeset version` whenever a range containing a `<` clause is about to be rewritten, so this doesn't ship silently.

Happy to provide the exact repo/commit shape from the real-world case if useful (kept generic here since the underlying repos are private).

Andarist:
fixes https://github.com/changesets/changesets/issues/2271

I agree the other ranges are more idiomatic, but we should try to respect the authored ranges as much as we can to avoid surprising results

### DIFF
--- a/.changeset/calm-ranges-stay.md
+++ b/.changeset/calm-ranges-stay.md
@@ -1,0 +1,6 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/cli": patch
+---
+
+Keep upper bounds when updating internal dependency ranges such as `>=1.0.0 <2.0.0`.
--- a/packages/apply-release-plan/src/index.test.ts
+++ b/packages/apply-release-plan/src/index.test.ts
@@ -403,6 +403,120 @@ describe("apply release plan", () => {
       });
     });
 
+    test.each([
+      [">=1.0.0 <2.0.0", "1.0.0", "2.0.0", "major", ">=2.0.0 <3.0.0"],
+      // Preserve the original comparator order to minimize the manifest diff.
+      ["<2.0.0 >=1.0.0", "1.0.0", "2.0.0", "major", "<3.0.0 >=2.0.0"],
+      // The refreshed lower bound has to include the new version, so `>` becomes `>=`.
+      [">1.0.0 <2.0.0", "1.0.0", "2.0.0", "major", ">=2.0.0 <3.0.0"],
+      // There is no finite `<=` equivalent for a complete major range, so it becomes `<`.
+      [">=1.0.0 <=1.9.9", "1.0.0", "2.0.0", "major", ">=2.0.0 <3.0.0"],
+      [">=1.0.0 <3.0.0", "1.0.0", "2.0.0", "major", ">=2.0.0 <3.0.0"],
+      [">=1.2.0 <1.3.0", "1.2.0", "1.3.0", "minor", ">=1.3.0 <1.4.0"],
+      [">=1.2.3 <1.2.4", "1.2.3", "1.2.4", "patch", ">=1.2.4 <1.2.5"],
+    ] as const)(
+      "should update the bounded dependency range %s",
+      async (versionRange, oldVersion, newVersion, type, expected) => {
+        const releasePlan = new FakeReleasePlan(
+          [
+            {
+              id: "some-id",
+              releases: [{ name: "pkg-b", type }],
+              summary: "a very useful summary",
+            },
+          ],
+          [
+            {
+              changesets: ["some-id"],
+              name: "pkg-b",
+              newVersion,
+              oldVersion,
+              type,
+            },
+          ],
+        );
+        const { changedFiles } = await testSetup(
+          {
+            "package.json": JSON.stringify({
+              private: true,
+              workspaces: ["packages/*"],
+            }),
+            "package-lock.json": "",
+            "packages/pkg-a/package.json": JSON.stringify({
+              name: "pkg-a",
+              version: "1.0.0",
+              dependencies: {
+                "pkg-b": versionRange,
+              },
+            }),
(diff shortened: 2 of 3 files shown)


beeequeue:
while i do agree that we shouldn't break the ranges, i'm not sure updating them is the best solution...

i think that when somone defines a complex range like this it is for a specific reason, e.g. supporting all except the latest major versions, so it getting changed would probably not be desired

maybe we should just warn that we saw a complex version range and that they should review and update it themselves? if it would be too invisible in the version command we could show it somewhere in `changeset add`

Andarist:
I think it's OK to infer probable intent for simple ranges like this one here. But the general problem ofc stays unsolved by this PR - so I agree that further warnings etc for more complex scenarios would be good to add

The baseline is that we establish a dependency between a package and its dependent - so it's implied the current package should be in its range (lower bound should really accept it). The upper bound is usually just forward-looking in some sense. Because u can't really know what you are going to release in a few months. But we need to preserve the intention and not drop it. It gets iffy when we end up having to change that upper bound... but then we just infer a sensible behavior

Such ranges make way more sense for packages outside of the user's control - and especially for already *published* versions of packages (and not for versions that *might* get released in the future)

.changeset/calm-ranges-stay.md
```diff
@@ -0,0 +1,6 @@
+---
+"@changesets/apply-release-plan": patch
+"@changesets/cli": patch
+---
+
+Keep upper bounds when updating internal dependency ranges such as `>=1.0.0 <2.0.0`.
```
beeequeue:
```suggestion
Fixed semver ranges (such as `>=1.0.0 <2.0.0`) getting cut off (`>=2.0.0`) when updating internal dependencies.
```

beeequeue:
> Such ranges make way more sense for packages outside of the user's control

that makes sense, this is internal packages only. :smile:

beeequeue (approved):

PR #2304, opened 2026-09-16
Prevent GitHub GraphQL query injection

Andarist:


beeequeue (commented):
i assume this happened because we don't use proper [arguments](https://graphql.com/learn/arguments/#querying-with-arguments)?

could we do that instead of doing brittle-ish sanitizing?

PR #2305, opened 2026-09-16
Use GraphQL variables for query arguments

Andarist:


beeequeue (approved):
the only thing i would want to change is to use a query builder of some kind rather than building a string, but im not sure there is one for this use case. 😄

